Issue #170, opened 2017-07-28
Path prefix matching for routes


mjg123:
It'd be great to be able to treat a route as a prefix when matching, rather than having to match exactly.  For example:

```
  /r/my_app/some-path-prefix/<anything>/<at>/<all>  ->  myfn
```

This doesn't mandate that path params to be evaluated by the fn platform, as long as the entire request URL is passed into the function we can do that ourselves.

This request would be especially useful for https://github.com/bbilger/jrestlesss - we can define an application with a base-url and many sub-resources hosted in a single function.  This is hard work at the moment as we need to create a lot of routes in the fn platform, and are not able to dynamically add a route at runtime.

Empty2k12:
One example of a use case for this is a function that serves a static file: You'd expect the filepath to look something like `domain.local/static/main.css` but this is currently not possible with `fn`. Instead I have to resort to hacky file serving aka `domain.local/static?file=main.css` which works but is semantically bad.

Another example is a function serving a Single Page Application that works with client-side routing á la React Router. I want the `root` function to be able to read the path and respond based on that (e.g. `domain.local/show/all` or `domain.local/show/1` handled by JavaScript and not server side)

PR #313, opened 2017-09-10
WIP : Supporting Path globs in routes

zootalures:
For context: We have a demo that puts a REST (Jersey) Java Web container inside a function.
One issue we had was that you can't bind a whole sub-path to a function route  - for demo we had to bind every possible route to the same image (which worked but means path variables are not an option). 

I know there used to be path variable support but I don't see a way to bring it back without making the database layer substantially more complex. 

I had a thought for how do this with less database impact  outlined here.  

Allow route paths of the form `/path/*` meaning  any route prefixed by `/path/` 

This does an  IN query of order of the number of components in the path  see here (assuming path is indexed on route this should be fast) 

So  `/app/path/to/fn` is searched as 

``` 
SELECT ... FROM routes where  app  WHERE app_name='app' AND  path IN ('/path/to/fn', '/*','/path/*','/path/to/*','/path/to/fn/*') ORDER BY LENGTH(path) DESC LIMIT 1
```

Main changes: 
* Add 'MatchRoute' to datastore : "Give me the right route for this URL path"  (vs GetRoute which is always based on the unique route) 
*  Allow Routes paths to contain a trailing '*' (not previously validated out, but now meaningful)  - that indicates  that the route will match any 
* This also removes the old Param code which (to my knowledge) did not work. 

MatchRoute picks the longest route that matches a subpath of the incoming route -  after caching routes are always resolved deterministically   but it's possible to define one route that takes precedence over others - figure that can be a user choice though. 

I'm not sure what the best option is for exact matches,  suppose you have route a) `/path/to/fn` and route b) `/path/to/fn/*`  - currently  a request `/path/to/fn` will match b) but a) seems more intuitive. 

I'm not sure about introducing the new syntax for routes. An alternative approach (with more change and a client change) - was having a 'wildcard' flag or similar on the function that that has the same effect of adding a '*' on to the path. 

Anyway would appreciate thoughts if anybody has any time - would be good for us but is not critical.


hibooboo2:
# Oracle CLA Bot
	Thank you for your submission, we really appreciate it. Like many open source projects,
	we ask that you sign our [Contributor License Agreement](http://www.oracle.com/technetwork/community/oca-486395.html) before we can accept your contribution.
	
- @zootalures

hibooboo2:
# Oracle CLA Bot
	Thank you for your submission, we really appreciate it. Like many open source projects,
	we ask that you sign our [Contributor License Agreement](http://www.oracle.com/technetwork/community/oca-486395.html) before we can accept your contribution.
	
- @zootalures

hibooboo2:
# CLA BotThank you for your submission!
It appears that the following authors have not signed our Contributor License Agreement:


* [](https://api.github.com/users/zootalures)


Please do so now by following [this link](http://www.oracle.com/technetwork/community/oca-486395.html). We are working on modernizing the CLA process into a digital signature but it isn’t quite ready yet.
Once complete, let us know in our [community Slack ](slack.fnproject.io) and we’ll send you an Fn T-shirt.
Thank you for being a part of the Fn Community!

hibooboo2:
# CLA Bot
Thank you for your submission!
It appears that the following authors have not signed our Contributor License Agreement:


* [](https://api.github.com/users/zootalures)


Please do so now by following [this link](http://www.oracle.com/technetwork/community/oca-486395.html). We are working on modernizing the CLA process into a digital signature but it isn’t quite ready yet.
Once complete, let us know in our [community Slack ](slack.fnproject.io) and we’ll send you an Fn T-shirt.
Thank you for being a part of the Fn Community!

carimura:
@zootalures please add Oracle as a public association to your Github account to pass this CLA check... thanks.

carimura:
oh this is old. well still needs to be done though.

rdallman:
should we move this to an issue that links to this? the code is out of date, I think there's still discussion / desire for this as a feature.

zootalures:
closes #170

Issue #259, opened 2017-08-29
JSON format prototype


treeder:
Prototype JSON format for I/O, it should be a lot easier for most languages to parse and deal with, rather than HTTP.

PR #364, opened 2017-09-26
[FEATURE] JSON hot format

denismakogon:
This feature adds new hot format: JSON.
Comparing to HTTP format, JSON takes request payload and adds certain fields that might appear useful for function and send sends it to function's STDIN.

Closes: #259


rdallman:
check this out: https://play.golang.org/p/bLZskdfB6T

easy peasy

denismakogon:
@rdallman you've mentioned the alternative that i've been testing as an alternative to what i've made at this moment, that's why i left that code in place. Don't think that i'm arguing regarding single write event, i think we'll be able to find proper ways to read JSON, no matter if that would be a byte-by-byte read, as you said, protocol dictates structure, not frame. So, feel free to review this one again.

rdallman:
ok. seems like everything still getting shoved in buffers here, not sure if you pushed. lmk when updated

denismakogon:
@rdallman there's only one buffer for request body everything else is not buffered and gets written just into STDIN. So, it's obvious why i did this for body - it needs to be encoded properly. Latest changes were pushed.

api/models/route.go
```diff
@@ -95,7 +95,7 @@ func (r *Route) Validate() error {
 		return ErrRoutesInvalidType
 	}
 
-	if r.Format != FormatDefault && r.Format != FormatHTTP {
+	if r.Format != FormatDefault && r.Format != FormatHTTP && r.Format != FormatJSON {
```
rdallman: at least note to self (not blocking), we could add `[]possibleFormats` to declarations in models so this falls out in the future

api/agent/protocol/json.go
```diff
@@ -0,0 +1,133 @@
...
+		err = json.NewEncoder(w).Encode(jout)
+		if err != nil {
+			return respondWithError(
+				w, fmt.Errorf("error writing function response: %s", err.Error()))
+		}
+	}
+	return nil
+}
+
+func respondWithError(w io.Writer, err error) error {
```
rdallman: i think we should probably just return the error from `Dispatch` rather than writing it back to the user / logs from in here. the `http` format follows this convention, and I think now that we have multiple 'hot' formats going on it makes sense for us to follow that so that we can possibly format errors to the user in a way that will be friendly to them in one place (rather than having to basically make a different `respondWithError` for every format). so I think get rid of this `respondWithError` function and just return the error in `Dispatch` -- `Dispatch`'s caller is hooked up to handle receiving the error and returning it to the user, so we don't need to do it in `Dispatch` (and caller will return json object to user or hopefully in soon future `async` will add this error to the `Call` struct -- don't worry about doing that here)

docs/function-format.md
```diff
@@ -2,7 +2,7 @@
 
 This document will describe the details of how a function works, inputs/outputs, etc.
 
-## Formats
+## Input Formats
```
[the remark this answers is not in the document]
rdallman: +1

docs/function-format.md
```diff
@@ -58,42 +62,44 @@ The header keys and values would be populated with information about the functio
 
 Pros:
 
-* Streamable
+* Supports streaming
```
rdallman: our use of the word 'streamable' is confusing, simply because from a user perspective it's not really true. our JSON format is 1 object, which is not really streamable. i.e. we're going to read the entire request into a json object and then plop it over to the function input, we're not going to write each blob the user sends over into the container as we get it (i.e. as it's own json object or however we want to frame it). 

i know that internally we use this verbiage to delineate hot from cold, doesn't really make it any less confusing from a user perspective for us to expose this semantic alias. 'hot' and 'streamable' aren't interchangeable imo, we should say 'supports hot' and not 'streamable'.

docs/function-format.md
```diff
@@ -58,42 +62,44 @@ The header keys and values would be populated with information about the functio
...
-Eg:
+An easy to parse JSON structure.
 
-```
+```json
 {
-  "request_url":"http://....",
-  "params": {
-    "blog_name": "yeezy"
+  "body": {
```
rdallman: this is out of line with what the code implemented, also we should add every field here (headers, others?)

docs/function-format.md
rdallman: oh, this is the http section. true here, but not what was intended really ;)

docs/function-format.md
```diff
@@ -58,42 +62,44 @@ The header keys and values would be populated with information about the functio
...
 Pros:
 
 * Streamable
-* Easy to parse headers
+* Easy to parse
 
 Cons:
 
-* New, unknown format
+* ???
```
rdallman: not streamable, actually, as we have defined it

docs/function-format.md
treeder: +1

docs/function-format.md
treeder: Sure, maybe say hot capable or something then.

api/agent/protocol/json.go
```diff
@@ -0,0 +1,133 @@
...
+		// - status code
+		// - body
+		// - headers
+		for k, vs := range jout.Headers {
+			for _, v := range vs {
+				rw.Header().Add(k, v) // on top of any specified on the route
+			}
+		}
+		rw.WriteHeader(jout.StatusCode)
+		_, err = rw.Write([]byte(jout.Body)) // TODO timeout
```
rdallman: i'm pretty sure that most implementers of `http.ResponseWriter` implement the `WriteString` method, so doing `io.WriteString(rw, jout.Body)` saves a pretty large alloc here

api/agent/protocol/json.go
```diff
@@ -0,0 +1,133 @@
...
+		_, err = bb.ReadFrom(req.Body)
+		if err != nil {
+			return err
+		}
+
+		err = stdin.Encode(bb.String())
+		if err != nil {
+			return err
+		}
+		_, err = io.WriteString(h.in, `",`)
```
rdallman: this and the accompanying `"` seem incorrect, see: https://play.golang.org/p/hgOXveCxhJ -- this adds extra delimiting quotes around the body's value string. i'm not sure how this manages to pass any tests written for this patch, perhaps we could write a failing test based on this discovery and make it pass.

api/agent/protocol/json.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func (h *JSONProtocol) DumpJSON(w io.Writer, req *http.Request) error {
+	stdin := json.NewEncoder(h.in)
+	_, err := io.WriteString(h.in, `{`)
+	if err != nil {
+		// this shouldn't happen
+		return err
+	}
+	if req.Body != nil {
```
rdallman: i don't think body will ever be nil outside of any test cases written, but it seems like including it as an empty string is probably preferable to omitting it as part of our defined api here

rdallman: i.e. an `http.Request` will always have a non-nil body, even if it's of length 0

api/agent/protocol/json.go
```diff
@@ -0,0 +1,133 @@
...
+		err = stdin.Encode(bb.String())
+		if err != nil {
+			return err
+		}
+		_, err = io.WriteString(h.in, `",`)
+		if err != nil {
+			// this shouldn't happen
+			return err
+		}
+		defer bb.Reset()
```
rdallman: this is basically just going to zero out this memory which we are never re-using, so is a wasted instruction and only adds confusion here. would remove

api/agent/protocol/json.go
```diff
@@ -0,0 +1,133 @@
...
+		defer req.Body.Close()
```
rdallman: if we are going to do this, it will be in the caller, as there are many dispatch methods and all of them assume consuming the `req.Body`. conveniently, for an http server, we don't need to call this, however I think in the async case we do need to. still, in caller imo

docs/function-format.md
rdallman: also, can we define the proper output format for the user here, as well? we seem to be expecting from the user's function `{"status_code":1, "headers":{"yo":["dawg"]}, "body":"sup"}` but not explicitly defining it here. for HTTP we can just say `HTTP/1.1` easily enough but we need to define the json object we are expecting explicitly.

api/agent/protocol/json.go
```diff
@@ -0,0 +1,133 @@
...
+		// this shouldn't happen
+		return err
+	}
+	if req.Body != nil {
+		_, err := io.WriteString(h.in, `"body":"`)
+		if err != nil {
+			// this shouldn't happen
+			return err
+		}
+		bb := new(bytes.Buffer)
```
rdallman: we'll just figure this out later ;)

api/agent/protocol/json.go
denismakogon: So, you're saying that there's no need check body at this place, just attempt to read it and see if it's an empty string?

rdallman: seems like it does what you would expect: https://play.golang.org/p/FTBdbti-OA -- so yes

rdallman: no need to check, just encode buffer

api/agent/protocol/json.go
```diff
@@ -0,0 +1,114 @@
...
+		return err
+	}
+
+	bb := new(bytes.Buffer)
+	_, err = bb.ReadFrom(req.Body)
+	if err != nil {
+		return err
+	}
+	reqData := bb.String()
+	if reqData != "" {
```
rdallman: yea, just skip this, it seems to do the right thing (see other comment)

api/agent/protocol/json.go
```diff
@@ -0,0 +1,114 @@
...
+		// this shouldn't happen
+		return err
+	}
+	_, err = io.WriteString(h.in, `"}`)
+	if err != nil {
+		// this shouldn't happen
+		return err
+	}
+	return nil
+}
```
rdallman: the 'this shouldn't happen' comments can all be removed, as well, they just take up space. if they shouldn't happen then don't check them. 

in particular, encoding the request body can throw an error from encoding and other reasons, and the headers could maybe throw an error since we're injecting some of them. reading from the request body can also timeout (i/o read timeout on the conn) and the pipe for `h.in` could also be closed, so that could fail. 

a somewhat nice pattern to mitigate error handling is to do something like:

```go
func writeString(err error, dst io.Writer, str string) error {
  if err != nil {
     return err
  }
  _, err = io.WriteString(dst, str)
  return err
}

func Dispatch(...) error {
// ...
err = writeString(err, h.in, "{")
err = writeString(err, h.in, `"body":`)
// ...
return err
```

this writeString func is small enough for the compiler to inline it. a function closure is nicer looking but induces a function allocation, so it's better not to do that for this particular piece of code. anyway, at least remove the comments, but it most definitely _can_ happen and we do need to try to catch the error if we can, even though we could probably try to continue writing to a broken pipe and just return the last error, it's not foolproof

api/agent/protocol/json.go
```diff
@@ -0,0 +1,114 @@
...
+type JSONProtocol struct {
+	in  io.Writer
+	out io.Reader
+}
+
+func (p *JSONProtocol) IsStreamable() bool {
+	return true
+}
+
+type RequestEncoder struct {
```
rdallman: can delete this now, unused

api/agent/protocol/json.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"io"
+	"net/http"
+)
+
+// This is sent into the function
+// All HTTP request headers should be set in env
+type JSONIO struct {
```
rdallman: this can be lower cased, since it's only used locally for encoding / decoding and the caller does not see this. as long as the fields are exported, JSON will behave correctly.

api/agent/protocol/json.go
```diff
@@ -0,0 +1,114 @@
...
+	*json.Encoder
+}
+
+func (h *JSONProtocol) DumpJSON(w io.Writer, req *http.Request) error {
+	stdin := json.NewEncoder(h.in)
+	_, err := io.WriteString(h.in, `{`)
+	if err != nil {
+		// this shouldn't happen
+		return err
+	}
```
rdallman: nit: i would move this down below reading the body, then all of the construction of the json is contiguous

docs/function-format.md
```diff
@@ -58,42 +62,68 @@ The header keys and values would be populated with information about the functio
...
 ```
+
+Internally function receives data in following format:
+
+```json
 {
-  "request_url":"http://....",
-  "params": {
-    "blog_name": "yeezy"
+  "body": "{\"some\": \"input\"}",
```
rdallman: this is incorrect, provided your example body it will be:

```
"body":"{\n\t\"body\": {\n\t\t\"some\": \"input\"\n\t}\n}\n"
```

we could probably simplify a little, here ya go https://play.golang.org/p/srantfAg5f ;)

api/agent/protocol/json_test.go
```diff
@@ -0,0 +1,102 @@
...
+	bb := new(bytes.Buffer)
+
+	_, err := bb.ReadFrom(r)
+	if err != nil {
+		t.Error(err.Error())
+	}
+	err = json.Unmarshal(bb.Bytes(), incomingReq)
+	if err != nil {
+		t.Error(err.Error())
+	}
```
rdallman: can we check that the output is what we expect it to be? for this and the other test. otherwise we're just checking that we don't get errors.

api/agent/protocol/json.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (h *JSONProtocol) DumpJSON(req *http.Request) error {
+	stdin := json.NewEncoder(h.in)
+	bb := new(bytes.Buffer)
+	_, err := bb.ReadFrom(req.Body)
+	err = writeString(err, h.in, "{")
+	err = writeString(err, h.in, `"body":`)
+	err = stdin.Encode(bb.String())
+	err = writeString(err, h.in, ",")
+	defer bb.Reset()
```
rdallman: delete

api/agent/protocol/json.go
```diff
@@ -0,0 +1,84 @@
...
+	return err
+}
+
+func (h *JSONProtocol) DumpJSON(req *http.Request) error {
+	stdin := json.NewEncoder(h.in)
+	bb := new(bytes.Buffer)
+	_, err := bb.ReadFrom(req.Body)
+	err = writeString(err, h.in, "{")
+	err = writeString(err, h.in, `"body":`)
+	err = stdin.Encode(bb.String())
```
rdallman: this will mask any prior errors, so we need to check the error before doing this or do similar to what `writeString` is for `stdin.Encode`

api/agent/protocol/json.go
```diff
@@ -0,0 +1,84 @@
...
+	err = writeString(err, h.in, ",")
+	defer bb.Reset()
+	err = writeString(err, h.in, `"headers":`)
+	err = stdin.Encode(req.Header)
```
rdallman: ditto, same as other stdin.Encode

docs/function-format.md
```diff
@@ -58,42 +62,70 @@ The header keys and values would be populated with information about the functio
...
 ```
+
+Internally function receives data in following format:
+
+```json
 {
-  "request_url":"http://....",
-  "params": {
-    "blog_name": "yeezy"
+  "body": {
```
rdallman: still incorrect here, we shove request data into body strong, so `"body"` will be in the value here

denismakogon: With the last commit it's right.

Here's what DumpRequest returns:
```
	json_test.go:57: {"body":"{\"a\":\"a\"}\n"
		,"headers":{"Content-Type":["application/json"],"Host":["localhost:8080"],"User-Agent":["curl/7.51.0"]}
		}

```

rdallman: your test is different than what you have in the docs. if you want to test that, you need to encode a struct like this:

```go
x := struct{
   Body struct{
       A string `json:"a"`
   } `json:"body"`
}{
  Body: struct{
    A: "a"
  }
}
```

so, the example in the docs is still incorrect. please address

rdallman:
the docs are still wrong, see https://github.com/fnproject/fn/pull/364#discussion_r143313043 (outdated but still wrong)

api/agent/protocol/json.go
```diff
@@ -0,0 +1,93 @@
...
+	stdin := json.NewEncoder(h.in)
+	bb := new(bytes.Buffer)
+	_, err := bb.ReadFrom(req.Body)
+	if err != nil {
+		return err
+	}
+	err = writeString(err, h.in, "{")
+	err = writeString(err, h.in, `"body":`)
+	err = stdin.Encode(bb.String())
+	if err != nil {
```
rdallman: this needs to be in front of the `stdin.Encode` call in order to avoid overwriting any prior errors, not after it

api/agent/protocol/json.go
```diff
@@ -0,0 +1,93 @@
...
+		return err
+	}
+	err = writeString(err, h.in, ",")
+	defer bb.Reset()
+	err = writeString(err, h.in, `"headers":`)
+	err = stdin.Encode(req.Header)
+	if err != nil {
```
rdallman: ditto

api/agent/protocol/json.go
```diff
@@ -0,0 +1,93 @@
...
+		return err
+	}
+	err = writeString(err, h.in, "{")
+	err = writeString(err, h.in, `"body":`)
+	err = stdin.Encode(bb.String())
+	if err != nil {
+		return err
+	}
+	err = writeString(err, h.in, ",")
+	defer bb.Reset()
```
rdallman: delete

api/agent/protocol/json.go
```diff
@@ -0,0 +1,93 @@
...
+		return err
+	}
+	err = writeString(err, h.in, "{")
+	err = writeString(err, h.in, `"body":`)
+	if err != nil {
+		return err
+	}
+	err = stdin.Encode(bb.String())
+	err = writeString(err, h.in, ",")
+	defer bb.Reset()
```
rdallman: ... remove :)

rdallman:
if user returns json object but status code is 0 should we rewrite it to 200 probably? this will return a 0 atm :/ empty headers is fine, but status one is weird.

api/agent/protocol/json.go
denismakogon: done

rdallman:
cool, seems much better now. want to test with the examples and your python library and make sure everything is good then we can merge? can optimize the buffer thing out in a bit, need to fiddle with it.

denismakogon:
@rdallman python JSON parser is not ready, that would be done later, so i excluded python's example from this PR and will push that later.

rdallman:
works for me

rdallman:
also in future we should probably parse out the response json ourselves since atm for output we are decoding into a [gigantic] buffer and then for logs re-encoding it, and for the http one we can just copy the raw body out of the json response (respString[x:y]) instead of allocating a gigantic string and copying it over. but since we don't care about that for now, moving on...

PR #370, opened 2017-09-28
make headers / env consistent

### LINKED ISSUE #329: Malformed HTTP headers for async hot functions
1. When logging incoming request headers in async hot function content length defined as `Fn_header_content_length` instead of `Content-Length`. Moreover all HTTP 1.0 spec headers are prefixed with `FN_`:
```
Fn_header_accept
Fn_header_content_type
Fn_header_content_length
```
2. App config included into headers
App info:
```
{
	"config": {
		"pg_db": "emokognition",
		"pg_host": "10.0.2.15",
		"pg_port": "5432",
		"pg_pswd": "postgres",
		"pg_user": "postgres"
	},
	"name": "emokognition"
}
```
http headers logged inside hot func
```
Pg_host: [10.0.2.15]
Pg_user
Pg_db: [emokognition]
Pg_port: [5432]
```

If this meant to be expected behaviour it doesn't look like documented.
When running same hot function but sync everything works okay.

rdallman:
these were becoming really hard to manage / brittle and the other day i had
the bright idea that we should just make our own type that can build all the
different kinds of maps we need on demand.

currently, we have a 'base' set of env vars (for hot), a 'full' set of env
vars (for cold & for storage regardless of format) and then request headers.
These different buckets are either map[string]string or map[string][]string.
For any given call type, all 3 are never necessary at the same time, and each
of them can expose different casing. this will end up storing less for async,
as we were storing the values twice, and for many request types will require
less memory overhead.

the biggest issue is that for async they were being stored in env format and
shoved back into headers, so formatting was inconsistent whether a hot
function was async or sync. so this patch does a few things:

* no more `FN_HEADER` prefix in env or headers across all types/formats.
* no more upper casing and replacing `-` with `_` for any keys in env/headers
* simplify building the env vars / headers
* headers / env vars match casing / formatting in all formats (http headers or
env vars), where env vars have comma separated list for any []string > 1

I'm open (begging, really) for concrete formatting suggestions, as I don't
really care so much as long as everything is reasonable & consistent. I think
overall wrt upper-casing and using underscores to have 'proper' env vars, a
case could be made, but we'd need to change our api to probably reflect this
-- i.e. when user sets config, we do this formatting for them, so that we
avoid case conflicts / rewriting `-` to `_` conflicts.

I looked at lambda and they for real have a different object altogether where
you define javascript to do mapping from http to your functions so that you
can format things however the hell you want to. since we don't really have the
luxury of javascript all the way down (/s), not really a great option.

closes #329


api/agent/call.go
```diff
@@ -178,22 +111,52 @@ func FromRequest(appName, path string, req *http.Request) CallOpt {
...
+		env.SetBase(k, v)
+	}
+	for k, v := range route.Config {
+		env.SetBase(k, v)
+	}
+
+	env.SetBase("FN_FORMAT", route.Format)
+	env.SetBase("FN_APP_NAME", app.Name)
+	env.SetBase("FN_PATH", route.Path)
+	// TODO: might be a good idea to pass in: "FN_BASE_PATH" = fmt.Sprintf("/r/%s", appName) || "/" if using DNS entries per app
```
denismakogon: Can you cover this TODO as well?

rdallman: i don't really agree with it, so no

treeder: The only way a user could know this would be to parse the request URL, remove the FN_PATH part of it and figure it out on their own. A little cumbersome for basic information, but anyways, certainly not essential for this PR.

rdallman:
I've updated this patch to be in line with the comments from https://github.com/fnproject/fn/pull/370#issuecomment-332833908 (thanks!), see this commit: https://github.com/fnproject/fn/pull/370/commits/b2974f1cc4f04f9b190adebe0d541d84f717a709 -- this seems pretty reasonable to me anyways, even if all the different ways we're exposing vars is pretty implementationally challenged.

would be nice to close out https://github.com/fnproject/fn/issues/329

@zootalures could I get your thoughts again here please? also cc @denismakogon for updates

zootalures:
Sorry was on vacation will look now

zootalures:
This looks good

My only thought was whether Fn_Method is now superflous in HTTP headers given it's present in the request but not a biggy.

rdallman:
thanks @zootalures -- true, we could simply delete it, I don't think it hurts for consistency since if a user say, switches between json and http (for w/e reason) and/or is using the fdk's then they may not directly have a handle on the request method anyway and be using `FN_METHOD`.

let me know (all) when ready to merge, i suspect anyone running async stuff will run into casing issue and not see their old vars (but that's the point of this patch). can notify those not following closely.

denismakogon:
@rdallman I suppose this PR is still valid, so can you please rebase it?

rdallman:
@denismakogon updated this.

treeder:
Some thoughts now that JSON is merged (this may already be known, but just to reiterate). 

For JSON format specifically:

* No headers should be in env vars 
* The only thing that should be in env vars are
  * Config vars (user defined), should be env vars that are exactly as the user defined them (no prefix, including case
  * Our special vars that ARE prefixed with `FN_`, eg: "FN_FORMAT", "FN_APP_NAME", etc.

rdallman:
i'm gonna close this until all the rules are sorted out

PR #379, opened 2017-10-01
Updating Python hot HTTP/JSON sample

denismakogon:
- using latest fdk 0.0.1 release


denismakogon:
https://github.com/fnproject/fdk-python/pull/1 <--- slightly modified version of hotfn-py.

rdallman:
@denismakogon take it this not ready since we changed library names and so on, don't see that reflected here. also needs rebase, not sure if part of last patch. want to update or should we close this ?

denismakogon:
Yeah, need to release fdk in first place.

denismakogon:
@treeder samples update with respect to first fdk release.

examples/formats/http/python/func.py
```diff
@@ -1,42 +1,35 @@
...
-        rs = response.RawResponse(
-            http_version, 200, "OK",
-            response_data=req_data)
-        print(rs.dump(), file=sys.stdout, flush=True)
-        super(MyProtocol, self).data_received(data)
-
-    def connection_lost(self, exc):
-        print('pipe closed', file=sys.stderr, flush=True)
-        super(MyProtocol, self).connection_lost(exc)
+# All Rights Reserved.
```
treeder: Remove license from this file.

examples/formats/json/python/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.6.2
```
treeder: Does it need it's own Dockerfile?  Looks like it's doing the basics that `fn build` already covers.

examples/formats/json/python/func.py
```diff
@@ -0,0 +1,34 @@
+# All Rights Reserved.
```
treeder: Remove all of these.

examples/formats/json/python/func.yaml
```diff
@@ -0,0 +1,7 @@
+name: denismakogon/hot-json-python:0.0.1
+version: 0.0.1
+runtime: docker
```
treeder: Why not just python?

examples/formats/json/python/func.py
denismakogon: They are required by apache 2.0 licence, no?

examples/formats/json/python/func.yaml
denismakogon: `python3.5.2` or `python3.6.2` to be more precise.

PR #389, opened 2017-10-04
UPT: Consistency of the if statement.

AlphaWong:


denismakogon:
That doesn't look like an improvement or bug fix.

PR #390, opened 2017-10-04
Disable docker update in CI

denismakogon:
docker version in CI is okay for us
 seems like there's certain problems with Ubuntu
 package repos which is not kinda temporary


denismakogon:
Seems like issue gone away. But i'm not sure if this would never come back again, so leaving this up to those ones who would review this PR.

rdallman:
let's close, this doesn't fix gpg keys or anything it's just a dirty work around

PR #415, opened 2017-10-10
Rules of fn core

treeder:
This defines what should go in this repo and what shouldn't. Feedback welcome. 

I will be following up with examples of easy ways to add extensions and rebuild the server.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,32 @@
 
 We welcome all contributions!
 
+## Rules of Fn core (ie: what belongs here and what doesn't)
+
+Fn server (core) is only for the essential parts of the entire Fn ecosystem. 
+These include:
+
+- The core API (apps, routes, calls, logs)
+- Executing functions (sync and async)
+- Extension points (callbacks, middleware, API additions)
+
+That's it. Everything else should be built as an extension.
+
+This does not include:
+
+- authentication
+- stats/metrics
+- special/optional features such as triggers, fdk's, workflows, event sources, etc.
+- could be argued that additional I/O formats beyond the basic ones we support should be built as plugins too
+
+Rule of thumb: If it could be built as an extension, then build it as an extension. 
+
+We WILL accept any reasonable additions to extension points in order to support building extensions. 
+
+We WILL do whatever we can to make it easy for users to add extensions (easy builds or use Go plugins). 
+
+Graduation: Some extensions can graduate into core if they become commonplace in the community (ie: majority of users are using it). 
+
 ## How to contribute
 
 * Fork the repo


treeder:
@rdallman @derekschultz @denismakogon @carimura

CONTRIBUTING.md
```diff
@@ -2,6 +2,32 @@
...
+- Executing functions (sync and async)
+- Extension points (callbacks, middleware, API additions)
+
+That's it. Everything else should be built as an extension.
+
+This does not include:
+
+- authentication
+- stats/metrics
+- special/optional features such as triggers, fdk's, workflows, event sources, etc.
```
denismakogon: maybe we should graduate certain FDKs to core?

nigeldeakin:
Moving the implementation code for stats and Prometheus metrics out of "core" sounds good, though of course this will require core to provide the required extension points (e.g. at server start when the Prom client needs to be initialised, or when a task is enqueued,dequeued, run and completed). Let me know if I can help.

Also, we want to give users a simple out-of-the-box experience for starting fn with, say, the stats/metrics extension. How will users specify what extensions they want? By means of parameters to `fn start` which automatically build a suitable docker image?

nigeldeakin:
What about tracing? Calls to the OpenTracing API should probably remain part of the core, but the OpenTracing client implementation (i.e. Zipkin), and the configuration of the trace collectors (the Http one already being used, and the new custom collector to send spans to Prometheus) should probably be add-ons.

treeder:
@nigeldeakin spot on, would be great if you could pull the prom stuff out into it's own extension. And check out this PR for and the example linked from there for building extensions: https://github.com/fnproject/fn/pull/425

CONTRIBUTING.md
```diff
@@ -2,6 +2,32 @@
...
+- could be argued that additional I/O formats beyond the basic ones we support should be built as plugins too
+
```
ottobackwards: Would using the terms plugins or extensions consistently in the documents make things more clear?

CONTRIBUTING.md
treeder: FDK's don't need to be in core.

treeder: By core, I mean this repo.

rdallman:
> What about tracing? Calls to the OpenTracing API should probably remain part of the core, but the OpenTracing client implementation (i.e. Zipkin), and the configuration of the trace collectors (the Http one already being used, and the new custom collector to send spans to Prometheus) should probably be add-ons.

+1 I agree we should standardize on the OpenTracing API for metrics logging in our code and then make it easy to plug in different backends to get that data out. 

I like this overall, as long as it's really easy to build with plugins (seems that way).

Issue #331, opened 2017-09-17
Add reason attribute to call object


denismakogon:
For instance, if you run async function that fails somehow (OOM, etc.) would be nice to have some sort of explanation what went wrong with function.

PR #424, opened 2017-10-12
CallListener - replaces RunnerListener

treeder:


api/extenders/listeners.go
```diff
@@ -0,0 +1,31 @@
+package extenders
```
treeder: Not a fan of this package name, any ideas?

derekschultz: why not just extensions?

treeder: Could do that (think that's what I had at first, but then it felt like it would be a directory of extensions, not just the interfaces). Would be nice to drop the whole `api` dir too, poorly named and unnecessary.

api/agent/agent.go
```diff
@@ -108,14 +109,15 @@ type Agent interface {
 	// TODO this should be derived from our metrics
 	Stats() Stats
 
-	// Return the http.Handler used to handle Prometheus metric requests
+	AddCallListener(extenders.CallListener)
```
rdallman: I think it would be nice if we made something that implements `agent.Agent` interface to do this. obviously, it's thin, but this is just one-more-thing and there will be other one-more-things to tack onto the `*agent.agent` (lower case... sigh), so it would be nice to avoid one-more-thing where easily possible. e.g.

```go
type listenerAgent struct {
  agent.Agent
  cls []extenders.CallListener
}

// don't implement GetCall, use underlying Agent's

func (l *listenerAgent) Submit(call Call) error {
  m := call.Model()
  for _, cl := range l.cls {
    cl.BeforeCall(ctx, m)
  }

  // actual agent submit
  err := l.Agent.Submit(call)

  m := call.Model()
  for _, cl := range l.cls {
    cl.AfterCall(ctx, m)
  }
}

// agent/agent.go initialization
func New(...) agent.Agent {
  agent = newAgent(...)
  return listener.New(agent)
}
```

this also means that users aren't expressly calling methods against the `Agent` (don't like exposing this, want to hide it), the `*server.Server` can expose `AddCallListener`. one downside i guess is maneuvering the timeout in the `context.Context` here, would need to move it from where we're currently setting it (to account for `BeforeCall`, specifically)

anyway, possibly a terrible idea, also willing to punt, just throwing ideas around.

rdallman: i guess the other thing this doesn't account for is that BeforeCall will be called before finding a 'slot', so BeforeCall has a higher chance of being called even when the call doesn't run (maybe not an issue? idk)

api/agent/agent.go
```diff
@@ -204,6 +206,11 @@ func (a *agent) Submit(callI Call) error {
 	// to make this remove the container asynchronously?
 	defer slot.Close() // notify our slot is free once we're done
 
+	err = a.fireBeforeCall(ctx, call.Model())
```
rdallman: please push this into `call.Start` after the other bits there, to keep `Submit` small and also `Start` may block the call from running, in which case we may not want `BeforeCall` to be called (?)

api/agent/agent.go
```diff
@@ -231,6 +238,11 @@ func (a *agent) Submit(callI Call) error {
 	ctx = opentracing.ContextWithSpan(context.Background(), span)
 	call.End(ctx, err)
 
+	err = a.fireAfterCall(ctx, call.Model())
```
rdallman: would shove this into `End`, maybe we need to make it return an `error` again (fine w/ me)

api/agent/call.go
```diff
@@ -331,7 +331,7 @@ func (c *call) End(ctx context.Context, err error) {
 	case context.DeadlineExceeded:
 		c.Status = "timeout"
 	default:
-		// XXX (reed): should we append the error to logs? Error field?
+		// XXX (reed): should we append the error to logs? Error field? (TR) think so, otherwise it's lost looks like?
```
rdallman: yea, want to do this https://github.com/fnproject/fn/issues/331 -- could do both, but it seems weird to add things to user logs that may not come from their code. if it were me, i'd go digging through all the code looking for some error string only to not find it and possibly go insane (possibly already insane?)

(want to add migrations to db so we don't bust everyone's db any time we add a field)

api/agent/agent.go
rdallman: leaving, but kind of leaning on just shoving into `call.Start` and `call.End`, that was the idea of those methods anyway. it brings up a point that we can do very similar to what's proposed here with agent with the `call.Call` interface, i.e. implement the `call.Call` interface (`Start`, `End`) with the `extender.CallListener` stuff, which makes sense too (maybe plumbing challenge?). just throwing ideas around again.

api/extenders/listeners.go
rdallman: `extensions` / `listeners` / `wrappers` all make sense. don't like wrappers, but maybe some synonym there.

api/agent/agent.go
```diff
@@ -231,6 +238,11 @@ func (a *agent) Submit(callI Call) error {
...
+	if err != nil {
+		return fmt.Errorf("AfterCall: %v", err)
```
rdallman: this could throw things off a little, I guess, since an error from `exec` may get masked by this function. say, timeout or docker auth failed. I guess we want to leave the option for users to be able to return their own errors to clients, though, so this makes sense too. we can still do https://github.com/fnproject/fn/pull/424/files#r144922405 and users can access underlying error.

api/agent/agent.go
treeder: I tried to leave this out of the Agent interface, but couldn't see a way to do it. And I'm guessing there'll be more callbacks within the agent down the road so I think we'll just have to live with it having these things. Which also makes breaking them apart, maybe kind of pointless?

Regarding shoving into call.Start() and call.End(), do you mean pass the listeners into those functions or pass the agent in?  Maybe we can figure out the optimal spot as a follow up. 

And it looks like it finds the slot right before the listeners are called?

api/agent/agent.go
treeder: Ok, doing this

api/agent/call.go
treeder: Perhaps we could prefix the error with something about where it came from.

treeder:
@rdallman can you review again, thanks.

api/agent/agent.go
```diff
@@ -205,7 +208,7 @@ func (a *agent) Submit(callI Call) error {
 	defer slot.Close() // notify our slot is free once we're done
 
 	// TODO Start is checking the timer now, we could do it here, too.
-	err = call.Start(ctx)
+	err = call.Start(ctx, a)
```
rdallman: probably just shove these onto `call.Listeners` when we make a call (add new field), to keep `Start`/`End` clean of parameters. can do this later.

api/agent/call.go
```diff
@@ -344,14 +350,21 @@ func (c *call) End(ctx context.Context, err error) {
 	// TODO: this should be update, really
 	if err := c.ds.InsertCall(ctx, c.Call); err != nil {
 		common.Logger(ctx).WithError(err).Error("error inserting call into datastore")
+		return err
```
rdallman: we need to explicitly not do this, so that the error sent into `End` is the one returned (generally), this is for timeouts and cancellations, primarily, but also, the task could have ran successfully and been a sync task, and we don't want to return a `500` simply because we couldn't store it in the db. We also want to try to store the logs, even if storing the call failed. I removed the comment lines about this, should probably add them back since it's not obvious.

api/agent/call.go
```diff
@@ -344,14 +350,21 @@ func (c *call) End(ctx context.Context, err error) {
...
 	}
 
 	if err := c.ds.InsertLog(ctx, c.AppName, c.ID, c.stderr); err != nil {
 		common.Logger(ctx).WithError(err).Error("error uploading log")
+		return err
```
rdallman: also delete, see previous comment

api/agent/call.go
```diff
@@ -344,14 +350,21 @@ func (c *call) End(ctx context.Context, err error) {
...
 	}
 
 	// NOTE call this after InsertLog or the buffer will get reset
 	c.stderr.Close()
+	err := t.fireAfterCall(ctx, c.Model())
+	if err != nil {
+		return fmt.Errorf("AfterCall: %v", err)
```
rdallman: imo we shouldn't rewrite this error with `fmt.Errorf` so that users can implement `models.APIError` in the returned error and have this returned from function invocations in the http response. we could log it so we know where it's from instead?

api/agent/call.go
```diff
@@ -316,22 +316,28 @@ func (c *call) Start(ctx context.Context) error {
 			return err // let another thread try this
 		}
 	}
+
+	err := t.fireBeforeCall(ctx, c.Model())
+	if err != nil {
+		return fmt.Errorf("BeforeCall: %v", err)
```
rdallman: ditto here about rewriting errors https://github.com/fnproject/fn/pull/424/files#r146943835

api/agent/listeners.go
```diff
@@ -0,0 +1,37 @@
+package agent
+
+import (
+	"context"
+
+	"github.com/fnproject/fn/api/extensions"
+	"github.com/fnproject/fn/api/models"
+)
+
+type callTrigger interface {
```
rdallman: would be kinda sexy to have this just be a `CallListener` that has a `[]CallListener` and it calls each in `Before` and `After`, but an unimportant nit; it would remove an extra type of `callTrigger` that is getting carried around, at least.

api/agent/call.go
treeder: I think we need to consider using an error wrapping/chaining lib, we lose so much context which make it really hard to debug. eg: https://github.com/hashicorp/errwrap

I just spent an hour trying to find where a JSON parsing error was happening for instance.

api/agent/listeners.go
treeder: There's already the CallListener interface.

treeder:
Removed error returns, please review.

api/agent/call.go
rdallman: the downsides are... it makes it really hard to bubble up typed errors, what I mean is probably better shown by example: https://play.golang.org/p/297H5RAolz

i.e. `models.APIError` would need to turn into... something else, or we have to explicitly make our own error wrapper that can implement `models.APIError`, or any other error interfaces we may be using / want to use (hairy, we attempted this on the runner project and you end up having to add tests for errors you expect because it's really brittle -- some errors are not so easy to generate, either)

makes ya miss exceptions (ooh it hurts to say) :(

I agree it's kind of gross / impossible to log at the root of every error site, need something.

api/agent/listeners.go
rdallman: right, I guess to be clear, it seems that `callTrigger` could pretty easily implement `CallListener` to do its work through those methods rather than having the additional `fireBeforeDispatch` and `fireAfterDispatch` methods (they are similar in nature).

api/agent/agent.go
rdallman: github reviews collapsing + beehive email delays will be the death of me (coronary). it probably makes sense to put the agent on the call for it's fields, we're atm selectively dropping in the datastore and some friends. makes sense just passing in additional fields we need so that nobody tries anything too weird, as well. i don't like my `listenerAgent` idea I left here ex-post-facto, using `Start` and `End` as we are now seems much better, but yea, think we should shove `[]CallListeners` on the `agent.Call`

api/agent/call.go
treeder: Well that's why these wrapping libs have things like https://godoc.org/github.com/hashicorp/errwrap#ContainsType so you can check for a type through the chain.

PR #425, opened 2017-10-12
Support for docker extension building

treeder:
This uses Docker's ONBUILD command to allow people to easily add extensions and build a new functions image with the extension(s). 

Based off #424. Note: this is mostly just the Dockerfile, the rest is from #424. 

Example here: https://github.com/treeder/fn-plugin-example - try commands in method 2 to use this. 

Note that this doesn't do the multi-stage build anymore since supporting the ONBUILD commands required Go to be in the image. Couldn't figure out a better way of doing it.


Dockerfile
```diff
@@ -1,25 +1,24 @@
-# build stage
-FROM golang:alpine AS build-env
-RUN apk --no-cache add build-base git bzr mercurial gcc
+FROM fnproject/dind
```
treeder: Note that this will create a bigger image since I couldn't get multi-stage to work in this model.

PR #426, opened 2017-10-12
JSON protocol updating

denismakogon:
this patch adds HTTP query string into payload (see more TODOs in code)
 adds one more test to verify query

### DIFF
--- a/api/agent/protocol/json.go
+++ b/api/agent/protocol/json.go
@@ -33,6 +33,8 @@ func writeString(err error, dst io.Writer, str string) error {
 	return err
 }
 
+// TODO(xxx): headers, query parameters, body - what else should we add to func's payload?
+// TODO(xxx): get rid of request body buffering somehow
 func (h *JSONProtocol) DumpJSON(req *http.Request) error {
 	stdin := json.NewEncoder(h.in)
 	bb := new(bytes.Buffer)
@@ -52,6 +54,12 @@ func (h *JSONProtocol) DumpJSON(req *http.Request) error {
 		return err
 	}
 	err = stdin.Encode(req.Header)
+	err = writeString(err, h.in, ",")
+	err = writeString(err, h.in, `"query_parameters":`)
+	if err != nil {
+		return err
+	}
+	err = stdin.Encode(req.URL.RawQuery)
 	err = writeString(err, h.in, "}")
 	return err
 }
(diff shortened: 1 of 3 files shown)


api/agent/protocol/json.go
```diff
@@ -33,6 +33,8 @@ func writeString(err error, dst io.Writer, str string) error {
 	return err
 }
 
+// TODO(xxx): headers, query parameters, body - what else should we add to func's payload?
```
hibooboo2: Request url request method shouldnt they be included?

api/agent/protocol/json_test.go
```diff
@@ -38,10 +40,17 @@ func TestJSONProtocolDumpJSONRequestWithData(t *testing.T) {
 		Host: "localhost:8080",
 	}
 	var buf bytes.Buffer
-	rDataBefore := RequestData{A: "a"}
-	json.NewEncoder(&buf).Encode(rDataBefore)
+
+	if data != nil {
+		_ = json.NewEncoder(&buf).Encode(data)
```
hibooboo2: What happens if this is not able to encode / errors? shouldnt you handle?

api/agent/protocol/json.go
denismakogon: They are part of env: FN_METHOD, FN_REQUEST_URL, etc.

hibooboo2: K

api/agent/protocol/json_test.go
denismakogon: Well, data being encoded here is test data, it should not fail.

api/agent/protocol/json.go
hibooboo2: Just made sense to have at top level to me but it's the either way so :).

api/agent/protocol/json_test.go
hibooboo2: Right. Gotcha 👍

api/agent/protocol/json.go
denismakogon: Other thing is auth: user/password from URL structure

treeder: Request url should be in Json since it can change every request.

denismakogon: @treeder most parts of URL already in env like HTTP method, request URL, which parts of it should also be included?

PR #433, opened 2017-10-17
Added missing -runtime go to fn init

martinahogg:
Fixing fn init command in go tutorial.

### DIFF
--- a/examples/tutorial/hello/go/README.md
+++ b/examples/tutorial/hello/go/README.md
@@ -6,7 +6,7 @@ This example will show you how to test and deploy Go (Golang) code to Fn. It wil
 
 ```sh
 # Initialize your function creating a func.yaml file
-fn init --name hello-go
+fn init --name hello-go --runtime go
 
 # Test your function. This will run inside a container exactly how it will on the server
 fn run


rdallman:
thanks @martinahogg -- did auto-detect runtime not work? if so, will debug that!

martinahogg:
Following the steps in the tutorial exactly, I ran fn init in a empty directory. So there wasn't a func.go for the autodetect to find. So I think this is purely a small issues in the README.md rather than a bug in fn init.

rdallman:
ah, that makes sense. thanks!

rdallman:
I think our tutorials are lacking clarification that one should check out the repo and be in that directory before running them. upon inspection, this tutorial will not build any go files, so the succeeding `fn run` call won't do anything I don't think? was this your experience or did the tutorial seem to work?

martinahogg:
Aaah. I see. I certainly didn't realise I needed to checkout the repo and be in the directory. That would explain why there was no sample.payload.json file too.

I think it would be better if the tutorials could all start from any empty directory and build up from there. This way the person following them knows there's no "magic" going on and they follow every step along the way. 

Thoughts?

rdallman:
lots of tutorial rewriting to be done in that case :) I don't disagree, and it would probably aid learning better, I just think unfortunately most of the ones we have expect the user to checkout the repo. shouldn't be too bad to add disclaimer at top of each README vs having to add more content to each tutorial on how to build all the files necessary.

carimura:
cc @michael-w-williams on this conversation as he's going to be working on tutorial/docs

carimura:
@martinahogg makes sense to build up tutorials from scratch. In fact looks like the state of our tutorials is not great shape -- i guess they got changed... The GoLang one seems overly complicated now, the passing params one doesn't have a README, etc.

@michael-w-williams I'd say next group of work is to go through the whole tutorials series with fresh eyes and document. Include a note at top of readme's specifying to checkout the repo first.

michael-w-williams:
I'll start on it tomorrow.

PR #434, opened 2017-10-18
add docker health check waiter to start

### LINKED ISSUE #336: hook in to docker health check api
for hot containers, they may have to initially download a lot stuff in the beginning that can take a lot of time and it would be best not to route functions to them until after this is done.

docker has a health check api https://docs.docker.com/engine/reference/builder/#healthcheck that we can use that if user containers implement we can hook in with (our driver has support for this too). after calling `docker.Start` we can try the health check until healthy and then return from `cookie.Run` so that this behavior falls out.

we could potentially hook into this health check api also to stop routing functions to containers but we can probably punt on this for now. 

atm we also have a herd issue when going from 0->1 containers up and it's exaggerated with this kind of work load (i.e. ends up launching a lot of containers instead of gracefully scaling). already an issue for this here: #274 -- ping ponging around to other `fn` servers won't really make this any better, seems best to just let everyone pile up in the 0->1 case.

rdallman:
before returning the cookie in the driver, wait for health checks
https://docs.docker.com/engine/reference/builder/#healthcheck if provided.
for images that don't have health checks, this will have no affect (an added
call to inspect container, for hot it's small potatoes).

this will be useful for containers so that they can pull large files or do
setup that takes a while before accepting tasks. since this is before start,
it won't run into the idle timeout. we could likely use these for hot
containers in general and check between runs or something, but didn't do that
here.

one nascient concern is that for hot if the containers never become healthy
I don't think we will ever kill them and the slot will 'leak'. this is true
for this and for other cases (pulling image) I think, we should probably
recycle hot containers every hour or something which would also close this.
anyway, not a huge blocker I don't think, there will likely be 1 user of this
feature for a bit, it's not documented since we're not sure we want to support
it.

closes #336

### DIFF
--- a/api/agent/drivers/docker/docker.go
+++ b/api/agent/drivers/docker/docker.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path"
 	"strings"
+	"time"
 
 	"github.com/fnproject/fn/api/agent/drivers"
 	"github.com/fnproject/fn/api/common"
@@ -426,6 +427,35 @@ func (drv *DockerDriver) startTask(ctx context.Context, container string) error 
 		} else {
 			return err
 		}
+	}
+
+	// see if there's any healthcheck, and if so, wait for it to complete
+	return drv.awaitHealthcheck(ctx, container)
+}
+
+func (drv *DockerDriver) awaitHealthcheck(ctx context.Context, container string) error {
+	// inspect the container and check if there is any health check presented,
+	// if there is, then wait for it to move to healthy before returning.
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+
+		cont, err := drv.docker.InspectContainerWithContext(container, ctx)
+		if err != nil {
+			// TODO unknown fiddling to be had
+			return err
+		}
+
+		// if no health check for this image (""), or it's healthy, then stop waiting.
+		// state machine is "starting" -> "healthy" | "unhealthy"
+		if cont.State.Health.Status == "" || cont.State.Health.Status == "healthy" {
+			break
+		}
+
+		time.Sleep(100 * time.Millisecond) // avoid spin loop in case docker is actually fast
 	}
 	return nil
 }
--- a/api/agent/drivers/docker/docker_client.go
+++ b/api/agent/drivers/docker/docker_client.go
@@ -36,6 +36,7 @@ type dockerClient interface {
 	RemoveContainer(opts docker.RemoveContainerOptions) error
 	PullImage(opts docker.PullImageOptions, auth docker.AuthConfiguration) error
 	InspectImage(ctx context.Context, name string) (*docker.Image, error)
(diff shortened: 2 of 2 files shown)


rdallman:
cc @vishvananda

rdallman:
I should mention that we explicitly do not set any retries limits / timeouts / interval here and inherit from the docker HEALTHCHECK command from the Dockerfile. so if a user does:

```
HEALTHCHECK --interval=5m --timeout=3s \
  CMD curl -f http://localhost/ || exit 1
```

then the container will not be 'healthy' for 5 minutes (again, this ignores the `idle_timeout` since it's before the container has 'started', in its eyes -- so waiting 5m works). otherwise, it uses the defaults, specified in their docs. but if no health check is provided then nothing changes for users (all users, atm, probably)

rdallman:
bump @treeder or @denismakogon

api/agent/drivers/docker/docker_client.go
```diff
@@ -36,6 +36,7 @@ type dockerClient interface {
 	RemoveContainer(opts docker.RemoveContainerOptions) error
 	PullImage(opts docker.PullImageOptions, auth docker.AuthConfiguration) error
 	InspectImage(ctx context.Context, name string) (*docker.Image, error)
+	InspectContainerWithContext(container string, ctx context.Context) (*docker.Container, error)
```
treeder: nit/curious: why are the method signatures on this interface like this?  Why isn't it just `InspectContainer(ctx context.Context, ...)`?

rdallman: we're just wrapping the docker driver client we use [to add timeouts / retries / some error filtering] https://godoc.org/github.com/fsouza/go-dockerclient so wanted to keep signatures consistent, fsouza definitely tacked on the 'WithContext' methods -- there is a bug filed to switch to the real docker client bindings but it will take a bit.

api/agent/drivers/docker/docker.go
```diff
@@ -427,6 +428,35 @@ func (drv *DockerDriver) startTask(ctx context.Context, container string) error
...
+
+		cont, err := drv.docker.InspectContainerWithContext(container, ctx)
+		if err != nil {
+			// TODO unknown fiddling to be had
+			return err
+		}
+
+		// if no health check for this image (""), or it's healthy, then stop waiting.
+		// state machine is "starting" -> "healthy" | "unhealthy"
+		if cont.State.Health.Status == "" || cont.State.Health.Status == "healthy" {
```
treeder: Will this loop forever if it goes to `unhealthy`?

rdallman: yep (for hot). i added a note in commit message for possible solution. didn't seem reasonable to bound to timeout / idle timeout (for hot), since it's only related to before the first call. cold it is bounded to timeout, which must encapsulate the time-to-become-healthy. open to idears.

rdallman:
bump @treeder

rdallman:
thanks

PR #435, opened 2017-10-18
Update func.go

carimura:
removing log


carimura:
@treeder fyi. not sure what your intention was here but it's weird extra lines in what's supposed to be a first hello world tutorial.

PR #442, opened 2017-10-19
Revert swagger back with some fixes

denismakogon:
Swagger generator is a source of a mystery. It doesn't like what swagger.io (no matter would that be OpenAPI v2 or v3) docs saying =/
Even
.
.
.
.
P.S.: @treeder @rdallman i'm unable to to update all clients because don't have appropriate rights to  push code

### DIFF
--- a/docs/swagger.yml
+++ b/docs/swagger.yml
@@ -454,42 +454,36 @@ paths:
       tags:
         - Call
       parameters:
-        - in: path
-          name: app
+        - name: app
           description: App name.
           required: true
-          schema:
-            type: string
-        - in: query
-          name: path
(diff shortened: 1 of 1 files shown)


rdallman:
thinking about it, would be really cool if we could check when commits to our swagger spec show up in CI (on master) and then generate clients from there, then they're never behind (unless something breaks / swagger spec doesn't get updated). just thinking out loud, could open issue

denismakogon:
CI check is almost there https://github.com/fnproject/fn/pull/443

treeder:
Made issue for @rdallman comment, #449 .

PR #443, opened 2017-10-19
Swagger linter for CI

denismakogon:


denismakogon:
This is so weird. Master branch swagger doc passes this check:
```
docker run --rm  quay.io/goswagger/swagger validate https://raw.githubusercontent.com/fnproject/fn/master/docs/swagger.yml
The swagger spec at "https://raw.githubusercontent.com/fnproject/fn/master/docs/swagger.yml" is valid against swagger specification 2.0
```
but swagger-lint branch swagger doc that is exact the same, doesn't pass.
```
docker run --rm  quay.io/goswagger/swagger validate https://raw.githubusercontent.com/fnproject/fn/swagger-lint/docs/swagger.yml
```

denismakogon:
hmm, seems like Github delays on updating files (actually, noticed that couple times).

denismakogon:
@rdallman can you review please?

treeder:
This seems to break builds, @denismakogon can you fix please: https://circleci.com/gh/fnproject/fn/1019?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

rdallman:
@treeder sadness is that we need to wait and then they pass once github file cache starts kicking. imo we can detect if `swagger.yml` has been updated in the branch (`git diff master --stat | grep "swagger.yml"`?) and only run this check then, @denismakogon  ?

denismakogon:
@rdallman unfortunately it doesn’t look like a fix rather than cosmetic change. We still depend on github cache. Maybe we better publish our swagger doc somewhere else and then validate it? Maybe SwaggerHub?

rdallman:
yea, but at least we would only depend on github cache for when the patches when the swagger.yml changes (this will be extremely infrequent over the lifetime of this project). would like to keep it in repo, then we have to worry about versioning external thing over a stinking text file. can we not load the file into the swagger-validator docker image and validate the version from disk instead of telling it to download?

denismakogon:
I’ll investigate, don’t see any reasons why I can’t do that, at least at this moment.

PR #447, opened 2017-10-20
Tracing to prometheus

nigeldeakin:
This change causes all completed tracing spans to be made available as Prometheus metrics, with each span name corresponding to a separate Prometheus metric. Where a span is associated with a particular function call, the path and application name are added to the metric as labels. 

The top-level span that is is associated with a particular function call is agent_submit. I've added a couple of lines of code to save the path and application name in the span as a BaggageItem. This only needs to be done here, as all child spans will "inherit" this value.

The work to extract the span name and duration out of a completed span, and send it to Prometheus, is done in a new custom Zipkin collector. This is the only place which can access duration data. This is also the place that extracts the path and appname from the span and set appropriate labels on the Prometheus metric. However the collector cannot access the BaggageItem of the span: it can only access the tags. 

To work around this limitation, I have implemented a custom Span implementation (and a custom Trace implementation to create it), which intercepts the call to Span.Finish, extracts the path and appname from the BaggageItems, and saves it in the Span using tags. This makes this information available to the collector. (Note that I have to go via BaggageItems here: we can't simply set the path and appname as tags since the Span implementation can't access Tags, though they can set them). I think the result is reasonably clean.

I have created a second example Grafana dashboard to demonstrate some of these new metrics.

grafanadashboard2


nigeldeakin:
The CI build is failing in go-fmt.sh
```
#!/bin/bash -eo pipefail
./go-fmt.sh

./api/server/server.go
Exited with code 1
```

Any ideas?

rdallman:
try running `go fmt api/server/server.go` -- recommend setting up on editor save hook (I use `goimports` myself)

rdallman:
initial thoughts, this looks really great. will dig into the code today

nigeldeakin:
Now fails with the following: will investigate.
```
    --- PASS: TestRoutes/list-and-find-route (0.04s)
FAIL
FAIL	github.com/fnproject/fn/test/fn-api-tests	38.694s
?   	github.com/fnproject/fn/test/fn-api-tests/fn/formats/json/go	[no test files]
?   	github.com/fnproject/fn/test/fn-api-tests/fn/log	[no test files]
?   	github.com/fnproject/fn/test/fn-api-tests/fn/multi-log	[no test files]
?   	github.com/fnproject/fn/test/fn-api-tests/fn/timeout	[no test files]
Exited with code 1
```

api/agent/agent.go
```diff
@@ -188,6 +188,8 @@ func (a *agent) Submit(callI Call) error {
 	ctx := call.req.Context()
 
 	span, ctx := opentracing.StartSpanFromContext(ctx, "agent_submit")
+	span.SetBaggageItem("fn_appname", callI.Model().AppName)
+	span.SetBaggageItem("fn_path", callI.Model().Path)
```
rdallman: would  be nice to have these in enclosing context, too, and for every type of request. we're spawning spans for every requests [here](https://github.com/fnproject/fn/blob/master/api/server/server.go#L103) -- `loggerWrap` in that same file has logic for extracting the `app_name` and `path` from the `*http.Request` that we could reuse, and only set if non-empty (things like `/v1/apps` won't have `path`, e.g.)

api/server/fntracer.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/log"
+	"strings"
+)
+
+// FnTracer is a custom Tracer which wraps another another tracer
+// its main purpose is to wrap the underlying Span in a FnSpan,
+// which adds some extra behaviour required for sending tracing spans to prometheus
+type FnTracer struct {
+	wrappedTracer opentracing.Tracer
```
rdallman: one cool trick w/ go is that you can embed interfaces / structs and inherit their methods. for this implementation of an `opentracing.Tracer` we only need to override `StartSpan`, so if the `wrappedTracer` were embedded then we wouldn't need to have pass through methods for `Inject` or `Extract` since the embedded `opentracing.Tracer` would be seamlessly used. the embedded tracer can be addressed by doing `thisFnTracer.Tracer`.

api/server/fntracer.go
```diff
@@ -0,0 +1,130 @@
...
+
+// NewFnTracer returns a new FnTracer which wraps the specified Tracer
+func NewFnTracer(tracerToWrap opentracing.Tracer) opentracing.Tracer {
+	newTracer := &FnTracer{}
+	newTracer.wrappedTracer = tracerToWrap
+	return newTracer
+}
+
+// FnTracer implements opentracing.Tracer
+func (thisFnTracer FnTracer) StartSpan(operationName string, opts ...opentracing.StartSpanOption) opentracing.Span {
```
rdallman: go nit: typically method receivers use short names, for this, `f`, `t`, or `ft` seem fitting.

api/server/fntracer.go
```diff
@@ -0,0 +1,130 @@
...
+
+// FnTracer implements opentracing.Tracer
+func (thisFnTracer FnTracer) Extract(format interface{}, carrier interface{}) (opentracing.SpanContext, error) {
+	return thisFnTracer.wrappedTracer.Extract(format, carrier)
+}
+
+// FnSpan is a custom Span that wraps another span
+// which adds some extra behaviour required for sending tracing spans to prometheus
+type FnSpan struct {
+	wrappedSpan opentracing.Span
```
rdallman: same trick here as above and it's more clear we only need to override `Finish*` methods

api/server/prom_zip_collector.go
```diff
@@ -0,0 +1,80 @@
...
+	return pc, nil
+}
+
+// PrometheusCollector implements Collector.
+func (pc PrometheusCollector) Collect(span *zipkincore.Span) error {
+
+	// extract any label values from the span
+	labelKeys, labelValueMap := getLabels(span)
+
+	getHistogramVecForSpanName(span.GetName(), labelKeys).With(labelValueMap).Observe(float64(span.GetDuration()) / 1000000)
```
rdallman: would be great to get millisecond resolution, as many (most?) stats are likely in that range or smaller. would prefer to see a constant of `time.Second` instead of magic number `1000000` too (easier to understand without needing a comment)

api/server/prom_zip_collector.go
```diff
@@ -0,0 +1,80 @@
...
+		histogramVecMap[spanName] = thisHistogramVec
+		prometheus.MustRegister(thisHistogramVec)
+	}
+	return thisHistogramVec
+}
+
+// PrometheusCollector is a custom Collector
+// which sends ZipKin traces to Prometheus
+type PrometheusCollector struct {
+	a agent.Agent
```
rdallman: this field seems unused?

api/server/fntracer.go
```diff
@@ -0,0 +1,130 @@
...
+
+// FnSpan implements opentracing.Span
+func (thisFnSpan FnSpan) LogEventWithPayload(event string, payload interface{}) {
+	thisFnSpan.wrappedSpan.LogEventWithPayload(event, payload)
+}
+
+// FnSpan implements opentracing.Span
+func (thisFnSpan FnSpan) Log(data opentracing.LogData) {
+	thisFnSpan.wrappedSpan.Log(data)
+}
```
denismakogon: Agree with @rdallman, most of this methods would go away when you'd embed  `opentracing.Tracer`.

api/server/prom_zip_collector.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/fnproject/fn/api/agent"
+	"github.com/openzipkin/zipkin-go-opentracing"
+	"github.com/openzipkin/zipkin-go-opentracing/thrift/gen-go/zipkincore"
+	"github.com/prometheus/client_golang/prometheus"
+	"strings"
+)
+
+// Each span name is published as a separate Histogram metric
+// Using metric names of the form fn_span_<span-name>_duration_seconds
+var histogramVecMap = make(map[string]*prometheus.HistogramVec)
```
rdallman: is there a reason this is not a field of the `PrometheusCollector` struct?

Also, should we flush this every so often in case nobody is reading (prometheus is dead / not hooked up at all)? otherwise, it seems like this will grow unbounded. the zipkin HTTP collector seems to keep 1000 buffered before overflowing, but it's a 'push' model.

api/server/prom_zip_collector.go
denismakogon: yeah, it does seem agent is not used anywhere.

api/server/fntracer.go
nigeldeakin: Yes, that's very cool. Done.

api/server/fntracer.go
nigeldeakin: OK. Done.

api/server/fntracer.go
nigeldeakin: Indeed. Done.

api/server/prom_zip_collector.go
nigeldeakin: Changed  `float64(span.GetDuration()) / 1000000` to `span.GetDuration().Seconds()`, which is of course much better. 

I chose seconds because the Prometheus documentation [best practices for naming metrics and labels](https://prometheus.io/docs/practices/naming) that metrics should use base units (i.e. seconds in this case). Is that a good enough reason?

nigeldeakin: Changed  `float64(span.GetDuration()) / 1000000` to `(time.Duration(span.GetDuration())*time.Microsecond).Seconds()`. 

I chose seconds because the Prometheus documentation [best practices for naming metrics and labels](https://prometheus.io/docs/practices/naming) that metrics should use "base units" (seconds for time units).

api/server/prom_zip_collector.go
nigeldeakin: Good spot. Removed.

api/server/prom_zip_collector.go
nigeldeakin: Agreed, it's neater for `histogramVecMap` to be a field of the `PrometheusCollector` struct. Done.

This is map of Prometheus metrics, keyed by span name. Since span names are hardcoded, the number of keys is fixed and small. We're continually sending observations to each metric, but the Prometheus client is responsible for flushing old observations, not us.

api/server/fntracer.go
nigeldeakin: Yes, agreed. That's a very nice feature: no boilerplate proxying.

nigeldeakin:
I've committed and pushed all the updates so far, partly to trigger another CI run (I don't have permissions to re-trigger the tests). I ran the tests locally and they passed.

Not yet looked at the [first comment](https://github.com/fnproject/fn/pull/447#pullrequestreview-70969887) about setting the baggage items.

nigeldeakin:
The CI tests passed this time (nothing to do with my changes).

api/agent/agent.go
nigeldeakin: I'd like to clarify this suggestion. Are you asking for me to set the `fn_appname` and `fn_path` baggage items in the `serve_http` span as well, by extracting the info from the request? Yes, I could do that. 

However this would need to be additional to setting them on the `agent_submit` span, since that span is not a child span of `serve_http` span and so would not "inherit" the baggage items.

rdallman: > so would not "inherit" the baggage items.

good point, for async I think this is true. for sync, it should be a child of. I'm guessing that we can set in both `serve_http` and `agent_submit` without much penalty for the duplicate addition in the sync case for the latter? and yes, that is what I meant. `serve_http` will cover every endpoint so we can see e.g. who is spamming the calls list endpoint, think it will be convenient.

rdallman:
> partly to trigger another CI run (I don't have permissions to re-trigger the tests)

I had this issue the other day and what fixed it was going to the circleci dashboard page and making sure i'm logged in, and then clicking into my projects and then to this repo. then in circleci I was logged in and had permission, where previously that had not been the case. was weird, anyway, might try doing that. now when I open circleci from this repo I appear logged in every time.

api/server/prom_zip_collector.go
rdallman: `time.Duration.Seconds()` returns a float64 we should be able to work with just fine 👍 

since we're multiplying by `time.Microsecond` I guess `span.GetDuration()` returns a value in microseconds that we want to convert to nanoseconds to be a time.Duration and then convert to seconds? is this logic correct?

api/agent/agent.go
nigeldeakin: Right. The main issue is that some calls to `serve_http` have a appname and path, and some do not. Prometheus requires a given metric to have the same labels every time (otherwise the go client panics with "inconsistent label cardinality"). So this would force us to change the prometheus collector to set the labels `fn_appname` and `fn_path` on *every* metric, using an empty string when a value was not available from the span. That makes the collector less generic.

Here's what `traceWrap` would look like (for the `server_http` span). It's neat enough...
```
func traceWrap(c *gin.Context) {
	// try to grab a span from the request if made from another service, ignore err if not
	wireContext, _ := opentracing.GlobalTracer().Extract(
		opentracing.HTTPHeaders,
		opentracing.HTTPHeadersCarrier(c.Request.Header))

	// Create the span referring to the RPC client if available.
	// If wireContext == nil, a root span will be created.
	// TODO we should add more tags?
	serverSpan := opentracing.StartSpan("serve_http", ext.RPCServerOption(wireContext), opentracing.Tag{Key: "path", Value: c.Request.URL.Path})
	if appName := c.Param(api.CApp); appName != "" {
		serverSpan.SetBaggageItem("fn_appname", appName)
	}
	if routePath := c.Param(api.CRoute); routePath != "" {
		serverSpan.SetBaggageItem("fn_path", routePath)
	}
	defer serverSpan.Finish()

	ctx := opentracing.ContextWithSpan(c.Request.Context(), serverSpan)
	c.Request = c.Request.WithContext(ctx)
	c.Next()
}
```

And here's what the collector might look like, with hardcoded label names:
```
// Return the HistogramVec corresponding to the specified spanName.
// If a HistogramVec does not already exist for specified spanName then one is created and configured with the specified labels
// otherwise the labels parameter is ignored.
func (pc PrometheusCollector) getHistogramVecForSpanName(spanName string, labels []string) *prometheus.HistogramVec {
	thisHistogramVec, found := pc.histogramVecMap[spanName]
	if !found {
		thisHistogramVec = prometheus.NewHistogramVec(
			prometheus.HistogramOpts{
				Name: "fn_span_" + spanName + "_duration_seconds",
				Help: "Span " + spanName + " duration, by span name",
			},
			labels,
		)
		pc.histogramVecMap[spanName] = thisHistogramVec
		prometheus.MustRegister(thisHistogramVec)
	}
	return thisHistogramVec
}

// PrometheusCollector implements Collector.
func (pc PrometheusCollector) Collect(span *zipkincore.Span) error {

	// all metrics have the same label names
	labelValueMap := &map[string]string{
		"fn_appname": "",
		"fn_path":    "",
	}

	labelKeys := []string{"fn_appname", "fn_path"}

	// fill labelValueMap with any label values found in the span
	fillWithLabelValues(span, labelValueMap)

	pc.getHistogramVecForSpanName(span.GetName(), labelKeys).With(*labelValueMap).Observe((time.Duration(span.GetDuration()) * time.Microsecond).Seconds())
	return nil
}

func fillWithLabelValues(span *zipkincore.Span, labelMap *map[string]string) {

	var keys []string

	// extract any tags whose key matches one of the keys in labelMap, and update the corresponding value
	binaryAnnotations := span.GetBinaryAnnotations()
	for _, thisBinaryAnnotation := range binaryAnnotations {
		key := thisBinaryAnnotation.GetKey()
		if thisBinaryAnnotation.GetAnnotationType() == zipkincore.AnnotationType_STRING {
			if _, ok := (*labelMap)[key]; ok {
				keys = append(keys, key)
				value := string(thisBinaryAnnotation.GetValue()[:])
				(*labelMap)[key] = value
			}
		}
	}
}

Any thoughts before I commit and push these changes?
```

nigeldeakin: Seems to look OK in Grafana (currently the legend says fn_path(fn_appname) for the series where these labels are empty: I expect this could be changed to show something better)
grafana - fn tracing span data - mozilla firefox 24102017 184647

rdallman: > this would force us to change the prometheus collector to set the labels fn_appname and fn_path on every metric, using an empty string when a value was not available from the span.

while this is pretty unfortunate, this seems like a sensible solution to me. For child spans of `serve_http` they will be consistent, generally, I would assume (maybe some db spans this isn't true). I'm assuming we will be able to pretty easily ignore empty strings like this from whatever tool we're using to read the metrics, anyway, this seems fine to me. 

The alternative I guess is to add new spans to each of our HTTP handlers and there we can pick and choose whether the endpoint will have `path` and `app_name` available from the span and it will then be consistent, though it's somewhat concerning I guess that we couldn't have e.g. span `get_calls` with label `app_name` and no `path` and also `get_calls` with both `app_name` and `path`, as both of these are perfectly valid and use the same db methods and endpoints. so maybe we will need to inject empty strings anyway and what you have proposed is preferable?

nigeldeakin: I've realised that we can avoid the extra complexity of the collector by changing `traceWrap` to set `fn_appname` and `fn_path` to empty strings when there are no values to set them to. Then I think we can leave the collector as it is now (phew). 

```
	serverSpan := opentracing.StartSpan("serve_http", ext.RPCServerOption(wireContext), opentracing.Tag{Key: "path", Value: c.Request.URL.Path})
	if appName := c.Param(api.CApp); appName != "" {
		serverSpan.SetBaggageItem("fn_appname", appName)
	} else {
		serverSpan.SetBaggageItem("fn_appname", "")
	}
	if routePath := c.Param(api.CRoute); routePath != "" {
		serverSpan.SetBaggageItem("fn_path", routePath)
	} else {
		serverSpan.SetBaggageItem("fn_path", "")
	}
	defer serverSpan.Finish()
```

rdallman: +1 -- I guess `serverSpan.SetBaggageItem("fn_appname", c.Param(api.CApp))` would probably do then fwiw since we are setting in both cases

nigeldeakin: Yes, indeed.

nigeldeakin:
Latest commits:
* Added `fn_appname` and `fn_path` labels to the `serve_http` span. In the end this turned out to be very simple.
* Updated the Grafana dashboard to allow rate and duration graphs to be displayed for any number of spans, with the user using a pair of dropdowns to choose which spans they are interested in.  Here's a screenshot
grafanadashboard2

rdallman:
> Spans which have no app and path, are shown with a line identified in the legend as fn_path (fn_appname). 

not really important but probably would expect to see simply an aggregation of all `serve_http` across all `app_name` and `path` (empty or otherwise) than simply an aggregation of all blanks. iirc grafana allows this syntax with a `*`, but I'm rusty.

Issue #311, opened 2017-09-08
How should we handle CORS?


treeder:
Kind of thinking we should just wildcard enable it by default, seems like almost all modern apps do this now. 

And perhaps have a way for user to restrict it if they like, but default is open. 

Some info on Lambda's handling of it: https://stackoverflow.com/questions/31911898/configure-cors-response-headers-on-aws-lambda

DrewRay:
ideally you would want to echo back the whitelisted set of domains for CORS.  returning a * prevents you from using authorization in many browsers. 

FROM https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS 

> **Credentialed requests and wildcards**
> 
> When responding to a credentialed request, the server must specify an origin in the value of the Access-Control-Allow-Origin header, instead of specifying the "*" wildcard.
> 
> Because the request headers in the above example include a Cookie header, the request would fail if the value of the Access-Control-Allow-Origin header were "*". But it does not fail: Because the value of the Access-Control-Allow-Origin header is "http://foo.example" (an actual origin) rather than the "*" wildcard, the credential-cognizant content is returned to the invoking web content.
> 
> Note that the Set-Cookie response header in the example above also sets a further cookie. In case of failure, an exception—depending on the API used—is raised.
> 

is this something we could configure in the YAML for each function? would it just be an array of allowed options defaulting to *.

zootalures:
Repoening this as the change above doesn't impact routes, only the API - I think we need to deal with this more elegantly for function calls

johndevs:
I've also tried everything @cfsilence tried, also not getting the access control headers through.

My expectation was that:
```
 OutputEvent.fromBytes(response.bytes, status, contentType, Headers.fromMap([ 
     'Access-Control-Allow-Origin' : '*'
     'x-foo' : 'bar' 
]))
```
would do the trick, but somehow the headers get lost.

How are one supposed to use a) customer headers b) access control headers?

rdallman:
Thanks for the bug report, that should work as outlined: users should be able to set headers on the response from a function. There are some tests that verify this is possible and I have some confidence that works at the fn level - it smells like this issue is at the fdk or function level. can dig into this to debug, some questions that may help though: the code above is using an fdk to do this? an http trigger? which version of the fdk is being used?

johndevs:
@rdallman Tested with a  http trigger and the java fdk 1.0.72

zootalures:
[...]

PR #455, opened 2017-10-23
Add CORS support to fn api

adoublebarrel:
Context
======

The api doesn't currently support [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) which prevents web clients from making direct use of it. While this can be worked around as we have in the current fn [UI](https://github.com/fnproject/ui) I'm reliably informed that CORS support is what we want long term.

This PR adds a new optional env variable "API_CORS" that accepts a comma
separated list of Origins (e.g. http://localhost:8080). This list is then used to generate the appropriate header when requests from those Origins are made.

No default Origin is included, maybe there should be for a good "out of the box" experience.

The new behaviour doesn't have any test coverage. I had trouble locating a good example of starting fn with different env vars in the suite (I haven't done a lot of Go). Any steer on a good way to test this would be welcome.

The [Gin middleware](https://github.com/gin-contrib/cors) is being used if one or more Origins are specified. Default setup for each Origin is as follows:

- GET,POST, PUT, HEAD methods allowed
- Credentials share disabled
- Preflight requests cached for 12 hours

Which are the defaults [gin-contrib/cors](https://github.com/gin-contrib/cors) comes with out of the box. This is fine for my current purposes but happy to setup different defaults if need be.

Gin-cors will return a 403 if it gets a request with an Origin header that isn't on its' list. If no Origin header is specified then it will just return the servers response.

How to Test
=========

- Start fn with `API_CORS="http://localhost:4000, http://localhost:3000" make run`
- `curl -H "Origin: http://localhost:4000" -i http://localhost:8080/v1/apps`

You should see `Access-Control-Allow-Origin: http://localhost:4000` in the response headers.

- `curl -H "Origin: http://localhost:3000" -i http://localhost:8080/v1/apps`

Now you should see `Access-Control-Allow-Origin: http://localhost:3000` in the response headers.

- `curl -H "Origin: http://localhost:5000" -i http://localhost:8080/v1/apps`

Returns a 403 response with No Access-Control-Allow-Origin header.


treeder:
related: #311 (that's one for functions rather than the API, but I'm sure they can share things so just linking).

treeder:
@adoublebarrel thanks, definitely need something like this. Since we're using Gin, could you implement what you've done with the gin cors middleware https://github.com/gin-contrib/cors ?

adoublebarrel:
@treeder Thanks for the feedback. Sure @zootalures mentioned that earlier today as well but I'd already written up the PR 🤦‍♂️ I'm still pretty new to Go, have worked with Gorilla before but not Gin so missed that. Will try swapping in the gin middleware tomorrow.

adoublebarrel:
The Gin middleware is now being used if one or more Origins are specified. Default setup for each Origin is as follows:

- GET,POST, PUT, HEAD methods allowed
- Credentials share disabled
- Preflight requests cached for 12 hours

Which are the defaults gin-contrib/cors comes with out of the box. This is fine for my current purposes but happy to setup different defaults if need be.

The gin-contrib/cors implementation differs from the Gorilla one in the way it handles requests from Origins not in the list. Gin will return a 403 whereas Gorilla will return an api response sans the Access-Control-Allow-Origin header.

The gin implementation only returns a 403 if the http request has provided an Origin header. If no Origin header is specified then gin-cors will just return the servers response.

tteggel:
I tested this patch and it works as described. I like the 'secure by default' approach.

Please squash the commits and update the commit message to reflect the behaviour with the gin plugin.

I think that you might need to update patch to keep /vendor in sync as there are some more dep changes in the glide that are not synced to vendor in this patch? But I'm not a gopher so would like a review from someone else who knows what they're doing ;)

adoublebarrel:
Thanks @tteggel. I've updated from master and added a bunch of untracked dependencies, although I'm not convinced they relate to this PR.

Going to hold off squashing to make it easier for a more seasoned gopher to check the vendor deps update is ok.

treeder:
Something is wrong with this, 10,748 files. Where are all the file changes coming from?

treeder:
LGTM. 

FWIW, I kind of like the insecure by default approach for this case as I know a large percentage of the population will want it enabled (everyone making an app with a user interface) and we'll keep having to field questions about it.

PR #457, opened 2017-10-24
Explicitly specifies the namespace

mimizone:
Explicitly specifies the namespace to avoid issues when one is set in the context


mimizone:
will move to a branch and clarify the CI workflow before creating another pull request

mimizone:
another pull request fixes the issue by removing the namespacing in the configmap https://github.com/fnproject/fn/pull/387/files/3796e69ae364480357975fe67698d149686521ec

PR #460, opened 2017-10-26
Run linter with swagger doc from disk

denismakogon:


test.sh
```diff
@@ -48,3 +48,8 @@ cd ../..
 cd examples/extensions
 go build
 cd ../..
+
+swagger_modified=$(git diff master --stat | grep "swagger.yml")
+if [ "${swagger_modified}" ]; then
```
rdallman: for `test.sh` it would be nice to just do this unconditionally, so that when somebody runs the tests after making changes it will check this even if they have files in staging or working off of a fork. i think running against the on disk file will eliminate our flakiness issues anyway. +1 on moving it here tho

PR #463, opened 2017-10-26
Update json

treeder:
Added more fields to JSON and added blank line between objects.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -502,7 +502,7 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	go func() {
 		// TODO make sure stdin / stdout not blocked if container dies or we leak goroutine
 		// we have to make sure this gets shut down or 2 threads will be reading/writing in/out
-		errApp <- s.proto.Dispatch(call.w, call.req)
+		errApp <- s.proto.Dispatch(call.Call, call.w, call.req)
 	}()
 
 	select {
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -502,7 +502,8 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	go func() {
 		// TODO make sure stdin / stdout not blocked if container dies or we leak goroutine
 		// we have to make sure this gets shut down or 2 threads will be reading/writing in/out
-		errApp <- s.proto.Dispatch(call.Call, call.w, call.req)
+		ci := protocol.NewCallInfo(call.Model(), call.req)
+		errApp <- s.proto.Dispatch(ctx, ci, call.w)
 	}()
 
 	select {
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -509,6 +509,7 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	case err := <-s.errC: // error from container
 		return err
 	case err := <-errApp:
+		// would be great to be able to decipher what error is returning from here so we can show better messages
 		return err
 	case <-ctx.Done(): // call timeout
 		return ctx.Err()
--- a/api/agent/protocol/json.go
+++ b/api/agent/protocol/json.go
@@ -46,7 +46,7 @@ func writeString(err error, dst io.Writer, str string) error {
 
 // TODO(xxx): headers, query parameters, body - what else should we add to func's payload?
 // TODO(xxx): get rid of request body buffering somehow
-func (h *JSONProtocol) DumpJSON(call *models.Call, req *http.Request) error {
+func (h *JSONProtocol) writeJSONInput(call *models.Call, req *http.Request) error {
 	stdin := json.NewEncoder(h.in)
 	bb := new(bytes.Buffer)
 	_, err := bb.ReadFrom(req.Body)
@@ -116,7 +116,7 @@ func (h *JSONProtocol) DumpJSON(call *models.Call, req *http.Request) error {
 }
 
 func (h *JSONProtocol) Dispatch(call *models.Call, w io.Writer, req *http.Request) error {
-	err := h.DumpJSON(call, req)
+	err := h.writeJSONInput(call, req)
 	if err != nil {
 		return err
 	}
@@ -136,7 +136,7 @@ func (h *JSONProtocol) Dispatch(call *models.Call, w io.Writer, req *http.Reques
 		if jout.StatusCode != 0 {
 			rw.WriteHeader(jout.StatusCode)
 		} else {
-			rw.WriteHeader(200)
+			rw.WriteHeader(http.StatusOK)
 		}
 		_, err = io.WriteString(rw, jout.Body) // TODO timeout
 		if err != nil {
--- a/api/agent/protocol/json.go
+++ b/api/agent/protocol/json.go
@@ -5,14 +5,24 @@ import (
 	"encoding/json"
 	"io"
 	"net/http"
+
+	"github.com/fnproject/fn/api/models"
 )
 
 // This is sent into the function
 // All HTTP request headers should be set in env
 type jsonio struct {
-	Headers    http.Header `json:"headers,omitempty"`
-	Body       string      `json:"body"`
-	StatusCode int         `json:"status_code,omitempty"`
+	Headers http.Header `json:"headers,omitempty"`
+	Body    string      `json:"body"`
+}
+
+type jsonIn struct {
+	jsonio
+	Config map[string]string `json:"config,omitempty"`
+}
+type jsonOut struct {
+	jsonio
+	StatusCode int `json:"status_code,omitempty"`
 }
 
 // JSONProtocol converts stdin/stdout streams from HTTP into JSON format.
@@ -35,41 +45,69 @@ func writeString(err error, dst io.Writer, str string) error {
 
 // TODO(xxx): headers, query parameters, body - what else should we add to func's payload?
 // TODO(xxx): get rid of request body buffering somehow
-func (h *JSONProtocol) DumpJSON(req *http.Request) error {
+func (h *JSONProtocol) DumpJSON(call *models.Call, req *http.Request) error {
 	stdin := json.NewEncoder(h.in)
 	bb := new(bytes.Buffer)
 	_, err := bb.ReadFrom(req.Body)
+	// todo: better/simpler err handling
 	if err != nil {
 		return err
 	}
+	// open
 	err = writeString(err, h.in, "{")
+	if err != nil {
+		return err
+	}
+
+	// body
 	err = writeString(err, h.in, `"body":`)
 	if err != nil {
 		return err
 	}
 	err = stdin.Encode(bb.String())
+	if err != nil {
+		return err
+	}
+
+	// request URL
+	err = writeString(err, h.in, `"request_url":`)
+	if err != nil {
+		return err
+	}
+	err = stdin.Encode(req.URL.String())
+	if err != nil {
+		return err
+	}
+
+	// headers
 	err = writeString(err, h.in, ",")
 	err = writeString(err, h.in, `"headers":`)
 	if err != nil {
 		return err
 	}
 	err = stdin.Encode(req.Header)
-	err = writeString(err, h.in, ",")
-	err = writeString(err, h.in, `"query_parameters":`)
-	if err != nil {
-		return err
+
+	// config
+	if call.EnvVars != nil && len(call.EnvVars) > 0 {
+		err = writeString(err, h.in, ",")
+		err = writeString(err, h.in, `"config":`)
+		if err != nil {
+			return err
+		}
+		err = stdin.Encode(call.EnvVars)
 	}
-	err = stdin.Encode(req.URL.RawQuery)
+
+	// close
 	err = writeString(err, h.in, "}")
 	return err
 }
 
-func (h *JSONProtocol) Dispatch(w io.Writer, req *http.Request) error {
-	err := h.DumpJSON(req)
+func (h *JSONProtocol) Dispatch(call *models.Call, w io.Writer, req *http.Request) error {
+	err := h.DumpJSON(call, req)
 	if err != nil {
 		return err
 	}
-	jout := new(jsonio)
+	jout := new(jsonOut)
 	dec := json.NewDecoder(h.out)
 	if err := dec.Decode(jout); err != nil {
 		return err
--- a/api/agent/protocol/json.go
+++ b/api/agent/protocol/json.go
@@ -3,6 +3,7 @@ package protocol
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 
@@ -12,8 +13,8 @@ import (
 // This is sent into the function
 // All HTTP request headers should be set in env
 type jsonio struct {
-	Headers http.Header `json:"headers,omitempty"`
-	Body    string      `json:"body"`
+	Headers map[string]string `json:"headers,omitempty"`
+	Body    string            `json:"body"`
 }
 
 type jsonIn struct {
@@ -59,7 +60,18 @@ func (h *JSONProtocol) DumpJSON(call *models.Call, req *http.Request) error {
 		return err
 	}
 
+	// call_id
+	err = writeString(err, h.in, `"call_id":`)
+	if err != nil {
+		return err
+	}
+	err = stdin.Encode(call.ID)
+	if err != nil {
+		return err
+	}
+
 	// body
+	err = writeString(err, h.in, ",")
 	err = writeString(err, h.in, `"body":`)
 	if err != nil {
 		return err
@@ -70,6 +82,7 @@ func (h *JSONProtocol) DumpJSON(call *models.Call, req *http.Request) error {
 	}
 
 	// request URL
+	err = writeString(err, h.in, ",")
 	err = writeString(err, h.in, `"request_url":`)
 	if err != nil {
 		return err
@@ -98,7 +111,7 @@ func (h *JSONProtocol) DumpJSON(call *models.Call, req *http.Request) error {
 	}
 
 	// close
-	err = writeString(err, h.in, "}")
+	err = writeString(err, h.in, "}\n\n")
 	return err
 }
 
@@ -110,17 +123,15 @@ func (h *JSONProtocol) Dispatch(call *models.Call, w io.Writer, req *http.Reques
 	jout := new(jsonOut)
 	dec := json.NewDecoder(h.out)
 	if err := dec.Decode(jout); err != nil {
-		return err
+		return fmt.Errorf("error decoding JSON from user function: %v", err)
 	}
 	if rw, ok := w.(http.ResponseWriter); ok {
 		// this has to be done for pulling out:
 		// - status code
 		// - body
 		// - headers
-		for k, vs := range jout.Headers {
-			for _, v := range vs {
-				rw.Header().Add(k, v) // on top of any specified on the route
-			}
+		for k, v := range jout.Headers {
+			rw.Header().Set(k, v) // on top of any specified on the route
 		}
 		if jout.StatusCode != 0 {
 			rw.WriteHeader(jout.StatusCode)
--- a/api/agent/protocol/json.go
+++ b/api/agent/protocol/json.go
@@ -142,7 +142,7 @@ func (h *JSONProtocol) writeJSONToContainer(ci CallInfo) error {
 	err = writeString(err, h.in, "}")
 
 	// close
-	err = writeString(err, h.in, "}\n")
+	err = writeString(err, h.in, "\n}\n\n")
 	return err
 }
 
--- a/api/agent/protocol/json.go
+++ b/api/agent/protocol/json.go
@@ -73,7 +73,7 @@ func (h *JSONProtocol) writeJSONToContainer(ci CallInfo) error {
 		return err
 	}
 	// open
-	err = writeString(err, h.in, "{")
+	err = writeString(err, h.in, "{\n")
 	if err != nil {
 		return err
 	}
@@ -142,7 +142,7 @@ func (h *JSONProtocol) writeJSONToContainer(ci CallInfo) error {
 	err = writeString(err, h.in, "}")
 
 	// close
-	err = writeString(err, h.in, "}\n\n")
+	err = writeString(err, h.in, "}\n")
 	return err
 }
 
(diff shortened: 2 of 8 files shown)


rdallman:
not really a fan, fwiw, specifically wrt the changes to Dispatch and the JSON fields. also not sure when `query_parameters` snuck in there, that one seems like it needs to go.

all additional fields added to the json blob are already available via other places. I'll enumerate:

* request url is in `req.Header["FN_REQUEST_URL"]`, and, notably, is generated slightly differently than `req.URL.String()`
* call id is available in `req.Header["CALL_ID"]`
* `call.EnvVars` the intent is to no longer send these in the 'headers' / per request, as of https://github.com/fnproject/fn/pull/370 as it is the union of the base vars, which are available in from environment variables regardless of `route.Format` and the 'headers', in `req.Header`. while i kind of agree headers/env vars are a mess, I don't think making them available from 2 different places with potentially different formats/casing per `route.Format` is going to make things any less messy. in any event, there has been some discussion on #370 and input would be appreciated.
* the json format will now override response headers that are set on the `route.Header` if specified from a given call, as opposed to adding them, as the other formats do. this is hairy and should likely be in someplace else (my b for http)

I assume this is pairing with changes to make `fdk-ruby` & co (not safe to assume -- proceeding with assumption) perhaps easier to use without having to do too much magic in the fdk itself. I am tepid about adding `*models.Call` to the `Dispatch` signature for some of the kind of inconsistency between various formats highlighted above, namely, giving access to call fields like env vars, but others as well. I do agree that it very much sucks as a dev to have to do:

```rb
cid = headers["FN_CALL_ID"][0]
```

but on the bright side we're overriding this one so that there won't be > 0, and fdk's could likely parse out the set of 'guaranteed' headers like this one and make them more easily available as fields? not sure that's any better, though. At least, it seems like the json format, if we want this, could extract the additional things we are yanking out of the `*models.Call` from the `req.Header` instead into the more "friendly" format inside of `Dispatch` as opposed to adding to `Dispatch`'s signature; this would skirt the inconsistencies of writing both `env vars` and `headers`, and the `call_id` and `request_url` are easy to extract.

I'm mostly guessing, a rationale for this change set would help in clarification. but those are my thoughts on the matter to consider, if you please, anyway.

api/agent/protocol/json.go
```diff
@@ -3,16 +3,27 @@ package protocol
...
 
 // This is sent into the function
 // All HTTP request headers should be set in env
 type jsonio struct {
-	Headers    http.Header `json:"headers,omitempty"`
-	Body       string      `json:"body"`
-	StatusCode int         `json:"status_code,omitempty"`
+	Headers map[string]string `json:"headers,omitempty"`
+	Body    string            `json:"body"`
+}
```
denismakogon: why status code no longer required?

api/agent/protocol/json.go
```diff
@@ -35,54 +46,92 @@ func writeString(err error, dst io.Writer, str string) error {
...
 		return err
 	}
+	// open
 	err = writeString(err, h.in, "{")
+	if err != nil {
+		return err
+	}
+
+	// call_id
+	err = writeString(err, h.in, `"call_id":`)
```
denismakogon: call ID is in env, so this looks redundant

api/agent/protocol/json.go
```diff
@@ -35,54 +46,92 @@ func writeString(err error, dst io.Writer, str string) error {
...
+	err = writeString(err, h.in, `"headers":`)
 	if err != nil {
 		return err
 	}
-	err = stdin.Encode(req.URL.RawQuery)
-	err = writeString(err, h.in, "}")
+	err = stdin.Encode(req.Header)
+
+	// config
+	if call.EnvVars != nil && len(call.EnvVars) > 0 {
```
denismakogon: route and app config available through env in full, why do we need them one more time here?

api/agent/protocol/json.go
```diff
@@ -35,54 +46,92 @@ func writeString(err error, dst io.Writer, str string) error {
...
+		err = writeString(err, h.in, ",")
+		err = writeString(err, h.in, `"config":`)
+		if err != nil {
+			return err
+		}
+		err = stdin.Encode(call.EnvVars)
+	}
+
+	// close
+	err = writeString(err, h.in, "}\n\n")
```
denismakogon: why do we need to write `}\n\n`

If you are reading JSON payload byte-by-byte then you should do following:
```
    while True: # read from the stream until some of the following checks will break it
        c = stream.read(1)
        if c is None:
            continue # this is actual emulates blocking, and attempts to read one more time again and again.
        if len(c) == 0: # means that previous byte was the last in payload.
            return json(s)
        if c == "\n": # this one follows the way we write payload and expects that next byte will be the last before exiting loop
            line += c
            ret = True
        elif c == "}" and ret: # this is the last byte handler
            line += c
            return json(s)
        else: # if not - keep reading until next matched check.
            ret = False
            line += c
```
So, i'm not kinda agreed to do add more bytes payload.

api/agent/protocol/json.go
```diff
@@ -35,54 +46,92 @@ func writeString(err error, dst io.Writer, str string) error {
...
+	err := h.DumpJSON(call, req)
 	if err != nil {
 		return err
 	}
-	jout := new(jsonio)
+	jout := new(jsonOut)
 	dec := json.NewDecoder(h.out)
 	if err := dec.Decode(jout); err != nil {
-		return err
+		return fmt.Errorf("error decoding JSON from user function: %v", err)
```
denismakogon: this is error handling can be even trickier, what if request body malformed? We might check the body for JSON syntax errors.

api/agent/protocol/json.go
rdallman: it's still there, it uses a separate struct now to write output that embeds this one.

api/agent/protocol/json.go
rdallman: +1

api/agent/protocol/json.go
treeder: Call ID shouldn't be in env since this is a per request value.

api/agent/protocol/json.go
treeder: @denismakogon not sure what you mean?  Previously it just showed the parsing error, eg: "Invalid character blah, blah". No context, no nothing to know where this occurred. At least this gives some reference.

api/agent/protocol/json.go
treeder: Wasn't clear how it was being delineated, now it's very clear:

```
...OBJECT 1
}\n
\n
{
OBJECT 2...
```

api/agent/protocol/json.go
treeder: Good point (same comment as @rdallman and I agree, should be in envs).

api/agent/protocol/json.go
rdallman: env, as used in this patch currently, contains the set of headers (per request) and env vars (base env)

api/agent/protocol/json.go
rdallman: i believe he meant something along these lines https://github.com/fnproject/fn/pull/424#discussion_r147214118 -- I agree.

denismakogon: @rdallman thanks for reffering to that error comment, yes that's what mean.

treeder: Commented on #424 comment.

treeder:
@rdallman addressing your long comment: Regarding first two bullets, I don't agree grabbing critical values out of a map is the best way to do do this. Having the Call object available makes a lot of sense considering everything Dispatch is doing is related to a Call. Again, need to get off the "let's make everything raw HTTP" wagon. 

Regarding bullet 3, fine by me. Although, after our FDK convos, I'm thinking this should map more closely to what we plan to offer in the FDK's. See #476.

rdallman:
> Regarding first two bullets, I don't agree grabbing critical values out of a map is the best way to do do this. 

fdk's can re-build this into the client side context rather easily, it's just kind of smelly to have values exist in > 1 place and have snowflake format values. i don't think it really makes us all that http specific, just seems like a straw man attempt -- we're going to send in per request values, regardless of format, the fact that they're called 'headers' doesn't seem like a point about http. it's pretty trivial to convert to some version that pretends it isn't http when it's basically the same, seems like we're hung up on hiding http even though it's basically http -- we can call them 'ducklings' if it would avoid adding things for the sake of hiding http.

rdallman:
i've had a comment that would likely acquiesce the apparent itch of `import "http"`, we can very likely have:

```
type Input interface{
  ID() // please no
  Ducklings() map[string][]string // since the list is not approved of, don't get hung up on this
  Env() map[string]string
  Body() io.Reader

  // whatever else we decide we don't want to access by key-value strings?
}
```

which a thin wrapper around an `http.Request` basically implements and we can have some other input format that implements this that doesn't happen to be an `*http.Request` underneath eventually (say, a `*grpc.Request` or what have you), but we need direct access to `*http.Request` fields (if the request is from http) in order to stream things in without doing any copying. then the http import is eliminated but http is masked, since this is a sticking point (and for that, we get added indirection, yay. and future proofing; but-maybe-yagni).

treeder:
Let's leave FDK's out of this for the moment, we're really just talking about the protocols and the `Dispatch` method, which for some reason you're hesitant to provide all the appropriate information, eg: the Call object. Previous to this PR, the protocol handlers have to pull values out of the headers map in the request object that was modified by adding values from the Call object. 😨  

IMO, it should be the other way around. The request object shouldn't be modified and the call object should be readily accessible to the protocol handler. The HTTP handler should be the only thing that modifies the request object since it's the only one that will be passing that along to the container. The others should probably get the real, unmodified request object. 

And lastly, I'm not saying remove http, I don't care that there's an http import or that we're using http things, I'm just saying we shouldn't force every interface method to only take stdlib HTTP objects throughout the whole system and be afraid to pass non-stdlib object around, there's no value in doing that, it's just a hindrance.

rdallman:
> Previous to this PR, the protocol handlers have to pull values out of the headers map in the request object that was modified by adding values from the Call object. 

I would say prior to this patch no protocols were even attempting to do this, so it's kind of a moot point unless we mostly agree we should start doing this (for whatever reason, as proposed here). python is/was also handing things in as `headers` key which mapped to a `map[string][]string` so just passing it down was easy enough. now we have 2 formats so can figure out what makes sense. I think the arg is really for this format as well as other formats, what do we really want? we can optimize for http and have that one fall out, but that's going to have to be translated to a grpc or some tcp framing thing or some json blob or whatever format. i maintain we're basically serving http requests so it seems an ok one to optimize for (the current ideas around kafka atm are completely unscalable, to counter the counter there, I haven't gotten around to writing a 'nice' rebuttal). 

so, we have json, which is extracting all these fields from the call to shove into the json blob, so that the fdk's/clients can read them easier than from in the headers (can't avoid _why_ here)? for grpc or for amqp (other request-like objects?) the construction will be very similar to http, in that we have to build a request-like-object that will basically (basically) have a method, headers, and a body and maybe some protocol specific bits -- not some special body that has a call id, some env vars, some headers, {other_call_fields}, and a payload, as does json. I realize json is a little different, but it seems completely sane for json to have a field for just headers and body basically, since basically every kind of higher level request-response type protocol will be defined in some way analogous to that. what we're ultimately getting at here is yea, fdk's / clients are going to have to translate from e.g. grpc/amqp/http to some nice in-language object/model/data and for most of those things it's just not so fun (hence fdk's masking). we could change to have more structured data in the body across all formats but the tradeoffs there are also... icky.

the call object has to be constructed, as well as the http headers, and these two will never really be unrelated since there's overlap. the main thing was not having to re-construct the http request that we already had coming in to send to the container, which was/is brittle and inefficient, so yea, while it's unfortunate that we're modifying the headers in place, it is kind of a pros-outweigh-the-cons kind of thing too, imo (yay computers). what's proposed is to build a call and then build a request from that, which is kinda what I'm trying to avoid yea. if we can wrap request objects more nicely than a call and limit the surface area (as noted in prior comment) then it could be workable.

> And lastly, I'm not saying remove http, I don't care that there's an http import or that we're using http things, I'm just saying we shouldn't force every interface method to only take stdlib HTTP objects throughout the whole system and be afraid to pass non-stdlib object around, there's no value in doing that, it's just a hindrance

I believe I pointed out the value in this in the last P. have proposed the wrapper interface which could possibly build each kind of object for us. i.e. we have a thing that can take grpc->json, http->json, http->http, http->default, etc. where the thing on the left is the actual wire format for the incoming request (not to be forgotten!). in the case of async right now we are reconstructing an http request (sad), it would be nice to serialize the http request to a message in the MQ and then just deserialize that (TODO exist for this). if we have the transmogrifier (TODO for this, too) then it would be easy enough for any input to turn into http and back into whatever else we need, too. i realize we are all comfy building layer cakes of abstraction, it is the app development model now, but things like kafka get a lot of sugar by having their wire format and their backend work together nicely and it's at least more fun to think about than some of the other yak's facial hairs.

this addresses everything, mostly:
* we do get sugar from not having to re-build http request
* this patch attempts to make json a snowflake
* an attempt to make everyone hold hands and say nice things about each other. in public.

rdallman:
* let's flesh this out tues, maybe with all the header stuff

denismakogon:
Meanwhile working on FDK context I figured out that having call object in `Dispatch` is kinda required in order to distinguish what are request headers (that are stored in env) and what is an app and/or config.

Having both app and route config merged makes it a bit complicated to identify whether this particular environment var is HTTP request header or config var.

rdallman:
> Meanwhile working on FDK context I figured out that having call object in Dispatch is kinda required in order to distinguish what are request headers (that are stored in env) and what is an app and/or config.

I think the plan here is that app and route configs are only present in environment, so this isn't true if we do that.

denismakogon:
I'm okay keeping them in env, but still not clear how do I know if env var `A` is a request header, app config var or route config var? That's the point I'm trying to make. I feel a bit uncomfortable when I start considering `context.Config()` as i copy of `os.Env()`.

rdallman:
I guess the thinking is that they are in separate buckets in the fdk context. i.e. `ctx.Headers()` and `ctx.Config()` or something along those lines. this is fdk specific. without fdks, they are simply in different places to delineate (headers in headers, app/route config in env). as far as app vs. route config, it's true we don't say which is which and route overrides app but when we have versioning of routes this should be easy enough to look up wrt auditing.

treeder:
> i.e. dispatch should explicitly not have access to config and app vars (ideally)

@rdallman why is that ideal? Dispatch should have all information it needs to pass into the container, which includes config vars and call information as well as the unmodified Request object. It shouldn't have to extract this information from the Request objects headers. Again, I think the HTTP Dispatch should be the only thing that does anything with the headers since that is what it will pass into the container, none of the other protocols will be doing that.

treeder:
Ok, @rdallman and I came up with a good plan for this, we'll pass in an interface with the required data so protocol can't modify the call object with methods like `CallID(), Writer(), etc`. Also, only the http protocol will add modify request headers to pass into the container.

denismakogon:
@treeder sounds reasonable, of course, if the main reason for forbidding protocol Dispatch access and modify call object. Meanwhile, we talk here about modifying JSON protocol fields by extending one to something new, can't we make OpenEvent-compatible payload here? We're almost there. Thoughts?

denismakogon:
```
{
  "eventId": "66dfc31d-6844-42fd-b1a7-a489a49f65f3", # call ID
  "eventType": "google.storage.object.finalize", # let's hard code one to http_api_call or whatever looks reasonable
  "createdAt": "2017-09-28T19:03:12Z", # i think we can get one because timer started before dispatch
  "data": {}, # request payload
  "dataContentType": "application/json", # Content-Type
  "resource": { # this might be blank at this moment
    "type": "storage#object",
    "id": "storage.googleapis.com"
  },
  "specVersion": "0.1", # we're not versioning protocols yet, so might be black
  "attributes": {} # include headers here would those ones be in Go-http headers or HTTP 1.1 style
}
```

treeder:
Alright, revamped this a bit, please take a look. 

The format this generates and expects is documented here: https://github.com/fnproject/fn/blob/3a481d4325d7500710a06bdf7aeb8606b4fb656e/docs/function-format.md#input

Not the "protocol" map to hold protocol specific things, that user can ignore or use at his/her own risk. Also directly aligns with our FDK discussions.

api/agent/protocol/json.go
```diff
@@ -35,59 +63,125 @@ func writeString(err error, dst io.Writer, str string) error {
 
 // TODO(xxx): headers, query parameters, body - what else should we add to func's payload?
 // TODO(xxx): get rid of request body buffering somehow
-func (h *JSONProtocol) DumpJSON(req *http.Request) error {
+// @treeder: I don't know why we don't just JSON marshal this, this is rough...
```
denismakogon: because if request body, we still don't have a better idea how to avoid buffering and stream it right into a container.

api/agent/protocol/json_test.go
```diff
@@ -138,8 +142,8 @@ func TestJSONProtocolDumpJSONRequestWithQuery(t *testing.T) {
 	if err != nil {
 		t.Error(err.Error())
 	}
-	if incomingReq.QueryParameters != req.URL.RawQuery {
-		t.Errorf("Request query string assertion mismatch: expected: %s, got %s",
-			req.URL.RawQuery, incomingReq.QueryParameters)
-	}
+	// if incomingReq.QueryParameters != req.URL.RawQuery {
```
denismakogon: this should be changed to request URL assertion

api/agent/protocol/json.go
treeder: Does it matter?  If it's too big, it'll blow up inside the container, either way it's getting fully loaded into an object. I'd vote for readable code/maintainability over a fraction of a millisecond of time marshaling to JSON. I believe you did this before and Reed had you change it ya?  I'll leave it for now, but as we add more fields, this will get harder and harder to maintain/read than it already is. 

And if we do want to support super large bodies, we should (must?) implement another scheme for it, eg: https://github.com/fnproject/fn/issues/500

api/agent/protocol/json_test.go
treeder: Updating tests (and fixing circle errors).

hhexo:
I'll have a deeper look at this tomorrow, but for now I have two at-a-glance comments...

I'm a bit concerned at the level of escaping that will be required for the "body" field. What happens if the input already has strings containing quotes that are escaped? Then those will have to be double-escaped... do we have a reliable solution for encapsulating general json in a json string? (maybe golang can already do json reliably, I'm not an expert)

Secondly, I guess the "protocol" section could be handled as an arbitrary nested hashmap in most languages, though in Java it might require a bit of work. Still, shouldn't we have the protocol "type" in the output as well? The current example / spec doesn't have it.
Or is it enforced to be the same protocol as the input? It would make sense... Or could a function receive http and spit out a protocol section for, say, a kafka event as output? What's the behaviour in that case?

treeder:
@hhexo the body is a string, so quotes and things have to be escaped, as you see in the example, they don't need to be double escaped beyond what the example shows. The function gets a string in the body, and they can do with as they please. If using an FDK, the FDK may deal with it for you. 

The protocol section is an arbitrary map, the docs just show an example. And you're right, the docs don't have the type field, I'll add. I suppose the type field is the one thing that's not arbitrary.

hhexo:
If the type field is required, could it be at the same level of "protocol"? Maybe always have "protocol_type" and rename the map to "protocol_data"?

Regarding escaping, my doubts were just about what happens when I send some input json like this:
```
{
  "sentence": "\"Elementary, my dear Watson!\" he said."
}
```

Then the does the json received by the function look like this?

```
{
  "call_id": "123",
  "content_type": "application/json",
  "body": "{\n"sentence": \"\\\"Elementary, my dear Watson!\\\" he said.\"\n}",
  ...
}
```
(i.e. with escaped backslashes and special characters)
Maybe the json libraries for the various languages will just "do the right thing" when encoding/decoding.

Another question... I notice that for json format, the body is represented as a `string`, but what happens if I make an http request to the function and the request has binary data in the body? Do we forbid that?

treeder:
re: escaping, yes, that is what would happen, json libs should handle it just fine. 

re: type: Was thinking the same thing on the protocol type thing, might be better top level. or to keep top level cleaner:

```json
"protocol": {"type": "x", "data": "MAP"}
```

re: binary: they'd have to base64 encode (or some other encoding perhaps) the data, but that's a limitation of JSON itself so not much we can do there. Although, this could probably also be solved with something like #500 if we need to do it.

api/agent/protocol/json.go
rdallman: we have the ability to limit the size of the reader itself, that code is checked in as a server option; not really the issue. memory pressure / fragmentation is the real issue, json uses sync.Pool but it's just unnecessary overhead and those buffers are not properly sized and need expansion frequently, so it's really not doing such a great job across heterogeneous bodies. 

it would be nice if the json package had support for `MarshalTo(io.Writer) error` like normal writers, but it don't.

i can think of clever hacks to get mostly `json.Marshal` functionality, where we can put the fields in a temporary struct, marshal that, bite off the closing `}` and then copy that followed by `"body":""}` which may make the json dispatch slightly more palatable, but it seems the right place to optimize really.

api/agent/agent.go
```diff
@@ -502,13 +502,15 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
...
-		errApp <- s.proto.Dispatch(call.w, call.req)
+		ci := protocol.NewCallInfo(call.Model(), call.req)
+		errApp <- s.proto.Dispatch(ctx, ci, call.w)
 	}()
 
 	select {
 	case err := <-s.errC: // error from container
 		return err
 	case err := <-errApp:
+		// would be great to be able to decipher what error is returning from here so we can show better messages
```
rdallman: this gets processed in `call.End`, where we can pick from any of the sources of the error (a much better place)

api/agent/protocol/json.go
```diff
@@ -142,7 +142,7 @@ func (h *JSONProtocol) writeJSONToContainer(ci CallInfo) error {
 	err = writeString(err, h.in, "}")
 
 	// close
-	err = writeString(err, h.in, "}\n\n")
+	err = writeString(err, h.in, "}\n")
```
denismakogon: I still think writing `\n` in the end is not necessary. The result of `writeString` would be the same as you expect to get with `}\n`
If you call `writeString(err, h.it, "}")` it will look like:
```
}

```
on the next line new request data will appear, no additional padding really needed. The thing we actually get from this padding (line 76 and 145) is visibility for debugging purpose, nothing else. At this particular moment, there's no problem with distinguishing two JSON requests without padding. So, i still feel uncomfortable with padding (+2 bytes, but who really counts ;-)  )

Issue #464, opened 2017-10-26
Fn  calls  failing with "inconsistent label cardinality"  prometheus errors 


zootalures:
As of about 0.3.163 or 0.3.164  APi calls are panicing with errors: 

```
time="2017-10-26T13:07:12Z" level=info msg="starting call" action="server.handleRunnerRequest)-fm" app=travel container_id=01BXCCPBR867WG400000000000 id=01BXCCPBN367WG200000000000 route=/trip
panic: inconsistent label cardinality

goroutine 158 [running]:
github.com/fnproject/fn/vendor/github.com/prometheus/client_golang/prometheus.(*MetricVec).With(0xc420650000, 0xc4201d3020, 0xc42019afe0, 0xc420011980)
	/go/src/github.com/fnproject/fn/vendor/github.com/prometheus/client_golang/prometheus/vec.go:146 +0x76
github.com/fnproject/fn/vendor/github.com/prometheus/client_golang/prometheus.(*HistogramVec).With(0xc4204fa598, 0xc4201d3020, 0x14, 0xc42019afe0)
	/go/src/github.com/fnproject/fn/vendor/github.com/prometheus/client_golang/prometheus/histogram.go:344 +0x38
github.com/fnproject/fn/api/server.PrometheusCollector.Collect(0xc4204cb890, 0xc4203c3800, 0x3b9c6a843, 0x1a5c7a0)
	/go/src/github.com/fnproject/fn/api/server/prom_zip_collector.go:51 +0x94
github.com/fnproject/fn/api/server.(*PrometheusCollector).Collect(0xc420542198, 0xc4203c3800, 0x3b9c6a843, 0x1a5c7a0)
	<autogenerated>:1 +0x46
github.com/fnproject/fn/vendor/github.com/openzipkin/zipkin-go-opentracing.(*Recorder).RecordSpan(0xc42019c1e0, 0x23b3d946e6f61be3, 0x3e40767a31786250, 0x1ac25441f6f41e58, 0x1, 0xc4201491a0, 0xc420335190, 0x0, 0x1, 0x11a652d, ...)
	/go/src/github.com/fnproject/fn/vendor/github.com/openzipkin/zipkin-go-opentracing/zipkin-recorder.go:186 +0x6ee
[...]

PR #465, opened 2017-10-26
Add protection against inconsistent cardinaliity errors

nigeldeakin:
This PR is intended to provide a fix to https://github.com/fnproject/fn/issues/464

That bug reports that in some (unknown) circumstances, the Prometheus go client is complaining that the Fn server is attempting to record a metric which has a different number of labels than it did when it was first sent. 

I haven't been able to reproduce this. The only way I can think of that might cause this error is that spans of a particular name are sometimes created with a parent span which has fn_appname and fn_path set, and are sometimes created without a parent span. However I haven't found how this might happen. 

Whatever the cause, this PR is intended to prevent something like this causing "inconsistent cardinaliity errors". When a metric is first created (e.g. the first time a span with a particular name is ended), we now record the label names that are specified. Then, when subsequent spans with the same name are ended, a new block of code ensures that an identical set of label names are used. If a subsequent span contains a label key which was not specified initially then it is ignored. And if a label key was specified initially, but is found to be missing from a subsequent span, then a value of "" is used.

This means that the labels used for a given span are determined by those that are "set" via baggage items (either directly or inherited) on the very first span to complete. This is potentially non-deterministic, but the only way to avoid this would be to hardcode the label names used for every span name. That sound undesirable as currently most span names simply inherit their parent labels and don't have to contain any special code.

### DIFF
--- a/api/server/prom_zip_collector.go
+++ b/api/server/prom_zip_collector.go
@@ -14,41 +14,63 @@ type PrometheusCollector struct {
 
 	// Each span name is published as a separate Histogram metric
 	// Using metric names of the form fn_span_<span-name>_duration_seconds
+
+	// In this map, the key is the name of a tracing span,
+	// and the corresponding value is a HistogramVec metric used to report the duration of spans with this name to Prometheus
 	histogramVecMap map[string]*prometheus.HistogramVec
+
+	// In this map, the key is the name of a tracing span,
+	// and the corresponding value is an array containing the label keys that were specified when the HistogramVec metric was created
+	registeredLabelKeysMap map[string][]string
 }
 
 // NewPrometheusCollector returns a new PrometheusCollector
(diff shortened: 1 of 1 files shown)


rdallman:
This label cardinality thing is quite a pain! I wonder if another prometheus client may not take such an issue with this? maybe prometheus will end up not liking it though, anyway?

For the load balancer the plan is to send a span in to `fn` servers, as well, I wonder if this could be / would also trigger the cardinality issue. We do want to add some tags onto that span, and then have `fn` continue that span.

For what it's worth, I ran with this patch included for testing all yesterday without seeing this, so I'm not sure it's anything within `fn` alone but it could be lingering somewhere I don't normally hit.

Thanks for the patch, let's get this in to avoid the panics at least.

PR #466, opened 2017-10-27
Call setTracer() before Agent creation

nigeldeakin:
This PR fixes an intermittent race condition which sometimes caused the following  panic at startup:
```
panic: interface conversion: opentracing.SpanContext is opentracing.noopSpanContext, not zipkintracer.SpanContext
goroutine 74 [running]:
github.com/fnproject/fn/vendor/github.com/openzipkin/zipkin-go-opentracing.(*tracerImpl).startSpanWithOptions()
	/go/src/github.com/fnproject/fn/vendor/github.com/openzipkin/zipkin-go-opentracing/tracer.go:309  
github.com/fnproject/fn/vendor/github.com/openzipkin/zipkin-go-opentracing.(*tracerImpl).StartSpan()
	/go/src/github.com/fnproject/fn/vendor/github.com/openzipkin/zipkin-go-opentracing/tracer.go:268  
github.com/fnproject/fn/api/server.FnTracer.StartSpan()
	/go/src/github.com/fnproject/fn/api/server/fntracer.go:23  
github.com/fnproject/fn/api/server.(*FnTracer).StartSpan()
	<autogenerated>:1  
github.com/fnproject/fn/vendor/github.com/opentracing/opentracing-go.startSpanFromContextWithTracer()
	/go/src/github.com/fnproject/fn/vendor/github.com/opentracing/opentracing-go/gocontext.go:52  
github.com/fnproject/fn/vendor/github.com/opentracing/opentracing-go.StartSpanFromContext()
	/go/src/github.com/fnproject/fn/vendor/github.com/opentracing/opentracing-go/gocontext.go:44  
github.com/fnproject/fn/api/agent/drivers/docker.(*dockerWrap).CreateContainer()
	/go/src/github.com/fnproject/fn/api/agent/drivers/docker/docker_client.go:213 
github.com/fnproject/fn/api/agent/drivers/docker.(*DockerDriver).Prepare
	/go/src/github.com/fnproject/fn/api/agent/drivers/docker/docker.go:136  
github.com/fnproject/fn/api/agent.(*agent).prepCold()
	/go/src/github.com/fnproject/fn/api/agent/agent.go:560  
github.com/fnproject/fn/api/agent.(*agent).launch.func1()
	/go/src/github.com/fnproject/fn/api/agent/agent.go:529 
created by github.com/fnproject/fn/api/agent.(*agent).launch
	/go/src/github.com/fnproject/fn/api/agent/agent.go:528  
```

This appears to be caused by an attempt to set a Span created using the default Tracer to be configured as the parent of a Span created using the Zipkin Tracer. It seems likely that this was because the parent span was created before the Zipkin Tracer was configured.

To fix this, this PR changes the `New` function in `api/server/server.go` to set the global tracer before creating the agent and server.

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -77,6 +77,9 @@ func NewFromEnv(ctx context.Context, opts ...ServerOption) *Server {
 
 // New creates a new Functions server with the passed in datastore, message queue and API URL
 func New(ctx context.Context, ds models.Datastore, mq models.MessageQueue, logDB models.LogStore, opts ...ServerOption) *Server {
+
+	setTracer()
+	
 	s := &Server{
 		Agent:     agent.New(cache.Wrap(ds), mq), // only add datastore caching to agent
 		Router:    gin.New(),
@@ -86,7 +89,6 @@ func New(ctx context.Context, ds models.Datastore, mq models.MessageQueue, logDB
 	}
 
 	setMachineId()
-	s.setTracer()
 	s.Router.Use(loggerWrap, traceWrap, panicWrap)
 	s.bindHandlers(ctx)
 
@@ -119,7 +121,7 @@ func traceWrap(c *gin.Context) {
 	c.Next()
 }
 
-func (s *Server) setTracer() {
+func setTracer() {
 	var (
 		debugMode          = false
 		serviceName        = "fn-server"
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -79,7 +79,7 @@ func NewFromEnv(ctx context.Context, opts ...ServerOption) *Server {
 func New(ctx context.Context, ds models.Datastore, mq models.MessageQueue, logDB models.LogStore, opts ...ServerOption) *Server {
 
 	setTracer()
-	
+
 	s := &Server{
 		Agent:     agent.New(cache.Wrap(ds), mq), // only add datastore caching to agent
 		Router:    gin.New(),


nigeldeakin:
Had a brief discussion with @hhexo about this: essentially he was pointing out that there are some operations performed before this function is called, and that since the global tracer is global we can't prevent these performing tracing, even though they don't appear to do currently. The code in question is in server.go:

```
func NewFromEnv(ctx context.Context, opts ...ServerOption) *Server {
	ds, err := datastore.New(viper.GetString(EnvDBURL))
	if err != nil {
		logrus.WithError(err).Fatalln("Error initializing datastore.")
	}

	mq, err := mqs.New(viper.GetString(EnvMQURL))
	if err != nil {
		logrus.WithError(err).Fatal("Error initializing message queue.")
	}

	var logDB models.LogStore = ds
	if ldb := viper.GetString(EnvLOGDBURL); ldb != "" && ldb != viper.GetString(EnvDBURL) {
		logDB, err = logs.New(viper.GetString(EnvLOGDBURL))
		if err != nil {
			logrus.WithError(err).Fatal("Error initializing logs store.")
		}
	}

	return New(ctx, ds, mq, logDB, opts...)
}
```
However the change in this PR certainly doesn't make things any worse, and should in practice fix the panics I observed (which occurred in threads started by the agent).

rdallman:
thanks. initialization always ends up being kind of sad in my experience, just have to comment so that people don't shuffle things around and mysteriously break things down the road :) -- doing this first in server init is probably as good as it's gonna get since we're going to need a handle on the config, and in theory without agent running (pulling calls from MQ) or the http server running, nothing should really be going on before that anyway so this should do the trick.

PR #484, opened 2017-11-07
link broken, changed to iron-io/logspout-statsd

stevencasey:
searched in github and could only find iron-io with same project name.


treeder:
Forked: https://github.com/treeder/logspout-statsd

Should probably remove that whole line though as we're doing metrics differently now.

PR #486, opened 2017-11-08
Docker stats to Prometheus

nigeldeakin:
This change makes statistics from the running docker container available as Prometheus metrics. 

Here's a summary of the various changes made:

There is already a mechanism for extracting stats from the running docker containers. Here is a brief summary:

* Each time a container is started, a goroutine is started which calls the function `docker.Stats`. This is part of the docker client API and causes statistics to be fetched every second from the running container. 
* Whenever statistics are received, the existing function `cherryPick` is used to extract selected statistics and pass them to a callback method `WriteStat`. This function was empty and didn't do anything.
* The collection of statistics stops when the container is terminated.

I have changed this as follows:

* The `WriteStat` function now starts a span (called `docker_stats`), logs to this span the various metric values extracted by `cherryPick`, and then ends the span. 
  * Note that we start and end a new span for every call to `WriteStat` (once a second or so). I did consider logging the docker metrics to the parent span, since this continues until the container is terminated. However since the collector is only invoked when a span ends, that would have meant that no data would have been sent to Prometheus until the container terminated - which is probably undesirable for long-lived or hot functions.

* Previously the collection of statistics ended as soon as the container had started. This was clearly incorrect so I have changed this so that the collection of statistics continues until the container is terminated.

In addition, 

* I have extended the Prometheus collector to extract the logged metrics from the span and send them to Prometheus as HistogramVec metrics.

* I have created a third Grafana dashboard to demonstrate these new metrics and updated the README accordingly.

* In testing I discovered an issue with hot functions which meant that the span created in agent.Submit (which is where the appname and path name are saved in the span as baggage items) wasn't configured as an ancestor to the new span created in `WriteStat`. This meant that the metrics didn't have any fn_app and fn_path labels. I discovered that this was because `agent.runHot` deliberately ignores the context of its caller and creates a completely new context, in order to avoid propagating the caller context's timeout. I fixed this by extracting the span from the caller's context and setting it on the new context.

Here are the new metrics. As with the span metrics, these "inherit" the labels `fn_app` and `fn_path` from the parent span.

| Prometheus metric name |
| ------------- |
| `fn_docker_stats_fn_cpu_kernel` |
| `fn_docker_stats_fn_cpu_kernel`  |
| `fn_docker_stats_fn_cpu_user` |
| `fn_docker_stats_fn_disk_read` |
| `fn_docker_stats_fn_disk_write` |
| `fn_docker_stats_fn_mem_limit` |
| `fn_docker_stats_fn_mem_usage` |
| `fn_docker_stats_fn_net_rx` |
| `fn_docker_stats_fn_net_tx` | |

Caveats:

* Docker provides many more statistics than the ones extracted by `cherryPick`. I have not investigated whether additional metrics are needed.  Note that if we do need to extract more metrics from docker then `cherryPick1 is the only function that needs to be changed.

* If the container is very short-lived (i.e. for fast cold functions) then there is sometimes insufficient time to obtain statistics before the container terminates,  which means no metrics are published. I did consider delaying the termination of the container until at least one set of statistics had been obtained, but decided against doing this.


nigeldeakin:
I see CircleCI is failing with a
```
api/agent/drivers/docker/docker.go:275: the cancel function returned by context.WithTimeout 
should be called, not discarded, to avoid a context leak
```
This is caused by my refactoring of the agent/driver interface so that the context passed to `docker.collectStats()` is not cancelled until `agent.exec()` finishes (previously it was cancelled when `docker.run()` finished, which was too soon.

To fix that issue, I changed `docker.run()` so that instead of creating a new context, passing into `docker.collectStats()` and then cancelling it immediately `run()` returned, I used the context passed in from `agent.exec()` instead.

However  when `docker.run()`  created a context it also set a timeout, and I didn't want to lose that behaviour. I therefore changed  `docker.run()` to wrap the agent's context into a new context.WithTimeout and passed that into `docker.collectStats()`. However I didn't cancel that context when the method ended since that was exactly the behaviour I was trying to avoid. 

However this has been detected by the `go vet` run in CircleCI, which considers this a potential resource leak. 

I'm still pondering how to fix this. My first thought was to create the timeout context in the agent, but that didn't work since the agent doesn't have visibility to access the timeout value.

nigeldeakin:
Having spent the day working on this, I think that it is probably a bad idea for `docker.run()` to depend on the exact point at which the calling function (`agent.exec()`) cancels the context. It is particularly inappropriate given that  `docker.run()` returns a `WaitResult`, which `agent.exec()` waits on before closing the context. It's clearly better if  `docker.run()` makes us of the  `WaitResult` directly.

This however means we need to have two separate threads waiting on the same  `WaitResult`, and I have confirmed that this doesn't work. I will therefore need to extend `WaitResult` to support multiple waiters. This is looking good so far, expect an updated PR tomorrow.

nigeldeakin:
I've changed `docker.Run()` so that it doesn't make any assumptions about when the caller's context will be cancelled. Instead it now uses the `WaitResult` that this function itself creates, and waits for that to determine when the container has stopped and the collection of docker statistics should be terminated. 

This required changes to the implementation of `WaitResult` to allow multiple threads (two in practice) to call `Wait` concurrently. The only issue I needed to solve was that the second or subsequent thread to close the `docker.CloseWaiter` would get a panic.  This panic is caught and the thread waits for the first thread to complete before itself completing. 

I have added a simple test case to this as well, using ten threads simultaneously calling `Wait` on the same `WaitResult`.

I think this is now ready to review.

rdallman:
> One thing to solve - metrics from cold functions.

this handles both as it is.

api/agent/drivers/docker/docker.go
```diff
@@ -269,55 +269,98 @@ func dockerMsg(derr *docker.Error) string {
...
 // The docker driver will attempt to cast the task to a Auther. If that succeeds, private image support is available. See the Auther interface for how to implement this.
-func (drv *DockerDriver) run(ctx context.Context, container string, task drivers.ContainerTask) (drivers.WaitResult, error) {
+func (drv *DockerDriver) run(ctx0 context.Context, container string, task drivers.ContainerTask) (drivers.WaitResult, error) {
 	timeout := task.Timeout()
 
-	var cancel context.CancelFunc
+	// we create and use the following contexts in this func
+	// ctx0        - parent context
+	// ctxForStats - new child context (with timeout if specified) - passed to collectStats and cancelled when the waiter completes
+	// ctx2        - new child context (with timeout if specified) - used for remaining calls and cancelled at end of func
```
rdallman: i believe most of these were created here simply due to the confusion made by me of leaving the `defer cancel()` and `go drv.collectStats` lines in after heavily modifying `Run`'s contract -- apologies. 

I believe we can simplify this greatly. I kinda wrote out a blurb but I think it's easier to illustrate.

```go
func (drv *DockerDriver) run(ctx context.Context, container string, task drivers.ContainerTask) (drivers.WaitResult, error) {
        mwOut, mwErr := task.Logger()

        waiter, err := drv.docker.AttachToContainerNonBlocking(ctx, docker.AttachToContainerOptions{
                Container: container, OutputStream: mwOut, ErrorStream: mwErr,
                Stream: true, Logs: true, Stdout: true, Stderr: true,
                Stdin: true, InputStream: task.Input()})
        if err != nil && ctx.Err() == nil {
                // ignore if ctx has errored, rewrite status lay below
                return nil, err
        }

        err = drv.startTask(ctx, container)
        if err != nil && ctx.Err() == nil {
                // if there's just a timeout making the docker calls, drv.wait below will rewrite it to timeout
                return nil, err
        }

        ctx, cancel := context.WithCancel(ctx) // to shut down collectStats once container exits
        go drv.collectStats(ctx, container, task)

        return &waitResult{
                container: container,
                waiter:    waiter,
                drv:       drv,
                cancel: cancel,
        }, nil
}

// implements drivers.WaitResult
type waitResult struct {
        container string
        waiter    docker.CloseWaiter
        drv       *DockerDriver
        cancel   func()
}

func (w *waitResult) Wait(ctx context.Context) (drivers.RunResult, error) {
        defer func() {
                w.waiter.Close()
                w.waiter.Wait() // make sure we gather all logs
                w.cancel()
        }()

        status, err := w.drv.wait(ctx, w.container)
        return &runResult{
                status: status,
                err:    err,
        }, nil
}
```

we should remove `Timeout()` from the `drivers.ContainerTask` interface imo, we are not using it really. we are effectively waiting for the task to finish or for the enclosing context to be up, at which point we're nuking the container. we're not handing the timeout off as a parameter to the container at all, the implementation in `run` left around is now just confusing. 

the only addition is that maybe putting `collectStats` below `startTask` is too late for quick [cold, hot is fine] tasks, in which case we can move it back up but in any enclosing context that calls `cookie.Run()` we just need to make sure we cancel the given context once the waiter returns. the only confusing bit is... well, the contract with `Run()` is such that a user _could_ call it, run off and not wait for it, and then `collectStats` would run forever. I think this is why the additional goroutine was added inside of `Run`. `Run`, now, seems to have a weird contract, but in any event, we are always waiting, so I think we can expect that (we could demand it, in the contract, i.e. 'Wait.Wait` must always be called or resources will be leaked; this seems wise to me anyway).

too many words I think, but anyway, lmk what you think / what this may be missing.

api/agent/drivers/docker/docker.go
```diff
@@ -269,55 +269,98 @@ func dockerMsg(derr *docker.Error) string {
...
+// waitResult implements drivers.WaitResult
+// This implementation allows multiple concurrent calls to Wait()
 func (w *waitResult) Wait(ctx context.Context) (drivers.RunResult, error) {
 	defer func() {
+		// Only one caller can call w.waiter.Close() without getting a panic.
+		// This caller will then call w.waiter.Wait() and then close closeSignal to tell the other threads that processing is finished
+		// Other callers will get a panic.
+		// They will recover() here and wait for closeSignal to be closed before returning
+		defer func() {
+			if recover() != nil {
```
rdallman: I'm hoping we can get rid of this as alluded to in previous comment.

api/agent/drivers/docker/docker.go
```diff
@@ -355,7 +402,11 @@ func (drv *DockerDriver) collectStats(ctx context.Context, container string, tas
 			if !ok {
 				return
 			}
-			task.WriteStat(cherryPick(ds))
+			stats := cherryPick(ds)
+			if !stats.Timestamp.IsZero() {
```
rdallman: thanks, good idea

examples/grafana/README.md
```diff
@@ -109,4 +109,27 @@ A second example dashboard `fn_grafana_dashboard2.json` in this example's direct
...
+| ------------- |
+| `fn_docker_stats_fn_cpu_kernel` |
+| `fn_docker_stats_fn_cpu_kernel`  |
+| `fn_docker_stats_fn_cpu_user` |
+| `fn_docker_stats_fn_disk_read` |
+| `fn_docker_stats_fn_disk_write` |
+| `fn_docker_stats_fn_mem_limit` |
+| `fn_docker_stats_fn_mem_usage` |
+| `fn_docker_stats_fn_net_rx` |
+| `fn_docker_stats_fn_net_tx` |
```
rdallman: these seem to have redundant `fn_` bits, perhaps we could remove the latter? `docker_stats_net_rx` seems pretty well scoped, regardless of the `fn_` prefix.

rdallman:
one additional thing to think about is that we will very likely want to swap in call ids to the child span that is reporting these statistics in hot functions. we already have to swap out stdin/stdout/stderr, so we can likely colocate this kind of swapping with that, but it's not immediately clear a very clean way to do this. as it stands, we are going to get a lot of stats about a container, have timestamps of logs of when that call ran in that container, and have to correlate that, but the span itself is not making it apparent which call is the source of any hot function's docker stats.

api/agent/agent.go
```diff
@@ -451,7 +452,9 @@ func (s *coldSlot) exec(ctx context.Context, call *call) error {
 	span, ctx := opentracing.StartSpanFromContext(ctx, "agent_cold_exec")
 	defer span.Finish()
 
-	waiter, err := s.cookie.Run(ctx)
+	ctxForRun, cancel := context.WithCancel(ctx)
+	defer cancel()
```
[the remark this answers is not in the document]
nigeldeakin: General comment: I agree that there is a bit of a proliferation of contexts here. The general principle behind having them is an attempt to retain exact the same behaviour as before, especially regarding context timeouts. Some of the previous behaviour might have been accidental and doesn't need to be preserved, or I may have made errors. In any case I am keen to simplify this if we can.

api/agent/agent.go
```diff
@@ -572,8 +575,15 @@ func (a *agent) prepCold(ctx context.Context, slots chan<- slot, call *call, tok
...
-// TODO add ctx back but be careful to only use for logs/spans
-func (a *agent) runHot(slots chan<- slot, call *call, tok Token) error {
+func (a *agent) runHot(ctxArg context.Context, slots chan<- slot, call *call, tok Token) error {
+	// We must be careful to only use ctxArg for logs/spans
+
+	// create a span from ctxArg but ignore the new Context
+	// instead we will create a new Context below and explicitly set its span
+	span, _ := opentracing.StartSpanFromContext(ctxArg, "docker_run_hot")
+	defer span.Finish()
+	// note won't use the orifinal or returned Context. Ind
```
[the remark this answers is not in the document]
nigeldeakin: Me neither.  I'll remove it.

api/agent/agent.go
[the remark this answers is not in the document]
nigeldeakin: As for `ctxForRun`: Hmm, yes, I can't see why this is needed.

I suspect this is left over from my first attempt at fixing the problem of the `collectStats` goroutine being ended too early. This involved passing a new context,  `ctxForRun`, into `Run` which would be  cancelled by `exec` after it had finished waiting for the container to finish. `Run` would use that to used to signal to the `collectStats` goroutine that it should terminate. As I explained in comments here, I decided that this was a bad design and so I replaced this with a better solution. I think that means that  `ctxForRun` no longer servs any useful purpose, so I will remove it. Good spot.

examples/grafana/README.md
nigeldeakin: The second half of these metric names (e.g. `fn_mem_limit`) is the name of the key by which the value in question is logged to the span. The Prometheus collector uses this to decide which of the values logged with the span to convert to Prometheus metrics (we don't want it turning every random logged value into a metric).

However it's easy to make the Prometheus collector strip this off when it constructs the metric name. I'll do this.

api/agent/drivers/docker/docker.go
nigeldeakin: That certainly looks simpler. Obvious you have some additional information that I didn't have, that two aspects of the existing implementation (which I was preserving in the new version) are not actually needed:
* honouring the timeout set in the `ContainerTask`, and 
* ensuring that the context passed to `AttachToContainerNonBlocking` and `startTask` is cancelled at the end of `Run`. 
I'm happy to take your advice and take the extra contexts out.

As you say above, your suggested new implementation relies on the caller calling `WaitResult`. I was trying to avoid this coupling. However if we're happy with this coupling then I agree we don't need to have `Run() ` spawn a goroutine to call `WaitResult`. This also avoids the need to make  `WaitResult` support multiple calls to `Wait`. 

Passing in the cancel function to `waitResult` is a nice way to avoid depending on the parent context being cancelled.

However I'm unsure whether we should be moving the call to `collectStats` to after `startTask`. As you suggest, this makes it more likely that the container will start and stop before we have had time to collect any stats.  In a few quick experiments I found that with a cold function that  performed a 300ms sleep, I observed one set of stats being received if `collectStats` was called before `startTask`, but not if it was called after it.

However It's not trivial to move it back to its original position before `startTask`. Can you clarify what you had in mind in this case? I think we would need to create a new context (so we could cancel it without cancelling the parent context), but we would need to ensure that it wasn't cancelled until the container had exited. However `go vet` would then complain that it was possible to return without cancelling the new context. 

(I think `vet` is being inconsistent here, since exactly the same issue applies if we put the call to `collectStats` after `startTask` and reuse the `ctx` variable. It was the need to avoid a vet error like this that led me into the more complex implementation I proposed.)

api/agent/drivers/docker/docker.go
```diff
@@ -269,55 +269,98 @@ func dockerMsg(derr *docker.Error) string {
...
-		waiter:    waiter,
-		drv:       drv,
-	}, nil
+	waitResult := newWaitResult(container, waiter, drv)
+
+	go func() {
+		// wait for the container to have stopped and then cancel the collection of statistics
+		waitResult.Wait(ctx0)
+		cancelCtxForStats()
+	}()
```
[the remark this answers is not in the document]
nigeldeakin: Agreed (as noted above)

api/agent/drivers/docker/docker.go
nigeldeakin: Agreed (as noted above)

api/agent/drivers/docker/docker.go
rdallman: > As you say above, your suggested new implementation relies on the caller calling WaitResult. I was trying to avoid this coupling. However if we're happy with this coupling then I agree we don't need to have Run() spawn a goroutine to call WaitResult. This also avoids the need to make WaitResult support multiple calls to Wait.

this seems a reasonable contract to me, +1. it seems like we already demand `Wait` be called, we just need to add that it may only be called once to the contract there. 

> However I'm unsure whether we should be moving the call to collectStats to after startTask. As you suggest, this makes it more likely that the container will start and stop before we have had time to collect any stats. In a few quick experiments I found that with a cold function that performed a 300ms sleep, I observed one set of stats being received if collectStats was called before startTask, but not if it was called after it.

would forcing collectStats to collect >= 1 data point fix this? I am unsure of plumbing, obviously we don't want to have to wait on this just to return the call result to the user. We are already suffering in that many functions run too fast to collect metrics (few ms), so this seems something we should fix anyway. 

> It was the need to avoid a vet error like this that led me into the more complex implementation I proposed.

if vet is going to be a nazi we can turn it off, it's nice that it enforces good practices robotically vs us having to explain to contributors some subjectively better way of doing things but meh, we can go back to the old fashioned way if it's a hindrance.

examples/grafana/README.md
rdallman: seems reasonable. 

> we don't want it turning every random logged value into a metric

actually, i think this is okay / desirable. it's just us logging metric events I think, would be nice to have. we have stderr logging for all else.

nigeldeakin:
The latest commits incorporate all the simplifications suggested, but keep the call to `collectStats` *before* the call to `startTask`, for the reason mentioned above. 

This means the tests fail in `go vet` as expected.

I'm wondering whether we should try to exclude this failure. I don't think this is a context leak, as the context will be cancelled when the parent context is cancelled. I think we need to create a new context here because it needs to be cancelled (in this function) when the `WaitResult` discovers that the container has stopped. 

I considered simply using the parent context, but I am concerned that cancelling it might have unexpected effects on any other code in the parent that uses this context. In any case, it  would be a misuse of the parent context, as we're not trying to cancel what it is doing. We're just trying to cancel something that *this* function is doing.

As for delaying the shutdown of the container until at least one set of statistics have been obtained: that sounds a potentially significant change. We should discuss with @treeder and @carimura

rdallman:
> As for delaying the shutdown of the container until at least one set of statistics have been obtained: that sounds a potentially significant change. We should discuss with @treeder and @carimura

we can just do things too ;) i'm actually saying the contrary, have no interest in waiting on stats collection to hold up returning the call; am asking whether the plumbing would require it. collectStats itself is asking for a stream of stats from docker for a given container, and it seems to me that it immediately tries to get one, and then uses the interval after that, however I'm not sure this is the case. at least, it's something to think about, because for many calls as it stands it seems like they will have 0 container metrics on them.

rdallman:
> I'm wondering whether we should try to exclude this failure. I don't think this is a context leak, as the context will be cancelled when the parent context is cancelled. I think we need to create a new context here because it needs to be cancelled (in this function) when the WaitResult discovers that the container has stopped.

hmm. yea, it's tricky. as you say, if we don't have a child context then we have to rely on the parent context getting canceled. we're calling `Run` in 2 spots, unfortunately, too.

there are alternatives, we can simply use a `chan struct{}` that we close in `Wait` to shut down the stats collection, we should additionally listen for the parent context. illustration:

```go
func collectStats(ctx context.Context, done <-chan struct{}, ...) {
for {
    select {
    case <-ctx.Done():
      return
    case <-done:
      return
    default:
    }
  
    // normal docker stats code
  }
}

func (w waitResult) Wait() {
  defer func() {
    // ...
    close(w.done) // shut down stats stream
  }
  // ...
}
```

i can copy in details if this is unclear (it's some go trickery for sure)

nigeldeakin:
I've updated `docker.Run` to use a channel rather than a context to terminate the `collectStats` goroutine, exactly as suggested.

PR #493, opened 2017-11-11
fn: fnlb: enhancements and new grouper tests

skinowski:
*) added healthy threshold (default: 1)
*) grouper now logs when servers switch between healthy/unhealthy state
*) moved DB code out of grouper
*) run health check immediately at start (don't wait until hcInterval)
*) optional shutdown timeout (default: 0) & mgmt port (default: 8081)
*) hot path List() in grouper now uses atomic ptr Load
*) consistent router: moved closure to a new function
*) bugfix: grouper is now using configured hcEndpoint for version checks
*) bugfix: version parsing from fn servers should not panic fnlb
*) bugfix: when servers removed from DB, stayed in healthy list
*) bugfix: if DB is down, health checker should not stop monitoring
*) basic new tests for grouper (add/rm/unhealthy/healthy) server

### DIFF
--- a/fnlb/lb/allgrouper.go
+++ b/fnlb/lb/allgrouper.go
@@ -8,7 +8,6 @@ import (
 	"net/http"
 	"sort"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"fmt"
@@ -22,8 +21,9 @@ import (
 // pass/fail and exposes endpoints for adding, removing and listing nodes.
 func NewAllGrouper(conf Config, db DBStore) (Grouper, error) {
 	a := &allGrouper{
-		nodeList: make(map[string]*nodeState),
-		db:       db,
+		nodeList:        make(map[string]nodeState),
+		nodeHealthyList: make([]string, 0),
+		db:              db,
 
 		// XXX (reed): need to be reconfigurable at some point
 		hcInterval:    time.Duration(conf.HealthcheckInterval) * time.Second,
@@ -36,9 +36,6 @@ func NewAllGrouper(conf Config, db DBStore) (Grouper, error) {
 		// for health checks
 		httpClient: &http.Client{Transport: conf.Transport},
 	}
-
-	empty_list := make([]string, 0)
-	a.healthyList.Store(&empty_list)
 
 	for _, n := range conf.Nodes {
 		err := a.add(n)
@@ -53,14 +50,6 @@ func NewAllGrouper(conf Config, db DBStore) (Grouper, error) {
 
 // nodeState is used to store success/fail counts and other health related data.
 type nodeState struct {
-	// name/address of the node
-	name string
-
-	// used to purge/delete old nodes when synching the nodes with DB
-	intervalID uint64
-
-	// exclusion for multiple go routine pingers
-	lock sync.Mutex
 
 	// num of consecutive successes & failures
 	success uint64
@@ -82,11 +71,9 @@ type nodeState struct {
 type allGrouper struct {
 
 	// health checker state and lock
-	nodeLock sync.Mutex
-	nodeList map[string]*nodeState
-
-	// current atomic store/load protected healthy list
-	healthyList atomic.Value
+	nodeLock        sync.RWMutex
+	nodeList        map[string]nodeState
+	nodeHealthyList []string
 
 	db DBStore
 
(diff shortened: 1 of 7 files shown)


skinowski:
Lots of things got piled up in this PR and it is long and hard to read. Please let me know if I should split these up.

fnlb/main.go
```diff
@@ -64,27 +72,57 @@ func main() {
...
+	logrus.WithFields(logrus.Fields{"signal": sig}).Info("received signal")
+
+	for i := 0; i < len(servers); i++ {
+
+		ctx := context.Background()
+
+		if conf.ShutdownTimeout > 0 {
+			tmpCtx, cancel := context.WithTimeout(context.Background(), time.Duration(conf.ShutdownTimeout)*time.Second)
+			ctx = tmpCtx
+			defer cancel()
```
rdallman: super nit: `ctx, _ = context.WithTimeout(ctx, ...)` and removing var assign / cancel seems fine here since we're shutting down, not too worried about context leaking.

fnlb/lb/allgrouper.go
```diff
@@ -415,20 +370,26 @@ func (a *allGrouper) removeNode(w http.ResponseWriter, r *http.Request) {
...
+	for _, value := range a.nodeList {
+		run_list = append(run_list, value)
+	}
+	a.nodeLock.Unlock()
+
+	out := make(map[string]string, len(run_list))
+	for _, node := range run_list {
+
+		node.lock.Lock()
+		healthy := node.healthy
```
rdallman: would be nice to do this in the previous `nodeList` iteration if we get rid of granular locks.

skinowski:
single rw lock sounds good, we can always pick & optimize later if needed.

fnlb/main.go
skinowski: "go vet" complains about leaked context... I'll retain 'defer cancel()' if that's ok.

fnlb/lb/allgrouper.go
```diff
@@ -117,69 +104,87 @@ func (a *allGrouper) remove(ded string) error {
...
+	ret := make([]string, len(a.nodeHealthyList))
+	copy(ret, a.nodeHealthyList)
+	a.nodeLock.RUnlock()
 
-	ret := make([]string, len(*ptr))
-	copy(ret, *ptr)
-	return ret, nil
+	var err error
+	if len(ret) == 0 {
+		err = ErrNoNodes
```
rdallman: missed this the first go round -- I think returning an empty list here without an error is fine, the API uses this endpoint and it's not an error really, it's just an empty set.

fnlb/lb/allgrouper.go
```diff
@@ -117,69 +104,87 @@ func (a *allGrouper) remove(ded string) error {
...
-				a.nodeList[node].intervalID = *iteration
-			} else {
+			if !ok {
 				// add new node
-				a.nodeList[node] = &nodeState{
-					name:       node,
-					intervalID: *iteration,
-					healthy:    true,
+				a.nodeList[node] = nodeState{
+					healthy: true,
```
rdallman: we should probably start nodes in an unhealthy state until they pass a health check? prevent sending to a bad node if e.g. the wrong address got put in for a node.

rdallman: ah, i guess implicitly we're not adding to the healthy list which we have separate, so this is the case in practice even if this bit is flipped to true.

rdallman: well, after another think, it's confusing I guess. one of the pingers which changes state could end up updating the list of healthy nodes from publishHealth and include this node even before it was pinged, since this bit is true -- so it seems like we should flip it off until a check passes.

skinowski: So, this is actually something I wanted to ask, add() operation runs 1 ping, then adds it to DB. I think we can reevaluate these and how we should handle, eg. case of fnlb crash/restart, case where healthy threshold is configured higher than 1 with long interval (potentially can take minutes to become healthy), perhaps default healthy/unhealthy preference can be in DB? etc... or we could choose to always gate servers for one run of ping for all these cases to ensure quick start up time... I don't know. :-)

skinowski: do you think it's reasonable to always guard initial state with 1 ping? That gives us reasonable restart fnlb time and guards against routing requests to faulty servers. And if fnlb is configured with eg. healthy threshold of 2 and check interval of 5 secs, it nicely guards frequent state flip/flops as configured, but also ensures fast restart/start.

rdallman: that's a good point about restart. it seems like we should be able to handle these cases separately with some maneuvering (concretely, somewhat as pointed out): 

* on startup we can just load in the entire list and do one health check immediately and go? ignore healthy count / interval.
* when a user detects in `runHealthcheck` that a new node has come in, go through the normal health checks to assert health.

i have been trying to consciously avoid adding anything more than the node list to the db, if possible, but not for any good reason beyond having to manage shared state between multiple processes (not a bad reason) as well as handling partitioning. one issue with requiring a ping is that on startup, the backing list we use for consistent hashing will be modified basically N times and could lead to a small amount of thrashing -- we could add a tiny buffer in recovery time before we allow serving traffic to mitigate this. 

in any case, I do think it's a good idea to guard with 1 ping. even if we manage 'health' in a db, we could have broken links between some nodes that other load balancers may not have. I agree, I think we'll have pretty reasonable restart times doing that and we can do the proper health checking from operating state.

skinowski: sending a PR soon. (I meant: not health state in DB, but the config option of default health state for new node in DB, as config option.)

skinowski: https://github.com/fnproject/fn/pull/508

PR #502, opened 2017-11-15
Removed a bunch of old examples.

treeder:


carimura:
nice. looks good.

nigeldeakin:
What's your intention in renaming  `fn/examples/grafana` to `fn/examples/UPDATE-OR-DELETE/grafana`? I added this recently, and have a PR open to update it.

treeder:
No intention, thought it was old. I'll move it back. Although that looks like something that would be better in the docs, rather than examples. Examples are for example functions.

treeder:
Moved it back @nigeldeakin

PR #508, opened 2017-11-16
fn: fnlb: default health state for new nodes

skinowski:
*) Any new node now by default is in unknown state.
*) One health check is required for unknown state to move in
   to healthy/unhealthy states, then actual interval and
   thresholds apply.
*) add() no longer runs health check as this is now handled
   with the new logic.

This means during restarts fnlb will run one health check
immediately for nodes to switch to healthy/unhealthy state
to ensure speedy start, but guard against routing traffic to
unhealthy servers. After this initial state, nodes are
subjected to regular interval and thresholds.

### DIFF
--- a/fnlb/lb/allgrouper.go
+++ b/fnlb/lb/allgrouper.go
@@ -48,10 +48,12 @@ func NewAllGrouper(conf Config, db DBStore) (Grouper, error) {
 	return a, nil
 }
 
+type healthState uint8
+
 const (
-	UNKNOWN   = iota
-	HEALTHY   = iota
-	UNHEALTHY = iota
+	StateUnknown = iota
+	StateHealthy
+	StateUnhealthy
 )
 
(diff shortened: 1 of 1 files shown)


fnlb/lb/allgrouper.go
```diff
@@ -48,6 +48,12 @@ func NewAllGrouper(conf Config, db DBStore) (Grouper, error) {
 	return a, nil
 }
 
+const (
+	UNKNOWN   = iota
+	HEALTHY   = iota
+	UNHEALTHY = iota
```
rdallman: go nit, would make a type so it's obvious what this is (vs an int) and also consts in go are generally camel cased (as unfortunate as it may be, it is the idiom)

```go
type state uint8

const (
  StateUnknown state = iota
  StateHealthy
  StateUnhealthy
)
```

PR #510, opened 2017-11-17
fn: prometheus collector concurrent map access

skinowski:
*) Added mutex to guard against concurrent access to maps

### DIFF
--- a/api/server/prom_zip_collector.go
+++ b/api/server/prom_zip_collector.go
@@ -6,13 +6,14 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 )
 
 // PrometheusCollector is a custom Collector
 // which sends ZipKin traces to Prometheus
 type PrometheusCollector struct {
-
+	lock sync.Mutex
 	// Each span name is published as a separate Histogram metric
 	// Using metric names of the form fn_span_<span-name>_duration_seconds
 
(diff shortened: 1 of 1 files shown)


skinowski:
is the lock too big here?

rdallman:
thanks. go has a new concurrent map I'm keen to try out that may work here, as well: https://golang.org/pkg/sync/#Map -- I've yet to use it, there's some type casting involved but is an option at least. I am not really sure if we need to protect the nodes inside of the map as well as the map itself, in which case holding the lock over the whole thing seems like something we might have to do. gonna defer to @nigeldeakin

nigeldeakin:
Thanks for picking this up. I too have  seen occasional concurrency failures in this map.

The change to using pointer receivers looks good to me. (Go newbie here)

I'm pretty sure the collector is called asynchronously to `span.Finish()` so performance isn't absolutely critical. 

However I think we only need to lock `getHistogramVec`, since this is the only place that accesses `registeredLabelKeysMap` and `histogramVecMap` (and missing keys are lazily added within the same call). 

We only need to claim the lock before the lookup:
```
	histogramVec, found := pc.histogramVecMap[metricName]
``` 
and release it after we have added the new value
```
		pc.histogramVecMap[metricName] = histogramVec
```
and at the top of the `else` block.

If we changed this to use a concurrent map we would still need to hold a lock as described above, so I would stick with a lock.

skinowski:
Thanks @nigeldeakin , moved the mutex into getHistogramVec() now.

PR #511, opened 2017-11-17
S3 loggyloo

### LINKED ISSUE #473: S3 API compatible log storage implementation
BMC has one of these we can use, but we can use any S3 library we want really (probably not aws's), we can just swap out the url and the arn rules may be funky.

### LINKED ISSUE #481: remove DELETE /v1/apps/:app/calls/:call/log
If we expose the ability for users to delete logs then it makes auditing relatively challenging in cases where some user actually used it (without us getting them to admit to it!). I could see the value in e.g. some sensitive data slipped into the log and I want to get rid of it, fwiw we don't have a call delete API and there should likely be some symmetry there. if we leave this then we could have cases where... a call happened, it had a log, the user deleted the log, so there's a call record laying around that indicates it should probably have a log (and definitely the call ran), and it kind of looks like a bug to the user if they don't remember deleting it or they're using fn with peers and don't audit every action.

I think if we have a cleaner for old logs and calls baked into fn then we don't really need to expose this at the api level -- users can just turn on the cleaner with some duration (e.g. `--cleaner 7d`). we should likely default to retain forever for the OSS one, just because people might be working on it over the course of a few weeks[/months/years] on their laptop, but different thread. it's also possible to clean out from the db / log store directly and is at least an option, albeit a bad one.

not really challenging to do, just need consensus.

rdallman:
minio has an s3 compatible api and is an open source project and, notably, is
not amazon, so it seems best to use their client (fwiw the aws-sdk-go is a
giant hair ball of things we don't need, too). it was pretty easy and seems
to work, so rolling with it. also, minio is a totally feasible option for fn
installs in prod / for demos / for local.

* adds 's3' package for s3 compatible log storage api, for use with storing
logs from calls and retrieving them.
* removes DELETE /v1/apps/:app/calls/:call/log endpoint
* removes internal log deletion api
* changes the GetLog API to use an io.Reader, which is a backwards step atm
due to the json api for logs, I have another branch lined up to make a plain
text log API and this will be much more efficient (also want to gzip)
* hooked up minio to the test suite and fixed up the test suite
* add how to run minio docs and point fn at it docs
* updates swagger
* updates deps & adds minio lib

some notes: notably we aren't cleaning up these logs. there is a ticket
already to make a Mr. Clean who wakes up periodically and nukes old stuff, so
am punting any api design around some kind of TTL deletion of logs. there are
a lot of options really for Mr. Clean, we can notably defer to him when apps
are deleted, too, so that app deletion is fast and then Mr. Clean will just
clean them up later (seems like a good option).

have not tested against BMC object store, which has an s3 compatible API. but
in theory it 'just works' (the reason for doing this). in any event, that's
part of the service land to figure out.

closes #481
closes #473


api/logs/s3/s3.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (s *store) InsertLog(ctx context.Context, appName, callID string, callLog io.Reader) error {
+	objectName := path(appName, callID)
+	_, err := s.client.PutObjectWithContext(ctx, s.bucket, objectName, callLog, -1, minio.PutObjectOptions{ContentType: contentType})
+	return err
+}
+
+func (s *store) GetLog(ctx context.Context, appName, callID string) (io.Reader, error) {
+	objectName := path(appName, callID)
+	return s.client.GetObjectWithContext(ctx, s.bucket, objectName, minio.GetObjectOptions{})
```
denismakogon: Shouldn't we really check error type here?

api/logs/s3/s3_test.go
```diff
@@ -0,0 +1,28 @@
...
+	"os"
+	"testing"
+
+	logTesting "github.com/fnproject/fn/api/logs/testing"
+)
+
+func TestS3(t *testing.T) {
+	minio := os.Getenv("MINIO_URL")
+	if minio == "" {
+		t.Skip()
```
denismakogon: worth adding a reason for skipping tests

api/server/call_logs.go
```diff
@@ -12,35 +14,25 @@ func (s *Server) handleCallLogGet(c *gin.Context) {
...
-func (s *Server) handleCallLogDelete(c *gin.Context) {
-	ctx := c.Request.Context()
-
-	appName := c.MustGet(api.AppName).(string)
-	callID := c.Param(api.Call)
-	_, err := s.Datastore.GetCall(ctx, appName, callID)
-	if err != nil {
-		handleErrorResponse(c, err)
-		return
+	// TODO this API needs to change to text/plain / gzip anyway, punting
```
denismakogon: Isn't this it the right time to do so?

rdallman: as stated in parent, I have another WIP patch working on this. but having s3 api is blocking for the service piece and this isn't really critical.

api/logs/s3/s3_test.go
rdallman: looks pretty obvious to me, this is set up in the test script

api/logs/s3/s3.go
rdallman: yea, should probably add `LogNotFound`, thanks.

rdallman:
@denismakogon updated per your review, thanks

rdallman:
thanks!

PR #512, opened 2017-11-17
fn: move memory/token code into resource

skinowski:
*) bugfix: fix nil ptr access in docker registry RoundTrip
*) move async and ram token related code into resource.go


skinowski:
no functionality change (move code) (except for docker registry null ptr fix)

PR #515, opened 2017-11-17
Update CONTRIBUTING.md

nigeldeakin:
Add a note to tell people to use go 1.9.1 or newer.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,6 +52,8 @@ asking for suggestions on how to address the documentation part.
 
 ### Build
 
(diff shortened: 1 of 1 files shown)


nigeldeakin:
If you try to use an older version,  you will get the following error when trying to build `Fn`:
```
github.com/fnproject/fn/vendor/github.com/rdallman/migrate/database/mysql: 
no buildable Go source files in
<whatever>/go/src/github.com/fnproject/fn/vendor/github.com/rdallman/migrate/database/mysql
```

PR #518, opened 2017-11-17
fn: introducing 503 responses for out of capacity case

### LINKED ISSUE #509: 503 with retry-after header if timeout occurs during scheduling
As opposed to generic 504 during token wait in agent, if context timeout occurs during token wait, then we can safely retry this request. In this case, we can return a 503 with retry-after header. (What should be the value of the retry-header in these cases?)

skinowski:
*) Adding 503 with Retry-After header case if request failed
during waiting for slots.
*) Increased MaxMemory for routes to 100TB.
*) TODO: return 503 without Retry-After if the request can
never be met by this fn server.

closes #509

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -99,9 +99,8 @@ type Agent interface {
 	// immediately before and after the Call is executed, respectively. An error
 	// will be returned if there is an issue executing the call or the error
 	// may be from the call's execution itself (if, say, the container dies,
-	// or the call times out). submitState returned will show the last execution flow
-	// progress Submit() call was able to reach.
-	Submit(Call) (error, submitState)
+	// or the call times out).
+	Submit(Call) error
 
 	// Close will wait for any outstanding calls to complete and then exit.
 	// Close is not safe to be called from multiple threads.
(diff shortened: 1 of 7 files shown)


api/agent/agent.go
```diff
@@ -99,8 +99,9 @@ type Agent interface {
 	// immediately before and after the Call is executed, respectively. An error
 	// will be returned if there is an issue executing the call or the error
 	// may be from the call's execution itself (if, say, the container dies,
-	// or the call times out).
-	Submit(Call) error
+	// or the call times out). submitState returned will show the last execution flow
+	// progress Submit() call was able to reach.
+	Submit(Call) (error, submitState)
```
rdallman: could we encode this in the `error` instead? an error type of `submitState` should do. e.g.

```
if s, ok := err.(submitState); ok && s.State() == SubmitStateInitial {
 // ...
}
```

there are a few ways to skin the cat.

rdallman: there is really only one state we are interested in, which is `SubmitStatesWaitSlot`. the rest seem like added complexity. in this case, we should probably just return some kind of `WaitError` type that has a real error on it we can extract, and then we can know that we hit this case by getting a `WaitError` type. then we only have 1 state to handle. beware that embedding errors has the side effect of losing hidden interface implementations (sadness), sorry I'm being intentionally vague to see if you can come up with something better than the rest of us gophers to date >.<

skinowski: no worries, I'll do another pass on this.

PR #523, opened 2017-11-20
Fix for Prometheus/Grafana quickstart guide

### LINKED ISSUE #520: prom/grafana example not working for me
i am on a mac, might be the difference....

```
docker run --name=prometheus -d -p 9090:9090 \
  --mount type=bind,source=`pwd`/prometheus.yml,target=/etc/prometheus/prometheus.yml \
  --add-host="localhost:`route | grep default | awk '{print $2}'`" prom/prometheus
usage: route [-dnqtv] command [[modifiers] args]
invalid argument "localhost:" for "--add-host" flag: invalid IP address in add-host: ""
See 'docker run --help'.
```

nigeldeakin:
This PR Fixes #520. This changes the way that we tell Prometheus (running in a container) which host the Fn server is running on, and how we tell Grafana (also running in a container) which host Prometheus is running on. 

The fix is to replace `docker --addHost` with `docker --link`. This removes the need for a complicated shell expression that, it appears, doesn't work on MacOS (but worked just fine when I tested it on Ubuntu...)

### DIFF
--- a/examples/grafana/README.md
+++ b/examples/grafana/README.md
@@ -6,10 +6,13 @@ The Fn server exports metrics using [Prometheus](https://prometheus.io/). This a
 
 ## Start an Fn server and deploy some functions
 
-This example requires an Fn server to be running and that you have deployed one or more functions. 
-See the [front page](/README.md) or any of the other examples for instructions. 
+Start an fn server
 
-The steps below assume that the Fn server is running at `localhost:8080`.
+```sh
+fn start
+```
+Deploy one or more functions as required.
+See the [front page](/README.md) or any of the other examples for instructions. 
 
 ## Examine the endpoint used to export metrics to Prometheus
 
@@ -22,7 +25,7 @@ This will display the metrics in prometheus format.
 
 Open a terminal window and navigate to the directory containing this example.
 
-Examine the provised Prometheus configuration file:
+Examine the provided Prometheus configuration file:
 
 ```
 cat prometheus.yml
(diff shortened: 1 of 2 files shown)


examples/grafana/README.md
```diff
@@ -57,11 +60,11 @@ If you are running a cluster of Fn servers then you can specify them all here.
 
 Now start Prometheus, specifying this config file:
 ```
-docker run --name=prometheus -d -p 9090:9090 \
-  --mount type=bind,source=`pwd`/prometheus.yml,target=/etc/prometheus/prometheus.yml \
-  --add-host="localhost:`route | grep default | awk '{print $2}'`" prom/prometheus
+  docker run --name=prometheus -d -p 9090:9090 \
+    --mount type=bind,source=`pwd`/prometheus.yml,target=/etc/prometheus/prometheus.yml \
```
denismakogon: can't `--mount` can be replaced with volume attached to `/etc/prometheus`:

```
-v `pwd`:/etc/prometheus
```

carimura: also in prometheus.yml, need to use docker.for.mac.localhost instead of localhost.

```
    static_configs:
      # Specify all the fn servers from which metrics will be scraped
      - targets: ['docker.for.mac.localhost:8080'] # Uses /metrics by default
```

denismakogon: Better to use docket dns names instead. The thing you suggest is very dind-specific. We can rely on dockets flat network.

carimura: @denismakogon ok how? it didn't work out of the box for me, which means it probably won't work for others out of the box.

nigeldeakin: Since the instructions have only been discovered to not work on Mac this PR will need to be completed by someone with the required device.

nigeldeakin:
I've tested the latest changes on Ubuntu and it looks good to me. Just need @denismakogon to approve.

Issue #19, opened 2017-07-27
Simple per function monitoring/metrics


rdallman:
In gitlab by @treeder on May 11, 2017, 17:25

- CPU
- memory

rdallman:
we're gonna do this, dk why it got closed

denismakogon:
@rdallman Probably better include CPU, RAM into `/calls` response?

rdallman:
yea, not clear from description but I agree, they're also getting spurted to prometheus now too.

Issue #524, opened 2017-11-20
Short lived functions have zeroed docker container metrics


denismakogon:
Assume you have hello-world hot function. Run it in a loop for 10 to 100 times. See spans in Prometheus or log container stats manually.

Short-lived call ends (sends `context.Done`) before docker was able to ship container stats, this causes metrics to be zeroed.

nigeldeakin:
If the container is very short-lived then it is possible that it will exit before there has been time to obtain any statistics. That's a particular issue with cold functions (I would have though it was no an issue with hot functions since by their nature they will continue running for longer)

We need to decide whether we should prevent a short-lived container from terminating until at least one set of stats has been obtained. Currently we don't, on the basis that stats collection should not be intrusive in this way. But we could decide otherwise...

rdallman:
we could not stream them, via https://godoc.org/github.com/fsouza/go-dockerclient#StatsOptions and manually set the interval and attempt to immediately get one data point and do our own polling of stats. fwiw trying this in the past the overhead of this docker request from multiple clients (30, when tested) completely choked docker, but it has been > 1 year since last tested and possibly changed. streaming made the load bearable (we use this now).

Issue #522, opened 2017-11-20
Hot path polls for container stats between calls


denismakogon:
Docker stats are being polled during whole container lifecycle. But Fn should only care about max memory consumption during the specific call. Unfortunately, it's not enough to ask for max memory consumption from Docker, we should collect all stats within call and then find maximum among them.

nigeldeakin:
Currently docker stats relate to the running container, for as long as it is running. 

Adding a new set of stats which are specific to each individual invocation of a hot function sounds like a good idea. So I'd say the current behaviour is by design, but you have raised a good feature request.

rdallman:
these metrics could prove useful and the overhead of opening the docker stats stream is relatively high to the point of not being worth it. it's really easy to ignore the data points in between... closing

PR #525, opened 2017-11-20
Call metrics

denismakogon:
Adding RAM/CPU metrics to call object

This PR splits call object creation at the backend in three steps:
1. Fn creates call object with `running` status when agent calls `call.Start`
2. Fn updates call object when status and completed_at available when agent calls `call.End`
3. Docker driver writes updates call object with CPU and RAM usage.

The idea of this PR to provide basic call metrics in addition to those presented such as status, created|started|completed_at timestamps.

Workaround for #522:
Write call metrics once then unset call ID to prevent metrics being overridden.

Known issue: #524 short-lived calls would not get their portion of stats due to races between docker stats and call context completion.

Closes: #19

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -475,6 +475,9 @@ func (a *agent) prepCold(ctx context.Context, slots chan<- slot, call *call, tok
 		stdin:   call.req.Body,
 		stdout:  call.w,
 		stderr:  call.stderr,
+
+		ds:      a.ds,
+		appName: call.AppName,
 	}
 
 	// pull & create container before we return a slot, so as to be friendly
@@ -541,6 +544,9 @@ func (a *agent) runHot(ctxArg context.Context, slots chan<- slot, call *call, to
 		stdin:  stdinRead,
 		stdout: stdoutWrite,
 		stderr: &ghostWriter{inner: stderr},
+
+		ds:      a.ds,
+		appName: call.AppName,
 	}
 
 	logger := logrus.WithFields(logrus.Fields{"id": container.id, "app": call.AppName, "route": call.Path, "image": call.Image, "memory": call.Memory, "format": call.Format, "idle_timeout": call.IdleTimeout})
@@ -621,6 +627,9 @@ type container struct {
 	stdin  io.Reader
 	stdout io.Writer
 	stderr io.Writer
+
+	ds      models.Datastore
+	appName string
 }
 
 func (c *container) swap(stderr io.Writer) {
@@ -652,6 +661,14 @@ func (c *container) WriteStat(ctx context.Context, stat drivers.Stat) {
 	for key, value := range stat.Metrics {
 		span.LogFields(log.Uint64("fn_"+key, value))
 	}
+	callStat := models.CallStat{
+		CallID:      c.Id(),
+		CPUUsage:    stat.Metrics["cpu_user"],
+		MemoryUsage: stat.Metrics["mem_usage"],
+		AppName:     c.appName,
+	}
+
+	_ = c.ds.InsertCallStat(ctx, callStat)
 }
 
 //func (c *container) DockerAuth() (docker.AuthConfiguration, error) {
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -416,7 +416,7 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	common.Logger(ctx).WithField("container_id", s.container.id).Info("starting call")
 
 	// swap in the new stderr logger
-	s.container.swap(call.stderr)
+	s.container.swap(call.ID, call.stderr)
 
 	errApp := make(chan error, 1)
 	go func() {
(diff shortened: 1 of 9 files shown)


denismakogon:
TODOs:

- [ ] swagger doc update
- [ ] sql migrations

api/agent/agent.go
```diff
@@ -652,6 +668,20 @@ func (c *container) WriteStat(ctx context.Context, stat drivers.Stat) {
 	for key, value := range stat.Metrics {
 		span.LogFields(log.Uint64("fn_"+key, value))
 	}
+	// Fn polls for stats as long as container lives,
+	// so we have to write stats only once
+	if c.callID != "" {
+		err := c.ds.UpdateCallMetrics(ctx,
```
rdallman: I think this present layering issues here, as well as being untenable from a db perspective -- this is too many writes to the db. we should aggregate stats inside of `WriteStat` and then write them to the db once inside of `call.End` -- so aside from moving this function and changing it's contract to write a list of stats, `WriteStat` will need to accumulate data points somewhere.

api/agent/agent.go
```diff
@@ -541,6 +546,11 @@ func (a *agent) runHot(ctxArg context.Context, slots chan<- slot, call *call, to
 		stdin:  stdinRead,
 		stdout: stdoutWrite,
 		stderr: &ghostWriter{inner: stderr},
+
+		ds:      a.ds,
+		appName: call.AppName,
+		path:    call.Path,
+		callID:  call.ID,
```
rdallman: none of these should be on the container, really. the context has a logger with these values, but adding a note to be thorough, see comment below about contract for writing stats.

api/agent/call.go
```diff
@@ -322,6 +322,10 @@ func (c *call) Start(ctx context.Context, t callTrigger) error {
 		return fmt.Errorf("BeforeCall: %v", err)
 	}
 
+	if errInsert := c.ds.InsertCall(ctx, c.Model()); errInsert != nil {
+		common.Logger(ctx).WithError(err).Error("error inserting call during call.Start")
+	}
```
rdallman: to date, we've intentionally not done this because it adds latency for hot functions. our docs explicitly state we don't have any writes to db in the hot path for functions, and it's true that a user can get this behavior without this patch [if they use hot and reply in their container correctly] -- we are adding the call afterwards, but this can be after a client gets a response if done properly [by the user]. 

anyway, I think for async we should do this, but not sync. this patch does unconditionally

api/agent/agent.go
```diff
@@ -652,6 +668,20 @@ func (c *container) WriteStat(ctx context.Context, stat drivers.Stat) {
 	for key, value := range stat.Metrics {
 		span.LogFields(log.Uint64("fn_"+key, value))
 	}
+	// Fn polls for stats as long as container lives,
+	// so we have to write stats only once
+	if c.callID != "" {
+		err := c.ds.UpdateCallMetrics(ctx,
+			c.appName, c.callID,
+			stat.Metrics["cpu_user"],
+			uint64(stat.Metrics["mem_max_usage"]/(1024*1024)),
```
rdallman: would like to store all stats from `cherryPick`, aside from the brittleness of doing this. imo we should store json blob in db or something.

api/datastore/internal/datastoreutil/metrics.go
```diff
@@ -83,6 +84,18 @@ func (m *metricds) InsertCall(ctx context.Context, call *models.Call) error {
...
+func (m *metricds) UpdateCallStatus(ctx context.Context, appName, callID, status string, completedAt strfmt.DateTime) error {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "ds_update_call_status")
+	defer span.Finish()
+	return m.ds.UpdateCallStatus(ctx, appName, callID, status, completedAt)
+}
+
+func (m *metricds) UpdateCallMetrics(ctx context.Context, appName, callID string, CPUUsage, MemoryUsage uint64) error {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "ds_update_call_metrics")
+	defer span.Finish()
+	return m.ds.UpdateCallMetrics(ctx, appName, callID, CPUUsage, MemoryUsage)
```
rdallman: can we just have a `map[string]uint64` to store? this api will require frequent breaking for adding any stats.

api/datastore/internal/datastoreutil/metrics.go
```diff
@@ -83,6 +84,18 @@ func (m *metricds) InsertCall(ctx context.Context, call *models.Call) error {
 	return m.ds.InsertCall(ctx, call)
 }
 
+func (m *metricds) UpdateCallStatus(ctx context.Context, appName, callID, status string, completedAt strfmt.DateTime) error {
```
rdallman: this is unnecessary and should be done in `InsertCall`

api/agent/agent.go
rdallman: this isn't correctly computing max. e.g. if Java does GC and then stats are collected, this value is lower than the previous -- we should store the histogram and later do processing, it's way better than restricting at the base.

PR #526, opened 2017-11-20
Docker compose file for simpler dev env

denismakogon:
TODO: grafana + prometheus

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,15 +62,6 @@ fn start
 
 This will start Fn in single server mode, using an embedded database and message queue. You can find all the
 configuration options [here](docs/operating/options.md). If you are on Windows, check [here](docs/operating/windows.md).
-
-### Quickstart with docker-compose
-
-With help of [compose file](./docker-compose.yml) it's easy to bootstrap whole development stack that includes:
-
- - instance of MySQL server for database
- - instance of Redis server for message queue
- - instance of Fn server with scale group (supported only for Docker Swarm)
- - instance of Fn UI
  
 
 ### Your First Function
--- a/README.md
+++ b/README.md
@@ -63,6 +63,16 @@ fn start
 This will start Fn in single server mode, using an embedded database and message queue. You can find all the
 configuration options [here](docs/operating/options.md). If you are on Windows, check [here](docs/operating/windows.md).
 
+### Quickstart with docker-compose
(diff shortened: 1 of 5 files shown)


denismakogon:
@derekschultz can you review this one?

README.md
```diff
@@ -63,6 +63,16 @@ fn start
 This will start Fn in single server mode, using an embedded database and message queue. You can find all the
 configuration options [here](docs/operating/options.md). If you are on Windows, check [here](docs/operating/windows.md).
 
+### Quickstart with docker-compose
```
derekschultz: Not sure if we should put this in the main readme. From my perspective, it complicates the quickstart, which aims to be as minimalistic as possible. Perhaps it would be better to break it out into docs/operating? Optionally, we could also add a note to the main readme that links to docs/operating, so people know where to find the compose file and helm chart.

README.md
```diff
@@ -63,6 +63,16 @@ fn start
...
+
+With help of [compose file](./docker-compose.yml) it's easy to bootstrap whole development stack that includes:
```
derekschultz: suggestion on language: With Docker Compose it's easy to bootstrap the full Fn development stack that includes the following components:

README.md
```diff
@@ -63,6 +63,16 @@ fn start
...
+
+ - instance of MySQL server for database
```
derekschultz: i think we can remove "instance of" for each of these components

README.md
```diff
@@ -63,6 +63,16 @@ fn start
...
+ - instance of Redis server for message queue
+ - instance of Fn server with scale group (supported only for Docker Swarm)
+ - instance of Fn UI
+
```
derekschultz: We're missing instructions on what command to run to spin this up. Assume the user knows nothing about Docker Compose, even if it's just `docker-compose up` it should be noted.

README.md
carimura: @denismakogon that's exactly what I said yesterday...

denismakogon:
@carimura @derekschultz i addressed your comments. Hopefully, this is what you've been asking for, thanks.

denismakogon:
@derekschultz addressed

derekschultz:
@denismakogon thanks!

PR #537, opened 2017-11-26
Fix container name referenced in UI startup

skwashd:
The README references a container named "functions", but `fn start` creates a container with a name "fnserver". When trying to start the UI using the documented command it fails because the expected container doesn't exist.


zootalures:
Hey @skwashd  Thanks for raising this - this has been fixed as part of #569 (sorry I missed this PR!)  -

PR #539, opened 2017-11-27
add error to call model

### LINKED ISSUE #331: Add reason attribute to call object

rdallman:
closes #331

previously, for async this error was being masked completely even if it was
something useful like the image not existing. for sync, the error was returned
in the http request but now it's also being stored. this error itself can
cover a lot of landscape, it could be an error in getting a slot, pulling an
image, running a container, among other things. anyway, no longer being
masked. we can likely improve it in certain cases we run into in the future,
but it's open ended at the moment and not being masked like some errors in
sync http request returns (503 non-models.APIError) for now.

* tucks in callTrigger stuff to keep api clean
* adds swagger
* adds migration
* adds tests for datastore and agent to ensure behavior

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -28,7 +28,11 @@ docker rm -fv func-mysql-test || echo No prev mysql test db container
 docker run --name func-mysql-test -p 3306:3306 -e MYSQL_DATABASE=funcs -e MYSQL_ROOT_PASSWORD=root -d mysql
 docker rm -fv func-minio-test || echo No prev minio test container
 docker run -d -p 9000:9000 --name func-minio-test -e "MINIO_ACCESS_KEY=admin" -e "MINIO_SECRET_KEY=password" minio/minio server /data
-sleep 5
+
+# pull all images used in tests so that tests don't time out and fail spuriously
+docker pull fnproject/sleeper
+docker pull fnproject/error
+docker pull fnproject/hello
 
 MYSQL_HOST=`host func-mysql-test`
 MYSQL_PORT=3306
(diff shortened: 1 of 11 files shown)


rdallman:
docker hub can't pull `fnproject/error` fast enough right now. the tests pass for me locally, i'll retrigger this periodically today and try to do #540 to get circle builds working...

rdallman:
this error doesn't necessarily come from a point in the call lifecycle where the logs were "open". for example, if the image does not exist. it also seems confusing to append things to a user's log, as they'll have to examine their code to find which part of the code emitted that log line when it was us (we could try to make this obvious, but my experience in attempting to make things obvious to users has never really worked). in general this field will be empty in the db, as it's nullable, so it's not taking up space, either.

rdallman:
thanks!

test.sh
```diff
@@ -28,7 +28,11 @@ docker rm -fv func-mysql-test || echo No prev mysql test db container
 docker run --name func-mysql-test -p 3306:3306 -e MYSQL_DATABASE=funcs -e MYSQL_ROOT_PASSWORD=root -d mysql
 docker rm -fv func-minio-test || echo No prev minio test container
 docker run -d -p 9000:9000 --name func-minio-test -e "MINIO_ACCESS_KEY=admin" -e "MINIO_SECRET_KEY=password" minio/minio server /data
-sleep 5
+
+# pull all images used in tests so that tests don't time out and fail spuriously
+docker pull fnproject/sleeper
+docker pull fnproject/error
+docker pull fnproject/hello
```
skinowski: yay! :-) 👍

PR #541, opened 2017-11-27
Container memory tracking related changes

skinowski:
The biggest change here is the separate tracking of sync/async memory. Here the code handles this as if these are separate pools, but sync requests can steal resources from async pool. The logic here is not significantly different than before, but it adds env vars to control the behavior. The default calculations are as follows:

1) calculate mem as before without magic numbers such as tooBig (300GB) There was a small bug in here where a user can configure a higher than avail ram in cgroups (eg 250GB cgroup limit on 50GB actual ram), which the previous code accidentally picked up (as opposed to min(cgroupLimit, ram))
2) subtract an estimate where this headroom is 10% of ram, but clamped with min: 256MB max: 5GB
3) use 20% of this value for sync only reservation, remaining 80% for both async or sync reservation
4) within that 80% reservation, use 20% of mem as async low water mark. This is because for async we do not know the mem requirement until we dequeue the request.

For example, for a 400GB linux server, assuming /proc/meminfo & cgroups return 400GB;

headroom = clamp(400GB * 0.1, 256MB, 5GB) => 5GB
ram = 400GB - 5GB => 395GB
sync mem = 0.2 * 395GB => 79GB
async mem = 0.8 * 395GB => 315GB
async low water mark = 0.2 * 315GB => 63GB

which means, this always keeps 79GB dedicated for sync requests, where remaining 315GB is for both async and sync. If async has free space >= 63GB, then we attempt to dequeue an async request from the queue.

The env variables allow customizations (even if fn server is not in a docker container.)

The default calculations can be improved, but given that a deployment can use the env vars to fine tune these, the above defaults might be good enough.

The headroom has some magic hardcoded numbers, which is a concern still. But users can get around it using the env variables.

If the OS is not linux, then the defaults are:

sync mem = 1GB
async mem = 1GB
async low water mark = 512MB

### DIFF
--- a/api/agent/async.go
+++ b/api/agent/async.go
@@ -15,16 +15,11 @@ func (a *agent) asyncDequeue() {
 		select {
 		case <-a.shutdown:
 			return
-		case <-a.resources.WaitAsyncResource():
-			// TODO we _could_ return a token here to reserve the ram so that there's
-			// not a race between here and Submit but we're single threaded
-			// dequeueing and retries handled gracefully inside of Submit if we run
-			// out of RAM so..
+		default:
 		}
 
-		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second) // TODO ???
+		ctx := context.Background()
 		model, err := a.mq.Reserve(ctx)
-		cancel()
 		if err != nil || model == nil {
 			if err != nil {
 				logrus.WithError(err).Error("error fetching queued calls")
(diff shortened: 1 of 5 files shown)


skinowski:
If this is the direction we want to go, I want to add cpu shares to this impl (another pull req).

api/server/init.go
```diff
@@ -26,6 +27,18 @@ func init() {
...
+
+	if runtime.GOOS != "linux" {
+		// This still lets 10-20 sync/async functions execute concurrently assuming a 2GB machine.
+		viper.SetDefault(EnvFnSyncMemory, 1*1024)
+		viper.SetDefault(EnvFnAsyncMemory, 1*1024)
+		viper.SetDefault(EnvFnAsyncLWMemory, 512) // 512MB for async low water mark
+	} else {
+		viper.SetDefault(EnvFnSyncMemory, 0)    // unlimited (fallback to 20% of docker --memory if possible)
+		viper.SetDefault(EnvFnAsyncMemory, 0)   // unlimited (fallback to 80% of docker --memory if possible)
+		viper.SetDefault(EnvFnAsyncLWMemory, 0) // not set (fallback to 20% of async memory)
```
rdallman: since async and sync pools are partitioned, this is to leave additional room for sync to steal from? is there a good reason to expose this to the user? (it seems complex, i'm not big on knobs but that's me, personally)

I can see the appeal of exposing these for people not on linux who don't want to eat our defaults _and_ don't want to run in docker, I guess (I'm also okay with not caring about said people [until the day comes]). I'm just not so keen on exposing any of these as it means we are basically contracted to support them and it seems like we may one day find a better policy than simply chopping things up by ram blocks for sync and async. since docker is a dependency at this point, I think 99% of cases are covered with `docker --memory` for setting max and then just a 'ratio' knob that may be more straightforward than a user having to calculate the ratios to concrete values a priori, but i'm not big on exposing the knob (supporting the knob) for ratio. it would simplify the configuration quite a bit as well and I think with the same outcome. it wouldn't account for the low watermark but I think that this all moves to a point where we calculate 'usable' ram on the fly every time we request a slot and it would be better to just pass over async requests (with no idea of watermark) if after we get them we see we don't have enough ram to run them, in a 2 phase manner, rather than needing to know we have enough to run one before hand (i think these cases will be only really when people run a really small vm to test out fn and try to run a huge image, which has pebkac issues above all else and we can just spew logs about the situation rather than trying to be precise).

skinowski:
Got rid of the config knobs and settled with some headroom + %20 remaining ram for sync reserved, the rest for servicing both sync + async. For non-linux, hard coded this to 512MB + 1.5GB.

For async code & MQ, there's some more work todo since the new dequeue/wait-resource now could take up to 900 secs and this is not going to work well with current 1 min MQ timeout. (This didn't work well before also.) After 900 secs, it attempts a final Submit() with call.Timeout enforced, then if all fails, the request fails. But async code now waits for the precise resource and then reserves that slot.

api/agent/async.go
```diff
@@ -33,17 +28,47 @@ func (a *agent) asyncDequeue() {
...
+		select {
+		case <-a.shutdown:
+			cancel()
+			return
+		case slot, isOpen := <-ch:
+			if isOpen {
+				// we got a slot, reserve/set it for this call to be used again by Submit() below
+				call.slot = slot
 			}
+		}
```
rdallman: this logic appears to be the same as what's in `Submit`. I'm not sure we're getting any sugar by moving it up here also, but maybe I'm missing something? If the job times out in `Submit` we should be able to handle the error that's being returned now when a job times out but hasn't been ran yet and not delete the message for it so it can be run again?

rdallman: the only issue with doing this without checking if there's _any_ ram available before trying to `Submit` or get a slot is that we're going to hold onto the message for a call for a while (likely timing out) when it could be being ran elsewhere, hence `WaitAsyncResources`.

skinowski: Yes, Call.Timeout doesn't apply the time wait in queue (or waiting for resources) basically. Except for this 900 sec wait, there's no benefit. It waits for 900 to reserve a lot. And this waiting is in one thread.

rdallman: I see, that makes sense. as alluded to in comment we're going over message timeout, I guess it seems likely that if a node hits saturation we could pretty easily run a task twice with this strategy yea?

skinowski:
ok, async is like the old logic a little bit now. We can revisit if needed if MQ side evolves in the next few weeks.

PR #551, opened 2017-11-30
fn: remove 100 msec sleep for hot containers

skinowski:
*) this reduces 100 msec response time in first
time launched hot containers
*) subsequent requests may trigger launch of
more hot containers, but this drawback is ok since
this means existing hot containers were not able
to meet the demand.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -2,12 +2,9 @@ package agent
 
 import (
 	"context"
-	"crypto/sha1"
 	"errors"
-	"fmt"
 	"io"
 	"net/http"
-	"sort"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


rdallman:
this ends up making a pretty big herd [against docker] for short lived tasks. the theory with the 100ms is that since we have to eat 300ms+ of docker startup time if we have to start a container, it's better to try to wait a little to find a free slot than to start a container (and, well, 100ms < 300ms), and if we do have to start a container then waiting 100ms isn't going to be more than 25% added at worst to latency. the policy [on master] ends up being problematic in the 0->1 case, but I have some ideas (use the `singleflight` mechanism we have, perhaps). anyway I think leaving the 100ms is advisable just to do a little herd dampening, but agree that it's a pretty naive policy to begin with.

carimura:
just for context Vish working on some AI/ML stuff on top of Fn was running into issues due to the 100ms sleep so @skinowski volunteered to re-look at strategy. Maybe he can describe what he ran into in this issue or IRL.

rdallman:
iirc the issue was too many containers launching on startup? this likely in practice is the same wrt that.

skinowski:
I'll submit another PR later for this. I have some draft changes to improve this.

PR #552, opened 2017-11-30
Move getEnv funcs to fncommon and export

nigeldeakin:
This change moves getEnv and getEnvInt from  to api/common and makes them public to that they can be used by extensions.


api/fncommon/fncommon.go
```diff
@@ -0,0 +1,26 @@
+package fncommon
```
denismakogon: Why can't this be just a file inside API package? I see no reason for creating a package for 2 functions.

nigeldeakin: This was requested by @treeder
https://fnproject.slack.com/archives/C7FKW83GV/p1511971979000041?thread_ts=1511947888.000097&cid=C7FKW83GV

denismakogon: I kinda disagree with the placement. I see no point to define the whole package for 2 functions, I'm not seeing more fn parts be exposed outside except API extensions. So, all these changes eventually land in `api/common.go`.

nigeldeakin: I'll leave you and @treeder to discuss this.

denismakogon: @treeder, what’s the specific reason to define yet one more package for two function?

nigeldeakin: Just to be clear, I don't have a strong view on this and am happy to go along with whatever @denismakogon  and @treeder want. 

For these two functions, I need to know:
* package name and * directory name. 

This PR (based on what I think @treeder asked for, to avoid possible name clashes in extensions that already have a `common` package) has
* package=`fncommon`
* directory= `api/fncommon`

I think @denismakogon is suggesting they stay where they are,:
* package=`common`
* directory= `api/common`

I suppose it partly depends on whether we intend to add additional API, for use by extensions, here in the future.

treeder: We could throw these in fnext package that's part of this PR: https://github.com/fnproject/fn/pull/554 

Basically a package meant for extension developers to use.

denismakogon:
To restore this discussion, i'd better do what this issue asks for 
 https://github.com/fnproject/fn/issues/556 so we don't have to add more packages for reusable code.

rdallman:
@nigeldeakin feel free to resurrect, closing for now

PR #554, opened 2017-11-30
Middleware upgrade

treeder:
* Adds root level middleware (was just API middleware)
* Adds control flow mechanism for middleware - [example](https://github.com/fnproject/fn/compare/root-middleware?expand=1#diff-370c8a94414d32d5ba9c77dedf9f318aR91)
* Moves extension interfaces to fnext package

### DIFF
--- a/api/server/middleware.go
+++ b/api/server/middleware.go
@@ -2,67 +2,128 @@ package server
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 
 	"github.com/fnproject/fn/api/common"
+	"github.com/fnproject/fn/fnext"
 	"github.com/gin-gonic/gin"
 )
 
-// Middleware just takes a http.Handler and returns one. So the next middle ware must be called
-// within the returned handler or it would be ignored.
-type Middleware interface {
-	Chain(next http.Handler) http.Handler
+type middlewareController struct {
+	// NOTE: I tried to make this work as if it were a normal context, but it just doesn't work right. If someone
+	// does something like context.WithValue, then the return is a new &context.valueCtx{} which can't be cast. So now stuffing it into a value instead.
+	// context.Context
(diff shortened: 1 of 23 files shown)


api/server/middleware.go
```diff
@@ -2,67 +2,128 @@ package server
...
+		err := recover()
+		if err != nil {
+			common.Logger(c.Request.Context()).WithField("MiddleWarePanicRecovery:", err).Errorln("A panic occurred during middleware.")
+			handleErrorResponse(c, ErrInternalServerError)
+		}
+	}()
+
+	ctx := context.WithValue(c.Request.Context(), fnext.MiddlewareControllerKey, s.newMiddlewareController(c))
+	last := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Println("final function called")
```
rdallman: see a few `Println` littered around. should we use debug logger maybe so it's there but with the flexibility to remove it? (also see that old impl used println, but easy enough to fix while we're here)

api/server/runner.go
```diff
@@ -20,24 +20,35 @@ type runnerResponse struct {
...
-
-	r, routeExists := c.Get(api.Path)
-	if !routeExists {
-		r = "/"
+// handleFunctionCall executes the function.
+// Requires the following in the context:
+// * "app_name"
+// * "path"
+func (s *Server) handleFunctionCall(c *gin.Context) {
+	// @treeder: Is this necessary? An app could have this prefix too. Leaving here for review.
```
rdallman: it seems like a pretty silly guard, so +1 remove. the front end had some funny business where this could have potentially been called from somewhere where this were true, because it was falling back any request where the gin router didn't match to try to run a function call with that path, but this has been removed now.

PR #557, opened 2017-12-01
fn: sigterm graceful shutdown handling

skinowski:


carimura:
nice nice. thanks.

```
time="2017-12-02T15:56:08Z" level=info msg="Serving Functions API on address `:8080`"
^C2017/12/02 07:56:12 interrupt caught, exiting
time="2017-12-02T15:56:12Z" level=info msg=Halting...
time="2017-12-02T15:56:12Z" level=info msg="server stopped"
```

PR #559, opened 2017-12-02
MySQL group replication for compose deployment

denismakogon:


denismakogon:
@carimura compose file updated, at this moment looking at https://github.com/kubernetes/kompose tool, i suppose there's a chance to run the same compose file against k8ts, will test one on minikube.

denismakogon:
Shame, unfortunately, kompose does not support volumes from compose file =/

denismakogon:
@carimura i switched to fnproject/fnserver images for simpler bootstrap flow, so compose would not attempts to build images from source.

@nigeldeakin @derekschultz please review, thanks.

rdallman:
are we officially supporting this? I don't want to mislead anyone, since we're not using this for prod.

denismakogon:
I see no objections supporting this, the term "officially" is not something that we should apply to this type of deployment. From Fn perspective, there's no difference from in running over single instance MySQL or MySQL group replication. But we'd like to should people who want to run their own deployment that we recommend (not strongly insisting) this type of deployment as long as they decide to work with MySQL instead of PG or whatever Fn supports except this two.

rdallman:
this just seems to imply we support this pretty strongly (officially, even), carrying it around in the top level of the repo. the changes introduced here are not exactly a quick getting started guide as this adds out of the box mysql group replication using innodb clustering, with images we don't control that aren't officially supported by innodb or mysql either, the registry page even states they are unofficial. there's a pretty big difference just in that, to me. if we explicitly have a docker compose to help users start up all the things really quickly just to get running, that's ok, but this seems like we're adding a lot of complexity to what is a top level docker compose file in our repo, that's likely to be used simply for cursory testing only of a clustered mysql backing, which we aren't even going to use ourselves very likely (just for figuring out config params and such). 

anyway, not very keen on this, this file being where it is. if we tuck it away in the ops docs and explicitly state its purpose, that seems more okay, and at that point i'm not sure that the only option for a compose file should be clustered mysql, either.

derekschultz:
I share the same concerns as @rdallman on this one. While I'm fine with having a compose example, I think the file should be moved from the root of the repo and into the operating docs, similar to the Helm stuff. Also, would be helpful to include a doc that describes what it does and how to utilize it, even if it's just as simple as `docker-compose up`. The point is we can't assume everyone is familiar with this tool, so it's on us to educate as best as we can, especially since there are several ways to roll your own Fn deployment.

denismakogon:
Gonna close this, compose file would never become production deployment guide, so leaving it as is totally fine.

PR #561, opened 2017-12-04
More metrics

nigeldeakin:
This PR modifies the tracing spans, and hence Prometheus metrics for the `agent_submit` operation, which essentially encompasses the execution of a cold or hot function. 

There are now three separate tracing spans (Prometheus metrics) with different baggage items (Prometheus labels) as follows:
* `agent_submit_route` - has the labels `fn_appname` and `fn_path` set
* `agent_submit_app` - has the label `fn_appname` set (but not `fn_path`)
* `agent_submit_global` - neither label set

These three separate metrics are required to support the queries needed by the new statistics API provided by the `ext-metrics` extension.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -195,19 +195,19 @@ func (a *agent) Submit(callI Call) error {
 
 	call := callI.(*call)
 	ctx := call.req.Context()
-
-	span, ctx := opentracing.StartSpanFromContext(ctx, "agent_submit")
-	span.SetBaggageItem("fn_appname", callI.Model().AppName)
-	span.SetBaggageItem("fn_path", callI.Model().Path)
-	defer span.Finish()
+	
+	// start spans in the correct order so each span is given, and inherits, the correct baggage
+	span_global, ctx := opentracing.StartSpanFromContext(ctx, "agent_submit_global")
+	defer span_global.Finish()
 	
 	span_app,_ := opentracing.StartSpanFromContext(ctx, "agent_submit_app")
 	span_app.SetBaggageItem("fn_appname", callI.Model().AppName)
 	defer span_app.Finish()
 
-	span_global, _ := opentracing.StartSpanFromContext(ctx, "agent_submit_global")
-	defer span_global.Finish()
-
+	span, ctx := opentracing.StartSpanFromContext(ctx, "agent_submit")
+	span.SetBaggageItem("fn_path", callI.Model().Path)
+	defer span.Finish()
+	
 	// start the timer STAT! TODO add some wiggle room
 	ctx, cancel := context.WithTimeout(ctx, time.Duration(call.Timeout)*time.Second)
 	call.req = call.req.WithContext(ctx)
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -195,19 +195,19 @@ func (a *agent) Submit(callI Call) error {
 
 	call := callI.(*call)
 	ctx := call.req.Context()
-	
+
 	// start spans in the correct order so each span is given, and inherits, the correct baggage
 	span_global, ctx := opentracing.StartSpanFromContext(ctx, "agent_submit_global")
 	defer span_global.Finish()
-	
-	span_app,_ := opentracing.StartSpanFromContext(ctx, "agent_submit_app")
+
+	span_app, ctx := opentracing.StartSpanFromContext(ctx, "agent_submit_app")
 	span_app.SetBaggageItem("fn_appname", callI.Model().AppName)
 	defer span_app.Finish()
 
 	span, ctx := opentracing.StartSpanFromContext(ctx, "agent_submit")
 	span.SetBaggageItem("fn_path", callI.Model().Path)
 	defer span.Finish()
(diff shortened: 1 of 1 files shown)


nigeldeakin:
For a route-scoped query (where both app and route are specified), the underlying Prometheus query is
```
http://localhost:9090/api/v1/query_range?query=rate(fn_span_agent_cold_exec_duration_seconds_sum{fn_appname="hello-async-a",fn_path="/hello-async-a1"}[1m])/rate(fn_span_agent_cold_exec_duration_seconds_count{fn_appname="hello-async-a",fn_path="/hello-async-a1"}[1m])
```
This works fine and returns a single range.

However for an app-scoped query (where only a app is specified) or a global query (where neither are specified) then the above query returns multiple ranges (one for each combination of label values), which is not what is wanted. In order to obtain a single range, we need to perform a sum across all label values that we're not explicitly constraining, so that the underlying query becomes
```
rate(sum(fn_span_agent_cold_exec_duration_seconds_sum{fn_appname="hello-async-a"})[1m])/rate(sum(fn_span_agent_cold_exec_duration_seconds_count{fn_appname="hello-async-a"})[1m])
```
However Prometheus doesn't you to calculate a rate across a sum, and gives an error. Removing the need to perform the sum works around this.

api/agent/agent.go
```diff
@@ -196,8 +196,15 @@ func (a *agent) Submit(callI Call) error {
 	call := callI.(*call)
 	ctx := call.req.Context()
 
+	// start spans in the correct order so each span is given, and inherits, the correct baggage
+	span_global, ctx := opentracing.StartSpanFromContext(ctx, "agent_submit_global")
```
rdallman: stuff like this seems like in prometheus we should be able to filter by `agent_submit.*` to get across-all-apps data? (same for by route)? and for digging into a specific app `agent_submit.yodawg` -- I am lacking knowledge on prometheus data structuring and this change seems to be reversing exactly that behavior, would appreciate your enlightening me.

rdallman: sorry, PR page hadn't updated to include https://github.com/fnproject/fn/pull/561#issuecomment-349090240 (yay firefox). it seems like we should probably have a root level split of all spans then to have 1 that is global, 1 that is per app and 1 per route then instead of only the agent ones?

this kind of blows if that's how prometheus works, but in the returned query from prometheus that returns multiple ranges can we not select the one we're interested in based on the parameters we gave it? e.g. `span = spans(query)[0]` ? the overhead of returning multiple ranges all the time kind of blows, but storing them seems worse I guess (prom api to specify '.first()' or something?)

rdallman: I think I misinterpreted, I guess we get one for each. This seems problematic if there are 2 routes with the same name under separate apps since we're summing based on path alone, yea? Should we make the route tag be "app/route" instead of just the route name? Then this is fixed, partially? I guess we still need the global one.

nigeldeakin: Thanks for reviewing this. 

I know I'm labouring the point a little here, but here we go:

With this PR there would be three separate tracing spans (Prometheus metrics) with different baggage items (Prometheus labels) as follows:
* `agent_submit` (Not `agent_submit_route` as I incorrectly stated initially, This metric already exists, and I'm suggesting we stick with the current name) - has the labels `fn_appname` and `fn_path` set
* `agent_submit_app` - has the label `fn_appname` set (but not `fn_path`)
* `agent_submit_global` - neither label set

Note that I'm creating the three spans in the opposite order to above, with each trace the child of the previous, so I don't need to set the app name explicitly on the `agent_submit` span: it will be inherited from the `agent_submit_app`  span.

Here's why I think we need all three:

If we are specifying a "route-scoped" query (i.e. the rolling mean duration across a specific route in a given app) then the Prom query specifies both the app label and the route label, using a metric which has both labels set (i.e. the metric that exists now), and obtains a single range in response.  Just to be clear: the user would never specify just a route. If they want to specify a route then they must also specify the app.

If we are specifying what I'm calling a "app-scoped" query (i.e. the rolling mean duration across all routes in a given app) then the Prom query specifies the app name. This needs to use a metric which has just an appname label so that the query returns a single range across all routes in that app. We can't use a metric with both route and app labels since if it did this would return multiple ranges, one for each route in the specified app, which is expressly not what we are asking for here, and it's not possible for the query expression to sum them. So we need a metric which has an app label but not a route label.

If we are specifying what I'm calling a "global-scoped" query (i.e. the rolling mean duration across all routes and all apps) then the Prom query doesn't specify an app name or route name. This needs to use a metric which no labels so that the query returns a single range across all routes in all apps. We can't use a metric with any labels since if it did this would return multiple ranges, one for each unique combination of label values, which is expressly not what we are asking for here, and it's not possible for the query expression to sum them. So we need a metric which has no labels.

It may help to see the intro to the new statistics API at https://github.com/fnproject/ext-metrics . This lists the three types of query I need to support, and the information that needs to be returned.

rdallman: thanks for reiteration, clicked that it's not separate spans but child of spans, seems fine.

Issue #274, opened 2017-09-01
figure out timeout & launch policy for hot & cold functions


rdallman:
at present we have:

cold
-----
* wait up to the call timeout to launch a container (and don't run if `> timeout / 2` elapsed)

hot
-----
* wait 100ms for a free container
* launch a container & wait for any free container up to the call timeout (and don't run if `> timeout / 2` elapsed

this policy requires synthesizing 2 things:

one is, under what conditions should we even launch hot containers? 

with the current policy if nothing is free in 100ms (hot), we attempt to launch unconditionally. this can lead to launching a lot of containers for a route, but eventually there is an equilibrium with the offered load to where there are free slots when requests come in and it's no longer necessary to launch containers. 100ms is just a magic number to apply some back pressure.

the other thing is, how long should we wait for ram to become available / for a slot to become free? since we are stopping things from executing if `> timeout / 2` has elapsed, it seems like we should probably at most only wait this long to begin with. when a machine is at capacity ram-wise, there are likely hot containers up for certain routes, so immediately timing out all tasks that can't find a slot immediately seems wrong, too (there's at least some point in time where it would be faster to wait than to retry against another fn server).

and another thing to keep in mind is that as alluded to, we want to timeout as quickly as possible if we have a good approximation that we aren't going to be able to run a certain call within a certain time window, so that the call could run on another fn server as quickly as possible.

not necessarily digging for code here, think we should discuss the policy at a high level and then go implement it after we all/mostly agree.

skinowski:
Should we close this? The wait policy was changed to the following in the last few weeks:

- 100 ms is removed
- Now we aggressively launch hot containers based on slots.isNewContainerNeeded(), but this
aggressiveness is dampened by LIFO queue which times out (IdleTimeout) idle (AKA unnecessary) containers.

isNewContainerNeeded() potentially needs more work and fine tuning going forward. Currently it considers a few metrics to decide if a new container is needed:

1) num of waiters
2) num of "starting" containers to avoid starting too many in parallel
3) num of currently running containers
4) wait latency
5) wait latency compared to estimated run latency given running container count
5) container start latency

Maybe there needs to be a optimization towards low latency versus optimization towards low consumption. Today, this is geared more towards low latency, but if a user is choosing hot containers, that implicitly means low latency. So, this policy may be OK.

NOTE: isNewContainerNeeded() is run by one go routine per slot.

rdallman:
[...]

PR #562, opened 2017-12-04
slot tracking improvements

skinowski:
*) moved slot management to its own file
*) slots are now implemented with LIFO semantics, this is important since we do
   not want to round robin hot containers. Idle hot containers should timeout properly.
*) each slot queue now stores a few basic stats such as avg time a call spent in a given
   state and number of running/launching containers, number of waiting calls in those states.
*) first metrics in these basic stats are discarded to avoid initial docker pull/start spikes.
*) agent now records/updates slot queue state and how much time a call stayed in that state.
*) waitHotSlot() replaces the previous wait 100 msec logic where it sends a msg to hot slot go routine launchHot() and waits for a slot
*) launchHot() is now a go routine for tracking containers in hot slots, it determines if a new containers is needed based on slot queue stats.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -267,7 +267,7 @@ launchLoop:
 				if !ok {
 					return nil, errors.New("slot shut down while waiting for hot slot")
 				}
-				if s.acquireSignal() {
+				if s.acquireSlot() {
 					if s.slot.Error() != nil {
 						s.slot.Close()
 						return nil, s.slot.Error()
@@ -276,10 +276,8 @@ launchLoop:
 				}
 
 				// we failed to take ownership of the token (eg. container idle timeout)
-				// reset & close state and try launching again
-				s.slot.Close()
+				// try launching again
 				continue launchLoop
-
 			case <-ctx.Done():
 				tokenCancel()
 				return nil, ctx.Err()
@@ -293,7 +291,7 @@ launchLoop:
 			if !ok {
 				return nil, errors.New("slot shut down while waiting for hot slot")
 			}
-			if s.acquireSignal() {
+			if s.acquireSlot() {
 				if s.slot.Error() != nil {
 					s.slot.Close()
 					return nil, s.slot.Error()
@@ -302,8 +300,7 @@ launchLoop:
 			}
 
 			// we failed to take ownership of the token (eg. container idle timeout)
-			// reset & close state and try launching again
-			s.slot.Close()
+			// try launching again
 		case <-ctx.Done():
 			return nil, ctx.Err()
 		case <-time.After(time.Duration(200) * time.Millisecond):
@@ -332,8 +329,7 @@ func (a *agent) launchCold(ctx context.Context, call *call) (Slot, error) {
 		if !ok {
 			return nil, errors.New("cold slot shut down while waiting for launch")
 		}
-		if !s.acquireSignal() {
-			s.slot.Close()
+		if !s.acquireSlot() {
 			return nil, errors.New("cannot acquire cold slot")
 		}
 		if s.slot.Error() != nil {
@@ -567,6 +563,14 @@ func (a *agent) runHot(ctxArg context.Context, call *call, tok ResourceToken) {
 
 	go func() {
 		for {
+			select {
+			case <-ctx.Done(): // container shutdown
+				return
+			case <-a.shutdown: // server shutdown
+				return
+			default:
+			}
+
 			done := make(chan struct{})
 			s := call.slots.queueSlot(&hotSlot{done, proto, errC, container, nil})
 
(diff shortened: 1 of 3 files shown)


api/agent/agent.go
```diff
@@ -595,10 +628,10 @@ func (a *agent) runHot(ctxArg context.Context, slots chan<- slot, call *call, to
 			}
 
 			done := make(chan struct{})
-			slot := &hotSlot{done, proto, errC, container}
+			qdone := call.slots.queueSlot(&hotSlot{done, proto, errC, container})
 
 			select {
-			case slots <- slot:
+			case <-qdone:
```
rdallman: does this leave us in a position where we have queued the slot, the slot remains queued, yet here we hit the idle timeout and bail but a future call could acquire this slot even though the container is dead?

api/agent/agent.go
```diff
@@ -244,35 +240,141 @@ func (a *agent) Submit(callI Call) error {
...
+func (a *agent) waitHotSlot(ctx context.Context, call *call) (Slot, error) {
+
+	// The loop executes in 'sleep' msec intervals; starting fast (25), then slowing down (100)
+	maxSleep := time.Duration(100) * time.Millisecond
+	stepSleep := time.Duration(10) * time.Millisecond
+	curSleep := time.Duration(25) * time.Millisecond
+
+	// let's start with call.Timeout as an estimate of latency tolerance. This is
+	// the amount of time we can tolerate to spend on waiting for slots, worst case.
+	maxWaitTolerance := uint64(call.Timeout) * 1000
```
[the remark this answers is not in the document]
rdallman: ah, this is seconds. would you mind doing `uint64(time.Second)` instead of `1000` for clarity?

skinowski:
Running a simple test against the old logic and the impact of round robin on hot containers is visible. Once you spin up a bunch of hot containers, even a small load keeps all of them active since idle_timeout is not triggered.

rdallman:
awesome. should ask flow team to run their benchmarks again and see if things improve w/ this

skinowski:
Seeing some suspicious 5xx, which I'll look into tomorrow, but overall latencies/numOfContainers looks good. I had to tweak a few things & fix some tracking bugs.

skinowski:
@rdallman I think this is a bit in better shape now. I noticed that the 5xx issues were popping up due to 50% time check in call.Start() which I was violating. I preserved this threshold as 50%, but added it's counterpart and made them two variables where we spend up to 40% of timeout on wait/queue, and 50% in after Start().

isWaitTolerable() is aggressive about launching containers. But this is dampened by a last-minute slot check and seems to work well with LIFO idle timeouts. I really tried to make isWaitTolerable() and waitHotSlot() as simple as I can since it's really hard to tune it. (feedback loop, various work loads, scenarios...)

slot latency trackers throw away first measurements to improve docker pull/start initial delay.

I tested this on a echo server test case and observed no differences between error rates between old/new logic, but new code tends to stabilize to a lower & constant number of containers to sustain that workload. Due to LIFO, this is significantly less than previous container count, which due to round robin kept alive in docker. There's some risk in the new code for different workloads, but LIFO idle timeout improvement is really big.

skinowski:
Ran another test with gamma distribution of (0 to 9000 msec) sleep time on containers which worked OK.

skinowski:
Hmm, I see a bug in getSlot()... I need to do one more pass.

api/agent/agent.go
```diff
@@ -597,18 +580,33 @@ func (a *agent) runHot(ctxArg context.Context, slots chan<- slot, call *call, to
...
+			// if we fail to eject the slot, it means that a consumer
+			// just dequeued this and acquired the slot. In that case,
+			// we will wait for that slot to complete (<-done) and
+			// perform exit/return at the beginning of this loop.
+			if (tryShut || tryTerm || tryIdle) && call.slots.ejectSlot(s) {
+				if tryIdle {
+					logger.Info("Canceling inactive hot function")
+				}
+				if tryShut || tryIdle {
+					shutdownContainer()
```
rdallman: prob help simplify a little: I think we can just defer this from directly after we acquire it instead of carrying it around down here. we won't run defers until we exit the loop and at that point either the container is dead already (won't hurt, noop basically) or needs to be shut down. seems less brittle

api/agent/agent.go
```diff
@@ -597,18 +580,33 @@ func (a *agent) runHot(ctxArg context.Context, slots chan<- slot, call *call, to
...
+				tryShut = true
+			}
+
+			// if we fail to eject the slot, it means that a consumer
+			// just dequeued this and acquired the slot. In that case,
+			// we will wait for that slot to complete (<-done) and
+			// perform exit/return at the beginning of this loop.
+			if (tryShut || tryTerm || tryIdle) && call.slots.ejectSlot(s) {
+				if tryIdle {
+					logger.Info("Canceling inactive hot function")
```
rdallman: same for this, we can just defer this and say hot function is shut down regardless of reason. then won't need any of these vars

skinowski:
related to: #274

PR #563, opened 2017-12-05
Fnlb/k8s grouper

jan-g:
K8s support for the fnlb.

This adds a k8sGrouper (as proposed in fn#553). It pulls in the k8s.io/client-go requirement (at their recommended v4.0.0).

Each fnlb, if run in a k8s container, can be configured to watch for the existence of pods matching a selector spec. The shared DB requirement disappears in this configuration (as the k8s master replaces it).

The management API still exists at the GET endpoint to list status.

There's a lot of similarity between allGrouper and k8sGrouper. Further refactoring can be done, but want to put a stake in the ground so that helm work can proceed; the refactor can wait until fnlb lives in its own repo.


skinowski:
In k8s's case, add()/remove() are probably not applicable (and should return perm denied?). Rest of allGrouper logic seems identical, so we should be able to replace that 'db' DBStore with a custom k8s "db" layer that implements DBStore iface right?

rdallman:
forgot that was there, +1 to using DBStore (apologies for misleading in #563)

jan-g:
That's pretty much what i was thinking, yeah. Want a first stab that can go into the helm charts.


-- Excuse typos. Phone screens don't make great ketbroads. 

-------- Original message --------

fnlb/lb/k8s_store.go
```diff
@@ -0,0 +1,181 @@
...
+			logrus.Info("Stopped watching for pod changes")
+			pods.Stop()
+			return
+		case event := <-pods.ResultChan():
+			if pod, ok := event.Object.(*v1.Pod); ok {
+				logrus.
+					WithField("Event", event.Type).
+					WithField("Pod", pod.SelfLink).
+					WithField("PodPhase", pod.Status.Phase).
+					WithField("Address", pod.Status.PodIP).
```
rdallman: logrus tip/nit: `WithFields(logrus.Fields{"key":value, "key2":value})` may be slightly less verbose when there are a lot of fields

fnlb/lb/k8s_store.go
```diff
@@ -0,0 +1,181 @@
...
+					Debug("Change detected")
+				switch event.Type {
+				case watch.Added:
+					fallthrough
+				case watch.Modified:
```
rdallman: go feature: switching on multiple values is supported as long as the concretely typed value inside of the case block isn't needed. i.e. `case watch.Added, watch.Modified:`

PR #567, opened 2017-12-05
Switch to new dep tool from glide, remove vendor dir.

treeder:
I realize some people might not like the idea of dropping the vendor dir, but if a dep tool works good, then we shouldn't have vendor in git. If nothing else, this shows that the dep tool is OK to use now. 

I just did the following:

*`rm -rf vendor`
*`dep init`
* set the same constraints as we had in glide files in Gopkg.toml
*`dep ensure`
* fixed one change introduced here: https://github.com/docker/distribution/blame/8cedd7b3a49e719220c77584e42b7c7386fba811/registry/client/repository.go#L65


rdallman:
[most](https://github.com/moby/moby) [projects](https://github.com/kubernetes/kubernetes) [understand](https://github.com/prometheus/prometheus) [darkness](https://github.com/mholt/caddy)

we've had this discussion before, and dep doesn't really seem like it makes any substantial changes to vendoring at large in Go vs. glide, it's just one is going to be core [in Feb]. it's like anything, there are trade offs, and it sure seems like devs having in sync dependencies, having faster CI builds, and not relying on ad hoc VCS mirrors outweighs the extra 2s of github latency to click on 'changes' and having to visually ignore `/vendor` files at the end of PRs for most large OSS projects. go is still lacking is just the reality of the whole thing. if go ever gets serious about making deps better it seems like a good time to re-evaluate, but dep is basically just glide by another name. 

moving to dep will be nice when things change over in feb.

denismakogon:
+1 to switch to dep tool as well as waiting until 1.10 pre-release (see https://github.com/golang/dep/wiki/Roadmap).

api/agent/drivers/docker/registry.go
```diff
@@ -90,7 +90,7 @@ func CheckRegistry(ctx context.Context, image string, config docker.AuthConfigur
 
 	tran = &retryWrap{cm, tran}
 
-	repo, err := registry.NewRepository(ctx, repoNamed, regURL, tran)
+	repo, err := registry.NewRepository(repoNamed, regURL, tran)
```
skinowski: ctx removal looks like a bad merge?

PR #576, opened 2017-12-06
Now matches docker version instead of our own versions.

treeder:


images/dind/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM docker:17.09-dind
+FROM docker:17.11-dind
```
denismakogon: just wonder if the right place of this image files is in fnproject/dockers?

rdallman: don't think it matters really, but I'm not sure how generally useful this image really is anyway. it's very similar to the dind image however it basically enforces starting docker with very specific args since the normal dind image doesn't.

treeder: ya, could probably move to dockers repo.

PR #577, opened 2017-12-06
fn: fail count should include timeouts

skinowski:


nigeldeakin:
That was my bug. Looks good.

Issue #531, opened 2017-11-21
Hybrid Mode MVP


treeder:
There are 3 main parts that will work together to make hybrid work. 

1) Fn Server (API)
2) Fn LB
3) Fn Runners/Workers

At a high level, they interact like this:

Control Plane | &nbsp; | Work Plane
------------ | ---- | -------------
Fn Server | &larr; | Fn LB(s)
&darr; | &uarr; | &darr; 
Database | &larr; | Fn Runner(s)
MQ | &nbsp; | &darr; 
Blob Store | &nbsp; | ?Blob Store

Where everything in the Work Plane talks to the Fn Server to tell it what to do. 

As a bare minimum starting point, we should have 3 containers that we can run like this:

Start Control Plane:

```sh
docker run ... fnproject/fnserver
```

Start Work Plane

```sh
docker run ... fnproject/fnlb
docker run .... fnproject/fnrunner
```

## Fn Server

This probably doesn't really have to change much except:

* flag to disable async processing
* new API endpoints to support below

## Fn LB

* allow runners to register themselves (simple API on LB)

## Runners

* When fnrunner starts, it will:
  * register itself with the fnlb (k8s does this for us, since we're using this)
  * start asking the fnserver for async calls 
* For each call it gets, it will:
  * execute call
  * report status to fnserver
  * send logs to fnserver (this could store logs in local blob store, but for first attempt, let's assume this isn't the case)

rdallman:
>  allow runners to register themselves (how to do this with multiple LBs running?)

it's backed by a db. just start lb with `DB_URL=...` same as server, supports mysql/pg.

rdallman:
>  new API endpoints to support below

I have a sketched out design doc for this api for fn server that I think will work & be simple, i can try to put it into something more palatable. was attempting to add reasonable MQ semantics at the same time, but it's possible we could/should punt this.

zootalures:
Am i right in thinking: async will  go : 

Outside -> (0: push)  FnLB -> (1: push)  Server->  (2: push) MQ -> (3: (poss via API) pull) runner

If so then  direct MQ is a good start for 3 and then wrap in an API once that works I think

treeder:
That sounds about right.  I don't think we'd want runner to directly access MQ though, would add another layer of authentication and what not that we'd have to think about. Runner should be pretty dumb for the most part and just get work (sync) or ask for work (async), run it, report results, that's basically all it should think about. And hopefully just talk to one thing. 

[...]

PR #578, opened 2017-12-07
hybrid server side api

rdallman:
this implements the beginnings of the server side API for API nodes. this is relatively in flux API, and as such, haven't wrote a doc with the endpoints or made a client library for it as of yet. it would be nice to start lining up trains so we can start mashing some of this stuff together, even if everything is in flux, so here goes. 

follows from discussion in #531. there are stubs for UpdateCall atm, we should discuss how we want this to work based on functionality highlighted here -- discuss elsewhere, probably. also, cannot really test this reliably until we can shut off the agent, which is in another patch.

anyway, choo choo


rdallman:
@hhexo @msgodf could you review please ?

rdallman:
closing for #581

PR #582, opened 2017-12-07
Adjusting client generation script

denismakogon:


clients/build.rb
```diff
@@ -23,7 +23,7 @@ def clone(lang)
   Dir.chdir 'tmp'
   ldir = "fn_#{lang}"
   if !Dir.exist? ldir
-    cmd = "git clone https://github.com/fnproject/#{ldir}"
```
treeder: why'd you change this?

denismakogon: Personally, I don't like login/passwords, if I can do simpler thing nothing would stop me, that's why relying on SSH authorisation/authentication is much more preferable, because you don't have to keep in mind your passwords.
Even in case if you do this from your work VM you might just configure one more SSH key for this type of things.

denismakogon:
Well, i'd better use same workflow we do for Fn, no one except owners should have write access to a master branch, so having a dedicated branch for the newer feature is much more preferable. The other thing is it might just happen that the newer feature is not in Fn master branch yet, but fn_go patch already there (as well as CLI), you have all parts to

PR #585, opened 2017-12-11
Hybrid plumby

rdallman:
continuing work on #531 - this is ready to review/merge 

hybrid tasks work now for sync & async, with the current async messaging semantics (i.e. no behavior changes). 

* changes `server.Server` configuration to be completely `ServerOption` function based now, this should be much easier to maintain for programs using extensions / us (no signature break-age)
* changes `*hybrid.Client` to implement `agent.DataAccessLayer` with small modification, moved to `api/agent/hybrid`, and plumbs server configuration of the hybrid client
* agent exposes `Enqueue` method in order for runner nodes to use the hybrid client to queue async tasks (not a fan, but it's quick and dirty and solves it; we could push into Submit most likely)
* API nodes no longer configure an `agent.Agent` (and can no longer enqueue async tasks)
* runner nodes no longer default to configuring a `Datastore`, `MessageQueue`, or `LogStore` -- further plumbing is possible to block configuration altogether since they are unused, but not done here
* return full call from async dequeue endpoint, since we're storing a whole
call in the MQ we don't need to worry about caching of app/route [for now]
* fast safe shutdown of dequeue looper in runner / tidying of agent
* nice errors for path not found against /r/, /v1/ or other path not found
* removed some stale TODO in agent
* mq backends are only loud mouths in debug mode now


api/agent/agent_test.go
```diff
@@ -49,7 +49,7 @@ func TestCallConfigurationRequest(t *testing.T) {
 		}, nil,
 	)
 
-	a := New(ds, ds, new(mqs.Mock), AgentTypeFull)
+	a := New(NewCachedDataAccess(NewDirectDataAccess(ds, ds, new(mqs.Mock))))
```
rdallman: for tests we probably don't want the cache, since we may want to e.g. update a route, make sure it doesn't work after update. seems this doesn't break anything for the time being though

rdallman:
updates LGTM, thanks! (I can't approve this since I opened it, please approve if you approve)

rdallman:
thanks!

PR #586, opened 2017-12-12
Update README.md for Grafana example

nigeldeakin:
This corrects an error in the instructions for the Grafana example. The instructions told the user to configure a Grafana data source to connect to Prometheus using `http://prometheus:9090`, where `prometheus` is an alias configured using `--link`. However the instructions specified that the data source should use `direct` mode. This doesn't work, since the connection is then created from the browser, which doesn't know what `prometheus:9090` is. The instructions need to be changed to specify `proxy` mode, in which case the connection is created from Grafana, which knows about the `prometheus` alias.


nigeldeakin:
This change also removes an unnecessary `--link` which linked Grafana directly to fnserver.

PR #590, opened 2017-12-12
fn: exclude timeouts from failed error count

skinowski:


nigeldeakin:
I'm wondering whether the logic should be the same as in `transformTimeout` (called immediately after this), which checks the `isRetriable ` flag as well as the error.

( `transformTimeout` converts the `DeadlineExceeded ` to a `ErrCallTimeoutServerBusy` which then causes `HandleErrorResponse` to set `Retry-After` to `15`. However I don't see how this any effect on my simple test case)

skinowski:
Logic is the same since isRetriable flag is passed from the caller context. In these two if blocks, it's set to true.

Agent code does not retry in these cases except for async code. Is your test case an async case?

api/agent/agent.go
```diff
@@ -229,7 +229,11 @@ func (a *agent) Submit(callI Call) error {
 
 	slot, err := a.getSlot(ctx, call) // find ram available / running
 	if err != nil {
-		a.stats.DequeueAndFail(callI.Model().AppName, callI.Model().Path)
+		if err == context.DeadlineExceeded {
```
denismakogon: Looks good, just one comment, maybe it is worth of having dedicated method because it does look like a duplication, no?

Issue #373, opened 2017-09-28
add FN_DEADLINE to request headers


rdallman:
for any function it would be nice if we could supply the time left for the function to run to the function, so that functions could decide whether to attempt action or not. we can yank this out of the context really easily, small patch (lot of moving parts around headers atm, waiting)

PR #591, opened 2017-12-13
Fn deadline

skinowski:
For #373

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -491,6 +491,7 @@ func (a *agent) launch(ctx context.Context, slots chan<- slot, call *call, tok R
 		go func() {
 			err := a.prepCold(ctx, slots, call, tok)
 			if err != nil {
+				tok.Close()
 				ch <- err
 			}
 		}()
@@ -507,6 +508,12 @@ func (a *agent) launch(ctx context.Context, slots chan<- slot, call *call, tok R
 }
 
 func (a *agent) prepCold(ctx context.Context, slots chan<- slot, call *call, tok ResourceToken) error {
+
+	_, err := call.CheckUpdateDeadline(ctx)
+	if err != nil {
+		return err
+	}
+
 	container := &container{
 		id:      id.New().String(), // XXX we could just let docker generate ids...
 		image:   call.Image,
@@ -523,7 +530,10 @@ func (a *agent) prepCold(ctx context.Context, slots chan<- slot, call *call, tok
 	// about timing out if this takes a while...
 	cookie, err := a.driver.Prepare(ctx, container)
 	if err != nil {
+<<<<<<< HEAD
 		tok.Close()
+=======
+>>>>>>> fn: check and update deadline related headers
 		return err
 	}
 
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"sort"
+	"strconv"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


api/agent/agent.go
```diff
@@ -487,6 +488,16 @@ func (a *agent) launch(ctx context.Context, slots chan<- slot, call *call, tok R
 	ch := make(chan error, 1)
 
 	if !protocol.IsStreamable(protocol.Protocol(call.Format)) {
+
+		deadline, err := common.GetRemainingTime(ctx)
```
rdallman: this seems more like a `panic` to me as it means that somebody removed the timeout surrounding the call invocation (very bad). this env var is build-able in the early, shared stages of `getSlot` and we could do it before we have a handle on a token slot so that we don't have to be worried about leakage.

rdallman: will also cover hot functions, which we should add this to, as well

rdallman: ah, i see, hot is handled in `Start` separately. headers are a mess (i table flipped and gave up for now). wdyt about just setting both the env var and the header, for now, in the start of `getSlot` ? then it's in one place and we can move it around more easily and no handles on slots or anything. i still think it's a panic-able offense to not find a deadline, but it seems like we should just expect one or not set the env var if deadline is not there (instead of error/panic)

api/agent/agent.go
```diff
@@ -487,6 +488,16 @@ func (a *agent) launch(ctx context.Context, slots chan<- slot, call *call, tok R
...
+		if err != nil {
+			tok.Close()
+			ch <- err
+			return ch
+		}
+
+		call.EnvVars["FN_DEADLINE"] = strconv.FormatFloat(deadline.Seconds(), 'f', 0, 64)
```
rdallman: I think this is faulty because there's some latency between the user getting a handle on this and when we're setting it. what do you think about just sticking an RFC3339 time stamp in `FN_DEADLINE`, then there's no races, but I am uncertain how easy it is to use vs seconds across all languages (I suppose that users will do `time.Now().Add(fnDeadlineSeconds * time.Second)` to get the actual timeout in their code, in whatever lang they're using, am less sure of non-go language patterns for setting time based deadlines)

rdallman: hmm, goofed again, this is unix time. that works. i'm gonna get coffee, brain not working

rdallman: nvm, since we're converting to duration it's not unix time. okay, actually coffee now (sorry)

skinowski: RFC3339 sounds good, then I can just set it during instantiation just like 'CreatedAt'

api/common/ctx.go
```diff
@@ -30,3 +32,20 @@ func LoggerWithFields(ctx context.Context, fields logrus.Fields) (context.Contex
 	ctx = WithLogger(ctx, l)
 	return ctx, l
 }
+
+// GetRemainingTime extracts the deadline from the context and returns the
+// remaining time left.
+func GetRemainingTime(ctx context.Context) (*time.Duration, error) {
+	deadline, ok := ctx.Deadline()
+	if !ok {
+		return nil, errors.New("call without deadline provided")
```
denismakogon: is it really possible to get a call without deadline?

rdallman: yea, i think for now it should be impossible but I could see in the future people wanting to fork and remove the timeouts to run really long running tasks (certain users are really stubborn and don't want to shorten their tasks, always happens)

api/agent/call.go
```diff
@@ -264,6 +264,10 @@ func (a *agent) GetCall(opts ...CallOpt) (Call, error) {
 		c.w = c.stderr
 	}
 
+	deadline := strfmt.DateTime(time.Now().Add(time.Duration(c.Call.Timeout) * time.Second)).String()
```
rdallman: there's a slight mismatch with the actual deadline that will be enforced from the context created inside of `agent.Submit` but I think the disparity will be very slight so this is ok (and ++4--0 is sweet)

PR #594, opened 2017-12-13
Documenting hybrid mode options

denismakogon:


rdallman:
do we want to put all hybrid related docs in a separate file? i feel like this will only create confusion for most users that don't want to run hybrid. thanks, we need to doc hybrid stuff (@hhexo has begun, I believe, there is `docs/developers/hybrid.md` to append to possibly)

rdallman:
i foresee 'what's this?' -> 'oh ignore that you don't need it' in slack x10

carimura:
agree with @rdallman on this one... very few customers should be running in Hybrid mode and thus the options are confusing even me.

If we decide to keep them all together there should be a loud "only required in hybrid mode" note on those two options.

denismakogon:
Okay, i'll open issue for extending hybrid doc with all new options.

Issue #593, opened 2017-12-13
Runner panics when starting runner node with wrong/unreachable FN_RUNNER_API_URL


denismakogon:
When FN_RUNNER_API_URL is not reachable runner crashes with panic
```
FN_LOG_LEVEL=ERROR FN_NODE_TYPE=runner FN_PORT=8082 FN_RUNNER_API_URL=http://localhost:8081 ./fnserver
ERRO[0000] error from API server, retrying               error="Get http://localhost:8081/v1/runner/async: dial tcp [::1]:8081: getsockopt: connection refused"

        ______
       / ____/___
      / /_  / __ \
     / __/ / / / /
    /_/   /_/ /_/
        v0.3.235

panic: interface conversion: opentracing.SpanContext is opentracing.noopSpanContext, not zipkintracer.SpanContext

goroutine 54 [running]:
github.com/fnproject/fn/vendor/github.com/openzipkin/zipkin-go-opentracing.(*tracerImpl).startSpanWithOptions(0xc4200de100, 0x4f93691, 0x15, 0xc4203d3900, 0x1, 0x1, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/vendor/github.com/openzipkin/zipkin-go-opentracing/tracer.go:309 +0xb36
github.com/fnproject/fn/vendor/github.com/openzipkin/zipkin-go-opentracing.(*tracerImpl).StartSpan(0xc4200de100, 0x4f93691, 0x15, 0xc4203a2080, 0x1, 0x1, 0xc420496000, 0xc420509b90)
	/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/vendor/github.com/openzipkin/zipkin-go-opentracing/tracer.go:268 +0xdd
github.com/fnproject/fn/api/server.FnTracer.StartSpan(0x55c0900, 0xc4200de100, 0x4f93691, 0x15, 0xc4203a2080, 0x1, 0x1, 0xc4203a2080, 0x4e74900)
	/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/api/server/fntracer.go:23 +0x73
github.com/fnproject/fn/api/server.(*FnTracer).StartSpan(0xc4204f3340, 0x4f93691, 0x15, 0xc4203a2080, 0x1, 0x1, 0x0, 0x1)
	<autogenerated>:1 +0x82
[...]

PR #597, opened 2017-12-13
makes span use init() to run before server.New

rdallman:
from #593 (but not that issue, as described)


rdallman:
@denismakogon

denismakogon:
Problem fixed.

PR #604, opened 2017-12-15
Async hot hdr fix

### LINKED ISSUE #601: async hot request body issue
Async hot request body does not seem to make it into the containers. In async hot case, some http headers are missing.

Compare the following async and sync requests:

`
POST /r/test/foofn HTTP/1.1
Host: localhost:8080
Fn_app_name: test
Fn_call_id: 01C1AGQA7K47WJR00000000000
Fn_format: http
Fn_header_accept: */*
Fn_header_content_length: 8
Fn_header_content_type: application/x-www-form-urlencoded
Fn_header_user_agent: curl/7.54.0
Fn_memory: 128
Fn_method: POST
Fn_param_app: test
Fn_param_route: /foofn
Fn_path: /foofn
Fn_request_url: http://localhost:8080/r/test/foofn
Fn_type: async
`

`
POST /r/test/foofn HTTP/1.1
Host: localhost:8080
Connection: close
Accept: */*
Connection: close
Content-Length: 8
Content-Type: application/x-www-form-urlencoded
Fn_app_name: test
Fn_call_id: 01C1AJ3FMD47WMM00000000000
Fn_format: http
Fn_memory: 128
Fn_method: POST
Fn_param_app: test
Fn_param_route: /foofn
Fn_path: /foofn
Fn_request_url: http://localhost:8080/r/test/foofn
Fn_type: sync
`

skinowski:
Fixes #601

### DIFF
--- a/api/agent/agent_test.go
+++ b/api/agent/agent_test.go
@@ -17,6 +17,33 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+func checkExpectedHeaders(t *testing.T, expectedHeaders http.Header, receivedHeaders http.Header) {
+
+	checkMap := make([]string, 0, len(expectedHeaders))
+	for k, _ := range expectedHeaders {
+		checkMap = append(checkMap, k)
+	}
+
+	for k, vs := range receivedHeaders {
+		for i, v := range expectedHeaders[k] {
+			if i >= len(vs) || vs[i] != v {
+				t.Fatal("header mismatch", k, vs)
+			}
+		}
+
+		for i, _ := range checkMap {
+			if checkMap[i] == k {
+				checkMap = append(checkMap[:i], checkMap[i+1:]...)
+				break
+			}
+		}
+	}
+
+	if len(checkMap) > 0 {
+		t.Fatalf("expected headers not found=%v", checkMap)
+	}
+}
+
 func TestCallConfigurationRequest(t *testing.T) {
 	appName := "myapp"
 	path := "/sleeper"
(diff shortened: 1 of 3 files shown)


api/agent/agent_test.go
```diff
@@ -265,6 +270,10 @@ func TestCallConfigurationModel(t *testing.T) {
 		expectedHeaders.Add(k, v)
 	}
 
+	// These should be here based on payload length and/or fn_header_* original headers
+	expectedHeaders.Add("Content-Type", contentType)
```
denismakogon: `expectedHeaders.Set` is more appropriate, `.Add` will turn header value into a set

api/agent/call.go
```diff
@@ -205,11 +206,24 @@ func FromModel(mCall *models.Call) CallOpt {
...
+		// Hoist original request content type into async request
+		if req.Header.Get("Content-Type") == "" {
+			content_type, ok := c.EnvVars["Fn_header_content_type"]
+			if ok {
+				req.Header.Set("Content-Type", content_type)
+			}
+		}
+
+		// Ensure content-length in async requests for protocol/http DumpRequestTo()
+		if req.Header.Get("Content-Length") == "" {
```
denismakogon: `req.ContentLength` suites better

api/agent/call.go
```diff
@@ -205,11 +206,24 @@ func FromModel(mCall *models.Call) CallOpt {
...
+			req.Header.Set("Content-Length", strconv.FormatInt(int64(len(c.Payload)), 10))
+		}
```
denismakogon: no need to convert, just use `req.ContentLength`

api/agent/call.go
rdallman: since async is the only one using this (and tests) i think this will end up working but it seems brittle still to rewrite blank content length to a value (len(c.Payload)) that is always 0 for sync requests, this would break them most likely (client will think it's an empty body). 

am ok with just patching content length in wait of a more proper header fix, we really need to fix the headers (they shouldn't have `Fn_header_` for routes with "http" format)

api/agent/call.go
skinowski: req.ContentLength does not work. How these headers are synthesized into actual headers depends on how/who processes this request. In this case, DumpRequestTo() in protocol handles this and fails since it only parses the actual Header map.

api/server/runner_test.go
```diff
@@ -262,10 +261,8 @@ func TestRouteRunnerTimeout(t *testing.T) {
 		expectedCode    int
 		expectedHeaders map[string][]string
 	}{
-		// first request with large timeout, we let the docker pull go through...
```
skinowski: Not needed since test.sh now pulls all images in tests.

PR #609, opened 2017-12-19
fn: add tester image

skinowski:
New tester image that uses go-fdk for advanced test scenarios.
Right now, this is an enhanced 'hello/sleeper' rolled into one
that echos the received headers/env to allow writing test cases.


skinowski:
I'm not sure this is the way forward, but I need this image to test scenarios that we currently cannot test such as:

*) hot requests
*) to be able to test a new header (fn_deadline)

etc.

Since it'll be directly used by fn test cases, I thought this should really be in fn repo under ./images.

rdallman:
makes sense to me. seems like we need to add dep for `fdk-go` in glide for the build to work with this.

skinowski:
@rdallman ping.

images/tester/release.sh
```diff
@@ -0,0 +1,5 @@
+set -e
+
+docker build -t fnproject/tester:latest .
```
treeder: Might want to name this something more descriptive. I can see someone deleting this image since it's sounds kind of temporary.

treeder:
I might be missing something, but why does core need fdk-go dependency for this image?

skinowski:
I think I can skip the go test in test.sh to avoid this dep.

rdallman:
i went back and forth on suggesting that, it seems nice to know which version of `fdk-go` was used to build the `tester` image when/if we experience breakage (makes some assumptions about not changing files on disk, since users build the image and not CI atm)

images/tester/release.sh
skinowski: How about fn-test-utils?

treeder:
It would probably best for it to have it's own glide (or preferably dep) files and vendor dir (or preferably build [get deps on build](https://github.com/fnproject/cli/blob/master/Dockerfile)) so we don't just go adding a bunch of dependencies to core that it doesn't use. Just like any other image or function, has it's own dependencies.

rdallman:
> (or preferably build get deps on build)

+1

skinowski:
ping @rdallman @treeder

PR #610, opened 2017-12-19
Adds before/after app get/list. And some bug fixes/cleanup.

treeder:
This looks bigger than it is since there's a search and replace for `server.Database` -> `server.Database()`. 

Main gist is that it adds:

```
+	BeforeAppGet(ctx context.Context, appName string) error
+	AfterAppGet(ctx context.Context, app *models.App) error
+	BeforeAppsList(ctx context.Context, filter *models.AppFilter) error
+	AfterAppsList(ctx context.Context, apps []*models.App) error
```

And some changes to support this. 

Example where these changes are required is: https://github.com/treeder/fn-ext-auth/tree/master/simple

### DIFF
--- a/api/datastore/sql/sql.go
+++ b/api/datastore/sql/sql.go
@@ -362,7 +362,16 @@ func (ds *sqlStore) GetApp(ctx context.Context, name string) (*models.App, error
 // GetApps retrieves an array of apps according to a specific filter.
 func (ds *sqlStore) GetApps(ctx context.Context, filter *models.AppFilter) ([]*models.App, error) {
 	res := []*models.App{}
-	query, args := buildFilterAppQuery(filter)
+	if filter.NameIn != nil && len(filter.NameIn) == 0 { // this basically makes sure it doesn't return ALL apps
+		return res, nil
+	}
+	query, args, err := buildFilterAppQuery(filter)
+	if err != nil {
+		return nil, err
+	}
+	// fmt.Printf("QUERY: %v\n", query)
+	// fmt.Printf("ARGS: %v\n", args)
+	// hmm, should this have DISTINCT in it? shouldn't be possible to have two apps with same name
 	query = ds.db.Rebind(fmt.Sprintf("SELECT DISTINCT name, config FROM apps %s", query))
 	rows, err := ds.db.QueryxContext(ctx, query, args...)
 	if err != nil {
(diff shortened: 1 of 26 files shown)


api/datastore/sql/sql.go
```diff
@@ -362,7 +362,16 @@ func (ds *sqlStore) GetApp(ctx context.Context, name string) (*models.App, error
 // GetApps retrieves an array of apps according to a specific filter.
 func (ds *sqlStore) GetApps(ctx context.Context, filter *models.AppFilter) ([]*models.App, error) {
 	res := []*models.App{}
-	query, args := buildFilterAppQuery(filter)
+	if filter.NameIn != nil && len(filter.NameIn) == 0 { // this basically makes sure it doesn't return ALL apps
```
rdallman: this seems specific to the caller that was worked on in conjunction with this. passing a nil filter.NameIn returns all apps so this is kind of confusing here to change behavior based on the nil-ness of an empty list in either case.

treeder: Ya, I agree, didn't really know a good way to handle this. By default it's nil and would return all apps, as expected. If someone sets it, even if it's empty, then assuming it's supposed to be used.  I could see extension developers misusing this though and causing a gaping security hole (although they'd have to set something that let's the filter know not to return all regardless). Open to suggestions.

api/server/server.go
```diff
@@ -433,6 +434,10 @@ func (s *Server) bindHandlers(ctx context.Context) {
 	})
 }
 
+func (s *Server) Datastore() models.Datastore {
```
rdallman: can't quite figure out the use for this, it just looks out of place now next to the MQ / LogStore / etc. since this will be referenced still on the Server struct and point to the same thing, this just disallows modifying the Datastore on the Server struct? not adding thread safety or anything for initialization. anyway, not sure why this is required in this patch, it seems like we should pick one way or another for everything exported on the server and not mix and match.

treeder: https://github.com/fnproject/fn/pull/610/files#diff-8d588bd150fe648500c167137fcc0b65R42

rdallman: hmm, ok. seems like we could have kept both method and exported `Datastore` (for uniformity in our front-end among members of `Server`), but it's a quick find & replace -- we just need to implement the additional method to satisfy ExtServer

treeder:
I was wondering where you were going with that, first off, you can't have the method and the var with the same name, so we can't just keep it as it was unless we call it something else. Which is fine, I don't care, call it Datastore2 if you want, but it's also not used anywhere that requires it to be exported, so....

treeder:
Maybe we can come back to that if it needs to be exported for something?

treeder:
Even then, it might make sense to have it as a function so extensions could swap it out via the interface, ie: SetDatastore(...).

rdallman:
i see. meh, `GetDatastore` is fine. i'm just anal and everything else is exported on `*Server`, but for what reason i'm less sure of since we provide options for configuring it and it probably shouldn't change after configuration and as long as we provide access to those things (through a method or an exported field), then sure whatever.

rdallman:
anyway, approved.. i can make vanity patches with some tools for however we want to munge stuff. i'm leaning towards lower casing the other things like datastore is now (can do that separate)

PR #611, opened 2017-12-20
fn: tests: adding hot container timeout and huge memory cases

skinowski:
*) switching TestRouteRunnerTimeout to fn-test-utils to handle
    both hot and cold.
*) in server_test added content-length handling as protocol http
    does not create content-length if it is not present.

### DIFF
--- a/api/server/server_test.go
+++ b/api/server/server_test.go
@@ -38,14 +38,15 @@ func createRequest(t *testing.T, method, path string, body io.Reader) *http.Requ
 
 	bodyLen := int64(0)
 
-	// derive content-length since protocol/http does not content-length if they
-	// are not present.
+	// HACK: derive content-length since protocol/http does not add content-length
+	// if it's not present.
 	if body != nil {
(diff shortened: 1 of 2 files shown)


api/server/server_test.go
```diff
@@ -33,30 +34,52 @@ func testServer(ds models.Datastore, mq models.MessageQueue, logDB models.LogSto
...
+	}
+
 	req, err := http.NewRequest(method, "http://127.0.0.1:8080"+path, body)
 	if err != nil {
 		t.Fatalf("Test: Could not create %s request to %s: %v", method, path, err)
 	}
+
+	if body != nil {
+		req.ContentLength = bodyLen
+		req.Header.Set("Content-Length", strconv.FormatInt(bodyLen, 10))
```
rdallman: fwiw I think setting this to `-1` will allow most http clients to read it properly (somewhere down the line the client is jumping the gun to guess that no content length provided -> 0 content length) -- but it's just tests, this is ok too

PR #612, opened 2017-12-20
remove -v on go test

treeder:


treeder:
They'll still show up when there's a failure, but will be hidden otherwise.

PR #613, opened 2017-12-20
Slot mgr fixes

skinowski:
*) new inactivity time out for hot queue, we previously kept hot queues in memory forever.
*) each hot queue now has a hot launcher to monitor and launch hot containers
*) consumers now create a consumer channel with startDequeuer() that can be cancelled via context
*) consumers now ping (signal) hot launcher every 200 msecs until they get a slot
*) tests for slot queue & mgr

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -2,7 +2,6 @@ package agent
 
 import (
 	"context"
-	"errors"
 	"io"
 	"net/http"
 	"sync"
@@ -188,7 +187,7 @@ func (a *agent) Submit(callI Call) error {
 
 	select {
 	case <-a.shutdown:
-		return errors.New("agent shut down")
+		return models.ErrCallTimeoutServerBusy
 	default:
 	}
 
@@ -376,7 +375,7 @@ func (a *agent) waitHot(ctx context.Context, call *call) (Slot, error) {
 		case <-time.After(time.Duration(200) * time.Millisecond):
 			// ping dequeuer again
 		case <-a.shutdown: // server shutdown
-			return nil, errors.New("agent shut down")
+			return nil, models.ErrCallTimeoutServerBusy
 		}
 	}
 }
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -272,7 +272,7 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
 		call.slots, isNew = a.slotMgr.getHotSlotQueue(call)
 		if isNew {
 			a.wg.Add(1)
-			go a.launchHot(ctx, call)
+			go a.hotLauncher(ctx, call)
 		}
 		start := time.Now()
 
@@ -286,62 +286,53 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
 	return a.launchCold(ctx, call)
 }
 
-// launchHot is spawned in a go routine for each slot queue to monitor and launch hot
-// containers if needed. slots.pingDequeuer() activates the signaller channel.
-func (a *agent) launchHot(ctx context.Context, call *call) {
-
-	defer a.slotMgr.destroySlotQueue(call.slots)
-	defer a.wg.Done()
+// hotLauncher is spawned in a go routine for each slot queue to monitor stats and launch hot
+// containers if needed. Upon shutdown or activity timeout, hotLauncher exits and during exit,
+// it destroys the slot queue.
+func (a *agent) hotLauncher(ctx context.Context, call *call) {
 
 	// Let use 60 minutes or 2 * IdleTimeout as hot queue idle timeout, pick
-	// whichever is longer. If in this time, there's no activity in launchHot
-	// loop, then we destroy the hot queue.
+	// whichever is longer. If in this time, there's no activity, then
+	// we destroy the hot queue.
 	timeout := time.Duration(60) * time.Minute
 	idleTimeout := time.Duration(call.IdleTimeout) * time.Second * 2
 	if timeout < idleTimeout {
 		timeout = idleTimeout
 	}
 
-	logrus.WithField("timeout", timeout).Debug("Starting a new hot queue with launcher")
+	logger := logrus.WithFields(logrus.Fields{
+		"app":              call.AppName,
+		"route":            call.Path,
+		"image":            call.Image,
+		"memory":           call.Memory,
+		"format":           call.Format,
+		"idle_timeout":     call.IdleTimeout,
+		"launcher_timeout": timeout})
 
+	logger.Info("Hot function launcher starting")
 	isAsync := call.Type == models.TypeAsync
-	timeoutFired := false
 
 loop:
 	for {
-		if !timeoutFired {
-			select {
-			case <-a.shutdown: // server shutdown
-				return
-			case <-time.After(timeout):
-				timeoutFired = true
-			case _, ok := <-call.slots.signaller:
-				if !ok {
-					return
-				}
+		select {
+		case <-a.shutdown: // server shutdown
+			break loop
+		case <-time.After(timeout):
+			if call.slots.isIdle() {
+				logger.Info("Hot function launcher timed out")
+				break loop
 			}
+		case <-call.slots.signaller:
 		}
 
-		// Check/evaluate if we need to launch a new hot container
 		isNeeded, stats := call.slots.isNewContainerNeeded()
-		logrus.WithField("stats", stats).Debug("checking hot container launch ", isNeeded)
-
-		// if inactivity timeout fired, then if there's no waiters/runners/starters, we can return
-		if timeoutFired {
-			timeoutFired = false
-
-			party := stats.states[SlotQueueWaiter] + stats.states[SlotQueueStarter] + stats.states[SlotQueueRunner]
-			if party == 0 {
-				logrus.WithField("stats", stats).Debug("hot container slotQueue timeout ")
-				return
-			}
-		}
-
+		logger.WithField("stats", stats).Debug("Hot function launcher stats")
 		if !isNeeded {
 			continue loop
 		}
 
 		resourceCtx, cancel := context.WithCancel(context.Background())
+		logger.WithField("stats", stats).Info("Hot function launcher starting hot container")
 
 		select {
 		case tok, isOpen := <-a.resources.GetResourceToken(resourceCtx, call.Memory, isAsync):
@@ -349,15 +340,24 @@ loop:
 			if isOpen {
 				go a.runHot(ctx, call, tok)
 			} else {
+				// this means the resource was impossible to reserve (eg. memory size we can never satisfy)
 				call.slots.queueSlot(&hotSlot{done: make(chan struct{}), err: models.ErrCallTimeoutServerBusy})
 			}
 		case <-time.After(timeout):
-			timeoutFired = true
+			cancel()
+			if call.slots.isIdle() {
+				logger.Info("Hot function launcher timed out")
+				break loop
+			}
 		case <-a.shutdown: // server shutdown
+			cancel()
+			break loop
 		}
-
-		cancel()
 	}
+
+	logger.Info("Hot function launcher terminating")
+	a.slotMgr.destroySlotQueue(call.slots)
+	a.wg.Done()
 }
 
 // waitHot pings and waits for a hot container from the slot queue
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -266,12 +266,19 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
 
 	isHot := protocol.IsStreamable(protocol.Protocol(call.Format))
 	if isHot {
-		// For hot requests, we use a long lived slot queue, which we use to manage hot containers
-		call.slots = a.slotMgr.getHotSlotQueue(call)
+
+		// For hot requests, we use a long lived (60 minutes) slot queue, which we use to manage hot containers
+		call.slots = a.slotMgr.getHotSlotQueue(call, func(sig chan bool, slots *slotQueue) {
+			a.wg.Add(1)
+			hotQueueTimeout := time.Duration(60) * time.Minute
+			a.launchHot(sig, ctx, call, slots, hotQueueTimeout)
+			a.slotMgr.destroySlotQueue(slots)
+			a.wg.Done()
+		})
 		start := time.Now()
 
 		call.slots.enterState(SlotQueueWaiter)
-		s, err := a.launchHot(ctx, call)
+		s, err := a.waitHot(ctx, call)
 		call.slots.exitStateWithLatency(SlotQueueWaiter, uint64(time.Now().Sub(start).Seconds()*1000))
 
 		return s, err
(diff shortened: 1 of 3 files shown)


api/agent/agent.go
```diff
@@ -278,55 +284,95 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
...
 
-// launchHot checks with slot queue to see if a new container needs to be launched and waits
-// for available slots in the queue for hot request execution.
-func (a *agent) launchHot(ctx context.Context, call *call) (Slot, error) {
+// hotLauncher is spawned in a go routine for each slot queue to monitor stats and launch hot
+// containers if needed. Upon shutdown or activity timeout, hotLauncher exits and during exit,
+// it destroys the slot queue.
+func (a *agent) hotLauncher(ctx context.Context, call *call) {
+
+	// Let use 60 minutes or 2 * IdleTimeout as hot queue idle timeout, pick
```
denismakogon: IdleTimeout is quite tricky, it can be zero, so maybe wait for two default idle timeouts (which is 1 minute).
I think that 60 minutes is quite a big timeout, so there should be the ability to configure it from outside, i.e. through Fn opts.

api/agent/agent.go
```diff
@@ -278,55 +284,95 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
...
+	// whichever is longer. If in this time, there's no activity, then
+	// we destroy the hot queue.
+	timeout := time.Duration(60) * time.Minute
```
denismakogon: `time.Minute * 60` also works fine

api/agent/agent.go
```diff
@@ -278,55 +284,95 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
...
+	idleTimeout := time.Duration(call.IdleTimeout) * time.Second * 2
```
denismakogon: Okay, as mentioned before, it could be zero (which turns hot into the cold), so waiting for 1 hour for getting out of the queue is pretty long.

api/agent/agent.go
```diff
@@ -278,55 +284,95 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
...
+	if timeout < idleTimeout {
+		timeout = idleTimeout
+	}
+
+	logger := logrus.WithFields(logrus.Fields{
```
denismakogon: better use context-bound logger https://github.com/fnproject/fn/blob/0b594cfeff7576fc685fad88823ca552766223e8/api/common/ctx.go#L17

api/agent/agent.go
```diff
@@ -278,55 +284,95 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
...
+			}
+		case <-a.shutdown: // server shutdown
+			cancel()
+			break loop
+		}
+	}
+
+	logger.Info("Hot function launcher terminating")
+	a.slotMgr.destroySlotQueue(call.slots)
+	a.wg.Done()
```
denismakogon: shouldn't we call this with `defer`?

api/agent/agent.go
skinowski: 60 mins is good. Note that this is the timeout we cache hot container slot queue *and* the relevant stats and we use those stats to decide to launch new containers or not.

api/agent/agent.go
skinowski: I fail to see why defer is preferable here. Code layout is clear and there's one exit path out of the function.

skinowski: Note that defer is also slower.

api/agent/agent.go
skinowski: If we use 1-2 minutes, by the time we end up with useful stats (start/run/wait) for the hot queue, it'll expire.

denismakogon: In case of sync mode, would that mean that user would get `Service busy` and will silently die for async?

api/agent/agent.go
denismakogon: It really looks like there's only one (i think) exit path then yes, defer is redundant

api/agent/agent.go
skinowski: I could, but this is similar to runHot() where we do not use the call context (which will expire with timeout) except for spans. Should I do this? As this ctx will expire shortly, while the launcher will continue running for extended period. So, we'd be running the logger under a ctx that is expired. The code here is a bit confusing because it still uses 'ctx' and passes that to runHot(), which only uses it to start a span. I'm open to ideas here.

skinowski: runHot() for instance uses a new logrus instead of using ctx of the original call.

api/agent/agent.go
denismakogon: So, i'm not sure for how slow defer would be here, but i've seen reports that defer with anonymous function works faster:
```go
defer func(){ a.wg.Done() }()
```

api/agent/agent.go
skinowski: There's a general problem of hardcoded constants in fn. See resource.go, call.go, async.go. I could take all of that out to opts/cfg in a separate PR.

denismakogon: It's a good idea, but let's hear other voices on this.

api/agent/agent.go
denismakogon: we love spans, we love tracing, so why not?

api/agent/agent.go
skinowski: If we destroy the queues frequently, this erases those stats. In that case, isNewContainerNeeded() in slots.go becomes a less effective. The cost of making a bad decision is usually a timeout. It would be good to store the stats forever. But also need to avoid bloating the memory with unused hot queues. I think 60 mins is ok. But I agree it should be tune-able somewhere along with other (and more critical) constants.

api/agent/agent.go
skinowski: @denismakogon you mean like this below, right?

```
	logger := common.Logger(ctx).WithFields(logrus.Fields{
		"app":              call.AppName,
		"route":            call.Path,
		"image":            call.Image,
		"memory":           call.Memory,
		"format":           call.Format,
		"idle_timeout":     call.IdleTimeout,
		"launcher_timeout": timeout})
```

denismakogon: Yeah, it still looks ugly, but who really cares about the beauty (maybe one more logger method to hide all of this and build logger from ctx and call?).

skinowski: Let me see what I can do here. There's some common log patterns between here and in runHot(), I'll hoist them into a common helper function.

api/agent/agent.go
skinowski: IdleTimeout cannot be zero. Attempt to set it to zero will emit ErrRoutesInvalidTimeout. (see models/route.go)

api/agent/agent.go
rdallman: this logic looks fine wrt timing. there's more value in picking sane values for your users than having a bajillion fricking flags to configure every little piece of the server that they don't need to know details of.

rdallman: fwiw idle_timeout will be 30s 99% of the time so this will be 1 minute, as highlighted (perhaps clamp around 60 minutes?)

rdallman: nvm, it is. coffee

api/agent/agent.go
```diff
@@ -264,12 +264,18 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
 
 	isHot := protocol.IsStreamable(protocol.Protocol(call.Format))
 	if isHot {
+
 		// For hot requests, we use a long lived slot queue, which we use to manage hot containers
-		call.slots = a.slotMgr.getHotSlotQueue(call)
+		var isNew bool
+		call.slots, isNew = a.slotMgr.getHotSlotQueue(call)
+		if isNew {
+			a.wg.Add(1)
```
rdallman: this seems like it will be brittle to maintain the corresponding `wg.Done()` with this from here. we're already in Submit, which will handle holding up shutdown until after the job is done, which is kind of all we care about except for removing all containers on shutdown. it seems like it would be easier to just add one of these in `runHot`, which as long as we keep things simple should always be called from within the span of a `Submit` so we don't have to worry about races all that much.

rdallman:
>  *) each hot queue now has a hot launcher to monitor and launch hot containers

initial thinking is that this should be on-demand driven vs. polling (haven't gotten to details yet)

api/agent/agent.go
```diff
@@ -278,55 +284,88 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
...
+			cancel()
+			if isOpen {
+				go a.runHot(ctx, call, tok)
+			} else {
+				// this means the resource was impossible to reserve (eg. memory size we can never satisfy)
+				call.slots.queueSlot(&hotSlot{done: make(chan struct{}), err: models.ErrCallTimeoutServerBusy})
 			}
+		case <-time.After(timeout):
+			cancel()
+			if call.slots.isIdle() {
```
rdallman: is there a race here where a call could come in where this is shutting down and this thread wouldn't get re-launched?

rdallman:
fwiw we had a structure similar to this (one thread managing each pool of hot containers per route, doling them out) originally and there were a couple pieces that were hard to manage. the main one being that when calls are coming in we tend towards trying to ask for a lot of slots, but generally after they start launching the backlog clears pretty quickly and we don't want to launch anymore (and, mainly, we want to cancel threads asking for slots). this reduces it a good bit by making one thread per route, but I'm not sure it's closed if we have a very mixed workload of routes. 

we have a mix of a lot of channels going on now to get a slot out: `trigger, output, signaller, closer` and i'm a bit concerned this will be hard to evolve over time. i'm optimistic we can reduce most of the channels to just the slot channel, even if there's a thread sandwiched in the middle launching hot containers to hand out slots to that channel. 

i'll drink some egg nog and give this another go later with some bourbon in the system.

api/agent/agent.go
skinowski: nice catch, I can fix this up by increasing the scope of the 'waiting' state in:

https://github.com/skinowski/fn/blob/f92901e673af9c3b0d9c271fa46fee7169afc22e/api/agent/agent.go#L277

I'll move this before the lookup to cover it. If it times out when someone is about lookup the hot queue map, the slot queue won't be idle.

api/agent/agent.go
skinowski: ok, I'll remove this one, since runHot() really needs to have wg.Done().

skinowski:
@rdallman ping.

api/agent/agent.go
```diff
@@ -266,88 +266,114 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
...
-			case tok, isOpen := <-a.resources.GetResourceToken(ctxToken, call.Memory, isAsync):
-				tokenCancel()
-				if !isOpen {
-					return nil, models.ErrCallTimeoutServerBusy
-				}
+		resourceCtx, cancel := context.WithCancel(context.Background())
+		logger.WithField("stats", stats).Info("Hot function launcher starting hot container")
 
+		select {
+		case tok, isOpen := <-a.resources.GetResourceToken(resourceCtx, call.Memory, isAsync):
```
rdallman: not a deal breaker [at all], but this does have the noted issue (from adding the sandwich) that we can end up launching an extra container here if this takes a while (i.e. memory full) and the thread that needed a slot got one a while back / timed out.

api/agent/agent.go
```diff
@@ -266,88 +266,114 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
...
+			cancel()
+			if isOpen {
 				a.wg.Add(1)
 				go a.runHot(ctx, call, tok)
```
rdallman: while we're in here, guess i missed this in the last one, a minor nit: 

```go
a.wg.Add(1)
go func() {
  a.runHot(ctx, call, tok)
  a.wg.Done()
}()
```

if we keep the wg.Done() calls close to the wg.Add() calls it makes it much easier to keep track of them when/if we run into races / shuffle code around.

api/agent/slots.go
```diff
@@ -135,18 +100,23 @@ func (a *slotQueue) ejectSlot(s *slotToken) bool {
...
-	if isFound {
-		a.slots = a.slots[1:]
+	if foundIdx != -1 {
+		newArr := make([]*slotToken, 0, len(a.slots))
+		for idx, val := range a.slots {
+			if idx != foundIdx {
+				newArr = append(newArr, val)
+			}
+		}
+		a.slots = newArr
```
rdallman: handy page to bookmark https://github.com/golang/go/wiki/SliceTricks

```go
for i := 0; i < len(a.slots); i++ {
  val := a.slots[i]
  if val.id == s.id {
    s.slots = append(s.slots[:i], s.slots[i+1:]...)
    break
  }
}
```

in place append will re-use the underlying array instead of allocating a new one, copy is faster than iteration w/ a branch and one loop (under a lock)

api/agent/slots.go
```diff
@@ -158,19 +128,92 @@ func (a *slotQueue) ejectSlot(s *slotToken) bool {
 
 func (a *slotQueue) destroySlotQueue() {
 	doSignal := false
+	purgePool := make([]*slotToken, 0)
```
rdallman: nit: `var purgePool []*slotToken` won't allocate, and `range` statement will handle the `nil` value fine (and of course, the assignment statement will also still work) -- this method is called infrequently, not a big deal, just leaving a note

api/agent/agent.go
```diff
@@ -266,88 +266,114 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
...
+		}
 
-			// wait on token/slot/timeout whichever comes first
-			select {
-			case tok, isOpen := <-a.resources.GetResourceToken(ctxToken, call.Memory, isAsync):
-				tokenCancel()
-				if !isOpen {
-					return nil, models.ErrCallTimeoutServerBusy
-				}
+		resourceCtx, cancel := context.WithCancel(context.Background())
```
rdallman: could remove this and just use `a.resources.GetResourceToken(context.Background(), call.Memory, isAsync)`, which removes the 3 `cancel()` func calls in each select case below. 

it would be nice to keep some context's logger and span info, adding either of these won't require a `cancel` func either, though. this thread seems like it could add app / route / image to the logger and have its own span that is a `FollowsFrom` (terminology: https://github.com/opentracing/specification/blob/master/specification.md ) of the enclosing context's span.

api/agent/agent.go
```diff
@@ -266,88 +266,114 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
...
+	if timeout < idleTimeout {
+		timeout = idleTimeout
+	}
 
+	logger := common.Logger(ctx)
+	logger.WithField("launcher_timeout", timeout).Info("Hot function launcher starting")
 	isAsync := call.Type == models.TypeAsync
 
-launchLoop:
+	defer call.slots.destroySlotQueue()
```
rdallman: this works because the following loop invariant is(?):

* server shutdown (need to pin down?)
* destroySlotQueue will have removed this slot queue from the slot queue bucket, 

correct? we could just let the reference die but we also want to signal anyone waiting on the cond ? is that possible since we've confirmed that this channel is idle and removed the ability to get a reference to it [safely]?

for server shutdown, we want to wait for all tasks to finish and for all containers to shut down. i think at present we're in a weird situation where we can have calls come in, they wait for a slot, but the server got shutdown so no containers are going to launch and no slots are going to be handed out, so we just wait for the calls to timeout. seems like maybe in waitHot we should check server shutdown and handle this with an immediate timeout (as opposed to waiting, potentially 60 minutes for async tasks). we can punt on fixing the shutdown in this patch, if the above is true we should probably try to close it though.

api/agent/slots.go
```diff
@@ -158,19 +128,92 @@ func (a *slotQueue) ejectSlot(s *slotToken) bool {
...
+	loop:
+		for {
+			a.cond.L.Lock()
+			for len(a.slots) <= 0 && !a.isClosed && (ctx.Err() == nil) {
+				a.cond.Wait()
+			}
+
+			if a.isClosed || ctx.Err() != nil {
+				a.cond.L.Unlock()
+				break loop
```
rdallman: return & no loop label :)

api/agent/agent.go
skinowski: we could perhaps do an additional check on isNewContainerNeeded() here after acquiring the resource.

api/agent/agent.go
rdallman: each hot container launch would also be a `FollowsFrom` span child of the span for this thread (would be nice to track these so we can watch causal chain of incoming calls vs container launches)

api/agent/agent.go
skinowski: 1) in waitHot() we should check shutdown to avoid that timeout issue
2) I think we can get rid of destroySlotQueue and related close logic in it, with idle check, this is now dead code, nobody should be waiting.

skinowski: With shutdown() case, destroySlotQueue() does wake up other waiters, but I need to add 'close' channel in startDequeuer() again to make this work properly. With that, waitHot() will wake up and find that the channel is now closed. Which addresses the shutdown/timeout issue....

skinowski: So, I have two options I think:
1) wait for shutdown in waitHot and try to simplify destroySlotQueue since now we might need to wake up waiters.
2) add closing of channel in startDequeuer() to let waitHot() detect the shutdown case.

'1' seems better, which I'm looking into...

api/agent/agent.go
skinowski: Could you explain why we don't need cancel() in GetResourceToken() context? For two cases, timeout and shutdown, don't we need to cancel that go routine in GetResourceToken?

skinowski:
So, this slightly got less complicated since the removal of destroySlotQueue(), which no longer needs closer channel and isClosed bool. Dequeuing hotWait() waiters no longer needs this notification/wake up since they now check a.shutdown themselves.

api/agent/agent.go
rdallman: d'oh, that's right.

api/agent/agent.go
```diff
@@ -374,6 +375,8 @@ func (a *agent) waitHot(ctx context.Context, call *call) (Slot, error) {
 			return nil, ctx.Err()
 		case <-time.After(time.Duration(200) * time.Millisecond):
 			// ping dequeuer again
+		case <-a.shutdown: // server shutdown
+			return nil, errors.New("agent shut down")
```
rdallman: since we're waiting on a slot and haven't run the call, we can return some kind of timeout error here so that fnlb might try it elsewhere (fnlb does not do this currently, but may eventually) -- would make updating an fn cluster cause less errors (slightly, in this case)

rdallman:
updates LGTM too. one bugaboo about returning ctx.DeadlineCanceled or something in waitHot on shutdown

skinowski:
related to #274

PR #614, opened 2017-12-20
Timestamps on apps / routes

### LINKED ISSUE #367: add `created_at` and `updated_at` to routes & apps
need to store in db + add to api, useful for auditing stuff.

we should probably implement db migrations baked in so that we can make changes like this without breaking people's dbs, too.

rdallman:
closes #367 

this finishes the time stamp saga for apps & routes, with created_at and updated_at on each one. should be pretty straight forward, there's some minor changes to tests / route init logic.

there are tests to ensure that these things are getting set.

### DIFF
--- a/api/models/route.go
+++ b/api/models/route.go
@@ -152,38 +152,51 @@ func (r *Route) Clone() *Route {
 	return clone
 }
 
-func (r *Route) up() { r.UpdatedAt = strfmt.DateTime(time.Now()) }
+func (r1 *Route) Equals(r2 *Route) bool {
+	// start off equal, check equivalence of each field.
+	// the RHS of && won't eval if eq==false so config/headers checking is lazy
+
+	eq := true
+	eq = eq && r1.AppName == r2.AppName
+	eq = eq && r1.Path == r2.Path
+	eq = eq && r1.Image == r2.Image
+	eq = eq && r1.Memory == r2.Memory
+	eq = eq && r1.Headers.Equals(r2.Headers)
+	eq = eq && r1.Type == r2.Type
+	eq = eq && r1.Format == r2.Format
+	eq = eq && r1.Timeout == r2.Timeout
+	eq = eq && r1.IdleTimeout == r2.IdleTimeout
+	eq = eq && r1.Config.Equals(r2.Config)
+	eq = eq && r1.CreatedAt == r2.CreatedAt
+	eq = eq && r2.UpdatedAt == r2.UpdatedAt
+	return eq
+}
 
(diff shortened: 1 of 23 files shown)


rdallman:
Meh, i'll fix those `reflect.DeepEquals` calls breaking the tests. the new tests pass at least

api/models/route.go
```diff
@@ -121,35 +132,58 @@ func (r *Route) Validate() error {
...
+			// TODO technically, we need to deep copy this slice...
+			clone.Headers[k] = v
+		}
+	}
+	return clone
 }
 
-// Update updates fields in r with non-zero field values from new.
-// 0-length slice Header values, and empty-string Config values trigger removal of map entry.
+func (r *Route) up() { r.UpdatedAt = strfmt.DateTime(time.Now()) }
```
denismakogon: Maybe instead of calling this for each modified field, we'd better find the simpler way to identify if route changed (hash?) and only then update the timestamp. This will give us a chance to call `up` with `defer` in `Update` only once.

denismakogon:
One thing is missing here is the swagger doc.

api/models/route.go
rdallman: it's kind of contained, I couldn't really think of a better way without checking full equality. update resource always ends up being pretty gnarly, this doesn't seem all that bad really as it's just function calls (that will be inlined) so there's not really room for logic errors with updating the timestamp itself. using a hash would give us the benefit of knowing if changes occurred not just that fields were overwritten (but possibly not changed), however the downside is we'd be generating a hash inside of the db transaction which is somewhat expensive but this isn't really a hot path either (and, sadly, agent code is doing this but not with every field so we need a different hash here). using a `Route.Equals(Route)` would be ~expensive as we have to copy the maps before updating them and can't just do a shallow copy (see `.Clone()`), but I think it would be cheaper than generating a hash (cpu cycles, and perhaps memory) since we can precisely allocate and the code would be easier to follow. we would have to generate 2 hashes (which requires sorting the maps, which requires allocating 4 slices with all the key values for both headers and config). but calling `up()` multiple times is about the last of my worries. checking full equality seems useful and it seems like we're going to need it for the tests instead of reflect, so I'm gonna go ahead and try adding `.Equals` to routes and apps.

rdallman:
> One thing is missing here is the swagger doc.

blanked on this, thanks, will do

rdallman:
updates:

* added `y.Equals(x)` methods for `App`, `Route`, `Headers`, `Config` & uses them (looks nice imo)
* kills reflect usage (yay)
* adds swagger
* updated w/ master

@denismakogon

PR #615, opened 2017-12-20
Ship call logs to the user as text/plain instead of JSON

denismakogon:
Fn API changed.
Swagger updated.
Certain API tests were disabled to make CI happy.

### DIFF
--- a/api/server/call_logs.go
+++ b/api/server/call_logs.go
@@ -23,6 +23,5 @@ func (s *Server) handleCallLogGet(c *gin.Context) {
 	var b bytes.Buffer
 	b.ReadFrom(logReader)
 
-	c.Header("Content-Type", "text/plain")
-	c.JSON(http.StatusOK, b.String())
+	c.String(http.StatusOK, b.String())
 }
--- a/api/server/call_logs.go
+++ b/api/server/call_logs.go
@@ -5,8 +5,15 @@ import (
 	"net/http"
 
 	"github.com/fnproject/fn/api"
+	"github.com/fnproject/fn/api/models"
 	"github.com/gin-gonic/gin"
 )
+
+// note: for backward compatibility, will go away later
+type callLogResponse struct {
+	Message string          `json:"message"`
+	Log     *models.CallLog `json:"log"`
+}
 
 func (s *Server) handleCallLogGet(c *gin.Context) {
 	ctx := c.Request.Context()
@@ -23,5 +30,16 @@ func (s *Server) handleCallLogGet(c *gin.Context) {
 	var b bytes.Buffer
 	b.ReadFrom(logReader)
 
-	c.String(http.StatusOK, b.String())
+	switch c.Request.Header.Get("Accept") {
+	case "application/json":
+		callObj := models.CallLog{
+			CallID:  callID,
+			AppName: appName,
+			Log:     b.String(),
+		}
+
+		c.JSON(http.StatusOK, callLogResponse{"Successfully loaded log", &callObj})
+	default:
+		c.String(http.StatusOK, b.String())
+	}
 }
--- a/api/server/call_logs.go
+++ b/api/server/call_logs.go
@@ -30,16 +30,19 @@ func (s *Server) handleCallLogGet(c *gin.Context) {
 	var b bytes.Buffer
 	b.ReadFrom(logReader)
 
-	switch c.Request.Header.Get("Accept") {
-	case "application/json":
-		callObj := models.CallLog{
-			CallID:  callID,
-			AppName: appName,
-			Log:     b.String(),
-		}
+	mimeTypes, _ := c.Request.Header["Accept"]
 
-		c.JSON(http.StatusOK, callLogResponse{"Successfully loaded log", &callObj})
-	default:
-		c.String(http.StatusOK, b.String())
+	for _, mimeType := range mimeTypes {
+		switch mimeType {
+		case "text/plain":
+			c.String(http.StatusOK, b.String())
+		default:
+			c.JSON(http.StatusOK, callLogResponse{"Successfully loaded log",
+				&models.CallLog{
+					CallID:  callID,
+					AppName: appName,
+					Log:     b.String(),
+				}})
+		}
 	}
 }
--- a/api/server/call_logs.go
+++ b/api/server/call_logs.go
@@ -5,7 +5,6 @@ import (
 	"net/http"
 
 	"github.com/fnproject/fn/api"
-	"github.com/fnproject/fn/api/models"
 	"github.com/gin-gonic/gin"
 )
 
@@ -21,18 +20,9 @@ func (s *Server) handleCallLogGet(c *gin.Context) {
 		return
 	}
 
-	// TODO this API needs to change to text/plain / gzip anyway, punting
-	// optimization, but we can write this direct to the wire, too... seems like
-	// we should write some kind of writev json thing for go since we keep
-	// hitting this :(
 	var b bytes.Buffer
 	b.ReadFrom(logReader)
 
-	callObj := models.CallLog{
-		CallID:  callID,
-		AppName: appName,
-		Log:     b.String(),
-	}
-
-	c.JSON(http.StatusOK, callLogResponse{"Successfully loaded log", &callObj})
+	c.Header("Content-Type", "text/plain")
+	c.JSON(http.StatusOK, b.String())
 }
--- a/api/server/call_logs.go
+++ b/api/server/call_logs.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"bytes"
+	"io"
 	"net/http"
 
 	"github.com/fnproject/fn/api"
@@ -27,16 +28,16 @@ func (s *Server) handleCallLogGet(c *gin.Context) {
 		return
 	}
 
-	var b bytes.Buffer
-	b.ReadFrom(logReader)
-
 	mimeTypes, _ := c.Request.Header["Accept"]
 
 	for _, mimeType := range mimeTypes {
 		switch mimeType {
 		case "text/plain":
-			c.String(http.StatusOK, b.String())
+			io.Copy(c.Writer, logReader)
+			c.Status(http.StatusOK)
 		default:
+			var b bytes.Buffer
+			b.ReadFrom(logReader)
 			c.JSON(http.StatusOK, callLogResponse{"Successfully loaded log",
 				&models.CallLog{
 					CallID:  callID,
--- a/api/server/call_logs.go
+++ b/api/server/call_logs.go
@@ -63,7 +63,8 @@ func (s *Server) handleCallLogGet(c *gin.Context) {
 			return
 		}
 	}
+
 	// if we've reached this point it means that Fn didn't recognize Accepted content type
-	WriteError(c, c.Writer, http.StatusNotAcceptable,
-		errors.New("unable to respond within acceptable response content types"))
+	handleErrorResponse(c, models.NewAPIError(http.StatusNotAcceptable,
+		errors.New("unable to respond within acceptable response content types")))
 }
--- a/api/server/call_logs.go
+++ b/api/server/call_logs.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"net/http"
 
+	"errors"
 	"github.com/fnproject/fn/api"
 	"github.com/fnproject/fn/api/models"
 	"github.com/gin-gonic/gin"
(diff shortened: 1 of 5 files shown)


api/server/call_logs.go
```diff
@@ -21,18 +20,9 @@ func (s *Server) handleCallLogGet(c *gin.Context) {
...
 
-	callObj := models.CallLog{
-		CallID:  callID,
-		AppName: appName,
-		Log:     b.String(),
-	}
-
-	c.JSON(http.StatusOK, callLogResponse{"Successfully loaded log", &callObj})
+	c.Header("Content-Type", "text/plain")
+	c.JSON(http.StatusOK, b.String())
```
skinowski: From docs, it sounds like c.JSON() is going to set the content-type to json again?

[https://godoc.org/github.com/gin-gonic/gin#Context.JSON](https://godoc.org/github.com/gin-gonic/gin#Context.JSON)

maybe c.String() or c.Data() would work better?

denismakogon: Yeah, but it’s funny, line above that set content type to text worked, so the response content type at client side was text/plain, but not json

api/datastore/internal/datastoreutil/metrics.go
```diff
@@ -104,7 +106,24 @@ func (m *metricds) GetCalls(ctx context.Context, filter *models.CallFilter) ([]*
...
+	err = gzippedLog.Flush()
+	if err != nil {
+		return err
+	}
+	newCallLog, err := gzip.NewReader(&b)
+	if err != nil {
+		return err
+	}
+
+	return m.ds.InsertLog(ctx, appName, callID, newCallLog)
```
rdallman: let's punt on this (didn't get a reply from you). we should discuss. we want to thread gzip through the whole system really, it's not this straightforward. then we aren't gathering un-compressed bytes (in memory/files) and then compressing them just on upload. let's just fix plain text api here

rdallman: metric ds is kind of the wrong place to put this, too.

rdallman: anyway i ran into complications trying that (hence my abandonment). but now i know what was wrong (there was another since fixed bug in logs)

denismakogon: I'm not sure i understand your first comment.

So, if that is not metrics ds, so gzip compressing should be just in every logstore implementation, which doesn't seem reasonable.

rdallman: metric ds is not the only option. i don't think we need any implementation. i can dig up my patch and try to explain more concretely, am happy to push forward with the plaintext api meanwhile (based on Accept headers)

denismakogon: there's one problem - swagger, not sure that they support multiple kinds of responses in different formation: same response code, but the body is different.

denismakogon: i'd like to deal with gzip thing in first place, only then plaintext api thing.

rdallman: we can change swagger to plaintext, that's fine, and new clients will use it, but we need backwards compat for old clients using json still. the gzip isn't pressing, really, and there's code that has to be written for compatibility so it's not perfectly straightforward

denismakogon: Done, API remains compatible with older clients. So, i'd like to know your thoughts and how you see gzip'ing function's logs.

As far as i understand there's nothing much code needed to do actual compressing, right? But the bigger question to me is where it happens in order to be the same for all stores (sql and minion at this moment, obviously it should happen just right before going to actual implementations).

api/server/call_logs.go
```diff
@@ -21,18 +27,19 @@ func (s *Server) handleCallLogGet(c *gin.Context) {
...
-	// hitting this :(
 	var b bytes.Buffer
 	b.ReadFrom(logReader)
 
-	callObj := models.CallLog{
-		CallID:  callID,
-		AppName: appName,
-		Log:     b.String(),
+	switch c.Request.Header.Get("Accept") {
+	case "application/json":
```
rdallman: this only gets the first value from Accept, which may be a list, and we likely want to prefer plaintext

denismakogon: I do aware that Accept can be a list, but fn_go sends exactly one item (which is `application/json`). So the best option i see here is to loop over MIME types.

rdallman: ok, for the old endpoint that makes sense. do the updated swagger clients request `text/plain` properly for this endpoint?

denismakogon:
Okay, the swagger doc is updated. Will push new client(s) once swagger doc will be at the master.

api/server/call_logs.go
```diff
@@ -21,18 +27,22 @@ func (s *Server) handleCallLogGet(c *gin.Context) {
...
-	callObj := models.CallLog{
-		CallID:  callID,
-		AppName: appName,
-		Log:     b.String(),
+	mimeTypes, _ := c.Request.Header["Accept"]
+
+	for _, mimeType := range mimeTypes {
+		switch mimeType {
+		case "text/plain":
+			c.String(http.StatusOK, b.String())
```
rdallman: one of the nice benefits of having a plaintext api here is that we don't need to buffer and encode the log. we should only read the log into a buffer for the deprecated json path and for the new plaintext api, we can just copy the log directly to the response connection. for minio/s3 backend this will read directly from the s3 connection and copy it to the client response which means we aren't buffering it at all internally, which is pretty slick.

rdallman:
aside from fixing the buffer usage, would be great to know 2 things before approving:

* this code works with the old swagger clients, for sure (pick one, cli even)
* the `plaintext` path is executed with the newly generated swagger clients. if not, can we fix our swagger doc to make this possible?

denismakogon:
Answering your questions:
1. This code works with old swagger clients (API tests are passing and Logs API tests are there).
2. API works with the new client (patch for https://github.com/fnproject/cli/pull/128).

Regarding API test, okay, that's possible.

rdallman:
thanks. please fix https://github.com/fnproject/fn/pull/615#discussion_r158585026 -- for plaintext, the code should read:

```go
io.Copy(c.ResponseWriter, logReader)
```

with no buffer in between

treeder:
I see that a lot of code was rolled back in this PR, but why did/do we want to change the default response to plain/text instead of json like it was?  I get the reason that someone might want this, but I would find it strange that a particular endpoint returns plain text by default while the rest return JSON.

denismakogon:
So, there's at least one reason why we should not ship logs as JSON. It forces us to do buffering of the whole log object which can actually be pretty huge. The second reason is JSON that we ship at this moment is mostly useless because users would need just log entity, why do they need exactly the same call ID as they passed in the request (just remember we dropped app name from router response object). Without call ID current JSON response is nothing but a JSON with 1 item only. So, tell me what's the point of having a JSON response that forces us to buffer the log at the server and do JSON parsing at the client side, why not just ship log as text?

Regarding amendments, at 1st iteration, i decided to implement gzip thing for making logs smaller, but @rdallman had concerns about the implementation, so we decided to stick with 1st part of the original issue - ship log as text, and as the follow-up - use gzip to shrink logs.

treeder:
Well, consistency for one. Every endpoint returns JSON by default except one. 

And two, there's a high probability we could add more fields. 

I like leaving it up to the user to pass in text/plain if they want to change to another format, but default should probably remain JSON. If it's too big of a log (which would take a lot and we probably shouldn't be storing it anyways), we could return an error and tell them to pass in text/plain to retrieve it.

rdallman:
it seems fine to default to json, I guess users may expect this if they don't provide a content type. even with `jq` magic the logs are pretty hard to parse out (via eyeball) of json since the newline chars get escaped fwiw. at least in our clients we generate and in the cli we can send in `Accept: text/plain` and get back a plain text log, which should cover most usage (so we get the benefits). wrt fields, at a minimum we can maintain _some_ endpoint that just returns a plaintext log, whether it's the current one or not -- we have `.../call/:id` for additional call fields, for most stuff probably.

api/server/call_logs.go
```diff
@@ -27,16 +28,16 @@ func (s *Server) handleCallLogGet(c *gin.Context) {
...
-	b.ReadFrom(logReader)
-
 	mimeTypes, _ := c.Request.Header["Accept"]
 
 	for _, mimeType := range mimeTypes {
 		switch mimeType {
 		case "text/plain":
-			c.String(http.StatusOK, b.String())
+			io.Copy(c.Writer, logReader)
+			c.Status(http.StatusOK)
```
rdallman: the status code is already written at this point, can either remove this or move it before `io.Copy`, either works.

api/server/call_logs.go
```diff
@@ -21,18 +28,22 @@ func (s *Server) handleCallLogGet(c *gin.Context) {
...
-	var b bytes.Buffer
-	b.ReadFrom(logReader)
-
-	callObj := models.CallLog{
-		CallID:  callID,
-		AppName: appName,
-		Log:     b.String(),
+	mimeTypes, _ := c.Request.Header["Accept"]
+
+	for _, mimeType := range mimeTypes {
```
rdallman: this isn't the behavior we want. an example: 

`Accept: application/xml, text/plain` will return a json blob. the user clearly can handle `text/plain` but it's not explicit whether they can handle json, we should return `text/plain` since it was specified and json was not.

One suggestion might be switch on exact `Accept` headers in the loop, since we _do_ want to treat them as a prioritized list, and only default to json if we don't find any of the types we're expected. the HTML spec pretty clearly defines the behavior we should expose here: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.1 and we also need to handle `q` values (which won't be switched on properly in this code); for `q` values, I don't think we need to worry about quality reduction [for now/ever], just to parse them properly.

api/server/call_logs.go
```diff
@@ -2,13 +2,20 @@ package server
...
 	"bytes"
+	"io"
 	"net/http"
 
 	"github.com/fnproject/fn/api"
 	"github.com/fnproject/fn/api/models"
 	"github.com/gin-gonic/gin"
 )
 
+// note: for backward compatibility, will go away later
```
rdallman: i don't think this is really true, we should probably support json perpetually if somebody asks for it

api/server/call_logs.go
```diff
@@ -21,18 +29,37 @@ func (s *Server) handleCallLogGet(c *gin.Context) {
...
+				&models.CallLog{
+					CallID:  callID,
+					AppName: appName,
+					Log:     b.String(),
+				}})
+			return
+		}
+	}
+	// if we've reached this point it means that Fn didn't recognize Accepted content type
+	c.Status(http.StatusNotAcceptable)
```
rdallman: if no accept headers are given, we also want to return a json. this is only if they provide accept headers, none of which we can offer.

p.s. extract the function for the json logic, please

rdallman: please add a body to this response, as well

api/server/call_logs.go
```diff
@@ -63,5 +64,6 @@ func (s *Server) handleCallLogGet(c *gin.Context) {
 		}
 	}
 	// if we've reached this point it means that Fn didn't recognize Accepted content type
-	c.Status(http.StatusNotAcceptable)
+	WriteError(c, c.Writer, http.StatusNotAcceptable,
```
rdallman: every other call in the front end uses `handleErrorResponse`, with an error that implements `models.APIError`. please do the same here, so that we might log this in `handleErrorResponse`, add a metric for it, whatever else we may want to do in there so that we have telemetry around this and uniformity in the front end error handling.

api/server/call_logs.go
```diff
@@ -21,18 +41,29 @@ func (s *Server) handleCallLogGet(c *gin.Context) {
...
+			return
+
+		}
+		if strings.Contains(mimeType, "*/*") {
+			writeJSON(c, callID, appName, logReader)
+			return
+		}
+	}
+	// if we've reached this point it means that Fn didn't recognize Accepted content type
+	WriteError(c, c.Writer, http.StatusNotAcceptable,
```
rdallman: every other call in the front end uses `handleErrorResponse`, with an error that implements `models.APIError`. please do the same here, so that we might log this in `handleErrorResponse`, add a metric for it, whatever else we may want to do in there so that we have telemetry around this and uniformity in the front end error handling.

PR #618, opened 2017-12-21
unexport all data abstractions on Server

rdallman:
this patch has no behavior changes, changes are:

* server.Datastore() (uses in `api/server` only) -> server.datastore
* server.MQ -> server.mq
* server.LogDB -> server.logstore
* server.Agent -> server.agent

these were at a minimum not uniform. further, it's probably better to force
configuration through initialization in `server.New` to ensure thread safety
of referencing if someone does want to modify these as well as forcing things
into our initialization path and reducing the surface area of the Server
abstraction.

### DIFF
--- a/api/server/apps_create.go
+++ b/api/server/apps_create.go
@@ -34,7 +34,7 @@ func (s *Server) handleAppCreate(c *gin.Context) {
 		return
 	}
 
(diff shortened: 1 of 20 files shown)


rdallman:
follows discussion in #610

rdallman:
thanks!

PR #620, opened 2017-12-21
fn: test and release scripts now run some jobs in background

skinowski:


skinowski:
Not sure how I feel about this change myself due to the mess it introduces into shell scripts, but it does speed things a bit.

treeder:
ehhh, not sure how I feel about it either. Maybe we can wait until CI gets way out of control before we start adding this?  Also, Circle supports parallel steps that might make it easier to grok when trying to figure out where something went wrong: https://circleci.com/docs/2.0/parallelism-faster-jobs/

test.sh
```diff
@@ -23,19 +25,33 @@ function host {
...
-docker pull fnproject/hello
+( docker pull fnproject/sleeper && docker pull fnproject/error && docker pull fnproject/hello) &
+PID2=$!
+
+# pull db images
+( docker pull postgres && docker pull mysql && docker pull minio/minio) &
+PID3=$!
+
+# pull swagger image
+(docker pull quay.io/goswagger/swagger ) &
```
denismakogon: see my comments at related issue, this image huge, why not just download binary for releases?

rdallman:
seems like we got this down to 7 minutes for non-master builds in #624 and #629 will probably improve a little more once that's in, let's punt on this?

PR #624, opened 2017-12-22
Speed up API tests

denismakogon:
Refactoring API tests to take less time
    
- sqlite: tests 15s, overall time: 1m
- mysql: tests 15s, overall time: 59s

All time that CI takes: pulling images

### DIFF
--- a/api_test.sh
+++ b/api_test.sh
@@ -48,21 +48,22 @@ case "$1" in
     ;;
 esac
 
-test test/fn-api-tests/fn-api-tests.test
-status=`echo $?`
-rebuild="${3:-1}"
-circleci=`echo ${CIRCLECI}`
-cd test/fn-api-tests
-if [[ $status -ne 0 ]] || [[ $rebuild -ne 0 ]] ; then
-    if [[ $circleci == "true" ]]; then
-        # dirty things to make CI pass
-        ls -lah /usr/local/go/pkg/linux_amd64/runtime
-        sudo chown -R `whoami`:root /usr/local/go
-        sudo chmod -R 777 /usr/local/go/pkg/linux_amd64
-        ls -lah /usr/local/go/pkg/linux_amd64/runtime
-    fi
-    pwd
-    go test -i -a -o fn-api-tests.test
-fi
-pwd
-./fn-api-tests.test -test.v  -test.parallel ${2:-1} ./...; cd ../../
+#test test/fn-api-tests/fn-api-tests.test
+#status=`echo $?`
+#rebuild="${3:-1}"
+#circleci=`echo ${CIRCLECI}`
+#cd test/fn-api-tests
+#if [[ $status -ne 0 ]] || [[ $rebuild -ne 0 ]] ; then
+#    if [[ $circleci == "true" ]]; then
+#        # dirty things to make CI pass
+#        ls -lah /usr/local/go/pkg/linux_amd64/runtime
+#        sudo chown -R `whoami`:root /usr/local/go
+#        sudo chmod -R 777 /usr/local/go/pkg/linux_amd64
+#        ls -lah /usr/local/go/pkg/linux_amd64/runtime
+#    fi
+#    pwd
+#    go test -i -a -o fn-api-tests.test
+#fi
+#pwd
+#./fn-api-tests.test -test.v  -test.parallel ${2:-1} ./...; cd ../../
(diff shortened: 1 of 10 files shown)


denismakogon:
Let's see how much time CI will take to run them

denismakogon:
Unfortunately, compiling takes a lot of time.
```
time go test -c -i

real	1m18.562s
user	2m37.974s
sys	0m21.343s
```

rdallman:
> Unfortunately, compiling takes a lot of time.

go install before running 3x? then builds once

skinowski:
FYI, in #625 , I've fixed a typo. sqlite => sqlite3. Makefile/circleCI was calling api_test.sh with wrong arg for sqlite.

api_test.sh
```diff
@@ -30,7 +30,7 @@ case "$1" in
     DB_CONTAINER="func-mysql-test"
     docker rm -fv ${DB_CONTAINER} || echo No prev mysql test db container
     docker run --name ${DB_CONTAINER} -p 3306:3306 -e MYSQL_DATABASE=funcs -e MYSQL_ROOT_PASSWORD=root -d mysql
-    sleep 15
+    sleep 5
```
skinowski: we should "loop curl/connect with sleep 1" to wait for that TCP port instead of big sleep here IMHO.

denismakogon:
That's not really critical, the API tests will just go for default value which is sqlite3. At this moment i think about what @rdallman mentioned. Build test binary once and then run it whenever you need API tests.

api_test.sh
denismakogon: there's an alternative, start datastore container, build binary (takes 1m), by that time datastore container will definitely be up and running.

skinowski: sounds good, not a big deal, I wouldn't spend too much time on this. :-)

denismakogon:
Hey, @skinowski check the latest commit, it tries to accomplish what I was talking about: build binary during first API test run then other ones will reuse that binary later.

denismakogon:
Okay, there are some problems with CI at this moment, when trying to run `go test -i -a` CI fails with the following error:
```
go install runtime/internal/sys: open /usr/local/go/pkg/linux_amd64/runtime/internal/sys.a: permission denied
```
see latest [build log](https://circleci.com/gh/fnproject/fn/1701?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).
Any idea how can we overcome this? @rdallman @skinowski

denismakogon:
Got some ideas, will try to fix access rights to corresponding files and that should make it work, at least had the same problems with Gogland couple month ago. Changing ownership and rights helped.

denismakogon:
Dirty hack worked. One binary, 3 test suites.

API tests with MySQL: 22 seconds
API tests with PG: 17 seconds (pg becomes reachable faster than MySQL)
API tests with sqlite3: 12 seconds

api_test.sh
```diff
@@ -24,30 +24,45 @@ case "$1" in
...
+        # dirty things to make CI pass
+        ls -lah /usr/local/go/pkg/linux_amd64/runtime
+        sudo chown -R `whoami`:root /usr/local/go
+        sudo chmod -R 777 /usr/local/go/pkg/linux_amd64
+        ls -lah /usr/local/go/pkg/linux_amd64/runtime
+    fi
+    pwd
+    go test -i -a -o fn-api-tests.test
+fi
+pwd
```
rdallman: you can replace all of this with `cd test/fn-api-tests && go install` and get about 95% of the benefits. install will build all the static libraries for each package (takes time) and then test will just link them, which doesn't take a whole lot of time. not so keen on the chown/chmod shenanigans, this should avoid that

test/fn-api-tests/utils.go
```diff
@@ -224,3 +236,27 @@ func MyCaller() string {
...
+func retryDatastorePing(attempts int, sleep time.Duration, dbURL string) (err error) {
+	for i := 0; i < attempts; i++ {
+		_, err = datastore.New(dbURL)
+		if err == nil {
+			return nil
+		}
+		time.Sleep(sleep)
+	}
+	return err
+}
```
rdallman: same deal as commented in #629 -- perhaps fixed there

rdallman:
sweet, says down to 7 minutes with go install. looks like this pulls in #630 - made a comment there to discuss, the rest looks good.

rdallman:
update with #630 and we'll get this in

PR #625, opened 2017-12-23
fn: circleci and makefile adjustments

skinowski:
*) Moved more tasks into Makefile to allow for
parallelism and dependency checks.
*) Added cpu count in circleci make invocations
for parallelism
*) Fixed broken api-tests for sqlite3 (shell script ignored sqlite typo)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,14 +27,11 @@ jobs:
           sudo service docker stop
           curl -fsSL https://get.docker.com/ | sudo sh
       - run: docker version
-      - run: make docker-build
-      - run: make install
-      - run: make test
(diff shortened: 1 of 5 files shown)


skinowski:
Not sure if want to go this way, but moving the dep tree into Makefile (and eventually release.sh itself too) works ok both in circleci and linux/mac. One big Makefile sucks, but it does allow to build dep tree for incremental builds. If we absorb 'release.sh' in future in the Makefile, that'll cut down an extra 'make docker-build'

There's a critical fix here for sqlite3 typo, even if we don't like this patch, I'll carry that over with another PR since old api_test.sh used to skip it ignoring unset shell variable.

PR #627, opened 2017-12-23
allow user configured agent in full node

### LINKED ISSUE #623: Fix overwriting agent object in server.go New()
The relevant code:

https://github.com/fnproject/fn/blob/master/api/server/server.go#L220

This overwrites the agent if already present and in case of tests we end up with two agent instances.

rdallman:
this should keep the old default behavior but allow users to pass in a
configured agent to configure the server themselves, without having to worry
about a russian agent being a british agent.

also closes any agent given to an api node.

closes #623


rdallman:
thanks, fixed tests

PR #629, opened 2017-12-24
Use alpine images to make tests take less time

denismakogon:
* use PG alpine
 * use Minio alpine
 * no official alpine distro for MySQL, uhhh :(
 * install swagger tool instead of docker image
 * use retry func to confirm that datastore is okay before running tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,4 +32,3 @@ fnlb/fnlb
 /fnserver
 *iml
 target/
-fn-api-tests.test
--- a/helpers.sh
+++ b/helpers.sh
@@ -27,7 +27,7 @@ function remove_containers {
 }
 
 function install_swagger_tool {
-    if [[ ! -f ${GOPATH}/bin/swagger ]]; then
+    if [[ ! -f ./swagger ]]; then
         case "$(uname)" in
           Linux)
             curl -L https://github.com/go-swagger/go-swagger/releases/download/0.13.0/swagger_linux_amd64 -o ./swagger
--- a/helpers.sh
+++ b/helpers.sh
@@ -30,11 +30,11 @@ function install_swagger_tool {
     if [[ ! -f ${GOPATH}/bin/swagger ]]; then
         case "$(uname)" in
           Linux)
-            curl -L https://github.com/go-swagger/go-swagger/releases/download/0.13.0/swagger_linux_amd64 -o ${GOPATH}/bin/swagger
+            curl -L https://github.com/go-swagger/go-swagger/releases/download/0.13.0/swagger_linux_amd64 -o ./swagger
           ;;
           Darwin)
-            curl -L https://github.com/go-swagger/go-swagger/releases/download/0.13.0/swagger_darwin_amd64 -o ${GOPATH}/bin/swagger
+            curl -L https://github.com/go-swagger/go-swagger/releases/download/0.13.0/swagger_darwin_amd64 -o ./swagger
         esac
     fi
-    chmod +x ${GOPATH}/bin/swagger
+    chmod +x ./swagger
 }
--- a/helpers.sh
+++ b/helpers.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+set -exo pipefail
+
+function host {
+    case ${DOCKER_LOCATION:-localhost} in
+    localhost)
+        echo "localhost"
+        ;;
+    docker_ip)
+        if [[ !  -z  ${DOCKER_HOST}  ]]
+        then
+            DOCKER_IP=`echo ${DOCKER_HOST} | awk -F/ '{print $3}'| awk -F: '{print $1}'`
+        fi
+
+        echo ${DOCKER_IP}
+        ;;
+    container_ip)
+        echo "$(docker inspect -f '{{.NetworkSettings.IPAddress}}' ${1})"
+        ;;
+    esac
+}
+
+function remove_containers {
+    docker rm -fv func-postgres-test 2>/dev/null || true
+    docker rm -fv func-mysql-test 2>/dev/null || true
+    docker rm -fv func-minio-test 2>/dev/null || true
+}
+
+function install_swagger_tool {
+    if [[ ! -f ${GOPATH}/bin/swagger ]]; then
+        case "$(uname)" in
+          Linux)
+            curl -L https://github.com/go-swagger/go-swagger/releases/download/0.13.0/swagger_linux_amd64 -o ${GOPATH}/bin/swagger
+          ;;
+          Darwin)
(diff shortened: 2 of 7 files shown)


denismakogon:
Reverting minio image:
```
docker images | grep minio
minio/minio                   latest              4110ee55fc11        4 weeks ago         27.3MB
webcenter/alpine-minio        latest              49c8f4547226        7 weeks ago         82.3MB
```

denismakogon:
Not sure what we can do with mysql image, which is 408Mb, mysql docker community has done zero to build on alpine by this time.

PG alpine size made me happy, really:
```
docker images | grep postgres
postgres                      latest              ec61d13c8566        12 days ago         287MB
postgres                      9.3-alpine          0350272b4017        2 weeks ago         35.5MB
```

So, the time difference is not that huge at this moment, but a good thing to have along with #624.

helpers.sh
```diff
@@ -0,0 +1,40 @@
...
+    esac
+}
+
+function remove_containers {
+    docker rm -fv func-postgres-test 2>/dev/null || true
+    docker rm -fv func-mysql-test 2>/dev/null || true
+    docker rm -fv func-minio-test 2>/dev/null || true
+}
+
+function install_swagger_tool {
```
rdallman: https://github.com/swagger-api/validator-badge seems to outline a way to send a file to a service running a swagger validator. the badge would be a nice addition, but sending our file to a validator service seems much cleaner than us downloading a binary every build, if anything because in theory the service will ensure we keep up to date with the latest swagger spec and here we have to manually keep the binary version up to date.

api/datastore/sql/sql_test.go
```diff
@@ -40,6 +41,18 @@ func newWithMigrations(url *url.URL) (*sqlStore, error) {
 	return ds, nil
 }
 
+func retryDBConnection(t *testing.T, attempts int, sleep time.Duration, dbURL *url.URL) (err error) {
+	for i := 0; i < attempts; i++ {
+		t.Logf("[%v] - attempting to reach out datastore %v", i, dbURL.String())
+		_, err = newDS(dbURL)
```
rdallman: this generates a lot of new db instances, which isn't really cheap afaik and also doesn't really ensure that the datastore instance we're going to actually use is valid and adds extra code in the db initialization caller path to call retryDBConnection. is there any reason not to do just do this in our db initialization `ping` inside of the sql package?

rdallman:
> Not sure what we can do with mysql image, which is 408Mb, mysql docker community has done zero to build on alpine by this time.

i might piddle around with this, it shouldn't be too bad to get all the deps into alpine but sometimes it is.

rdallman:
> i might piddle around with this, it shouldn't be too bad to get all the deps into alpine but sometimes it is.

found https://github.com/docker-library/mysql/issues/179 -- we can try to poke the right people internally to get this supported, it appears to work now but need to have release track to follow.

helpers.sh
denismakogon: Well, the badge is a good thing, but it's not related to this PR, but what if I want to modify swagger how can know if that is okay?

I can add the badge, but we do know that github cache works really slow, we've seen that couple times with swagger that's why we use huge docker image with swagger tool only to validate our swagger doc through the mounted file:
```
docker run -v /Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn:/go/src/github.com/fnproject/fn --rm quay.io/goswagger/swagger validate /go/src/github.com/fnproject/fn/docs/swagger.yml
```
The point of downloading binary is to reduce the time of swagger doc validation.
FYI, the swagger tool image is 493MB, swagger tool binary ~2MB.

api/datastore/sql/sql_test.go
denismakogon: fixed, see #630, I'll rebase this PR to that branch.

denismakogon:
@rdallman i'll address all comments except swagger tool change. Regarding the MySQL alpine image, we might ask @mattlord for help.

helpers.sh
rdallman: Not worried about the badge so much here, it would be nice to have separate from this. We can send the file from on disk inside of CI to the swagger server listed from that link for validation (I think), which requires no binary download and then we don't have to worry about versioning of the binary. If we change swagger we had the same issue with the image previously, I don't think this is much of a concern vs us lagging behind swagger updates by manually downloading the binary.

denismakogon: Unfortunately, recommended method (posting swagger doc as the payload) doesn't work:
```
curl -X POST -d @docs/swagger.yml -H 'Content-Type:application/json' http://online.swagger.io/validator/debug
<html>
<head>
<meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
<title>Error 500 Internal Server Error</title>
</head>
<body><h2>HTTP ERROR 500</h2>
<p>Problem accessing /validator/debug. Reason:
<pre>    Internal Server Error</pre></p><hr><i><small>Powered by Jetty://</small></i><hr/>

</body>
</html>
```
so, leaving as is.

rdallman: Ok, next idea: can we build a smaller swagger image with that binary to use instead?

denismakogon:
@rdallman I pushed the image to my own docker repo: `denismakogon/swagger-tool:0.0.1`. Dockerfile you can find in-tree.
It's not 2Mb size, but 30Mb. Still better than 0.5Gb.

treeder:
How about add it to fnproject/dockers ?

denismakogon:
Sure, can submit there.

denismakogon:
@treeder https://github.com/fnproject/dockers/pull/7

skinowski:
Kind of unrelated, but we should eventually be able to get rid of sleeper, hello and error images as fn-test-utils can perform same functionality. I tried removing these with one PR, but I quickly realized it's more work than I anticipated.

.gitignore
```diff
@@ -33,3 +33,4 @@ fnlb/fnlb
 .idea/
 *iml
 target/
+swagger
```
rdallman: can remove this change now

denismakogon:
Rebased.

denismakogon:
Nice, we're at `07:46`, which is good. Is there anything else that we can speed up?

rdallman:
* we could play with caching the static library files between runs. 99.9% of the time they don't change, and `go install` takes about a minute, and should handle building new versions only when it needs to. 
* seems like reducing tests down to one image could save maybe 30s or so, per https://github.com/fnproject/fn/pull/629#issuecomment-354847459
* we're building the release docker image on branches even though we don't use it for anything, we only need to do this on master. that's 2 minutes. 

seems like we could get under 5 minutes pretty easily.

rdallman:
we're also installing new docker version every time, that's 1 minute. maybe we could cache that, too

skinowski:
+1 on "go install", that should shave extensions/middleware building time as well.

PR #630, opened 2017-12-26
Use retry func while trying to ping SQL datastore

denismakogon:
- implements retry func specifically for SQL datastore ping
 - fmt fixes
 - using sqlx.Db.PingContext instead of sqlx.Db.Ping
 - propogate context to SQL datastore

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -98,9 +98,9 @@ func NewFromEnv(ctx context.Context, opts ...ServerOption) *Server {
 		defaultDB = fmt.Sprintf("sqlite3://%s/data/fn.db", curDir)
 		defaultMQ = fmt.Sprintf("bolt://%s/data/fn.mq", curDir)
 	}
-	opts = append(opts, WithDBURL(getEnv(EnvDBURL, defaultDB)))
+	opts = append(opts, WithDBURL(ctx, getEnv(EnvDBURL, defaultDB)))
 	opts = append(opts, WithMQURL(getEnv(EnvMQURL, defaultMQ)))
-	opts = append(opts, WithLogURL(getEnv(EnvLOGDBURL, "")))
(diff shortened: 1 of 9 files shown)


api/server/server.go
```diff
@@ -115,9 +115,9 @@ func pwd() string {
 	return strings.Replace(cwd, "\\", "/", -1)
 }
 
-func WithDBURL(dbURL string) ServerOption {
+func WithDBURL(ctx context.Context, dbURL string) ServerOption {
```
rdallman: this seems more natural to put on `ServerOption`, since it's a func that will be called later on initialization. the func that returns a ServerOption func will be called from a different context and probably shouldn't really do any work (the idea is to be lazy). both the `WithDBURL` and `WithLogURL` should probably call `.New` for their respective things inside of a `ServerOption` func instead of in the enclosing func that returns them, I guess these two were something of a mistake in that sense, would you mind fixing this since it's now an issue?

rdallman: it seems reasonable to make `ServerOption` slightly more flexible in 2 ways, with a signature of:

```go
type ServerOption func(context.Context, *Server) error
```

then we can just bubble up errors and log / return them in `server.New()` instead of relying on `ServerOption` calling `log.Fatal` to terminate the program on errors. 

not sure this matters too much, maybe it's easier (it is simpler) just keeping it like it is. hopefully people aren't doing very complex things in initialization anyway

denismakogon:
Not sure what's going on with tests, dependency problem? Trying to figure it out.

api/server/server_test.go
```diff
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/docker/docker/api/server/router"
```
rdallman: delete this line to fix dep problem @denismakogon

denismakogon: thanks, haven't seen this line in a diff, my bad =/

Issue #405, opened 2017-10-08
[Feature Request] Limit Function CPU usage


kuashe:
### Why ?

All FaaS (AWS , Google , Azure ) limit CPU usage of a function.

In my opinion it's critical to implement such feature . 
At the moment a single function **can exhaust an entire server** with 100% CPU utilization using all cores even if it's for a short period of time.

I think Google CPU to Memory Ratio is very coherent as a default setting

| Memory | CPU    |
|--------|--------|
| 128    | 200MHz |
| 256    | 400MHz |
| 512MB  | 800MHz |
| 1024MB | 1.4GHz |
| 2048MB | 2.4GHz |

### Proposition
- Limit container to only one thread utilization
- Add a default CPU to Memory function ratio 
- Add a parameter to define arbitrary CPU (in **Megahertz**) to Memory ratio

Links :
[How to limit a container CPU usage](https://docs.docker.com/engine/admin/resource_constraints/#cpu)
[GCloud Compute to Memory Ratio](https://cloud.google.com/functions/pricing#compute_time)
[GoLang CPU Megahertz and Cores count Lib](https://github.com/shirou/gopsutil)

treeder:
We currently use CPU shares so 1 function won't take the entire CPU if others  are running, they will all get an equal share. If there's only 1 running, then it gets 100% of the share.

kuashe:
Wouldn't this lead to un-consistent performance ?

When there is little or no traffic , the functions will execute really fast.
When there is a lot of traffic , functions would run much slower. 

Limiting CPU by default would help to predict performance of a server.

rdallman:
yep, it is inconsistent. but at least, it makes it not a __critical__ issue since we have this in place, and we won't exhibit the behavior that 1 function can exhaust an entire server if > 1 function exists on it and needs cpu, since they will share equally. 

we had some success running prod workloads like this on a different product in the past since it ends up getting better utilization when not at capacity, many users like this feature, though obviously it's ideal to run at capacity, we found that most of the time users are not running at 100% capacity on their clusters (only spikes). it also makes it much easier to size nodes if you only talk in terms of memory ;)

[...]

PR #642, opened 2018-01-05
Cpu resources

skinowski:
As discussed in #405 , this change adds CPUs param to routes and calls. This is represented by a string in ether of the following formats:

*) as MilliCPUs eg. 100m which is similar to k8s which means 1/10 of a CPU core or
*) as floating-point eg. 0.5 which means 1/2 of a CPU core, similar to docker 'cpus'

Resource allocation for CPU is handled very similar to Memory except during initialization there's no headroom estimation for fn server and container run time (docker) cpu consumption itself.

Internally, CPUs is converted to millicpu units to set CFS period/quota via docker client API.

Mixed workloads (unlimited CPUs calls mixed with non-zero CPUs calls) is left open, so operators have to keep an eye on that either by not allowing zero CPUs calls at all or only allowing zero CPUs calls to disable CPU resource allocation.


api/agent/agent_test.go
```diff
@@ -412,6 +422,7 @@ func TestSubmitError(t *testing.T) {
 		"FN_APP_NAME": appName,
 		"FN_PATH":     path,
 		"FN_MEMORY":   strconv.Itoa(memory),
+		"FN_CPUS":     CPUs,
```
denismakogon: Overall looks good, but i do feel like there would be a problem with UX. How would user know the capacity for CPU and RAM? I do feel like there should be some sort of API to see the exact free/used capacity (think of flavors on AWS, OpenStack, etc.).

skinowski: Yes, I need to add these as stats (CPU, mem, slots in use, hot/cold container counts, etc. etc.) and maybe a API could also be useful. I'll open an issue to discuss and proceed.

api/agent/resource.go
```diff
@@ -68,78 +80,93 @@ func (t *resourceToken) Close() error {
...
 
 		a.ramAsyncUsed += asyncMem
 		a.ramSyncUsed += syncMem
+		a.cpuAsyncUsed += asyncCPU
+		a.cpuSyncUsed += syncCPU
 		c.L.Unlock()
 
 		t := &resourceToken{decrement: func() {
+
+			// Guard against multiple close
```
rdallman: wrapping the `.decrement()` call in https://golang.org/pkg/sync/#Once.Do inside of `.Close()` may be a little cleaner than using a variable from the closure's enclosing scope

api/agent/resource.go
```diff
@@ -286,24 +378,59 @@ func getMemoryHeadRoom(usableMemory uint64) (uint64, error) {
 	return headRoom, nil
 }
 
-func checkCgroup() (uint64, error) {
-	f, err := os.Open("/sys/fs/cgroup/memory/memory.limit_in_bytes")
+func readString(fileName string) (string, error) {
```
rdallman: ~dupe code: https://golang.org/pkg/io/ioutil/#ReadFile

api/models/route.go
```diff
@@ -128,9 +133,52 @@ func (r *Route) Validate() error {
...
+	// IMPORTANT: This is varchar(16) in datastore sql
+	// Enforce .2 precision and clean up padded zeros, preceding sign, etc.
+	sanitized := fmt.Sprintf("%.2f", cpu)
+
+	// now check if by sanitizing we ended up losing precision
+	// if we lost precision, then an unsupported format (eg. 0.22020)
+	// was submitted.
+	tmp, err := strconv.ParseFloat(sanitized, 64)
+	if err != nil || cpu != tmp {
+		return "", ErrRoutesInvalidCPUs
```
rdallman: can we just floor this? the error message doesn't make it clear that they must specify only up to 2 decimals precision, and it doesn't seem to warrant an error really to lop off any more precision. up to you, but need to make error clearly state what's wrong or floor / ceiling I think.

api/agent/call.go
```diff
@@ -392,6 +407,21 @@ func (c *call) End(ctx context.Context, errIn error) error {
 	return errIn // original error, important for use in sync call returns
 }
 
+// getCPUQuota converts CPUs to CPU quota by multiplying them by 100.
+// eg. 0.20 becomes 20, 2.0 becomes 200, which are easier to calculate
+// and manage as resources and translate to Linux CFS (cpu-quota and cpu-period)
+func getCPUQuota(CPUs string) (uint64, error) {
```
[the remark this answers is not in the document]
skinowski: So, the flow is more like string -> float -> int. I opted for string since we may extend it to k8s style milicpus, etc. formats. We end up converting all of this to uint64 at driver level:

https://godoc.org/github.com/moby/moby/api/types/container#Resources

Maybe we should stick with uint64 from the front and stay consistent using millicpu, where "100" means 100% (1CPU), etc.

api/models/route.go
skinowski: I was thinking about adding this to the documentation where expected format is 2 decimal points precision. We kind of do the same for other settings, eg. we don't have ErrMemoryTooLarge or TimeoutTooBig, so I stuck with one error code. I could round it silently too. Not sure.

api/agent/call.go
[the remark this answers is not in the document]
skinowski: Assuming we are ok with "string -> float -> int" flow, I'm going to clean up (and clarify this) further in the next commit.

api/models/route.go
rdallman: I think that's fine, just the error could be more specific about its format / bounds so that users could debug it quicker than having to go look up the docs (we'll get a slack message long before then). rounding for this seems much easier and pretty inconsequential.

api/agent/call.go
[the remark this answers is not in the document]
rdallman: it would be nice if it were either a float or an int, and then the type for it can add `UnmarshalJSON`, etc., if we want the format to be `"cpus":"100mCPU"` or something, it would be pretty flexible. i think we should just pick something, though, instead of having a string we have to parse out much later down the line. then the user is confronted with their poor choices when they're adding it to the route and there's no chance of a mishap (error) in the backend, we can just use the thing.

i.e. there'd be no string->float->int, just parse to an int or float immediately, and then the driver can do what it needs to with that. it's a little strange to see a number in a string, as it is, in a json api. I'd be okay with just picking 'milliCPUs' to be consistent with k8s now and calling it a day. if we end up regretting that it should be easy to work around, but I doubt it.

skinowski:
So, this version is interesting..

I've added the MilliCPUs type, but to fetch proper API errors, I had add a few checks every time we do json handling (in order to return specific error code and not generic Invalid JSON)

I ran into some dependency chaos, so I chose not to use MilliCPUs inside docker driver. And resource.go has been unrelated to models so far, so I kept it as is and ended up doing a few uint64 casts.

In DB, I modified this into 'int' column similar to memory, timeouts.

This handles float format which is 0.2 => 20% of 1 CPU as well as k8s millicpus, which is 1000m => 1 CPU, etc.  MilliCPUs have a suffix of 'm', similar to k8s.

rdallman:
> I've added the MilliCPUs type, but to fetch proper API errors, I had add a few checks every time we do json handling (in order to return specific error code and not generic Invalid JSON)

ah, crap that's right gin isn't very flexible in this dept. we should probably have our own json helper function instead of using gin's in handlers (that uses gin's, sure) since we need to return our own error type if things go awry.

api/models/config.go
```diff
@@ -125,3 +127,67 @@ func (h *Headers) Scan(value interface{}) error {
...
+	if c == 0 {
+		return ""
+	}
+	return fmt.Sprintf("%dm", c)
+}
+
+// implements json.Unmarshaler
+func (c *MilliCPUs) UnmarshalJSON(data []byte) error {
+
+	outer := strings.TrimSpace(string(data))
```
rdallman: not a huge deal, but useful to know, golang `strings` and `bytes` package have nearly identical helper functions for trim/suffix/prefix/etc work. https://golang.org/pkg/bytes/#TrimSpace

api/models/config.go
```diff
@@ -125,3 +127,67 @@ func (h *Headers) Scan(value interface{}) error {
...
+		*c = MilliCPUs(fCPU * 1000)
+	}
+
+	return nil
+}
+
+// implements json.Marshaler
+func (c *MilliCPUs) MarshalJSON() ([]byte, error) {
+
+	if *c < MinMilliCPUs || *c > MaxMilliCPUs {
```
rdallman: kind of pedantic but this check here and in unmarshal seem like they might belong in `route.Validate()` since they're not parsing errors. this might get rid of the noted json handling woes

api/models/error.go
```diff
@@ -206,6 +207,11 @@ func (e err) Code() int { return e.code }
 
 func NewAPIError(code int, e error) APIError { return err{code, e} }
 
+func IsAPIError(e error) bool {
+	_, ok := e.(err)
```
rdallman: nit: probably should be the interface. `_, ok := e.(APIError)`

api/models/config.go
rdallman: i guess not quite on the json woes, since they could still not provide a valid float / millicpu value

rdallman:
> I ran into some dependency chaos, so I chose not to use MilliCPUs inside docker driver. And resource.go has been unrelated to models so far, so I kept it as is and ended up doing a few uint64 casts.

seems like a sane boundary 👍

skinowski:
@rdallman bump. :-)

Issue #644, opened 2018-01-05
Logs should support specifying region when using S3-compatible object store


gviedma:
If the object storage region is not specified, the minio-go client will attempt to lookup the bucket for every get/put operation. This is not only slower but also causes problems with some object storage implementations.

PR #645, opened 2018-01-05
Logs should support specifying region when using S3-compatible object store 

gviedma:
Fixes issue #644

### DIFF
--- a/api/logs/s3/s3.go
+++ b/api/logs/s3/s3.go
@@ -30,6 +30,7 @@ type store struct {
 }
 
 // s3://access_key_id:secret_access_key@host/location/bucket_name?ssl=true
+// Note that access_key_id and secret_access_key must be URL encoded if they contain unsafe characters!
 func New(u *url.URL) (models.LogStore, error) {
 	endpoint := u.Host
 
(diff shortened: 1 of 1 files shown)


msgodf:
LGTM. Does it work? :)

msgodf:
Looks like the existing tests for this are out of date, as they only run if the `MINIO_URL` environment variable is set.

msgodf:
Also, might be worth noting that this is part of the work for 
https://github.com/fnproject/internal/issues/34

rdallman:
> Looks like the existing tests for this are out of date, as they only run if the MINIO_URL environment variable is set.

the test script sets this up properly https://github.com/fnproject/fn/blob/master/test.sh#L25 so the tests will run against minio. we can run minio locally so it seems the best way to test, and the tests pass here so I think it works.

@gviedma you need to make the oracle org public on github to auto pass the CLA bot, or alternatively signing the CLA should work too but it's slightly more involved.

gviedma:
I've tested running some calls against Oracle Object Store and uploading of log files is currently failing. It seems that OSS requires the actual content length to be provided with the request. I'll investigate further.

rdallman:
makes sense. we should be able to provide that relatively easily. somewhat related: https://github.com/fnproject/fn/issues/646 (I believe in progress atm)

PR #649, opened 2018-01-05
fn: async wait bugfix and low mem warnings

skinowski:


skinowski:
Ok, increased the warning to 256MB for time being..

PR #651, opened 2018-01-05
Fixes 404 responses from functions that go through NoRoute path.

treeder:
This fixes an issue where if the function call doesn't match a gin route, it would return a 404, even though the function ran fine. 

screen shot 2018-01-04 at 11 42 09 am

Also, while trying to debug this, I made some of the function executing methods return an error which cleans things up a bit (IMO), and allows callers to know if something went wrong.

### DIFF
--- a/api/server/runner.go
+++ b/api/server/runner.go
@@ -3,7 +3,6 @@ package server
 import (
 	"bytes"
 	"errors"
-	"fmt"
 	"net/http"
 	"path"
 	"time"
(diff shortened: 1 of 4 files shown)


docs/developers/apps.md
```diff
@@ -48,7 +48,7 @@ http://abc.io/users -> function in users/ directory
 fn deploy --all
 ```
 
-If you're just testing locally, you can speed it up with the `--local` flag.
+If you're just testing locally, you can speed it up with the `--local` flag. Or if you want to deploy to a different app, use the `--app APPNAME` flag.
```
treeder: Sorry, this stuff is unrelated...

treeder:
Don't review this yet...

treeder:
Ok, ready now.

api/server/runner.go
```diff
@@ -32,12 +39,15 @@ func (s *Server) handleFunctionCall(c *gin.Context) {
 	var a string
 	ai := ctx.Value(api.AppName)
 	if ai == nil {
-		handleErrorResponse(c, errors.New("app name not set"))
-		return
+		err := models.ErrAppsMissingName
+		return err
```
denismakogon: Why not just return that error?

rdallman:
i think we need to do this in the not-in-`NoRoute` path too, but this works for one case

treeder:
Doesn't seem to be an issue with other paths (ie: any route registered with gin).

Issue #326, opened 2017-09-14
Issue with new streaming responses


treeder:
I have a function that failed halfway through, my browser got half the response then it crapped out. Then an error json halfway through:

```html
<!DOCTYPE html>
--
  | <html>
  | <head>
  | <meta charset="UTF-8">
  | <title>My page</title>
  | </head>
  | <body>
  | <h1>My page</h1>
  | <p>This is my app. It may not be the best app, but it&#39;s my app.&lt;br/&gt;&lt;br/&gt;And it&#39;s multilingual!</p>
  | <div>
  | <div><a href="{"error":{"message":"container exit code 1"}}
```

It should have 50X'd back to the user rather than return anything at all. 

```
ERRO[2712] api error                                     action="server.handleRunnerRequest)-fm" app=helloapp code=502 error="container exit code 1"
```

Also, there's no call_id in that log line to pull up the container logs. 

 This should return:

- [ ] 500 internal server error
- [ ] error JSON only

treeder:
Still see this, pretty ugly (and unusable) when a function errors halfway through.

treeder:
![screen shot 2018-01-05 at 1 05 41 pm](https://user-images.githubusercontent.com/75826/34628465-514d166a-f219-11e7-85c4-6e74260e1e2a.png)

rdallman:
that one specifically i think the fix is separate [from buffering], we need to watch to see if bytes have been written and not write the error if they have. even if we have buffering we need that.

treeder:
To be clear here, the problem is that a failed function returns a `200` status and a garbage response. If a function fails at any point, it should be a 500 and return a JSON error response.

skinowski:
Aside from the 200/500 issue above, I'll probably end up a adding a configurable system wide setting (eg. buffer size of 1-10MB) along with stream of/on option per route.

PR #652, opened 2018-01-05
watch the writer to not double write

rdallman:
we're writing our error over a previously perfectly fine response from a call.
we need to just not do that. i didn't test the more exotic cases, this should
do the trick.

related to #326 but doesn't close entirely (just one case mentioned)

### DIFF
--- a/api/server/runner.go
+++ b/api/server/runner.go
@@ -55,8 +55,10 @@ func parseParams(params gin.Params) agent.Params {
 func (s *Server) serve(c *gin.Context, appName, path string) {
 	// GetCall can mod headers, assign an id, look up the route/app (cached),
 	// strip params, etc.
+	writer := &watchWriter{inner: c.Writer}
+
 	call, err := s.agent.GetCall(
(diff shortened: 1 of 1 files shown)


rdallman:
i think this will fix this one https://github.com/fnproject/fn/issues/326#issuecomment-355665645 -- i gotta run for today, should do the trick though. i can/should/will test more thoroughly (a testing on the case linked would be nice to confirm)

treeder:
This doesn't fix 326, it possibly makes it worse because now we're hiding the fact that the function failed even more. 

Needs to return a 500 and an error response if the function fails at any point.

treeder:
This example https://github.com/fnproject/fn/issues/326#issuecomment-355665645 output is very contrived btw, I threw in a panic at the end. That wouldn't be typical.

rdallman:
> This example #326 (comment) output is very contrived btw, I threw in a panic at the end. That wouldn't be typical.

ok, think this confused my interpretation of what was happening then.

Issue #535, opened 2017-11-24
JSON protocol writes HTTP path and query instead of full request URL


denismakogon:
`request.URL.RequestURI()` return path and query of the request URL, i.e. `/r/app/route?something=something&etc=etc`, but has to return full request URL, i.e. `fnhost.io/r/app/route?something=something&etc=etc`

skinowski:
I see the issue in api/agent/protocol. I'll fix this.

skinowski:
Closed with #657

PR #657, opened 2018-01-08
fn: URL parsing updates to fix json request_url

skinowski:
*) Updated fn-test-utils to latest fdk-go
*) Added hot-json to runner tests
*) Removed anon function in FromRequest which had
a side effect to set req.URL.Host. This is now more
explicit and eliminates some corresponding logic in
protocol http.
*) in gin, http request RequestURI is not set, removed
code that references this. (use Call.URL instead)

### DIFF
--- a/api/agent/protocol/http.go
+++ b/api/agent/protocol/http.go
@@ -27,7 +27,7 @@ func (p *HTTPProtocol) IsStreamable() bool { return true }
 // TODO maybe we should take io.Writer, io.Reader but then we have to
 // dump the request to a buffer again :(
 func (h *HTTPProtocol) Dispatch(ctx context.Context, ci CallInfo, w io.Writer) error {
-	err := DumpRequestTo(h.in, ci.Request()) // TODO timeout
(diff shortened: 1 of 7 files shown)


skinowski:
For #535

api/agent/protocol/http.go
```diff
@@ -70,32 +70,17 @@ func (h *HTTPProtocol) Dispatch(ctx context.Context, ci CallInfo, w io.Writer) e
...
-		if host == "" && req.URL != nil {
-			host = req.URL.Host
-		}
-
-		if host != "" {
-			fmt.Fprintf(w, "Host: %s\r\n", host)
-		}
+	absRequestURI := strings.HasPrefix(reqURI, "http://") || strings.HasPrefix(reqURI, "https://")
+	if !absRequestURI && req.URL.Host != "" {
+		fmt.Fprintf(w, "Host: %s\r\n", req.URL.Host)
```
rdallman: this was ~copied from https://github.com/golang/go/blob/master/src/net/http/httputil/dump.go#L66 fwiw

PR #658, opened 2018-01-08
fn: agent slot and execution wait correction

skinowski:
Since by policy we require timeout/2 remaining time
before we can execute the request, we should also
bound the slot wait time by timeout/2 to avoid
waiting for full timeout in slot wait phase.


skinowski:
See #274

PR #660, opened 2018-01-09
make headers quasi-consistent

### LINKED ISSUE #329: Malformed HTTP headers for async hot functions

rdallman:
possible breakages:

* `FN_HEADER` on cold are no longer `s/-/_/` -- this is so that cold functions
can rebuild the headers as they were when they came in on the request (fdks,
specifically), there's no guarantee that a reversal `s/_/-/` is the original
header on the request.
* app and route config no longer `s/-/_/` -- it seemed really weird to rewrite
the users config vars on these. should just pass them exactly as is to env.
* headers no longer contain the environment vars (previously, base config; app
config, route config, `FN_PATH`, etc.), these are still available in the
environment.

this gets rid of a lot of the code around headers, specifically the stuff that
shoved everything into headers when constructing a call to begin with. now we
just store the headers separately and add a few things, like FN_CALL_ID to
them, and build a separate 'config' now to store on the call. I thought
'config' was more aptly named, 'env' was confusing, though now 'config' is
exactly what 'base_vars' was, which is only the things being put into the env.
we weren't storing this field in the db, this doesn't break unless there are
messages in a queue from another version, anyway, don't think we're there and
don't expect any breakage for anybody with field name changes.

this makes the configuration stuff pretty straight forward, there's just two
separate buckets of things, and cold just needs to mash them together into the
env, and otherwise hot containers just need to put 'config' in the env, and then
hot format can shove 'headers' in however they'd like. this seems better than
my last idea about making this easier but worse (RIP).

this means:

* headers no longer contain all vars, the set of base vars can only be found
in the environment.
* headers is only the headers from request + call_id, deadline, method, url
* for cold, we simply add the headers to the environment, prepending
`FN_HEADER_` to them, BUT NOT upper casing or `s/-/_/`
* fixes issue where async hot functions would end up with `Fn_header_`
prefixed headers
* removes idea of 'base' vars and 'env'. this was a strange concept. now we just have
'config' which was base vars, and headers, which was base_env+headers; i.e.
they are disjoint now.
* casing for all headers will lean to be `My-Header` style, which should help
with consistency. notable exceptions for cold only are FN_CALL_ID, FN_METHOD,
and FN_REQUEST_URL -- this is simply to avoid breakage, in either hot format
they appear as `Fn_call_id` still.
* removes FN_PARAM stuff
* updated doc with behavior

weird things left:

`Fn_call_id` e.g. isn't a correctly formatted http header, it should likely be
`Fn-Call-Id` but I wanted to live to fight another day on this one, it would
add some breakage.

examples to be posted of each format below

closes #329

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io"
 	"net/http"
+	"strings"
 	"sync"
 	"time"
 
@@ -513,11 +514,25 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	// TODO we REALLY need to wait for dispatch to return before conceding our slot
 }
 
+func specialHeader(k string) bool {
+	return k == "Fn_call_id" || k == "Fn_method" || k == "Fn_request_url"
+}
+
 func (a *agent) prepCold(ctx context.Context, call *call, tok ResourceToken, ch chan Slot) {
+	// add additional headers to the config to shove everything into env vars for cold
+	for k, v := range call.Headers {
+		if !specialHeader(k) {
+			k = "FN_HEADER_" + k
+		} else {
+			k = strings.ToUpper(k) // for compat, FN_CALL_ID, etc. in env for cold
+		}
+		call.Config[k] = strings.Join(v, ", ")
+	}
+
 	container := &container{
 		id:      id.New().String(), // XXX we could just let docker generate ids...
 		image:   call.Image,
-		env:     call.EnvVars, // full env
+		env:     map[string]string(call.Config),
 		memory:  call.Memory,
 		timeout: time.Duration(call.Timeout) * time.Second, // this is unnecessary, but in case removal fails...
 		stdin:   call.req.Body,
@@ -577,7 +592,7 @@ func (a *agent) runHot(ctxArg context.Context, call *call, tok ResourceToken) {
 	container := &container{
 		id:     cid, // XXX we could just let docker generate ids...
 		image:  call.Image,
-		env:    call.BaseEnv, // only base env
+		env:    map[string]string(call.Config),
 		memory: call.Memory,
 		stdin:  stdinRead,
 		stdout: stdoutWrite,
--- a/api/agent/agent_test.go
+++ b/api/agent/agent_test.go
@@ -95,14 +95,9 @@ func TestCallConfigurationRequest(t *testing.T) {
 	req.Header.Add("Content-Length", contentLength)
 	req.Header.Add("FN_PATH", "thewrongroute") // ensures that this doesn't leak out, should be overwritten
 
-	// let's assume we got there params from the URL
-	params := make(Params, 0, 2)
-	params = append(params, Param{Key: "YOGURT", Value: "garlic"})
-	params = append(params, Param{Key: "LEGUME", Value: "garbanzo"})
-
 	call, err := a.GetCall(
 		WithWriter(w), // XXX (reed): order matters [for now]
-		FromRequest(appName, path, req, params),
+		FromRequest(appName, path, req),
 	)
 	if err != nil {
 		t.Fatal(err)
@@ -148,7 +143,7 @@ func TestCallConfigurationRequest(t *testing.T) {
 		t.Fatal("GetCall FromRequest should not fill payload, got non-empty payload", model.Payload)
 	}
 
-	expectedBase := map[string]string{
+	expectedConfig := map[string]string{
 		"FN_FORMAT":   format,
 		"FN_APP_NAME": appName,
 		"FN_PATH":     path,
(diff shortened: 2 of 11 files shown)


rdallman:
this is the revival of #370 - i think this is much simpler, i reviewed that PR for context extensively. this should have really minimal breakage compared to the previous attempt at this.

below is some example output that may help clarify, where `ENV` is from the environment (ignore the index printed), `HEADER` is from either stripping off `FN_HEADER_` (default), from request headers (http) or json protocol http header section (json) -- (the code uses fdk-go, for context), and `CONFIG` is basically `ENV`:

json:
```
ENV 0 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
ENV 1 HOSTNAME=debian
ENV 2 FN_TYPE=sync
ENV 3 FN_FORMAT=json
ENV 4 FN_PATH=/hello
ENV 5 FN_MEMORY=128
ENV 6 FN_APP_NAME=hot-app
ENV 7 HOME=/root
HEADER Fn_request_url [http://localhost:8080/r/hot-app/hello]
HEADER User-Agent [curl/7.57.0]
HEADER Accept [*/*]
HEADER Content-Length [17]
HEADER Content-Type [application/x-www-form-urlencoded]
HEADER Fn_call_id [01C2Z43XEJY7WG600000000000]
HEADER Fn_deadline [2018-01-03T15:01:52.002-08:00]
HEADER Fn_method [POST]
CONFIG FN_PATH /hello
CONFIG FN_MEMORY 128
CONFIG FN_APP_NAME hot-app
CONFIG HOME /root
CONFIG PATH /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
CONFIG HOSTNAME debian
CONFIG FN_TYPE sync
CONFIG FN_FORMAT json
```

http: 

```
ENV 0 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
ENV 1 HOSTNAME=debian
ENV 2 FN_PATH=/hello
ENV 3 FN_FORMAT=http
ENV 4 FN_APP_NAME=hot-app
ENV 5 FN_MEMORY=128
ENV 6 FN_TYPE=sync
ENV 7 HOME=/root
HEADER Content-Length [17]
HEADER Content-Type [application/x-www-form-urlencoded]
HEADER Accept [*/*]
HEADER Fn_call_id [01C2Z44TWPY7WGA00000000000]
HEADER Fn_deadline [2018-01-03T15:02:22.150-08:00]
HEADER Fn_method [POST]
HEADER Fn_request_url [http://localhost:8080/r/hot-app/hello]
HEADER User-Agent [curl/7.57.0]
CONFIG FN_PATH /hello
CONFIG FN_FORMAT http
CONFIG FN_APP_NAME hot-app
CONFIG FN_MEMORY 128
CONFIG FN_TYPE sync
CONFIG HOME /root
CONFIG PATH /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
CONFIG HOSTNAME debian
```

default:
```
ENV 0 PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
ENV 1 HOSTNAME=debian
ENV 2 FN_MEMORY=128
ENV 3 FN_HEADER_User-Agent=curl/7.57.0
ENV 4 FN_HEADER_Accept=*/*
ENV 5 FN_HEADER_Content-Length=17
ENV 6 FN_HEADER_Content-Type=application/x-www-form-urlencoded
ENV 7 FN_HEADER_Fn_deadline=2018-01-03T15:20:43.585-08:00
ENV 8 FN_CALL_ID=01C2Z56EGHY7WG200000000000
ENV 9 FN_FORMAT=default
ENV 10 FN_REQUEST_URL=http://localhost:8080/r/hot-app/hello
ENV 11 FN_METHOD=POST
ENV 12 FN_PATH=/hello
ENV 13 FN_TYPE=sync
ENV 14 FN_APP_NAME=hot-app
ENV 15 HOME=/root
HEADER Content-Type [application/x-www-form-urlencoded]
HEADER Fn_call_id [01C2Z56EGHY7WG200000000000]
HEADER Fn_method [POST]
HEADER User-Agent [curl/7.57.0]
HEADER Content-Length [17]
HEADER Fn_deadline [2018-01-03T15:20:43.585-08:00]
HEADER Fn_request_url [http://localhost:8080/r/hot-app/hello]
HEADER Accept [*/*]
CONFIG FN_HEADER_Content-Length 17
CONFIG FN_CALL_ID 01C2Z56EGHY7WG200000000000
CONFIG FN_MEMORY 128
CONFIG FN_HEADER_Content-Type application/x-www-form-urlencoded
CONFIG FN_HEADER_Fn_deadline 2018-01-03T15:20:43.585-08:00
CONFIG FN_REQUEST_URL http://localhost:8080/r/hot-app/hello
CONFIG FN_PATH /hello
CONFIG FN_HEADER_User-Agent curl/7.57.0
CONFIG FN_FORMAT default
CONFIG HOSTNAME debian
CONFIG FN_HEADER_Accept */*
CONFIG FN_METHOD POST
CONFIG FN_TYPE sync
CONFIG FN_APP_NAME hot-app
CONFIG HOME /root
CONFIG PATH /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

skinowski:
we get rid of FN_PARAMs because Fn_request_url already has those?

rdallman:
put some care into that, there's a test case for `FN_PATH` specifically https://github.com/fnproject/fn/blob/master/api/agent/agent_test.go#L96 which should cover most of our similar 'injected' config variables. the story would be, for stuff like this, in cold we prepend `FN_HEADER_` to avoid collision with existing `FN_PATH`, and in hot they are in separate places altogether, with one `FN_PATH` in the http headers, and one in the environment. we are overriding `FN_PATH` (and family) when building the config itself, overwriting any app and route config vars that may use one of our 'reserved' keys -- the prior code did similar. `FN_CALL_ID`(an 'injected header') just outright gets overridden with ours, as well. 

for app and route configs, we could likely disallow setting our 'reserved' set of keys to begin with, and avoid having to return an error at function runtime or hide those vars. if we hide them, though, for any in our reserved set it should be pretty easy to inspect the behavior (i.e. that user overriding doesn't work). 

I can't think of any other cases, which doesn't mean they don't exist. not sure what to do exactly to improve things, we could log any suspects, but I do think users will quickly figure out we're just overriding certain if they try to do that. if we had any header/config collisions, that would be concerning, but I can't think of any cases beyond what we have a test for now, could possibly add one for `FN_CALL_ID` since it's slightly different in nature. ideas?

rdallman:
let's just rip the band aid off, dupe work already going in for this... most people will not have issues with the `s/-/_/` replacement which is the biggest breaking change here, other than placement, which is easily fixed.

PR #663, opened 2018-01-09
add FN_LOG_DEST for logs, fixup init

rdallman:
* FN_LOG_DEST can point to a remote logging place (papertrail, whatever)
* FN_LOG_PREFIX can add a prefix onto each log line sent to FN_LOG_DEST

default remains stderr with no prefix. users need this to send to various
logging backends, though it could be done operationally, this is somewhat
simpler.

we were doing some configuration stuff inside of init() for some of the global
things. even though they're global, it's nice to keep them all in the normal
server init path.

we have had strange issues with the tracing setup, I tested the last repro of
this repeatedly and didn't have any luck reproducing it, though maybe it comes
back.

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -31,6 +31,8 @@ import (
 
 const (
 	EnvLogLevel  = "FN_LOG_LEVEL"
+	EnvLogDest   = "FN_LOG_DEST"
+	EnvLogPrefix = "FN_LOG_PREFIX"
 	EnvMQURL     = "FN_MQ_URL"
 	EnvDBURL     = "FN_DB_URL"
 	EnvLOGDBURL  = "FN_LOGSTORE_URL"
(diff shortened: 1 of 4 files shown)


api/server/server.go
```diff
@@ -206,6 +226,9 @@ func WithAgent(agent agent.Agent) ServerOption {
 // New creates a new Functions server with the opts given. For convenience, users may
 // prefer to use NewFromEnv but New is more flexible if needed.
 func New(ctx context.Context, opts ...ServerOption) *Server {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "server_init")
```
rdallman: this was to try to repro the span panic before the tracer is set up, which wasn't successful. seems useful though to leave, in case it takes forever to ping sql db or something in init (need to add sub spans in init for stuff like that)

rdallman:
cc @derekschultz

rdallman:
done. thanks

PR #664, opened 2018-01-09
Switch to dep from glide

treeder:
Similar to #567 but keeps vendor dir (for now). 

Upgrades a few libs like Docker too. 

Closes #567.

This breaks @rdallman 's licenses.sh thingy, but should be able to swap it out with dep I'm sure, maybe something along these lines: https://github.com/golang/dep#visualizing-dependencies


rdallman:
the license script will be easy to switch over. takin a look, thanks

PR #665, opened 2018-01-09
Sets default content-type in function calls to application/json

treeder:


rdallman:
feels kind of wrong to do this, even though users should set it and it gets overridden if they do, it's still magic behavior from us

treeder:
Should have been like this all along, was surprised to see we've been returning text/plain this whole time.

Issue #640, opened 2018-01-04
update swagger 2 -> openapi 3


rdallman:
swagger is now openapi 🎉 

docs on upgrading: https://blog.readme.io/an-example-filled-guide-to-swagger-3-2/

why we need this: https://swagger.io/docs/specification/media-types/ -- see thread https://github.com/fnproject/cli/pull/128 -- for logs, we can return a json or a plaintext now. if somebody updates the client (or CLI) with an old version of fn, it will return a json still, so the client needs to handle it. swagger 3 lets us do this, swagger 2 does not. we should probably update anyway, just because.

PR #667, opened 2018-01-09
OpenAPI v3.0.0

denismakogon:
Say hello to OpenAPI 3.0, say goodbye to Swagger 2.0

 Closes: #640


denismakogon:
Spec was auto-converted with help of https://mermade.org.uk/api

test.sh
```diff
@@ -29,4 +29,5 @@ go vet $(go list ./... | grep -v vendor)
 
 remove_containers
 
+curl -H 'Accept: application/json' -X GET https://mermade.org.uk/api/v1/validate?url=https://gist.githubusercontent.com/denismakogon/f1c306d771480bbe39b363de070e432f/raw/ad8ef852e88c0cd8fcfbe52d7cfaa6d85ee62d1f/gistfile1.txt | go run test/openapi-validate.go
```
denismakogon: That gist contains the same spec as this PR.

P.S.: i nearly crashed my brain to develop go tool to validate spec with API endpoints of the online validator, it just doesn't work with cURL (if i attempt to validate with the file as the payload, it works only in the browser).

denismakogon:
With all-new OpenAPI 3.0 we would need new client generator script, the tool is available here:
https://github.com/Mermade/openapi-codegen
That tool is a bit bad because it depends on templates from its source repo: https://github.com/Mermade/openapi-codegen/tree/master/templates

That means that we can't/wouldn't use nmp distro just to install that, but we can build our own docker images with all of that inside to keep dev env clean.

Thoughts?
@treeder @rdallman @skinowski

test.sh
```diff
@@ -29,4 +29,5 @@ go vet $(go list ./... | grep -v vendor)
 
 remove_containers
 
+curl -X POST https://mermade.org.uk/api/v1/validate -H "Accept: application/json" -H "Content-Type: multipart/form-data" -F "source=`cat docs/openapi.yml`" | go run test/openapi-validate.go
```
rdallman: 2 things: 

* can nuke the go file for python, grep or jq (jq likely not installed, first 2 / other preferably)
* can we use something that we control or that is offical to do this validation ? this is hosted by some dude on his site (granted, he looks reputable)

rdallman:
> https://github.com/swagger-api/swagger-codegen/issues/6598#issuecomment-356286657

seems like beta is this week. should we wait?

test.sh
denismakogon: There's no official validation tool yet released.

rdallman: is there any tool that we can run in a container instead of an http endpoint then? just don't want our builds to go down because some guy's personal server is offline for a day.

denismakogon: According to official OpenAPI doc, there are couple libraries we can use to build a validation tool, but nothing we can use just out of the box.

denismakogon: https://github.com/OAI/OpenAPI-Specification/blob/master/IMPLEMENTATIONS.md

denismakogon: Node lib: https://github.com/mermade/swagger2openapi

denismakogon: ruby lib looks simple: https://github.com/kevindew/openapi3_parser

denismakogon: @rdallman I made validation tool: https://github.com/fnproject/dockers/pull/8. This PR was updated, until that moment when fnproject/openapiv3-validator:0.0.1 will reach docker hub CI will remain red.

rdallman:
well, i guess openapi 3.0 is re-building the generators from scratch, basically only supports java https://github.com/swagger-api/swagger-codegen/releases/tag/v3.0.0-rc0 -- gee, maybe we should just get running with the template version from Mermade. makes me think we should stick with swagger 2 for as long as we can bear it... 3 is far from ready

MikeRalphson:
Happy to help (mermade guy).

zootalures:
I have some strong preferences around staying on swagger 2.0 for a while - largely around the maturity of generators,  would prefer not to make this change now.

treeder:
Closing for now until we decide to upgrade.

Issue #646, opened 2018-01-05
add metric for log bytes uploaded/downloaded, latency for all log stores


rdallman:
the datastore has a metrics datastore wrapper that adds metrics to each datastore call. we need to do the same for the log store since users (and us) may use a separate one from the datastore.

since `InsertLog` and `GetLog` take an `io.Reader`, we shouldn't read it just to count bytes (full copy), I think there are 2 options:

* make an `io.Reader` type that wraps the log `io.Reader` and counts bytes read from this. after the enclosed (real) `InsertLog`/`GetLog` call is completed and has read the `io.Reader`, we can then add a metric for the size of that log that was uploaded. 
* check that type assertion `sizer, ok := log.(interface{ Size() int64 })` passes. we're using a `bytes.Buffer` at present so this check will pass in `InsertLog` and we can get the size from that. this doesn't cover `GetLog`, however the s3 api does expose the length that we could snatch out, but then this is a plumbing problem.

1) seems more robust and preferable. latency is a pretty straight forward span addition. bytes uploaded / downloaded will be useful to see if we hit caps against any log store we might be using (latency will go up but throughput will stay flat).

gviedma:
Fixed by PR https://github.com/fnproject/fn/pull/670

PR #670, opened 2018-01-10
Replace minio-go with aws-sdk-go for s3-compatible log backend

gviedma:
Minio-go has a couple of deficiencies that are resolved by switching to aws-sdk-go for our s3 client driver.

1. There is a bug when uploading small files in streaming mode (size = -1) that results in an EOF error being thrown by the client. Related to this is the following bug https://github.com/minio/minio-go/issues/848 As a result, it is not possible to use a generic `io.Reader` interface, since the size of the uploaded object must be computed prior to making a request. This would preclude any kind of streaming readers in the future.
2. Minio-go uses a strict date parsing format which makes it throw an unrecoverable error when parsing the `Last-Modified` header in an HTTP response. This makes it *incompatible* with some major cloud vendor S3/Object Storage implementations, for example Oracle's.

Also fixes https://github.com/fnproject/fn/issues/646


api/logs/s3/s3.go
```diff
@@ -85,26 +124,51 @@ func path(appName, callID string) string {
...
+
+func (s *store) GetLog(ctx context.Context, appName, callID string) (io.Reader, error) {
+	span, ctx := opentracing.StartSpanFromContext(ctx, "s3_get_log")
+	defer span.Finish()
+
+	objectName := path(appName, callID)
+	logrus.WithFields(logrus.Fields{"bucketName": s.bucket, "key": objectName}).Debug("Downloading log")
+
+	// stream the logs to an in-memory buffer
+	target := &aws.WriteAtBuffer{}
```
rdallman: have used https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#S3.GetObject w/ some success in the past to avoid having to copy into a buffer here, am less familiar with the `s3Manager` api.

rdallman: happy to punt on optimizing here

gviedma: I think that should work as well. Will look at making the change tomorrow.

gviedma: @rdallman I've looked into replacing this with `s3.GetObject` and the challenge there is that we wouldn't be able to simply return the Body of `s3.GetObjectOutput` as the io.Reader result of `GetLog`. The reason is that the result body needs to be closed in order to not leak connections (see https://github.com/aws/aws-sdk-go/issues/408), so unfortunately we wouldn't be able to avoid copying into an intermediate buffer.

rdallman: gotcha. we could change `LogStore.GetLog` to return an `io.ReadCloser` ?

PR #671, opened 2018-01-10
Change basic stats to use opentracing rather than Prometheus API 

nigeldeakin:
This PR changes the way that the "basic metrics" (calls, queued, running, completed and failed) are sent to Prometheus.

Previously the code in `stats.go` called the Prometheus client API directly.

Now it calls the opentracing API instead. `stats.go` no longer contains any dependence on the Prometheus API. 

Our custom zipkin collector `prom_zip_collector.go`, which previously converted values logged in a tracing span to Prometheus histogram metrics, has now been extended to support Prometheus counter and gauge metrics as well. A prefix on the name of the logged value is used to determine which type of Prometheus metric it corresponds to.

Before this change, these metrics looked something like this (this is what you would see if you looked at `localhost:8080/metrics`):
```
fn_api_calls{app="hello-cold-async-a",path="/hello-cold-async-a1"} 33
fn_api_completed{app="hello-cold-async-a",path="/hello-cold-async-a1"} 3
fn_api_failed{app="hello-cold-async-a",path="/hello-cold-async-a1"} 27
fn_api_queued{app="hello-cold-async-a",path="/hello-cold-async-a1"} 0
fn_api_running{app="hello-cold-async-a",path="/hello-cold-async-a1"} 0
```
After this change, the app and path labels have been changed to be consistent with those used for the duration and docker stats metrics. This is necessary because these are now inherited from the parent span rather than set explicitly. The new metrics are as follows:
```
fn_calls{fn_appname="hello-cold-async-a",fn_path="/hello-cold-async-a1"} 33
fn_completed{fn_appname="hello-cold-async-a",fn_path="/hello-cold-async-a1"} 3
fn_failed{fn_appname="hello-cold-async-a",fn_path="/hello-cold-async-a1"} 27
fn_queued{fn_appname="hello-cold-async-a",fn_path="/hello-cold-async-a1"} 0
fn_running{fn_appname="hello-cold-async-a",fn_path="/hello-cold-async-a1"} 0
```

Note that in addition, the names of the five metrics have been changed to remove the characters `api_`. This is not essential, but renaming the metrics helps avoid confusion given that the label names have changed. In addition the  characters `api_` don't appear to have any purpose.

Anything that uses these five Prometheus metrics needs to be changed to use the new metric names and label names. The code in the `fn` repo that needs to be changed is the first of the three Grafana dashboards. This has been updated and is included in this PR.

Once this PR is in I will look for any other docs that needs to be updated, including my own medium blog post, and update these as well.

In addition the statistics API in the `ext-metricsapi` repo will need to be updated to use the new metric names. This will be done separately.

Please review and approve if appropriate. Please leave me to do the merge so I can make sure that the `ext-metricsapi` repo is updated as well.


nigeldeakin:
Note that this PR does not change the "legacy" stats returned by `localhost:8080/stats`. Although these are deprecated they cannot be removed until the Fn UI is changed to get its data from Prometheus.

nigeldeakin:
We probably in the future want to create a new API which encapsulates the details of setting metrics using open tracing. For now, if you want an example of how to set a counter or gauge metric, look at the functions `IncrementCounter`, `,DecrementCounter` and `IncrementGauge` in `stats.go`. If you want an example of how to set a histogram metric, look at the function `WriteStat` in `agent.go`.

api/agent/agent.go
```diff
@@ -726,14 +728,24 @@ func (c *container) Timeout() time.Duration         { return c.timeout }
...
+	// Spans are not processed by the collector until the span ends, so to prevent any delay
+	// in processing the stats when the function is long-lived we create a new span for every call
 	span, ctx := opentracing.StartSpanFromContext(ctx, "docker_stats")
 	defer span.Finish()
+
 	for key, value := range stat.Metrics {
-		span.LogFields(log.Uint64("fn_"+key, value))
+		// The field name we use is the metric name prepended with "fn_histogram_" to designate that it is a Prometheus histogram metric
+		// The collector will remove "histogram_" and use the result as the Prometheus metric name.
+		fieldname := "fn_histogram_" + "docker_stats_fn_" + key
```
rdallman: we could likely codify this (can shove in common) to make it a little less error prone, as was done with `IncrementGauge`, though every stats interface has its formatting idiosyncrasies:

```go
func Histogram(span opentracing.Span, key string, value int64) {
  span.LogFields(log.Uint64("fn_histogram_"+key, value))
}
```

api/agent/stats.go
```diff
@@ -200,3 +157,61 @@ func (s *stats) Stats() Stats {
 	s.mu.Unlock()
 	return stats
 }
+
+// Constants used when constructing span names and field keys for these metrics
+var queuedSuffix = "queued"
+var callsSuffix = "calls"
+var runningSuffix = "running"
+var completedSuffix = "completed"
+var failedSuffix = "failed"
```
rdallman: go thing: these look like consts more than vars:

```go
const (
  queuedSuffix = "queued"
  // ...
)
```

if passed as args (as I see, for these) that take string parameters, they will get turned into strings. constants in go are pretty cool under the hood, this blog post is pretty interesting https://blog.golang.org/constants

nigeldeakin:
This is now ready to merge. Before I do I will update the `ext-statsapi` repo to work with these changes.

Issue #669, opened 2018-01-10
regression: timeout errors have increased in normal operation


mjg123:
When running locally this error is seen intermittently:

```
{"error":{"message":"Timed out - server too busy"}}
```

This causes a 5xx error to the caller, but I suspect this is a recent regression as it didn't use to happen, and the server isn't actually very busy.

Seen by @carimura @shaunsmith

mjg123:
Possibly related to this patch? https://github.com/fnproject/fn/commit/14789ab

rdallman:
we could curtail the error message to just say 'timed out', it's not always the case that the server is too busy, we return this if it takes too long to pull the image, for example. we can't get rid of timeouts, but the regression here is we're giving more than we're used to I guess (changing issue name to that). the thinking is in the future a load balancer can retry these. anyway, can try to repro, but if it's image pulling, which I suspect, all we can really do is change the error message?

skinowski:
Can I have some info on this?

hot/cold? mixed case?
what is the load? how many req/sec? could you describe the case please?
what's the env? memory?

carimura:
anecdotally i think this has been coming up a bunch... one from community, from others on team, and from me yesterday (although I *was* logged into zoom while demo'ing).

I was just running Vista in Flow mode (locally) and got it.

rdallman:
corollary: maybe we should consider raising the default timeout if things are too tight for everyone. 

since we're exposed through an http endpoint, clients have to retry using their http client, so we can't just add retries in seamlessly in client libraries (since they don't exist for this endpoint). other ux ideas? we will never not have timeouts, need to make it look less like a bug to users somehow.

skinowski:
Hmm, my latest timeout adjustment #658  could be a factor. Now, the timeouts are more accurate since now, it includes couple more DB queries which were not being factored before.

skinowski:
#658 now starts the timer in GetCall() while before #658 the timers started in Submit(), after FromRequest() execution which includes two DB queries.

skinowski:
Closing this, let's reopen if we observe it again.

jan-g:
Reopening this: we are still seeing it (and can provoke it at will); this is blocking fdj-java ci.

It's possible to duplicate what the CI does by judiciously running integration-tests/run-local.sh, but to talk you through it:

- spawns the latest released fn server using /var/run/docker.sock passthrough

- spawns the latest flow server

[...]

PR #673, opened 2018-01-10
fn: hot container launcher adjustment

skinowski:
Latency stats are not always read-time updated and
if calls are stuck in waiting state, isNewContainerNeeded()
needs to be a bit more aggresive if the wait queue grows.

### DIFF
--- a/api/agent/slots.go
+++ b/api/agent/slots.go
@@ -204,11 +204,23 @@ func (a *slotQueue) isNewContainerNeeded() (bool, slotQueueStats) {
 		return false, stats
 	}
 
-	// no executors? Start a container now.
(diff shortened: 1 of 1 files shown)


skinowski:
Fixes: #669

rdallman:
hehe. launching one at a time probably slowed things up a bit, too. the one thread launching one at a time (if needed) should slow things up a little at least to dampen this.

skinowski:
When a burst arrives, the requests pile up in the queue, but none of them proceeds to running state quickly enough, so stats that we make the decisions here end up being very old. Because wait latency didn't update quick enough. So, the system (ignoring queue growth) thinks everything is fine (eg. wait latency is super low)

PR #675, opened 2018-01-11
Add simple test for calling several hot functions in parallel

hhexo:
This PR adds a test to verify we can ~easily call 50 simple hot~ run trivial hot functions in parallel.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,8 +46,6 @@ img-error:
 	docker pull fnproject/error
 img-hello:
 	docker pull fnproject/hello
-img-hot:
-	docker pull hhexo/hot
 img-mysql:
 	docker pull mysql
 img-postgres:
@@ -55,7 +53,7 @@ img-postgres:
 img-minio:
 	docker pull minio/minio
 
-pull-images: img-sleeper img-error img-hello img-hot img-mysql img-postgres img-minio
+pull-images: img-sleeper img-error img-hello img-mysql img-postgres img-minio
 
 test-datastore:
 	cd api/datastore && go test -v ./...
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,8 @@ img-error:
 	docker pull fnproject/error
 img-hello:
 	docker pull fnproject/hello
+img-hot:
+	docker pull hhexo/hot
 img-mysql:
 	docker pull mysql
 img-postgres:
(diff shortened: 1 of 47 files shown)


test/fn-api-tests/exec_test.go
```diff
@@ -66,6 +67,47 @@ func TestCanCallfunction(t *testing.T) {
...
+		s.RouteConfig, s.RouteHeaders)
+
+	u := url.URL{
+		Scheme: "http",
+		Host:   Host(),
+	}
+	u.Path = path.Join(u.Path, "r", s.AppName, s.RoutePath)
+
+	errs := make(chan error)
+	numCalls := 50
```
rdallman: we'll probably want to be a little more precise than this, specifically wrt how much memory do we have & how long do we expect this to take, since depending on the size of the instance in which this test suite is run this can change dramatically. for example, a dev using 1GB of RAM for their vm to dev may not be able to get this passing, even if they didn't 'break' this behavior -- one option is to skip under a certain size. since we're running inside of a CI container we'll likely have to work to make this test not flaky. 

wrt breaking other tests (as mentioned), one option is to move this into the agent or server tests, where you could call `agent.Close()` after the test is over to shut down all the hot containers.

Makefile
```diff
@@ -46,14 +46,16 @@ img-error:
...
+	docker pull hhexo/hot
 img-mysql:
 	docker pull mysql
 img-postgres:
 	docker pull postgres:9.3-alpine
 img-minio:
 	docker pull minio/minio
 
-pull-images: img-sleeper img-error img-hello img-mysql img-postgres img-minio
+pull-images: img-sleeper img-error img-hello img-hot img-mysql img-postgres img-minio
```
skinowski: Please use fn-test-utils.

skinowski:
I think this should be part of perf run, only to be invoked in circle CI (eg. a specific Makefile target called from circle CI), since there we have some sense of what size the machine is. But part of regular test suite, is going to fail in small machines.

Makefile
skinowski: Here's a hot call example:

https://github.com/fnproject/fn/blob/master/api/server/runner_test.go#L247

hhexo: Used fn-test-utils now.

hhexo:
I moved the test to the server tests, and changed it to be just a "minimal concurrency" test - more concerned with functionality than performance now. We'll still have to have a performance test somewhere at some point.

Even on a ridiculously underpowered machine / container, we should be able to run 4 hot functions in parallel within a timeout of 30 seconds.

PR #680, opened 2018-01-12
Use new metrics API for s3 log metrics

gviedma:


nigeldeakin:
Looks good. Please test before merging: you should be able to do 
```
curl localhost:8080/metrics | grep fn_s3_log_upload_size
```
and
```
curl localhost:8080/metrics | grep fn_s3_log_download_size
```
to see the two histogram metrics.

gviedma:
Tested new metrics successfully against `localhost:8080/metrics`

PR #681, opened 2018-01-12
clean up agent.Submit

rdallman:
this was getting bloated with various contexts and spans and stats
administrivia that obfuscated what was going on a lot. this makes some helper
methods to shove most of that stuff into, and simplifies the context handling
around getting a slot by moving it inside of slot acquisition code. also
removed most uses of `call.Model()` -- I'll kill this thing some day, but if a
reason is needed, then the overhead of dynamic dispatch is unnecessary, we're
inside of the implementee for the agent, we don't want to use the interface
methods inside of that.

the only behavior change is (should be) that the slot deadline no longer encloses call.Start(). right now this only changes a few fields and for async deletes the message. we could likely add additional padding like we do on call.End or limit it further just before hand, but it seems fine this way (as opposed to being under slot ctx deadline).


rdallman:
thanks

PR #682, opened 2018-01-12
fn: cpus documentation

skinowski:


skinowski:
I'm already a bit regretting adding 'floating-point' format to this. Maybe milliCPUs itself is good enough and easy to describe as opposed to having both of these two formats. But one could argue that docker uses "0.2" format and it's good to have this in as well as milliCPUs which is same as k8s millicores/milliCPUs.

skinowski:
swagger.yml is already up-to-date (has cpus) with previous PR, no?

PR #685, opened 2018-01-15
Fix for MTU problems in some k8s environments

jan-g:
This should probably be pushed back into the base image, and/or
upstream into dind's master.

In many k8s environments the host docker uses an overlay network
which'll take bytes away from the effective MTU of outer
containers; eg, vxlan needs 50 bytes, often leaving a 1450 MTU
on the container running dind and fn-api.

In such an arrangement, packets exceeding the smaller MTU may be
invisibly dropped as they travel across the dind's docker0
bridge. This mostly surfaces as a failure of functions to be able
to reliably talk to external services. (Note, the failure may be
intermittent depending on the profile of the resulting TCP
communication.)

A robust fix for this is to intercept the startup of the dind
dockerd and ensure that /etc/docker/daemon.json (currently
absent) contains the following setting:

{
  "mtu": 1450
}

(or whatever the MTU on the external interface may be). This
should be autosized so the container works in a variety of
deployments.

The problem does not arise when using an embedded
/var/run/docker.sock - or when running with dind on a host that
can supply 1500-byte MTUs to containers on the 'host' docker.

### DIFF
--- a/images/dind/preentry.sh
+++ b/images/dind/preentry.sh
@@ -6,6 +6,15 @@ if [ $fsdriver == "overlay" ]; then
   fsdriver="overlay2"
 fi
 
+mkdir -p /etc/docker
+/bin/cat > /etc/docker/daemon.json <<EOF
+{
+  "mtu": $(ip link show dev $(ip route |
+                              awk '$1 == "default" { print $NF }') |
+           awk '{for (i = 1; i <= NF; i++) if ($i == "mtu") print $(i+1)}')
+}
+EOF
+
 dockerd-entrypoint.sh --storage-driver=$fsdriver &
 
 # give docker a few seconds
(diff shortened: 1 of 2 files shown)


hhexo:
Can we just add the MTU setup to the preentry script in `images/dind/preentry.sh` instead?

I'm not sure how the base dind works either, probably @treeder knows more.

jan-g:
Done; that's cleaner, but it'll need the fnproject/dind image re-released, and the top-level Dockerfile updated to refer to it.

jan-g:
Right, I've just rebuilt and re-run the same test on my desktop using that patch. Good to merge I think.

hhexo:
LGTM, though I'd defer to @treeder for final approval.

images/dind/preentry.sh
```diff
@@ -6,6 +6,15 @@ if [ $fsdriver == "overlay" ]; then
   fsdriver="overlay2"
 fi
 
+mkdir -p /etc/docker
+/bin/cat > /etc/docker/daemon.json <<EOF
```
rdallman: similar to `--storage-driver`, docker has an option for this one we could pass below: `dockerd --help | grep mtu`

treeder:
@rdallman he updated based on your suggestion in #687 .

PR #687, opened 2018-01-16
Ensure we have the latest fnproject/dind

jan-g:
*To release this:*

Ensure that fnproject/dind:17.12 is rebuilt *and published* first.

At that point the release process should grab it and build a new fnproject/fnserver based off it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,7 @@ jobs:
           #docker version
       - run: docker version
       - run: nproc
+      - run: make clear-images
       - run: make full-test -j $(nproc)
       - deploy:
           command: |
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,14 @@ install:
 checkfmt:
 	./go-fmt.sh
 
+clear-images:
+	-docker images -q -f dangling=true | xargs docker rmi -f
+	for i in fnproject/fn-test-utils fnproject/hello fnproject/error fnproject/sleeper fnproject/error \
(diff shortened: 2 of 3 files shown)


release.sh
```diff
@@ -26,6 +26,7 @@ perl -i -pe 's/\d+\.\d+\.\K(\d+)/$1+1/e' $version_file
 version=$(grep -m1 -Eo "[0-9]+\.[0-9]+\.[0-9]+" $version_file)
 echo "Version: $version"
 
+docker pull fnproject/dind:17.12
```
treeder: It will always get the latest dind during release already, this is probably unecessary.

rdallman: +1 to remove, we have https://github.com/fnproject/fn/blob/master/Dockerfile#L9 -- this line risks going stale vs the Dockerfile and we'd eat the pull

jan-g: While staleness is a problem, are we guaranteed that the docker caches on the ci machines are cleared between runs?

If so, I've no problem with dropping this. On t'other hand, if it doesn't *and* if we shift to "rebuild these tools when changed" then we risk a change on a branch leaving a broken dind behind for main to consume.

jan-g: ... rather than ask I'll check it :-)

rdallman: with circleci they make you pay to have caching, and we're not ponying up, so I'm pretty sure the caches are cleared out (it has been a bit of a thorn, to date)

skinowski: We can explicitly clear them in Makefile (eg. clear-images target) and hook that up to circleci/config.yaml.

Issue #497, opened 2017-11-13
Hot sync function: 504 from fn whilst the container still runs.


jan-g:
I'm seeing the following behaviour from fn. (The run is triggered by flow but the behaviour's the same for functions called directly by the user.)

In this case the function makes a bunch of API calls over a period of (over) two minutes, before finally emitting a small amount of data. After two minutes, fn responds with a 504 timeout. However, I'm still seeing indications that the container continues to run after that 504 is received.

In my opinion (which I invite discussion on) this is error-prone: in this case, fn should probably stop the container (since it doesn't know if the function's run away or not) strictly *before* returning a 504. It's much harder to write correct software as a set of functions if I can't tell whether a timed-out function is going to attempt to trigger more side-effects after fn has returned to me.

Example log extract follows.

Function triggered (this log trace from flow service):

```
time="2017-11-10T12:04:40Z" level=info msg="Running Stage" faas_url="http://fn-service.fn:8080/r" function_id=t/flow-load-test-vista graph_id=6cd61dfb-2e88-47d6-9a9d-d441cca284c0 logger=executor_actor operation=thenCompose stage_id=2 
```

It generates some POSTs to the flow api:

```
[GIN] 2017/11/10 - 12:04:40 | 200 |  5.912722207s |      10.99.57.0 | POST     /graph/6cd61dfb-2e88-47d6-9a9d-d441cca284c0/stage/1/thenCompose
... about 20 lines elided ...
[GIN] 2017/11/10 - 12:05:55 | 200 |  1.776090655s |      10.99.57.0 | POST     /graph/6cd61dfb-2e88-47d6-9a9d-d441cca284c0/stage/28/thenApply
```

Then Fn gives up on the stage ever returning anything (timeout at 2 minutes):

```
time="2017-11-10T12:06:43Z" level=info msg="Fn load balancer delay" faas_url="http://fn-service.fn:8080/r" fn_lb_delay="81.197µs" function_id=t/flow-load-test-vista graph_id=6cd61dfb-2e88-47d6-9a9d-d441cca284c0 logger=executor_actor operation=thenCompose stage_id=2 
time="2017-11-10T12:06:43Z" level=error msg="Got non-200 error from FaaS endpoint" faas_url="http://fn-service.fn:8080/r" function_id=t/flow-load-test-vista graph_id=6cd61dfb-2e88-47d6-9a9d-d441cca284c0 http_status=504 logger=executor_actor operation=thenCompose stage_id=2 
```

[...]

PR #690, opened 2018-01-17
fn: hot container IO mgmt changes

skinowski:
*) Remove go routine protocol Dispatch() to avoid unsafe
concurrent access to http.ResponseWriter headers map.
*) Swap go routine Dispatch() with channel IO select
to detect timeouts and also notify hot container to
shutdown in case of context cancel/timeout.
*) Close read/write pipe ends in hot container exit
to purge lingering IO in pipes.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -361,7 +361,7 @@ func (a *agent) hotLauncher(ctx context.Context, callObj *call) {
 				}(ctx, callObj, tok)
 			} else {
 				// this means the resource was impossible to reserve (eg. memory size we can never satisfy)
-				callObj.slots.queueSlot(&hotSlot{done: make(chan struct{}), err: models.ErrCallTimeoutServerBusy})
+				callObj.slots.queueSlot(&hotSlot{done: make(chan error, 1), err: models.ErrCallTimeoutServerBusy})
 			}
 		case <-time.After(timeout):
 			cancel()
@@ -485,7 +485,7 @@ func (s *coldSlot) Close() error {
 
 // implements Slot
 type hotSlot struct {
-	done      chan<- struct{} // signal we are done with slot
+	done      chan error // signal we are done with slot
 	proto     protocol.ContainerIO
 	errC      <-chan error // container error
 	container *container   // TODO mask this
@@ -493,7 +493,10 @@ type hotSlot struct {
 }
 
 func (s *hotSlot) Close() error {
-	close(s.done)
+	select {
+	case s.done <- nil:
+	default:
+	}
 	return nil
 }
 
@@ -517,24 +520,28 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	oldStderr := s.container.swap(call.stderr, &call.Stats)
 	defer s.container.swap(oldStderr, nil) // once we're done, swap out in this scope to prevent races
 
-	errApp := make(chan error, 1)
+	containerError := make(chan error, 1)
+
 	go func() {
-		// TODO make sure stdin / stdout not blocked if container dies or we leak goroutine
-		// we have to make sure this gets shut down or 2 threads will be reading/writing in/out
-		ci := protocol.NewCallInfo(call.Call, call.req)
-		errApp <- s.proto.Dispatch(ctx, ci, call.w)
+		select {
+		case err := <-s.errC: // error from container
+			containerError <- err
+		case <-ctx.Done(): // timeout or cancel from ctx
+			containerError <- ctx.Err()
+			s.done <- ctx.Err()
+		}
 	}()
 
+	ci := protocol.NewCallInfo(call.Call, call.req)
+	dispatchError := s.proto.Dispatch(ci, call.ioWriter)
+
 	select {
-	case err := <-s.errC: // error from container
+	case err := <-containerError:
 		return err
-	case err := <-errApp: // from dispatch
-		return err
-	case <-ctx.Done(): // call timeout
-		return ctx.Err()
+	default:
 	}
 
-	// TODO we REALLY need to wait for dispatch to return before conceding our slot
+	return dispatchError
 }
 
 func specialHeader(k string) bool {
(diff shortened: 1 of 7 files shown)


skinowski:
Build failure from today:

```
fatal error: concurrent map writes

goroutine 1127 [running]:
runtime.throw(0x11f22f2, 0x15)
	/usr/local/go/src/runtime/panic.go:605 +0x95 fp=0xc4206b3ca0 sp=0xc4206b3c80 pc=0x436f75
runtime.mapassign_faststr(0x10e8b80, 0xc42058ab70, 0xc4204ac5f4, 0xc, 0x2)
	/usr/local/go/src/runtime/hashmap_fast.go:685 +0x4da fp=0xc4206b3d20 sp=0xc4206b3ca0 pc=0x41815a
net/textproto.MIMEHeader.Add(0xc42058ab70, 0xc4204ac5f4, 0xc, 0xc4204ac600, 0x10)
	/usr/local/go/src/net/textproto/header.go:15 +0xf0 fp=0xc4206b3d88 sp=0xc4206b3d20 pc=0x6b3150
net/http.Header.Add(0xc42058ab70, 0xc4204ac5f4, 0xc, 0xc4204ac600, 0x10)
	/usr/local/go/src/net/http/header.go:24 +0x53 fp=0xc4206b3dc0 sp=0xc4206b3d88 pc=0x715b73
github.com/fnproject/fn/api/agent/protocol.(*JSONProtocol).Dispatch(0xc420608f60, 0x1aa1dc0, 0xc42063c9f0, 0x1aa5140, 0xc4203b5950, 0x2b15b8b71910, 0xc420303d90, 0xc42070568f, 0x0)
	/home/circleci/go/src/github.com/fnproject/fn/api/agent/protocol/json.go:172 +0x234 fp=0xc4206b3f38 sp=0xc4206b3dc0 pc=0xd4cf84
github.com/fnproject/fn/api/agent.(*hotSlot).exec.func2(0xc420158ab0, 0xc4205cdda0, 0xc4205b9600, 0x1aa1dc0, 0xc42063c9f0)
	/home/circleci/go/src/github.com/fnproject/fn/api/agent/agent.go:525 +0xe6 fp=0xc4206b3fb8 sp=0xc4206b3f38 pc=0xd6dfb6
runtime.goexit()
	/usr/local/go/src/runtime/asm_amd64.s:2337 +0x1 fp=0xc4206b3fc0 sp=0xc4206b3fb8 pc=0x467fc1
created by github.com/fnproject/fn/api/agent.(*hotSlot).exec
	/home/circleci/go/src/github.com/fnproject/fn/api/agent/agent.go:521 +0x365
```

This is caused by the go routine in hot exec protocol.Dispatch().

api/server/runner_test.go
```diff
@@ -266,9 +266,9 @@ func TestRouteRunnerTimeout(t *testing.T) {
 	}{
 		{"/r/myapp/cold", `{"sleepTime": 0, "isDebug": true}`, "POST", http.StatusOK, nil},
 		{"/r/myapp/cold", `{"sleepTime": 5000, "isDebug": true}`, "POST", http.StatusGatewayTimeout, nil},
-		{"/r/myapp/hot", `{"sleepTime": 5000, "isDebug": true}`, "POST", http.StatusGatewayTimeout, nil},
+		{"/r/myapp/hot", `{"sleepTime": 50000, "isDebug": true}`, "POST", http.StatusGatewayTimeout, nil},
```
skinowski: sleep 50000 will cause subsequent requests to fail.

skinowski:
Related #497

api/agent/agent.go
```diff
@@ -517,24 +520,28 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
...
+		case err := <-s.errC: // error from container
+			containerError <- err
+		case <-ctx.Done(): // timeout or cancel from ctx
+			containerError <- ctx.Err()
+			s.done <- ctx.Err()
+		}
 	}()
 
+	ci := protocol.NewCallInfo(call.Call, call.req)
+	dispatchError := s.proto.Dispatch(ci, call.ioWriter)
```
rdallman: the issue here is that i/o could put this over the timeout, so this isn't really an option (sadly). i have a test that can repro this, repro'ing a different issue. 

as far as fixing the race to write to the header bucket -- it seems really easy to slide by making a thread safe `http.ResponseWriter` that we hand into the agent. this would be easily removed once we have fixed the root causal issue (opposed to shuffling things around twice, if the case)

api/agent/agent.go
```diff
@@ -690,7 +700,11 @@ func (a *agent) runHot(ctxArg context.Context, call *call, tok ResourceToken) {
 
 			// wait for this call to finish
 			// NOTE do NOT select with shutdown / other channels. slot handles this.
-			<-done
+			err := <-done
```
rdallman: this seems less than ideal, if a user's task times out then it causes the container to shut down. we don't need to do this, we just need to strip the pipes away from them -- right now, we're sharing pipes between tasks. if we shut down every timeout we increase container churn a lot even though it's not really necessary. actually started on fixing this yday, have a test that fusses now and was gonna fix it up.

api/agent/agent.go
skinowski: we catch the I/O timeout from ctx in the go routine. If dispatch takes forever, then go routine picks up timeout from ctx, then signals container to stop.

api/agent/agent.go
skinowski: so, the pipe swap will not fix this. If the container has a process that has failed, we can no longer pass new requests to it. Some examples; a broken function that continues to spit out output or reads partial input. A function that sleeps, etc. Similar example as a broken http req/resp, once the req/resp structure is broken, we can no longer issue requests through the same connection.

rdallman: this presumes the container's process will never get back to resuming correct operation. with deadline now, I think there's some hope that in the timeout case this isn't true (this only covers the sleep case if used properly).

the incorrect input/output spitting out is the case I was gunning for -- for this, the trade off is what's the likelihood that making another container of that image will simply do the same thing (i.e. not just a rare funny state)? it seems rather high, I think, and makes sense to not go about making a lot of them and instead keeping just one that goofs and errors. the goal here was to make this situation easier to debug, to be exact, right now it's not very clear.

if a container responds beyond the boundary with a response for a prior call, I can think of a couple things that might happen (need to test to see which). It seems like ideally... that write to stdout gets ignored, it resumes reading stdin and hops to the next task? 

gonna dig around on this, I think there's a possibility to recover at least in prominent cases.

api/agent/agent.go
rdallman: that makes sense, made this comment before the other (trying to avoid container shutdown, if possible)

skinowski:
Closing this as many relevant changes went in since this proposal.

PR #694, opened 2018-01-17
fn: cancellations in WaitAsyncResource

skinowski:
*) Locked broadcast to avoid missed wake-ups.
*) Added go context with cancel to wait async resource. Although
today, the only case for cancellation is shutdown, this cleans
up agent shutdown a little bit.

### DIFF
--- a/api/agent/async.go
+++ b/api/agent/async.go
@@ -16,7 +16,7 @@ func (a *agent) asyncDequeue() {
 	defer cancel()
 
 	for {
-		ch, cancelWait := a.resources.WaitAsyncResource(ctx)
+		ch, cancelWait := a.resources.WaitAsyncResource()
 		select {
 		case <-a.shutdown:
 			cancelWait()
--- a/api/agent/async.go
+++ b/api/agent/async.go
@@ -16,10 +16,13 @@ func (a *agent) asyncDequeue() {
 	defer cancel()
 
 	for {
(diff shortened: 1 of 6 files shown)


api/agent/async.go
```diff
@@ -16,10 +16,13 @@ func (a *agent) asyncDequeue() {
 	defer cancel()
 
 	for {
+		ch, cancelWait := a.resources.WaitAsyncResource()
```
rdallman: one go nit: this & startDequeuer both have interfaces that return a cancel func instead of taking a context. as far as familiar interfaces for gophers go, it looks like the context methods in general which seems sane, but would generally expect the function to take a context which the caller would cancel, with the callee checking `ctx.Done` as opposed to returning a cancel func that the caller must explicitly close themselves. the main difference being in the former, an enclosing thread of the caller could cancel the context and this would be propagated. I do think it would clean up some of the calls to cancel funcs that have to be slid into certain case statements inside of a for loop select. some shifting around of the special cancel funcs may be necessary (they'd turn into goroutines that wait to get a <-ctx.Done then do their business instead of being a cancel func)

skinowski: I agree, but I prefer removing them from args, because go condition variables and contexts do not mix well. If we pass the ctx to the function as arg, it gives the false impression that the function will cancel properly if that ctx was cancelled. But that's misleading since, we in reality require returned cancel() function to be called.

PR #696, opened 2018-01-18
fn: resource and slot cancel and broadcast improvements

skinowski:
*) Context argument does not wake up the waiters correctly upon
cancellation/timeout. Returning a cancel function clarifies
this and honors this contract better.
*) Avoid unnecessary broadcasts in slot and resource.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -344,15 +344,15 @@ func (a *agent) hotLauncher(ctx context.Context, callObj *call) {
 			continue
 		}
 
-		resourceCtx, cancel := context.WithCancel(context.Background())
+		ctxResource, cancelResource := context.WithCancel(context.Background())
 		logger.WithFields(logrus.Fields{
 			"currentStats":  curStats,
 			"previousStats": curStats,
 		}).Info("Hot function launcher starting hot container")
 
 		select {
-		case tok, isOpen := <-a.resources.GetResourceToken(resourceCtx, callObj.Memory, uint64(callObj.CPUs), isAsync):
-			cancel()
+		case tok, isOpen := <-a.resources.GetResourceToken(ctxResource, callObj.Memory, uint64(callObj.CPUs), isAsync):
+			cancelResource()
 			if isOpen {
 				a.wg.Add(1)
 				go func(ctx context.Context, call *call, tok ResourceToken) {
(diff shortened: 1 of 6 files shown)


skinowski:
same cond var is held by many waiters, signal may wake up an unrelated waiter, so it has to be broadcast in this case.

rdallman:
> same cond var is held by many waiters, signal may wake up an unrelated waiter, so it has to be broadcast in this case.

d'oh, that's right.

skinowski:
This one is a bit cleaner I think... Preserved the ctx arg and spawned a go routine...

rdallman:
thanks, it LGTM. we need to add logs/span to the asyncDequeue routine (separate, thinking out loud). 

one catch: I haven't examined every context of get resource token / wait resource but it seems like in cases where we find the resource token but the enclosing context has not yet finished we have a go routine hanging out until the context is finished. (apologies, I could have improved my suggestion the first go round). since everything has a timeout, it will go away, but it does mean we have more goroutines than we need active most of the time. i think something like this fixes, though I don't like having two separate contexts in the same scope generally, this does get rid of the `isWaiting` bit with the same behavior:

```go
func (a *resourceTracker) GetResourceToken(ctx context.Context, memory uint64, cpuQuota uint64, isAsync bool) <-chan ResourceToken {
  ch := make(chan ResourceToken)
  ctxWait, cancel := context.WithCancel(ctx)
  
  go func() {
    select {
    case <-ctx.Done():
    case <-ctxWait.Done():
      c.Broadcast()
    }
  }()

  go func() {
    defer cancel()
    // ... await resource
  }()
  return ch
}
```

skinowski:
the bool is more efficient.  we shouldn't get rid of it. The reason is that it bounds the broadcast blast to a much smaller window. defer here means we broadcast as long as the other go routine is running and as long as the consumer hasn't read the token.. we'd be waking up everyone.

If the caller does not want this ctx to survive as long then the caller should send a smaller ctx window. The caller could spin up a ctx for that context and defer cancel it within the hotWait or coldWait scopes as opposed to entire func execution ctx deadline.

We could... introduce the cancel() function back (and avoid that go routine) but also take ctx for logging, but all of these, it's callers responsibility to decide how long to wait.

rdallman:
yes, simply talking about managing the extra go routine we launch surviving the resource actually being fulfilled. made an edit that fixes the broadcast blast I think, since we wouldn't broadcast blast in the case we found a token*

skinowski:
I'll reduce the scope of those contexts.. changes are coming. :-)

skinowski:
should I merge?

rdallman:
yep, that method works too. push away

PR #698, opened 2018-01-18
Docs tutjava - Created a Java Hello World Tutorial

michael-w-williams:
Created a Java Hello World tutorial. This is fairly redundant with the current README.md for fdk-java. However, if we have this tutorial, this frees us up to focus on only FDK features in that README. Anyway, open to discussion on this.


examples/tutorial/hello/java/README.md
```diff
@@ -1,3 +1,188 @@
...
+
+* `version`: Automatically starting at 0.0.1.
+* `runtime`: Set to `java` from the command line.
+* `cmd`: Name of the method to invoke. In this case `handleRequest` of the `com.example.fn.HelloFunction` class.
+
+These fields are set by default when you run `init` on a function. For more details on [function files go here](../../../../docs/developers/function-file.md).
+
+## Add Fn Registry Environment Variable
+
+Before we start developing we need to set the `FN_REGISTRY` environment variable. Normally, set the value to your Docker Hub username. However, you can work with Fn locally.  Set the `FN_REGISTRY` variable to an invented value: `noreg`.
```
shaunsmith: We should say it's preferable to set FN_REGISTRY to your Docker Hub id but you can use an invented value if only working locally and not pushing to Docker Hub.

examples/tutorial/hello/java/README.md
```diff
@@ -1,3 +1,188 @@
...
+When execution is complete, the function returns output to `stout`. In this case, `fn run` returns:
+
+```txt
+Hello World!
+```
+
+To pass data to our function, pass input to `stdin`. You could pass text data to your function like this:
+
+>```sh
+>echo "Johnny" | fn run
```
shaunsmith: Use `echo -n "Johnny"` to avoid the addition of a \n to the input string.

michael-w-williams:
Updated the wording for FN_Registry and fixed echo -n.

michael-w-williams:
This is replaced by the language tutorials in the tutorials repo. Closing.

Issue #555, opened 2017-11-30
Move fnlb to it's own repo


treeder:
github.com/fnproject/lb

PR #702, opened 2018-01-18
Fnlb was moved to its own repo: fnproject/lb

denismakogon:
https://github.com/fnproject/lb

Closes: #555


rdallman:
and the ebb of the mono repo goes out... one day to flow right back in... :)

rdallman:
yea, this https://github.com/fnproject/fn/blob/master/release.sh#L51 good catch -- i grepped for `fnlb` and everything else looks innocuous

rdallman:
oh, there's a test harness that should be moved, too. everything else is innocuous*

rdallman:
can't link to frickin readme lines (wth github, it renders in markdown only) but there's also a link in the readme

denismakogon:
Weird things are happening, newer our docker client can't talk to docker somehow. I filed an issue against docker client we use: https://github.com/fsouza/go-dockerclient/issues/705

rdallman:
thanks. we could lock that version of the docker client to vendor until your update comes in, and then proceed here without any changes necessary to the docker driver.

denismakogon:
@rdallman was able to lock the docker client version and get newer deps, so the code works fine without any changes.

PR #703, opened 2018-01-18
add config placement info to docs

rdallman:
this behavior was recently cemented but was entirely omitted from the doc on
'how to write functions'


rdallman:
thx

PR #704, opened 2018-01-19
fn: agent slot improvements

skinowski:
*) Stopped using latency previous/current stats, this
was not working as expected. Fresh started agent usually has
these stats as zero for a long time, and initial samples
are high due to downloads, caches, etc.

*) New state to track: containers that are idle. In other
words, containers that have an unused token in the slot
queue.

*) Removed latency counts that used to ignore initial samples,
since these are not used in container start decision anymore.
Also simplifies logs.

*) isNewContainerNeeded() simplified to use idle count
to estimate effective waiters. Removed speculative
latency based logic and progress check comparison.
In agent, waitHot() delayed signalling compansates
for these changes. The estimation may fail, but
this should correct itself in the next 200 msec
signal.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -319,7 +319,6 @@ func (a *agent) hotLauncher(ctx context.Context, callObj *call) {
 	logger := common.Logger(ctx)
 	logger.WithField("launcher_timeout", timeout).Info("Hot function launcher starting")
 	isAsync := callObj.Type == models.TypeAsync
-	prevStats := callObj.slots.getStats()
 
 	for {
 		select {
@@ -334,11 +333,10 @@ func (a *agent) hotLauncher(ctx context.Context, callObj *call) {
 		}
 
 		curStats := callObj.slots.getStats()
-		isNeeded := isNewContainerNeeded(&curStats, &prevStats)
-		prevStats = curStats
+		isNeeded := isNewContainerNeeded(&curStats)
 		logger.WithFields(logrus.Fields{
-			"currentStats":  curStats,
-			"previousStats": curStats,
+			"currentStats": curStats,
+			"isNeeded":     isNeeded,
 		}).Debug("Hot function launcher stats")
 		if !isNeeded {
 			continue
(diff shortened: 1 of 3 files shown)


api/agent/slots.go
```diff
@@ -270,14 +240,7 @@ func (a *slotQueue) recordLatencyLocked(metricIdx SlotQueueMetricType, latency u
...
 	// some noise. For our purposes, newer observations are much more important
 	// than older, but we still would like to low pass some noise.
-	// first samples are ignored.
-	if a.stats.latencyCount[metricIdx] != 0 {
-		a.stats.latencies[metricIdx] = (a.stats.latencies[metricIdx]*5 + latency*5) / 10
-	}
-	a.stats.latencyCount[metricIdx] += 1
-	if a.stats.latencyCount[metricIdx] == 0 {
-		a.stats.latencyCount[metricIdx] += 1
-	}
```
skinowski: We used to ignore these first samples using these counters as a gate, but now with this commit, it's better to not hide them for visibility/logs. Initial samples will be smoothed over zero.

api/agent/agent.go
```diff
@@ -673,22 +671,28 @@ func (a *agent) runHot(ctxArg context.Context, call *call, tok ResourceToken) {
 			}
 
 			done := make(chan struct{})
+			start := time.Now()
+			call.slots.enterState(SlotQueueIdle)
 			s := call.slots.queueSlot(&hotSlot{done, proto, errC, container, nil})
 
 			select {
 			case <-s.trigger:
+				call.slots.exitStateWithLatency(SlotQueueIdle, uint64(time.Now().Sub(start).Seconds()*1000))
```
rdallman: super nit: every call to `exitStateWithLatency` uses `uint64(time.Now().Sub(start).Second()*1000`, seems like we should take a `time.Duration` to just be `time.Now().Sub(start)` (or alternatively, just take `start`, or something else idk) and do any conversion inside of `exitStateWithLatency`

skinowski: np, I'll fix this with separate pr.

skinowski:
initially, I was thinking about making waiter/launcher a bit more restrictive too. Historically, I think earlier versions of this code even limited this to 1 starter at a time. But when I flood my desktop with say 50 or 100 req at once, the thundering herd launch actually works fine. (In reality, that many containers will be launched anyway.) But Image pull phase is probably more problematic. In future, I could see us changing this logic into 1) if not pulled, then limit at x 2) if pulled already, then limit at y, but I just don't know what those numbers are yet...

PR #705, opened 2018-01-20
fix the http protocol dumper

rdallman:
we were using a modified httputil.DumpRequest when there is a perfectly good
req.Write method hanging out in the stdlib, that even does the chunked thing
that a few people ran into if they don't provide a content length:
https://golang.org/pkg/net/http/#Request.Write -- so we shouldn't run into
that issue again. I hit this in testing and it was not very fun to debug, so
added a test that repro'd it on master and fixes it here. of course, adding a
content length works too. tested this and it appears to work pretty well, also
cleaned up the control flow a little bit in http protocol.

### DIFF
--- a/api/agent/protocol/http.go
+++ b/api/agent/protocol/http.go
@@ -3,11 +3,8 @@ package protocol
 import (
 	"bufio"
 	"context"
-	"fmt"
 	"io"
 	"net/http"
-	"net/http/httputil"
-	"strings"
 )
 
 // HTTPProtocol converts stdin/stdout streams into HTTP/1.1 compliant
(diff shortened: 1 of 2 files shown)


api/agent/protocol/http.go
```diff
@@ -26,123 +23,40 @@ func (p *HTTPProtocol) IsStreamable() bool { return true }
...
-
-// Return value if nonempty, def otherwise.
-func valueOrDefault(value, def string) string {
-	if value != "" {
-		return value
+	if resp.StatusCode > 0 {
+		rw.WriteHeader(resp.StatusCode)
 	}
-	return def
+	io.Copy(rw, resp.Body)
```
skinowski: Also for the previous code (and json side too): we are not checking error cases from Write(), WriteHeader() and io.Copy().

PR #708, opened 2018-01-22
Selective releasing

jan-g:
This needs a bit of an eyeball.

We check to see which bits of the tree have changed (this is a broad-grained test) before
forcing rebuilds of any dependencies (DIND in this case).

When it comes to releasing, we only release bits that've changed. Bump tags and versions
in code where necessary.

The fnserver release is basically unchanged; fnlb and DIND modified to work around it.

(I've added a handful of relevant reviewers as this is a bit fiddly.)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,7 +30,27 @@ jobs:
       - run: docker version
       - run: nproc
       - run: make clear-images
-      - run: make full-test -j $(nproc)
+
+      # Work out what needs rebuilding and what has changed
+      - run: |
+          ( ./changed-since-last.sh
+            echo "export DIND_NEEDED DIND_TAG DIND_PREV"
+            echo "export FN_NEEDED FN_TAG FN_PREV"
+          ) >> $BASH_ENV
+
+      # Rebuild the build tools if necessary
+      # This leaves local docker images around
+      - run: |
+          if [[ -n "$DIND_NEEDED" ]]; then
+            make build-dind
+          fi
+
+      # Rebuild fnserver if necessary
+      - run: |
+          if [[ -n "$FN_NEEDED" ]]; then
+            make full-test -j $(nproc)
+          fi
+
       - deploy:
           command: |
             if [[ "${CIRCLE_BRANCH}" == "master" && -z "${CIRCLE_PR_REPONAME}" ]]; then
@@ -38,5 +58,10 @@ jobs:
               git config --global user.email "ci@fnproject.com"
               git config --global user.name "CI"
               git branch --set-upstream-to=origin/${CIRCLE_BRANCH} ${CIRCLE_BRANCH}
-              ./release.sh
+              if [[ -n "$DIND_NEEDED" ]]; then
+                make release-dind
+              fi
+              if [[ -n "$FN_NEEDED" ]]; then
+                make release-fnserver
+              fi
             fi
--- a/changed-since-last.sh
+++ b/changed-since-last.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 set -e
 
--- a/changed-since-last.sh
+++ b/changed-since-last.sh
@@ -1,0 +1,43 @@
+#!/bin/sh
+
+set -e
+
+# Has a subset of the tree changed since the last tag of a particular kind?
+# (If so, we'll need to rebuild and re-release, assuming the tests pass.)
+
+RELEASE_BRANCH=origin/master
+FIRST_COMMIT="$(git rev-list "$RELEASE_BRANCH" | tail -1)"
+
+FN_TAG="$(git tag --merged "$RELEASE_BRANCH" --sort='v:refname' '[0-9]*' | tail -1)"
+FN_PREV="$FN_TAG"
+[[ -z "$FN_TAG" ]] && FN_TAG="$FIRST_COMMIT"
+[[ -z "$FN_PREV" ]] && FN_PREV=0.0.0
+
+DIND_TAG="$(git tag --merged "$RELEASE_BRANCH" --sort='v:refname' 'dind-*' | tail -1)"
+DIND_PREV="$DIND_TAG"
+[[ -z "$DIND_TAG" ]] && DIND_TAG="$FIRST_COMMIT"
+[[ -z "$DIND_PREV" ]] && DIND_PREV=dind-0.0.0
+
+# Which pieces of the tree are changed since to each tag?
+# We are only interested in parts of the tree corresponding to each tag's aegis
+# We are *not* interested in solely-DIND changes if we're considering a release of fnserver
+
+# DIND bumps only if there are changes under images/dind.
+[[ -n "$(git diff --dirstat=files,0,cumulative "$DIND_TAG" | awk '$2 ~ /^(images\/dind)\/$/')" ]] && DIND_NEEDED=yes
+
+# FN bumps only if there are changes *other* than images/dind/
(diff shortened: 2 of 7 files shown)


treeder:
Think the LB has moved now, want to rebase this?

jan-g:
Finally got back to this. Rebased; it now selectively builds and releases DIND and fnserver.

.circleci/config.yml
```diff
@@ -30,13 +30,38 @@ jobs:
...
+      # Rebuild the build tools if necessary
+      # This leaves local docker images around
+      - run: |
+          if [[ -n "$DIND_CHANGED" ]]; then
+            make build-dind
+          fi
+
+      # Rebuild fnserver if necessary
+      - run: |
+          if [[ -n "$FNLB_CHANGED" ]]; then
```
rdallman: think this is either misleading, wrong or cruft from removal :)

rdallman: `FN_CHANGED` right?

rdallman: actually I think we should probably do `make full-test` unconditionally really ?

.circleci/config.yml
```diff
@@ -30,13 +30,38 @@ jobs:
...
             if [[ "${CIRCLE_BRANCH}" == "master" && -z "${CIRCLE_PR_REPONAME}" ]]; then
               docker login -u $DOCKER_USER -p $DOCKER_PASS
               git config --global user.email "ci@fnproject.com"
               git config --global user.name "CI"
               git branch --set-upstream-to=origin/${CIRCLE_BRANCH} ${CIRCLE_BRANCH}
-              ./release.sh
+              if [[ -n "$DIND_CHANGED" ]]; then
+                make release-dind
+              fi
+              if [[ -n "$FN_CHANGED" ]]; then
```
rdallman: this should probably be every time (unconditional) too? if dind changes, so does fn, otherwise it's just fn changes in theory (or a bug in change checking).

changed-since-last.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# Which pieces of the tree are changed since to each tag?
+# We are only interested in parts of the tree corresponding to each tag's aegis
+# We are *not* interested in solely-DIND or solely-FNLB changes if we're considering
+# a release of fnserver
+
+# DIND bumps only if there are changes under images/dind.
+[[ -n "$(git diff --dirstat=files,0,cumulative "$DIND_TAG" | awk '$2 ~ /^(images\/dind)\/$/')" ]] && DIND_CHANGED=yes
+
+# FN bumps only if there are changes *other* than fnlb/ or images/dind/
```
rdallman: `s/fnlb\/ or//`

changed-since-last.sh
```diff
@@ -0,0 +1,44 @@
...
+[[ -z "$FN_PREV" ]] && FN_PREV=0.0.0
+
+DIND_TAG="$(git tag --merged "$RELEASE_BRANCH" --sort='v:refname' 'dind-*' | tail -1)"
+DIND_PREV="$DIND_TAG"
+[[ -z "$DIND_TAG" ]] && DIND_TAG="$FIRST_COMMIT"
+[[ -z "$DIND_PREV" ]] && DIND_PREV=dind-0.0.0
+
+# Which pieces of the tree are changed since to each tag?
+# We are only interested in parts of the tree corresponding to each tag's aegis
+# We are *not* interested in solely-DIND or solely-FNLB changes if we're considering
```
rdallman: s/or solely-FNLB//

changed-since-last.sh
```diff
@@ -0,0 +1,44 @@
...
+# a release of fnserver
+
+# DIND bumps only if there are changes under images/dind.
+[[ -n "$(git diff --dirstat=files,0,cumulative "$DIND_TAG" | awk '$2 ~ /^(images\/dind)\/$/')" ]] && DIND_CHANGED=yes
+
+# FN bumps only if there are changes *other* than fnlb/ or images/dind/
+[[ -n "$(git diff --dirstat=files,0,cumulative "$FN_TAG" | awk '$2 !~ /^(images\/dind|fnlb)\/$/')" ]] && FN_CHANGED=yes
```
rdallman: s/|fnlb//

images/dind/build.sh
```diff
@@ -1,3 +1,17 @@
 set -ex
 
 docker build --build-arg HTTPS_PROXY --build-arg HTTP_PROXY -t fnproject/dind:latest .
+
+# Match version with Docker version
+version=$(docker run --rm -v "$PWD":/app treeder/bump  --extract --input "`docker -v`")
```
rdallman: this uses the docker version from `treeder/bump` ? is this right?

.circleci/config.yml
jan-g: The changed-since-last.sh script will set the one on the basis of the other. I'll rename these to make that more obvious.

changed-since-last.sh
jan-g: done

changed-since-last.sh
jan-g: done

images/dind/build.sh
jan-g: Ye-es. Ish. What we don't do (at the moment) is bump the dind:17.12 requirement in the fnserver dockerfile.

I *think* the thing to do here is to tag and push an fnproject/dind:latest - *and* to refer to that from the fnserver dockerfile.

jan-g: (pushed a second commit with that change too)

images/dind/release.sh
```diff
@@ -1,19 +1,44 @@
-set -ex
+#!/bin/bash
 
-# ./build.sh
```
treeder: Can we keep this in there?  All of these scripts are intended to be able to run and do their entire thing (ie: release.sh requires build.sh).

images/dind/release.sh
```diff
@@ -1,19 +1,44 @@
...
 
+# Calculate new release version
+DIND_NEW=$(echo "$DIND_PREV" | perl -pe 's/\d+\.\d+\.\K(\d+)/$1+1/e')
+
+# Add appropriate docker tags
 docker tag fnproject/dind:latest fnproject/dind:$version
 # be nice to have bump image do all of this tagging and pushing too (mount docker sock and do it all)
 docker tag fnproject/dind:$version fnproject/dind:$Mm
 docker tag fnproject/dind:$version fnproject/dind:$M
+docker tag fnproject/dind:$version fnproject/dind:release-$DIND_NEW
```
treeder: Curious what this tag is for?

images/dind/release.sh
jan-g: That was already commented out :-)

The point of this is to release automatically; we shouldn't need to run this manually at all - but if we do, `build.sh && release.sh` doesn't seem to hurt.

As it is, I'd rather that CI built things once, tested those then released the same artifacts. If you want a manual release.sh for use in extremis I don't mind having it call build.sh and (say) push-release.sh or something; ideally a flag with a longer character count than the characters in `build.sh &&` :-)

I'm a (soft, not dogmatic) -1 on the idea of rebuilding anything we've just tested in order to make a release.

images/dind/release.sh
jan-g: Historical traceability / reproducability. You can find the previous DIND tag for any fnserver release and get hold of the image it was built atop of.

(Obviously you could also dive in and grab the docker layers directly; this is just more explicit.)

jan-g:
Merged (can rejig the release.sh behaviour with a later patch if required). Let's get this moving forward for now.

PR #713, opened 2018-01-23
WIP: I/O through a port

treeder:
This enables a function `type` called `app`, which is intended for a function written as an app. 

It also, enables a wildcard route at the root level, ie: `path: /*` (re: #313).

This does NOT help with concurrency (these types of functions could handle concurrent requests). cc @skinowski, re your slack comment today.  

Simple ruby example func included and works. 

## Background

After seeing some similar things out there, there have been many discussions how that could look within Fn. `fn deploy` is such a nice and simple way to deploy a function, why not a containerized app? A hot function is essentially the same as a long running app, just that it can only do one thing and can only be reached by one endpoint. The way Fn scales and deals with hot functions could easily apply to an application too. Getting a lot of traffic?  Add more containers and load balance between them. Sitting idle for X seconds? Kill it. 

## TODO

* [ ] I was noticing sometimes that the first request to a new container would fail. 
* [ ] Automatically use format `port-http` if type is app (and only allow that combination).


skinowski:
I'm excited about this. I was expecting "port" to be an orthogonal attribute though (eg. like cpu, memory).
For example, an async route that uses json, but binds to a port. If port is enabled, we spawn an non-interactive container and all container needs to do is to bind/listen/accept, but then route that connection through json or http depending on protocol chosen. (Eventually perhaps all of this could go over http2 under the hood to cut down num of connections plus we could do server push.)

skinowski:
If we could use non-interactive containers, we might (a bit experimental it seems, but) be able to use:

https://criu.org/Docker

to checkpoint/start containers quickly. (They don't support interactive containers seems like.)

zootalures:
If you want to run long-lived containers  there are lots of things you ultimately want to take into account of  - ultimately you end up implementing something like kubernetes don't you? 

I would prefer to see a better integration between fn as it now and k8s than a cheap version of k8s creeping into fn. 

What if I need to kill the node that the container has started on - do I have to start this up again? 
Wr.t. this - In a distributed install behind a LB  with multiple fn servers where (no which host) would that port land?
would containers be able to talk to each other on these ports? 

I think solving wildcard at the gateway/ingress layer of fn (with everything else staying the same) looks better from a user perspective (i.e. solve the path-matching problem for current FN calls) is my preference.

skinowski:
@zootalures There's no change in hot container semantics. The port discussed in this PR (please correct me on this) is not a port that the clients actually talk to, but a (aka internal) port that fn server itself talks to.

api/agent/agent.go
```diff
@@ -632,10 +635,25 @@ func (a *agent) runHot(ctxArg context.Context, call *call, tok ResourceToken) {
...
 		stderr: &ghostWriter{inner: stderr},
 	}
-
+	if call.Type == models.TypeApp {
+		// then we'll open a port
+		max := 60000
+		min := 55000
+		cport := rand.Intn(max-min) + min // port for container
+		container.port = strconv.Itoa(cport)
+		call.Format = models.FormatPortHTTP
```
skinowski: we should let docker pick this port dynamically. (eg. publish-all, then find out actual port number)

api/datastore/sql/sql.go
```diff
@@ -560,7 +560,8 @@ func (ds *sqlStore) RemoveRoute(ctx context.Context, appName, routePath string)
 }
 
 func (ds *sqlStore) GetRoute(ctx context.Context, appName, routePath string) (*models.Route, error) {
-	rSelectCondition := "%s WHERE app_name=? AND path=?"
+	fmt.Println("in GetRoute")
+	rSelectCondition := "%s WHERE app_name=? AND (path=? OR path='/*')"
```
skinowski: GetRoute looks like it evaluates one row. In this case, we should prefer the non '/*' match (exact match) if both are present.

treeder:
@zootalures this is more along the lines of Azure Container Instances or Hyper.sh. 

> If you want to run long-lived containers there are lots of things you ultimately want to take into account of - ultimately you end up implementing something like kubernetes don't you?
> I would prefer to see a better integration between fn as it now and k8s than a cheap version of k8s creeping into fn.

I don't think Kubernetes is a great end user experience, definitely not as nice as `fn deploy` and you're done. And this is why things like Container Instances exist.

> What if I need to kill the node that the container has started on - do I have to start this up again? Wr.t. this - In a distributed install behind a LB with multiple fn servers where (no which host) would that port land? would containers be able to talk to each other on these ports?

Everything is exactly the same as hot functions, which is why this PR is pretty simple. App in use, keep it hot. App getting busy, scale it out. App not in use, kill it. 

> I think solving wildcard at the gateway/ingress layer of fn (with everything else staying the same) looks better from a user perspective (i.e. solve the path-matching problem for current FN calls) is my preference.

This does need to solved, but it's somewhat independent of this. Needs to be solved for functions too, not just for this.

treeder:
Going to close this for now, until we decide to move forward with something like this.

rosko:
Is there some news? What did you decide regarding things like that?

PR #716, opened 2018-01-23
additional ctx spans / maid service

rdallman:
* add root spans to async calls (oops) so we can see the span started in async land
* there were a few methods which had multiple contexts which existed in the same
scope (this doesn't end well, usually), flattened those out.
* loop bound context cancels now rely on defer (also was brittle)
* runHot had a lot of ctx shuffling, flattened that.
* added some additional spans in certain paths for added granularity
* linked up the hot launcher / run hot / wait hot to _a_ root span, the first
2 are follows from spans, but at least we can see the source of these and also
can see containers launched over a hot launcher's lifetime

I left TODO around the FollowsFrom because OpenCensus doesn't, at least at the
moment, appear to have any idea of FollowsFrom and it was an extra OpenTracing
method (we have to get the span out, start a new span with the option, then
add it to the context... some shuffling required). anyway, was on the fence
about adding at least.

i know @skinowski was wanting to add some of the hot launcher stats onto spans, was meddling with these to debug some timing stuff so just yanked them into separate pr to land them.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -304,82 +304,65 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
 // hotLauncher is spawned in a go routine for each slot queue to monitor stats and launch hot
 // containers if needed. Upon shutdown or activity timeout, hotLauncher exits and during exit,
 // it destroys the slot queue.
-func (a *agent) hotLauncher(ctx context.Context, callObj *call) {
-
+func (a *agent) hotLauncher(ctx context.Context, call *call) {
 	// Let use 60 minutes or 2 * IdleTimeout as hot queue idle timeout, pick
 	// whichever is longer. If in this time, there's no activity, then
 	// we destroy the hot queue.
 	timeout := time.Duration(60) * time.Minute
-	idleTimeout := time.Duration(callObj.IdleTimeout) * time.Second * 2
+	idleTimeout := time.Duration(call.IdleTimeout) * time.Second * 2
 	if timeout < idleTimeout {
 		timeout = idleTimeout
 	}
 
 	logger := common.Logger(ctx)
 	logger.WithField("launcher_timeout", timeout).Info("Hot function launcher starting")
-	isAsync := callObj.Type == models.TypeAsync
 
 	// IMPORTANT: get a context that has a child span / logger but NO timeout
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel() // shut down GetResourceToken if timeout/shutdown
 	// TODO this is a 'FollowsFrom'
-	ctx = opentracing.ContextWithSpan(common.WithLogger(ctx, logger), opentracing.SpanFromContext(ctx))
+	ctx = opentracing.ContextWithSpan(common.WithLogger(context.Background(), logger), opentracing.SpanFromContext(ctx))
 	span, ctx := opentracing.StartSpanFromContext(ctx, "agent_hot_launcher")
 	defer span.Finish()
 
 	for {
-		ctx, cancel = context.WithTimeout(ctx, timeout)
+		ctx, cancel := context.WithTimeout(ctx, timeout)
+		a.checkLaunch(ctx, call)
+
 		select {
 		case <-a.shutdown: // server shutdown
 			cancel()
 			return
-		case <-ctx.Done():
+		case <-ctx.Done(): // timed out
 			cancel()
-			if a.slotMgr.deleteSlotQueue(callObj.slots) {
+			if a.slotMgr.deleteSlotQueue(call.slots) {
 				logger.Info("Hot function launcher timed out")
 				return
 			}
-		case <-callObj.slots.signaller:
-		}
-		cancel()
-
-		curStats := callObj.slots.getStats()
-		isNeeded := isNewContainerNeeded(&curStats)
-		logger.WithFields(logrus.Fields{"currentStats": curStats, "isNeeded": isNeeded}).Debug("Hot function launcher stats")
-		if !isNeeded {
-			continue
-		}
-		logger.WithFields(logrus.Fields{"currentStats": curStats, "isNeeded": isNeeded}).Info("Hot function launcher starting hot container")
-
-		// if we get a timeout, we must call off get resource token
-		ctx, cancel = context.WithTimeout(ctx, timeout)
-
-		select {
-		case tok, isOpen := <-a.resources.GetResourceToken(ctx, callObj.Memory, uint64(callObj.CPUs), isAsync):
-			if isOpen {
-				a.wg.Add(1)
-				go func(ctx context.Context, call *call, tok ResourceToken) {
-					a.runHot(ctx, call, tok)
-					a.wg.Done()
-				}(ctx, callObj, tok)
-			} else {
-				cancel()
-				// this means the resource was impossible to reserve (eg. memory size we can never satisfy),
-				// return here so we don't infinite loop
-				callObj.slots.queueSlot(&hotSlot{done: make(chan struct{}), err: models.ErrCallTimeoutServerBusy})
-				return
-			}
+		case <-call.slots.signaller:
 			cancel()
-		case <-ctx.Done():
-			cancel()
-			if a.slotMgr.deleteSlotQueue(callObj.slots) {
-				logger.Info("Hot function launcher timed out")
-				return
-			}
-		case <-a.shutdown: // server shutdown
-			cancel()
-			return
 		}
+	}
+}
+
+func (a *agent) checkLaunch(ctx context.Context, call *call) {
+	curStats := call.slots.getStats()
+	isAsync := call.Type == models.TypeAsync
+	isNeeded := isNewContainerNeeded(&curStats)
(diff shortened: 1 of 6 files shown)


api/agent/async.go
```diff
@@ -15,14 +16,15 @@ func (a *agent) asyncDequeue() {
...
+
 	for {
-		ctxResource, cancelResource := context.WithCancel(context.Background())
 		select {
 		case <-a.shutdown:
-			cancelResource()
 			return
-		case <-a.resources.WaitAsyncResource(ctxResource):
-			cancelResource()
+		case <-a.resources.WaitAsyncResource(ctx):
```
skinowski: I think this is going to leak go routines. (See: resource go routine for cancel.)

skinowski: ping @rdallman

rdallman: I don't see anything, but maybe I'm missing it. In english:

select 1) we either are shutdown or we wait until async is ready
   * if we get a shutdown, return will call the deferred cancel to cancel the waiter's ctx
   * if we get a return from async resource it doesn't matter, we can reuse the ctx until we get a shutdown

select 2)
   * if we get a shutdown, return will call the deferred cancel to cancel asyncChew's ctx
   * if we get a message from asyncChew it doesn't matter, we can reuse the ctx until we get a shutdown
   * asyncRun only uses the ctx to extract the span

what'd I miss?

rdallman: right, inside of get resource token. inc.

rdallman: resolved (I think): https://github.com/fnproject/fn/pull/716/commits/e258e3ebc28e3c3bb1f27d8e33897cab8f62b477

api/agent/agent.go
```diff
@@ -417,14 +416,14 @@ func (a *agent) waitHot(ctx context.Context, call *call) (Slot, error) {
...
 	ch := make(chan Slot)
-	ctxResource, cancelResource := context.WithCancel(ctx)
-	defer cancelResource()
+
+	span, ctx := opentracing.StartSpanFromContext(ctx, "agent_launch_cold")
+	defer span.Finish()
 
 	select {
-	case tok, isOpen := <-a.resources.GetResourceToken(ctxResource, call.Memory, uint64(call.CPUs), isAsync):
+	case tok, isOpen := <-a.resources.GetResourceToken(ctx, call.Memory, uint64(call.CPUs), isAsync):
```
skinowski: side effect of resource cancel go routine to live longer. (before: go routine lived while waiting for resource, after: it waits call ctx duration)

rdallman: I believe this is also fixed here: https://github.com/fnproject/fn/commit/e258e3ebc28e3c3bb1f27d8e33897cab8f62b477 -- either we get a resource (and no goroutines live on) or we hit ctx.Done() in this scope (also shuts them down)

api/agent/agent.go
```diff
@@ -334,23 +341,14 @@ func (a *agent) hotLauncher(ctx context.Context, callObj *call) {
...
-		logger.WithFields(logrus.Fields{
-			"currentStats": curStats,
-			"isNeeded":     isNeeded,
-		}).Debug("Hot function launcher stats")
+		logger.WithFields(logrus.Fields{"currentStats": curStats, "isNeeded": isNeeded}).Debug("Hot function launcher stats")
 		if !isNeeded {
 			continue
 		}
-
-		ctxResource, cancelResource := context.WithCancel(context.Background())
```
skinowski: so, here this creates a shorter life span ctx intentionally since for-loop would fork a go routine every time. This ensures with each for-loop iteration we cleanup any go routines.

rdallman: the only exit paths are timeout, shutdown, or get resource. i think with get resource managing its own go routines now there's no longer an issue

api/agent/agent.go
```diff
@@ -334,23 +341,14 @@ func (a *agent) hotLauncher(ctx context.Context, callObj *call) {
...
-			"currentStats": curStats,
-			"isNeeded":     isNeeded,
-		}).Info("Hot function launcher starting hot container")
+		logger.WithFields(logrus.Fields{"currentStats": curStats, "isNeeded": isNeeded}).Info("Hot function launcher starting hot container")
 
 		select {
-		case tok, isOpen := <-a.resources.GetResourceToken(ctxResource, callObj.Memory, uint64(callObj.CPUs), isAsync):
-			cancelResource()
+		case tok, isOpen := <-a.resources.GetResourceToken(ctx, callObj.Memory, uint64(callObj.CPUs), isAsync):
 			if isOpen {
```
skinowski: unrelated, but there's a bug here... user first-time submits an impossible request (huge mem), then we spin infinite-loop

skinowski: I'll fix that one up, with another PR.

api/agent/agent.go
skinowski: so, timeout (slot busy) and impossible request (huge-mem with isOpen = false) would leak here though.

api/agent/agent.go
rdallman: good catch. i can add (need to return after queueing a poopy slot, right ?)

api/agent/agent.go
skinowski: (Currently, impossible req doesn't fork a go routine, but anyways. :-) )

api/agent/agent.go
skinowski: oh, nevermind, there's an else handling that case.. Sorry... If I can only read.. :-)

rdallman: i guess this assumes that we are getting memory from the route (i.e. for a given route, we'll never be able to satisfy the resource request locally since memory is static) -- if we have ability to change memory per call then this behavior is less savory.

skinowski: yes, if mem/cpu were different per call, then this logic will break...

skinowski: then slot key changes and results in entire new hot container.

rdallman: https://github.com/fnproject/fn/pull/716/commits/f058559e2a7f9561e84214bca5aae4d6f563893d

api/agent/agent.go
rdallman: https://github.com/fnproject/fn/pull/716/commits/b644bd67c6dbe801edf1158e9d984bea503798f0 -- good catch.

api/agent/agent.go
```diff
@@ -334,52 +341,52 @@ func (a *agent) hotLauncher(ctx context.Context, callObj *call) {
...
-		case tok, isOpen := <-a.resources.GetResourceToken(ctxResource, callObj.Memory, uint64(callObj.CPUs), isAsync):
-			cancelResource()
+		case tok, isOpen := <-a.resources.GetResourceToken(ctx, callObj.Memory, uint64(callObj.CPUs), isAsync):
+			cancel()
 			if isOpen {
 				a.wg.Add(1)
 				go func(ctx context.Context, call *call, tok ResourceToken) {
 					a.runHot(ctx, call, tok)
 					a.wg.Done()
 				}(ctx, callObj, tok)
```
skinowski: I think ctx here should be different, we are passing a cancelled ctx to runHot, right?

rdallman: we probably should to reduce confusion, but we're really just extracting out the span in `runHot`

rdallman: gonna fiddle with this block

api/agent/agent.go
```diff
@@ -334,52 +341,52 @@ func (a *agent) hotLauncher(ctx context.Context, callObj *call) {
...
 			} else {
-				// this means the resource was impossible to reserve (eg. memory size we can never satisfy)
+				// this means the resource was impossible to reserve (eg. memory size we can never satisfy),
+				// return here so we don't infinite loop
 				callObj.slots.queueSlot(&hotSlot{done: make(chan struct{}), err: models.ErrCallTimeoutServerBusy})
+				return
```
skinowski: we can punt on this for now and fix this later as well, but we can't return here without deleting the slot. I think for now, we can just queue the error without a return or maybe we could move this as a validation check much before we even bother launching this hot slot. We could probably call isResourcePossible() directly before even attempting getSlotQueue in getSlot() and return an err immediately..

rdallman: i've made some re-arrangements of the fine china once again https://github.com/fnproject/fn/pull/716/commits/c4374de0dbab92fe8a3e3d78587de2cf883c92c7 -- hopefully, there are no cracked teapots this time. GetResourceToken contract was kind of bugging me, now there is no possibility of us queueing an error slot in here since GetCall will block anybody that will try to ask for too much

rdallman:
hmm, minor adjustment needed. 1s

skinowski:
isResourcePossible() takes bytes, where GetResource() translates, then calls it.

rdallman:
thanks

rdallman:
may want to add units to force devs to have to consider the conversion (or stop imbeciles like me from programming computers) -- i'll list it. pushed fix. I can't seem to get the test failure circle is getting either, seems fishy.

skinowski:
thanks for the patience on this.. lots of tricky code there for sure, but it looks much cleaner now...

rdallman:
i think the error that ci is fussing up about is the i/o stuff we've been chatting about. I'm still WIP there, this was part of that and took longer than expected today! thanks for reviewing

impeccableassuredamericanbadger-max-1mb

PR #719, opened 2018-01-26
fn: better slot/container/request state tracking

skinowski:
*) in stats.go no need to lock span log (aka common/metrics.go calls)
*) tracker code to simplify agent side and also track request vs container states.
*) slots code a bit clearer since 'waiters' is actually queued requests, versus 'starter' means starting containers.
*) for cold cookie run moved into prepCold() as this is part of container spawn/start phase as std IO is tied up to the container, with this we consider waiting on 'waiter' channel period is the 'execution' phase. This is not entirely true, but a close enough approximation.
*) container states get logged in spans as histograms (for time), and gauge (for counts)

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -182,20 +182,12 @@ func (a *agent) Submit(callI Call) error {
 
 func (a *agent) startStateTrackers(ctx context.Context, call *call) {
 
-	if protocol.IsStreamable(protocol.Protocol(call.Format)) {
-		// For hot requests, we use a long lived slot queue, which we use to manage hot containers
-		var isNew bool
-		call.slots, isNew = a.slotMgr.getSlotQueue(call)
-		if isNew {
-			go a.hotLauncher(ctx, call)
-		}
-	} else {
+	if !protocol.IsStreamable(protocol.Protocol(call.Format)) {
 		// For cold containers, we track the container state in call
 		call.containerState = NewContainerState()
 	}
 
 	call.requestState = NewRequestState()
-	call.requestState.UpdateState(ctx, RequestStateWait, call.slots)
 }
 
 func (a *agent) endStateTrackers(ctx context.Context, call *call) {
@@ -314,10 +306,18 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
 	defer span.Finish()
 
 	if protocol.IsStreamable(protocol.Protocol(call.Format)) {
+		// For hot requests, we use a long lived slot queue, which we use to manage hot containers
+		var isNew bool
+		call.slots, isNew = a.slotMgr.getSlotQueue(call)
+		call.requestState.UpdateState(ctx, RequestStateWait, call.slots)
+		if isNew {
+			go a.hotLauncher(ctx, call)
+		}
 		s, err := a.waitHot(ctx, call)
 		return s, err
 	}
 
+	call.requestState.UpdateState(ctx, RequestStateWait, call.slots)
 	return a.launchCold(ctx, call)
 }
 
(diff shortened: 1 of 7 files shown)


api/agent/agent.go
```diff
@@ -453,12 +482,10 @@ func (s *coldSlot) exec(ctx context.Context, call *call) error {
 	span, ctx := opentracing.StartSpanFromContext(ctx, "agent_cold_exec")
 	defer span.Finish()
 
-	waiter, err := s.cookie.Run(ctx)
-	if err != nil {
-		return err
-	}
```
rdallman: not so sure about moving this into prepare, this is the execution. left comment at other location

api/agent/agent.go
```diff
@@ -567,16 +596,28 @@ func (a *agent) prepCold(ctx context.Context, call *call, tok ResourceToken, ch
...
 	// about timing out if this takes a while...
-	cookie, err := a.driver.Prepare(ctx, container)
-	slot := &coldSlot{cookie, tok, err}
+	var err error
+	var cookie drivers.Cookie
+	var waiter drivers.WaitResult
+
+	cookie, err = a.driver.Prepare(ctx, container)
+	if err == nil {
+		waiter, err = cookie.Run(ctx)
```
rdallman: the issue here is changing the ordering of this means that the call will be running before we have called `call.Start` -- which we want to use to manage the task state (specifically, for async, but also sync has some field updates). I'm not sure it's worth trying to dangle this down here, trying to keep it so that `submit` is 'prepare' -> 'start' -> 'exec' -> 'end' and running the container for cold should remain in exec imo. i could elaborate on finer details, but the timings are off for a few things (the task finishing before the start time is set seems... odd). in any case, it doesn't seem worth moving this for state management alone and we'll have to make a number of other adjustments.

wrt state management, I'm less sure of the implications. my mental model is not yet built, but it seems like we're kind of lying here anyway saying the container is idle after calling run. we should set to execution before calling cookie run I think (and this should be in exec where that happens, I think). at least, moving this back to its old location seems sane, and for cold there really is no 'idle' container transition, there is basically only... created but not yet running (not really idle, but kind of I guess), not sure if we'd like to skip it [for cold] or if we're relying on certain state transitions for correctness somewhere, but leaving the state transition to idle here and moving cookie run back also seems to make sense to me.

api/agent/agent.go
```diff
@@ -653,30 +690,27 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken) {
...
 					return
 				}
 			case <-ctx.Done(): // container shutdown
 				if call.slots.ejectSlot(s) {
-					call.slots.exitStateWithLatency(SlotQueueIdle, uint64(time.Now().Sub(start).Seconds()*1000))
 					return
 				}
 			case <-a.shutdown: // server shutdown
 				if call.slots.ejectSlot(s) {
-					call.slots.exitStateWithLatency(SlotQueueIdle, uint64(time.Now().Sub(start).Seconds()*1000))
```
rdallman: yay

api/agent/agent.go
```diff
@@ -180,9 +180,40 @@ func (a *agent) Submit(callI Call) error {
...
+func (a *agent) startStateTrackers(ctx context.Context, call *call) {
+
+	if protocol.IsStreamable(protocol.Protocol(call.Format)) {
+		// For hot requests, we use a long lived slot queue, which we use to manage hot containers
+		var isNew bool
+		call.slots, isNew = a.slotMgr.getSlotQueue(call)
+		if isNew {
+			go a.hotLauncher(ctx, call)
+		}
+	} else {
```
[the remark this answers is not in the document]
skinowski: yeah, this doesn't really have to be this way.. I think I got tempted to do this to clarify that call.slot is assigned here... And immediately after that assignment, hotLauncher must be launched.. It's a bit weird though.. Let me look one more time.

api/agent/agent.go
skinowski: I agree, this was a tricky move... Start() interaction, etc. I'll revert this one...

PR #721, opened 2018-01-27
pipe swapparoo each slot

rdallman:
previously, we made a pair of pipes for stdin and stdout for each container,
and then handed them out to each call (slot) to use. this meant that multiple
calls could have a handle on the same stdin pipe and stdout pipe to read/write
to/from from fn's perspective and could mix input/output and get garbage. this
also meant that each was blocked on the previous' reads.

now we make a new pipe every time we get a slot, and swap it out with the
previous ones. calls are no longer blocked from fn's perspective, and we don't
have to worry about timing out dispatch for any hot format. there is still the
issue that if a function does not finish reading the input from the previous
task, from its perspective, and reads the next call's it can error out the
second call. with fn deadline we provide the necessary tools to skirt this,
but without some additional coordination am not sure this is a closable hole
with our current protocols since terminating a previous calls input requires
some protocol specific bytes to go in (json in particular is tricky). anyway,
from fn's side fixing pipes was definitely a hole, but this client hole is
still hanging out. there was an attempt to send an io.EOF but the issue is
that will shut down docker's read on the stdin pipe (and the container). poop.

this adds a test for this behavior, and makes sure 2 containers don't get
launched. this test fails with master's agent.go (so this wasn't a complete
waste of time...)

this also closes the response writer header race a little, but not entirely, I
think there's still a chance that we read a full response from a function and
get a timeout while we're changing the headers. I guess we need a thread safe
header bucket, otherwise we have to rely on timings (racy). thinking on it.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -22,8 +22,6 @@ import (
 
 // TODO we should prob store async calls in db immediately since we're returning id (will 404 until post-execution)
 // TODO async calls need to add route.Headers as well
-// TODO need to shut off reads/writes in dispatch to the pipes when call times out so that
-// 2 calls don't have the same container's pipes...
 // TODO handle timeouts / no response in sync & async (sync is json+503 atm, not 504, async is empty log+status)
 // see also: server/runner.go wrapping the response writer there, but need to handle async too (push down?)
 // TODO storing logs / call can push call over the timeout
@@ -35,7 +33,6 @@ import (
 // dies). need coordination w/ db.
 // TODO if a cold call times out but container is created but hasn't replied, could
 // end up that the client doesn't get a reply until long after the timeout (b/c of container removal, async it?)
-// TODO between calls, logs and stderr can contain output/ids from previous call. need elegant solution. grossness.
 // TODO if async would store requests (or interchange format) it would be slick, but
 // if we're going to store full calls in db maybe we should only queue pointers to ids?
 // TODO examine cases where hot can't start a container and the user would never see an error
@@ -516,9 +513,8 @@ func (s *coldSlot) Close() error {
 // implements Slot
 type hotSlot struct {
 	done      chan<- struct{} // signal we are done with slot
-	proto     protocol.ContainerIO
-	errC      <-chan error // container error
-	container *container   // TODO mask this
+	errC      <-chan error    // container error
+	container *container      // TODO mask this
 	err       error
 }
 
@@ -540,16 +536,21 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	// link the container id and id in the logs [for us!]
 	common.Logger(ctx).WithField("container_id", s.container.id).Info("starting call")
 
-	// swap in the new stderr logger & stat accumulator
-	oldStderr := s.container.swap(call.stderr, &call.Stats)
-	defer s.container.swap(oldStderr, nil) // once we're done, swap out in this scope to prevent races
+	// swap in fresh pipes & stat accumulator to not interlace with other calls that used this slot [and timed out]
+	stdinRead, stdinWrite := io.Pipe()
+	stdoutRead, stdoutWrite := io.Pipe()
+	defer stdinRead.Close()
+	defer stdoutWrite.Close()
+
+	proto := protocol.New(protocol.Protocol(call.Format), stdinWrite, stdoutRead)
+
+	swapBack := s.container.swap(stdinRead, stdoutWrite, call.stderr, &call.Stats)
+	defer swapBack() // NOTE: it's important this runs before the pipes are closed.
 
 	errApp := make(chan error, 1)
 	go func() {
-		// TODO make sure stdin / stdout not blocked if container dies or we leak goroutine
-		// we have to make sure this gets shut down or 2 threads will be reading/writing in/out
 		ci := protocol.NewCallInfo(call.Call, call.req)
-		errApp <- s.proto.Dispatch(ctx, ci, call.w)
+		errApp <- proto.Dispatch(ctx, ci, call.w)
 	}()
 
 	select {
@@ -560,8 +561,6 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	case <-ctx.Done(): // call timeout
 		return ctx.Err()
 	}
-
-	// TODO we REALLY need to wait for dispatch to return before conceding our slot
 }
 
 func specialHeader(k string) bool {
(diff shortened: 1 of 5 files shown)


rdallman:
to clarify commit message, it does not close https://github.com/fnproject/fn/pull/690#discussion_r161963028 where if we sleep forever in the call it will make subsequent calls against that container fail. we could play with waiting for output for some set amount of time (idle timeout maybe?) before shutting down but meh. i agree shutting down the container is a clean break, and starting to think it's a good idea even given our current hot protocols but i'm worried about some seemingly not pathological cases (low timeout, bad image author). did a lot of fiddling here, but this really only ends up fixing i/o issues from fn's side, not the container's.

api/agent/agent.go
```diff
@@ -821,5 +825,78 @@ func (g *ghostWriter) Write(b []byte) (int, error) {
...
+		err = nil
+	}
+	return n, err
+}
+
+// waitReader returns io.EOF if anyone calls Read. don't call Read, this is a sentinel type
+type waitReader struct{}
+
+func (e *waitReader) Read([]byte) (int, error) {
+	panic("read on waitReader should not happen")
```
skinowski: But it could, right? After servicing a request and the agent is done, where container becomes idle, the user container could try to read it again out of blue? shouldn't this return 0, io.EOF?

api/agent/agent.go
```diff
@@ -540,16 +536,21 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
...
+	swapBack := s.container.swap(stdinRead, stdoutWrite, call.stderr, &call.Stats)
+	defer swapBack() // NOTE: it's important this runs before the pipes are closed.
 
 	errApp := make(chan error, 1)
 	go func() {
-		// TODO make sure stdin / stdout not blocked if container dies or we leak goroutine
-		// we have to make sure this gets shut down or 2 threads will be reading/writing in/out
 		ci := protocol.NewCallInfo(call.Call, call.req)
-		errApp <- s.proto.Dispatch(ctx, ci, call.w)
+		errApp <- proto.Dispatch(ctx, ci, call.w)
```
skinowski: I wonder if now we can swap this code inside the go routine with a ctx.Done() select, which can close the proper pipe to unblock main go routine waiting on proto.Dispatch().

api/agent/agent.go
```diff
@@ -801,8 +805,8 @@ func (c *container) WriteStat(ctx context.Context, stat drivers.Stat) {
 // TODO per call could implement this stored somewhere (vs. configured on host)
 //}
 
-// ghostWriter is a writer who will pass writes to an inner writer
-// (that may be changed at will).
+// ghostWriter is an io.Writer who will pass writes to an inner writer
+// that may be changed at will. it is thread safe to swap or write.
 type ghostWriter struct {
```
skinowski: it would be good to move these IO related types/code into its own .go file in future...

api/agent/agent.go
```diff
@@ -748,19 +746,25 @@ type container struct {
...
+	c.stdin.(*ghostReader).swap(stdin)
+	c.stdout.(*ghostWriter).swap(stdout)
+	c.stderr.(*ghostWriter).swap(stderr)
 	c.stats = cs
-	return old
+
+	return func() {
+		c.stdin.(*ghostReader).swap(ostdin)
+		c.stdout.(*ghostWriter).swap(ostdout)
+		c.stderr.(*ghostWriter).swap(ostderr)
```
skinowski: couple questions:
1) how about stats? shouldn't we set stats to nil here to stop accumulating stats via ?
2) shouldn't we lock/unlock in the function returned?
3) unrelated but 'lock' as inherited sync.Mutex in container struct makes this harder to understand IMHO, what does that lock protect? entire container? or just stats and io? We pass the container to Docker driver and that reads container fields without locking it... which means, we only protect IO and stats swap/append with it, right? (I'll fix/rename this later.)

skinowski: I think the lock here is only needed for stats for driver collect go routine (append) and during swap (set/unset) of stats member variable. There's no other go routine trying to swap these at the same time.

denismakogon: Aren't we allocating new slice for container stats for each call? As far as understand, we do that each time we call `task.WriteStat`

skinowski: is calling append() with nil safe? There's a nil check there... in WriteStat()

But yeah, I think we only need to protect stats here. Lock the mutex when setting/unsetting/append...

skinowski: In other words, if stats is nil, we shouldn't collect stats, right? That's what WriteStat() is trying to do I think.

api/agent/agent.go
skinowski: If we invert this go routine, then no need to worry about thread safe headers.

rdallman: hmm, once we close the pipes though Dispatch could still be stuck on the Read from the request body (not a pipe) and will not get an error until that Read call returns to write it to the pipe writer. I could not find a way around this, tinkered with some ideas, seems one of the pains of io.Copy. for this case, we do need to set https://golang.org/pkg/net/http/#Server.ReadTimeout which would close it. for the pipe for stdout of the container, closing this is good enough to get it to bail, but the header race still exists I think since the context could timeout after parsing the stdout but while we're adding the headers (we have no mutex around this). 

> which can close the proper pipe to unblock main go routine waiting on proto.Dispatch().

i think this is happening, but i'm not sure if i am understanding correctly. we are closing the pipes once the `exec` select ticks (ctx.Done() is a case).

api/agent/agent.go
rdallman: thanks, good catch. yes, we need to set stats to nil and have the mutex to swap just for stats I think. the ghosts have their own locks. 

> unrelated but 'lock' as inherited sync.Mutex in container struct makes this harder to understand IMHO, what does that lock protect? entire container? or just stats and io? 

think it's expanded and contracted a few times: id, stats, writers. now the reader and writers have their own locks. will make this more clear, thanks, it is definitely not a 'container' lock as implied currently.

api/agent/agent.go
skinowski: Right... ReadTimeout on client would be an issue.. We can think about it later, especially if we are going to buffer req/resp, then client read timeout issue will be moved somewhere else (earlier), and then we can safely invert the go routine like;

```
go func() {
    select {
        case <-ctx.Done():
               // close pipe to unblock Dispatch()
        case err := <- s.errC:
              // close pipe to unblock Dispatch()
   }
}

// block here (stays on this go routine to avoid Header lock issue)
return s.proto.Dispatch(...)
```

rdallman: got it. yea it would be nice if we could get dispatch to bail correctly on the ctx, reading the request body is an issue whether we put into a buffer or not I think (dang io.Copy) wrt timing out. spent some time playing with the io copy thing and really didn't have a great solution, but if we yanked the request into a buffer in this thread we at least wouldn't have the header race (is it worth it?)

api/agent/agent.go
skinowski: we can merge, and then tweak this as a followup as this should work fine.

rdallman: gonna fix the stats mu shortly, then i think we're safe

api/agent/agent.go
skinowski: I think we'll be buffering req/resp going forward, right? I do like this (despite increased mem footprint) since then we decouple these I/O (slow client keeping container busy, etc.) issues out of the picture.

skinowski: I know we'll buffer resp. But I think we should also buffer request.

skinowski: #326

skinowski: Oh, another gotcha for request buffering would be to wait until we have a container... (eg. we got resource token), otherwise we'd be piling up these in memory unnecessarily...

rdallman: good idea

rdallman:
thanks for review. there's a ticket to buffer already https://github.com/fnproject/fn/issues/326 just gonna link it up

rdallman:
and wrt https://github.com/fnproject/fn/pull/690 we've had a discussion ongoing. I think we probably need to delay it but there's possibly some finagling we can do to 'wait' for some amount of time for output to finish and stdin to open back up.

PR #728, opened 2018-01-31
return bad function http resp error

### LINKED ISSUE #355: If hot fn returns malformed response server should be more explicit on what went wrong
If function returns malformed HTTP response server returns to a user HTTP 500 Internal Server Error, but in server logs we see more verbose error:
```
ERRO[1283] internal server error                         action="server.handleRunnerRequest)-fm" app=testapp error="malformed HTTP status code \"established\"" id=01BTNMVEHJ08S0A00000000000 route=/hot-json stack="goroutine 1485 [running]:\nruntime/debug.Stack(0x0, 0x0, 0x0)\n\t/usr/local/go/src/runtime/debug/stack.go:24 +0xbe\ngithub.com/fnproject/fn/api/server.HandleErrorResponse(0x56a8500, 0xc42056d3b0, 0x5e71918, 0xc4203b38c0, 0x5697ac0, 0xc42043c060)\n\t/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/api/server/error_response.go:37 +0x4bf\ngithub.com/fnproject/fn/api/server.handleErrorResponse(0xc4203b38c0, 0x5697ac0, 0xc42043c060)\n\t/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/api/server/error_response.go:24 +0xae\ngithub.com/fnproject/fn/api/server.(*Server).serve(0xc420392090, 0xc4203b38c0, 0xc4204082a8, 0x7, 0xc4204082af, 0x9)\n\t/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/api/server/runner.go:109 +0x677\ngithub.com/fnproject/fn/api/server.(*Server).handleRequest(0xc420392090, 0xc4203b38c0)\n\t/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/api/server/runner.go:44 +0x248\ngithub.com/fnproject/fn/api/server.(*Server).handleRunnerRequest(0xc420392090, 0xc4203b38c0)\n\t/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/api/server/server.go:213 +0x35\ngithub.com/fnproject/fn/api/server.(*Server).(github.com/fnproject/fn/api/server.handleRunnerRequest)-fm(0xc4203b38c0)\n\t/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/api/server/server.go:298 +0x34\ngithub.com/fnproject/fn/vendor/github.com/gin-gonic/gin.(*Context).Next(0xc4203b38c0)\n\t/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/vendor/github.com/gin-gonic/gin/context.go:110 +0x43\ngithub.com/fnproject/fn/api/server.panicWrap(0xc4203b38c0)\n\t/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/api/server/server.go:192 +0x51\ngithub.com/fnproject/fn/vendor/github.com/gin-gonic/gin.(*Context).Next(0xc4203b38c0)\n\t/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/vendor/github.com/gin-gonic/gin/context.go:110 +0x43\ngithub.com/fnproject/fn/api/server.traceWrap(0xc4203b38c0)\n\t/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/api/server/server.go:114 +0x2e9\ngithub.com/fnproject/fn/vendor/github.com/gin-gonic/gin.(*Context).Next(0xc4203b38c0)\n\t/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/vendor/github.com/gin-gonic/gin/context.go:110 +0x43\ngithub.com/fnproject/fn/api/server.loggerWrap(0xc4203b38c0)\n\t/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/api/server/server.go:209 +0x17c\ngithub.com/fnproject/fn/vendor/github.com/gin-gonic/gin.(*Context).Next(0xc4203b38c0)\n\t/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/vendor/github.com/gin-gonic/gin/context.go:110 +0x43\ngithub.com/fnproject/fn/vendor/github.com/gin-gonic/gin.(*Engine).handleHTTPRequest(0xc42009a900, 0xc4203b38c0)\n\t/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/vendor/github.com/gin-gonic/gin/gin.go:344 +0x5c5\ngithub.com/fnproject/fn/vendor/github.com/gin-gonic/gin.(*Engine).ServeHTTP(0xc42009a900, 0x56a73c0, 0xc42062ea80, 0xc420138f00)\n\t/Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn/vendor/github.com/gin-gonic/gin/gin.go:311 +0x16b\nnet/http.serverHandler.ServeHTTP(0xc420080c30, 0x56a73c0, 0xc42062ea80, 0xc420138f00)\n\t/usr/local/go/src/net/http/server.go:2619 +0x1e2\nnet/http.(*conn).serve(0xc420132820, 0x56a8440, 0xc42056e440)\n\t/usr/local/go/src/net/http/server.go:1801 +0x125d\ncreated by net/http.(*Server).Serve\n\t/usr/local/go/src/net/http/server.go:2720 +0x734\n"
```

We might let users know if their function fails to be good one for http format.

rdallman:
this was being thrown into the fn server logs but it's relatively easy to get
this to crop up if a function user forgets that they left a `println` laying
around that gets written to stdout, it garbles the http (or json, in its case)
output and they just see 'internal server error'. for certain clients i could
see that we really do want to keep this as 'internal server error' but for
things like e.g. docker image not authorized we're showing that in the
response, so this seems apt.

json likely needs the same treatment, will file a bug.

as always, my error messages are rarely helpful enough, help me please :)

closes #355

### DIFF
--- a/api/agent/protocol/http.go
+++ b/api/agent/protocol/http.go
@@ -43,7 +43,7 @@ func (h *HTTPProtocol) Dispatch(ctx context.Context, ci CallInfo, w io.Writer) e
 	if err != nil {
 		// From here (boo stdlib): https://github.com/golang/go/blob/d3c1df712658398f29bc8bebd6767e7b3cac2d12/src/net/http/response.go#L174
 		if strings.Contains(err.Error(), "malformed HTTP") {
-			return models.NewAPIError(http.StatusInternalServerError, fmt.Errorf("invalid http response from function err:", err))
+			return models.NewAPIError(http.StatusInternalServerError, fmt.Errorf("invalid http response from function err: %v", err))
 		}
 		return err
 	}
--- a/api/agent/protocol/http.go
+++ b/api/agent/protocol/http.go
@@ -3,8 +3,12 @@ package protocol
 import (
 	"bufio"
 	"context"
+	"fmt"
 	"io"
 	"net/http"
+	"strings"
+
+	"github.com/fnproject/fn/api/models"
 )
 
 // HTTPProtocol converts stdin/stdout streams into HTTP/1.1 compliant
@@ -37,6 +41,10 @@ func (h *HTTPProtocol) Dispatch(ctx context.Context, ci CallInfo, w io.Writer) e
 
 	resp, err := http.ReadResponse(bufio.NewReader(h.out), ci.Request())
 	if err != nil {
+		// From here (boo stdlib): https://github.com/golang/go/blob/d3c1df712658398f29bc8bebd6767e7b3cac2d12/src/net/http/response.go#L174
+		if strings.Contains(err.Error(), "malformed HTTP") {
+			return models.NewAPIError(http.StatusInternalServerError, fmt.Errorf("invalid http response from function err:", err))
+		}
 		return err
 	}
 	defer resp.Body.Close()
--- a/api/agent/protocol/http.go
+++ b/api/agent/protocol/http.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"strings"
 
 	"github.com/fnproject/fn/api/models"
 )
@@ -41,11 +40,7 @@ func (h *HTTPProtocol) Dispatch(ctx context.Context, ci CallInfo, w io.Writer) e
 
 	resp, err := http.ReadResponse(bufio.NewReader(h.out), ci.Request())
 	if err != nil {
-		// From here (boo stdlib): https://github.com/golang/go/blob/d3c1df712658398f29bc8bebd6767e7b3cac2d12/src/net/http/response.go#L174
-		if strings.Contains(err.Error(), "malformed HTTP") {
-			return models.NewAPIError(http.StatusInternalServerError, fmt.Errorf("invalid http response from function err: %v", err))
-		}
-		return err
+		return models.NewAPIError(http.StatusBadGateway, fmt.Errorf("invalid http response from function err: %v", err))
 	}
 	defer resp.Body.Close()
 
--- a/api/server/runner_test.go
+++ b/api/server/runner_test.go
@@ -164,8 +164,7 @@ func TestRouteRunnerExecution(t *testing.T) {
 	}{
 		{"/r/myapp/", ok, "GET", http.StatusOK, expHeaders, ""},
 
-		// TODO: this should return 502 and description about container misbehaving
-		{"/r/myapp/myhot", badHttp, "GET", http.StatusInternalServerError, expHeaders, "internal server error"},
+		{"/r/myapp/myhot", badHttp, "GET", http.StatusBadGateway, expHeaders, "invalid http response"},
 		// hot container now back to normal, we should get OK
 		{"/r/myapp/myhot", ok, "GET", http.StatusOK, expHeaders, ""},
 
--- a/api/server/runner_test.go
+++ b/api/server/runner_test.go
@@ -132,6 +132,7 @@ func TestRouteRunnerExecution(t *testing.T) {
 		[]*models.Route{
 			{Path: "/", AppName: "myapp", Image: "fnproject/fn-test-utils", Type: "sync", Memory: 128, Timeout: 30, IdleTimeout: 30, Headers: map[string][]string{"X-Function": {"Test"}}},
 			{Path: "/myhot", AppName: "myapp", Image: "fnproject/fn-test-utils", Type: "sync", Format: "http", Memory: 128, Timeout: 30, IdleTimeout: 30, Headers: map[string][]string{"X-Function": {"Test"}}},
+			{Path: "/myhotjason", AppName: "myapp", Image: "fnproject/fn-test-utils", Type: "sync", Format: "json", Memory: 128, Timeout: 30, IdleTimeout: 30, Headers: map[string][]string{"X-Function": {"Test"}}},
 			{Path: "/myroute", AppName: "myapp", Image: "fnproject/fn-test-utils", Type: "sync", Memory: 128, Timeout: 30, IdleTimeout: 30, Headers: map[string][]string{"X-Function": {"Test"}}},
 			{Path: "/myerror", AppName: "myapp", Image: "fnproject/fn-test-utils", Type: "sync", Memory: 128, Timeout: 30, IdleTimeout: 30, Headers: map[string][]string{"X-Function": {"Test"}}},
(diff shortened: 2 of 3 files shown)


rdallman:
https://github.com/fnproject/fn/issues/729 for jason, it's a good first issue for somebody capable of growing sufficient beard to want to dig into the internals (really, I do not have the beard for this one...)

api/agent/protocol/http.go
```diff
@@ -33,6 +37,10 @@ func (h *HTTPProtocol) Dispatch(ctx context.Context, ci CallInfo, w io.Writer) e
 
 	resp, err := http.ReadResponse(bufio.NewReader(h.out), ci.Request()) // TODO timeout
 	if err != nil {
+		// From here (boo stdlib): https://github.com/golang/go/blob/d3c1df712658398f29bc8bebd6767e7b3cac2d12/src/net/http/response.go#L174
```
skinowski: I don't think we should cherry pick based on error string match. There are many error variations in the stdlib (3-4 cases of malformed http only) so I think we should do one of the following:

1) mask all errors coming from ReadResponse as "invalid response from function" (hides internals, but better than generic 500)
2) pass through all errors from  ReadResponse inside NewAPIError (more user friendly, but could expose internal details)

Should these errors be 502 also?

IMHO, I'd pick '1' but using 502 bad gateway. (Since we are using 504 for timeout already and acting as a gw.)

api/agent/protocol/http.go
```diff
@@ -33,6 +37,10 @@ func (h *HTTPProtocol) Dispatch(ctx context.Context, ci CallInfo, w io.Writer) e
 
 	resp, err := http.ReadResponse(bufio.NewReader(h.out), ci.Request()) // TODO timeout
```
skinowski: req.Write() failure perhaps should be a 502.

api/agent/protocol/http.go
rdallman: at least one exception is that if the pipe closes they'll get io.ErrClosedPipe, but in theory that error won't get plumbed up since it's past the timeout (at some point, it may on accident leak out). 

i was hoping for a typed error from stdlib, sadly it's not the case. 

502 makes sense, 1) makes sense. we can log the real error. the error we get back conveniently does have some info in it such as `invalid HTTP status code: "MYOOPSYLOGLINE"` which could tip a user off as to what they're doing wrong. 

I was thinking it's possible since we're basically only doing i/o that we could plumb up any error from Dispatch as a `models.APIError` but the surface area may be too large. it would cover the json case nicely and also cover `req.Write` failures, but it's less clear what the exact error should be. something along the lines of "container i/o failed".

skinowski: "invalid response from function", "cannot process function response" or something like that maybe? "container i/o failed" sounds good too except maybe that sounds like it could also be fn server issue.

rdallman: I had another thought about this the other day that in http / json we could probably handle cases where uses write accidental log lines to stdout before the actual http / json blob. i'm not sure how worth it that is, but it's definitely a lego in the foot at the moment. it may well be better to try to get them to correct their behavior than covering up their misbehaving. 'invalid function response' kind of only makes sense in the case handled in this patch.

skinowski:
Also, I'm seeing 502 from fn server for responses like this. Somewhere we must be replacing 500 with 502.

api/agent/protocol/http.go
denismakogon: agree with pt.1 - if function writes bad http response (whatever except well-structured http response) Fn should be more verbose for this cases and i'd like to get "invalid response from function" because it also indicates broken FDK as well as a user's faults.

rdallman: it's more a matter of where to put it at this point. if we catch every error and say that it's not necessarily true. we're running on a hijacked connection, any number of things could go wrong there that aren't the functions fault, hence checking for very specific errors.

denismakogon:
The only thing i miss here is the test that does the following: creates a function with json or http format but the function is capable to do only default protocol framing (dummy echo func would be more than enough here). Overall, looks good.

rdallman:
yea, could use the hello image, can't use the test-utils image without being able to inject bad behavior into it. been marinating on exactly what scope this should cover in light of discussion, open to thoughts

skinowski:
I'm working on fn-test-utils to do stuff like this.. PR coming soon.

skinowski:
For now, you can trigger malformed http via fn-test-utils request:

```
{
        "responseCode": -1,
        "echoContent": "hello",
        "isDebug": true
}
```

skinowski:
I've got a #755 PR that has an invalid http case for this change.

api/server/runner_test.go
```diff
@@ -164,21 +166,17 @@ func TestRouteRunnerExecution(t *testing.T) {
...
 		{"/r/myapp/myroute", ok, "GET", http.StatusOK, expHeaders, ""},
 		{"/r/myapp/myerror", crasher, "GET", http.StatusBadGateway, expHeaders, "container exit code 2"},
 		{"/r/myapp/mydne", ``, "GET", http.StatusNotFound, nil, "pull access denied"},
 		{"/r/myapp/mydnehot", ``, "GET", http.StatusNotFound, nil, "pull access denied"},
-
-		// Added same tests again to check if time is reduced by the auth cache
-		{"/r/myapp/", ok, "GET", http.StatusOK, expHeaders, ""},
-		{"/r/myapp/myroute", ok, "GET", http.StatusOK, expHeaders, ""},
-		{"/r/myapp/myerror", crasher, "GET", http.StatusBadGateway, expHeaders, "container exit code 2"},
-		{"/r/myapp/myerror", crasher, "GET", http.StatusBadGateway, expHeaders, "container exit code 2"},
```
rdallman: these were all duplicates except oom, it was time for these to go..

rdallman:
ok, i fixed up test and went ahead and fixed #729, thanks @skinowski for test image updates. should be ready to go... it does leak out the underlying error, in the prominent cases users keep hitting I do think it will help with specificity, even though extraneous cases are possible they still seem somewhat useful so that users can see where it went wrong.

rdallman:
thanks!

PR #733, opened 2018-02-01
Idle Hot Container Freeze/Preempt Support

skinowski:
1) Idle containers (which hasn't moved into busy state for 50 msecs) are frozen to save cpu and stop execution after function processing in idle state.
2) Containers that are idle for more than 1 second check if any active waiters are present for cpu & mem, if this is the case, they terminate themselves.


.circleci/config.yml
```diff
@@ -22,11 +22,11 @@ jobs:
...
-      #- run: |
-          #sudo rm -rf /var/cache/apt/archives && sudo ln -s ~/.apt-cache /var/cache/apt/archives && mkdir -p ~/.apt-cache/partial
-          #sudo service docker stop
-          #curl -fsSL https://get.docker.com/ | sudo sh
-          #docker version
+      - run: |
+          sudo rm -rf /var/cache/apt/archives && sudo ln -s ~/.apt-cache /var/cache/apt/archives && mkdir -p ~/.apt-cache/partial
+          sudo service docker stop
+          curl -fsSL https://get.docker.com/ | sudo sh
+          docker version
```
skinowski: had to enable this, 17.06 (old) docker fails and hangs when removing containers that are frozen.

skinowski:
It's also possible to move towards a much more strict freezer policy, where we can freeze all hot containers as soon as they become idle. The latency introduced by these extra calls s about 1-5 msec, but this approach can more robustly lock the container without leaving any holes (50 msec, etc.) And this approach can simplify the implementation reducing the complexity in hotRun() select block.

This PR currently only freezes if container is idle more than 50 msecs. We can decide on this policy if this is not suitable.

api/agent/agent.go
```diff
@@ -683,21 +685,60 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
...
-			case <-s.trigger:
-			case <-time.After(time.Duration(call.IdleTimeout) * time.Second):
-				if call.slots.ejectSlot(s) {
-					logger.Info("Canceling inactive hot function")
-					return
-				}
-			case <-ctx.Done(): // container shutdown
-				if call.slots.ejectSlot(s) {
-					return
+			isFrozen := false
```
skinowski: Here can we choose to freeze as soon as container becomes idle for a more strict/robust freezer policy. But at the cost of slowing down a busy slot queue (where dequeue takes < 50 msec.)

api/agent/agent.go
```diff
@@ -683,21 +685,60 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
...
 				}
-			case <-a.shutdown: // server shutdown
-				if call.slots.ejectSlot(s) {
+				break
+			}
+
+			// if we can eject token, that means we are here due to
+			// abort/shutdown/timeout, attempt to eject and terminate,
+			// otherwise continue processing the request
+			if call.slots.ejectSlot(s) {
```
rdallman: this seems a little dicey, but did some thinking and couldn't dig up any apparent issues with doing this other than relying on ejectSlot's behavior.

api/agent/drivers/driver.go
```diff
@@ -54,6 +54,12 @@ type Driver interface {
 	//
 	// The returned cookie should respect the task's timeout when it is run.
 	Prepare(ctx context.Context, task ContainerTask) (Cookie, error)
+
+	// Freeze the container to pause running processes
+	Freeze(ctx context.Context, task ContainerTask) error
+
+	// Unfreeze a frozen container to unpause frozen processes
+	Unfreeze(ctx context.Context, task ContainerTask) error
```
rdallman: these seem more apt on `*Cookie` fwiw, a `ContainerTask` is basically a configuration whereas a 'cookie' maps to a created container (withstanding of its status before/after cookie.Run)

api/agent/agent.go
```diff
@@ -683,21 +685,60 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
...
+
+						// if someone is waiting for resource in our slot queue, we must not terminate,
+						// otherwise, see if other slot queues have resource waiters.
+						stats := call.slots.getStats()
+						if stats.containerStates[ContainerStateWait] > 0 ||
+							a.resources.GetResourceTokenWaiterCount() <= 0 {
+							continue
+						}
+
+						logger.Debug("attempting hot function eject")
```
rdallman: maybe I need to eat something today, but this appears to attempt to eject the container after one ejectTicker interval after freezing rather than after the idle timeout? if so, why do we have idle timeout if we're just going to obey ejectTicker? should we set ejectTicker to idleTimeout or am I missing something here?

rdallman: sometimes i forget it's better to read pr comments afterwards. i see, we're ejecting only if nobody is hanging out. part of me thinks this defeats the purpose of idle to a certain degree. I do think we need to eject in certain cases (highlighted by the red TODO at the head of agent.go, among others) but this seems really aggressive. if i make 1 total api call every 2 seconds i'll have to eat the container start up time both times even though idle timeout is 30s, right?

skinowski: We are ejecting only:

1) if our own slot queue does not have any container-spawn code waiting on resources.
(eg. burst of route A calls should not trigger killing of route A containers.)
and
2) some other slot queue has someone blocked waiting on cpu & mem
(eg. burst of route B calls may kill route A containers, but only if route B blocks on cpu & memory condition variable and route A container was idling for 1 sec)

So, 1 total api call every 2 secs would only terminate that container if the timer fired and found out that there was some other route-call that was blocked on cpu & resources. If there is enough cpu & memory in the system, no call should block on the condition variable. They'd lock the mutex and get the resource... But if there's contention, then the call would likely to block on cond.wait().

api/agent/agent.go
```diff
@@ -670,6 +668,10 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
 	go func() {
 		defer shutdownContainer() // also close if we get an agent shutdown / idle timeout
 
+		freezeTicker := time.Duration(50) * time.Millisecond
```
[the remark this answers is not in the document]
skinowski: np, I'll plumb up a config option for this. I actually want this tune-able;
1) always freeze without waiting
2) freeze after x msecs
3) disable freeze

api/agent/agent.go
skinowski: This is a bit tricky code here. I attempted to simplify the loop/flow, but maybe it's worse now. It assumes if we can eject a token, then we are here due to timeouts, shutdowns, etc. Otherwise, the I/O select should not land here without the consumer pulling the trigger. If consumer pulled the trigger, then we should not be able to eject. Costs an atomic cas, so it's fine wrt perf IMHO.

.circleci/config.yml
skinowski: I agree, is it ok if I get that in with separate PR?

rdallman: that's fine. if the breakage is mysterious it may be better to add here so that we don't have to run slack support, most users don't update their docker very frequently (usually when something blows up and we tell them to)

api/agent/agent.go
rdallman: I take it that's the `a.resources.GetResourceTokenWaiterCount() <= 0`, blanked on that. that seems pretty slick to me.

api/agent/drivers/docker/docker_client.go
```diff
@@ -19,6 +19,7 @@ import (
 
 const (
 	retryTimeout = 10 * time.Minute
+	pauseTimeout = 5 * time.Second
```
rdallman: don't think we can avoid this really, but docker likes to fart and we'll hit this at some point for sure. for freezing or unfreezing if this does time out it likely makes the problem worse, since in the freeze case, it will go destroy the container (docker load) potentially requiring making a new one instead of just thawing. in the unfreeze case, we'll just have to launch another which will take longer & more docker load. just tossing this around as a thought experiment... in the unfreeze case specifically it means a call has a handle on the slot and ate a lot of time, which is bad. don't really want 2 phase slots so that we can have triggerThaw->handoutSlot, seems nasty and not sure it buys us much; we could unfreeze 2 and take the first one to unfreeze or potentially something to increase the probability that we can quickly get a thawed container. anyway, not sure it's critical until we start hitting such cases and maybe not worth talking about until we can repro/see some of that behavior.

api/agent/agent.go
rdallman: yea, I couldn't think up any case, since if the slot is taken this will fail I think it covers most of the races. it seems correct

api/agent/drivers/docker/docker_client.go
skinowski: Yeah, I think for now I just want to clamp on that timeout and if it takes more than 5 secs, we are done and we just tear down the container. The timeout is tricky as the longer it is, the longer we expose the container for executing code outside of the allowed thawed state. I'd like to be much more aggressive on it, but at the same time, there's docker/contaierd/runc in the middle, so probably 5 sec is OK for time being...

skinowski:
depends on #744

skinowski:
@rdallman ping

api/agent/agent.go
```diff
@@ -127,13 +131,25 @@ func New(da DataAccess) Agent {
...
 
+	freezeIdleMsecs, err := getEnvMsecs("FN_FREEZE_IDLE_MSECS", 50*time.Millisecond)
+	if err != nil {
+		logrus.WithError(err).Fatal("error initializing freeze idle delay")
+	}
+
+	ejectIdleMsecs, err := getEnvMsecs("FN_EJECT_IDLE_MSECS", 1000*time.Millisecond)
+	if err != nil {
+		logrus.WithError(err).Fatal("error initializing eject idle delay")
+	}
```
rdallman: looked and seems like we're not plumbing agent config stuff very well, no need here, but made a note to allow this to be plumbed from the *Server configuration more easily (seems like a strange issue to file)

rdallman: (there's other things we'll need to / already need to plumb in there)

PR #735, opened 2018-02-02
Extend /stats API to handle two routes with the same path in different apps

nigeldeakin:
The "legacy" stats API (`/stats`), which is currently used by the UI tool, has a bug. The bug is that the returned data is simply a list of route stats, keyed by path. So this would not distinguish between two routes, in different apps, which had the same path. It would simply merge the stats for the two routes together, which would be extremely misleading.

```
{
  "Queue": 0,
  "Running": 10,
  "Complete": 23,
  "Failed": 10,
  "FunctionStatsMap": {
    "/hello-cold-async-a1": {
      "Queue": 0,
      "Running": 0,
      "Complete": 0,
      "Failed": 9
    },
    "/hello-cold-async-a2": {
      "Queue": 0,
      "Running": 4,
      "Complete": 1,
      "Failed": 0
    },
...
```
(Full example [here](https://gist.github.com/nigeldeakin/662af33f366b539c0d855de7bc4c3847))

This PR fixes this bug by reorganising the data into two levels: app and then path. So it looks like this:
```
{
  "Queue": 1,
  "Running": 20,
  "Complete": 45,
  "Failed": 16,
  "Apps": {
    "hello-cold-async-a": {
      "Routes": {
        "/hello-cold-async-a1": {
          "Queue": 1,
          "Running": 1,
          "Complete": 0,
          "Failed": 16
        },
        "/hello-cold-async-a2": {
          "Queue": 0,
          "Running": 8,
          "Complete": 10,
          "Failed": 0
        },
        "/hello-cold-async-a3": {
          "Queue": 0,
          "Running": 4,
          "Complete": 5,
          "Failed": 0
        }
      }
    },
    "hello-cold-async-b": {
      "Routes": {
        "/hello-cold-async-b1": {
          "Queue": 0,
          "Running": 4,
          "Complete": 5,
          "Failed": 0
        },
...
```
(Full example [here](https://gist.github.com/nigeldeakin/b55beb98438f247b0bcb30f985eca1ce))

This is not part of the public API and is used only by the UI tool, which is being updated in parallel with this change.

You might be thinking that this API needs to be remove rather than improved. Yes, removal is still the plan, with the UI tool modified to get its information from either Prometheus or by scraping the `/metrics` endpoint. However that's a rather bigger task (whereas making this change to Fn is very simple). Consider this change an intermediate stage in this wider plan.

### DIFF
--- a/api/agent/stats.go
+++ b/api/agent/stats.go
@@ -10,7 +10,9 @@ import (
 // * hot containers active
 // * memory used / available
 
-// global statistics
+// stats holds statistics of all apps
+// an instance of this struc is used to maintain an in-memory record of the current stats
+// access must be synchronized using the Mutex
 type stats struct {
 	mu sync.Mutex
 	// statistics for all functions combined
@@ -18,11 +20,20 @@ type stats struct {
 	running  uint64
 	complete uint64
 	failed   uint64
-	// statistics for individual functions, keyed by function path
-	functionStatsMap map[string]*functionStats
+	// statistics for individual apps, keyed by appname
+	apps map[string]appStats
 }
 
-// statistics for an individual function
+// appStats holds statistics for an individual app, keyed by function path
+// instances of this struc are used to maintain an in-memory record of the current stats
+// access must be synchronized using the Mutex on the parent stats
+type appStats struct {
+	functions map[string]*functionStats
+}
+
+// functionStats holds statistics for an individual function
+// instances of this struc are used to maintain an in-memory record of the current stats
+// access must be synchronized using the Mutex on the parent stats
 type functionStats struct {
 	queue    uint64
 	running  uint64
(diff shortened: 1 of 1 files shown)


api/agent/stats.go
```diff
@@ -161,16 +184,22 @@ func IncrementTooBusy(ctx context.Context) {
...
 	s.mu.Lock()
 	stats.Running = s.running
 	stats.Complete = s.complete
 	stats.Queue = s.queue
 	stats.Failed = s.failed
-	stats.FunctionStatsMap = make(map[string]*FunctionStats)
-	for key, value := range s.functionStatsMap {
-		thisFunctionStats := &FunctionStats{Queue: value.queue, Running: value.running, Complete: value.complete, Failed: value.failed}
-		stats.FunctionStatsMap[key] = thisFunctionStats
+	stats.Apps = make(map[string]AppStats)
```
skinowski: not a big deal, but this make statement doesn't have in locked scope.

nigeldeakin: Noted, thanks.

Issue #656, opened 2018-01-08
remove sleeper, error, hello images from fn repo


skinowski:
fn-test-utils can (should) be able to simulate all of this functionality, therefore
remove/cleanup references to these images and decommission them from fn repo.

denismakogon:
+1 for this, fnproject/error is relatively huge:
```
fnproject/error           latest              3a928b32cc60        3 months ago        55.6MB
```
alpine pg image is 35megs, lol, we definitely need to come up with the only one image for testing.

As side problem, unknown (really) number of images our tests are depending on pollutes local registry blazingly fast, so having 1 tagged image makes a lot more sense. Along with that, the number of functions we made for tests really makes fn repo looks like a dump of stuff we don't really want to store locally or at our own repositories.

denismakogon:
Yeah, agreed. But not sure that that image is the last one lest there. Quick search gave at least one more image to substitute: `fnproject/multi-log` that is used at API tests as well

skinowski:
multi-log is gone... moved to fn-test-utils.

PR #736, opened 2018-02-02
fn: sleeper image is now deprecated

skinowski:
Please use fn-test-utils instead for testing.


skinowski:
Partially fixes: #656

PR #740, opened 2018-02-05
Fix default setting

### LINKED ISSUE #738: `created_at` set to `0` when application is created from a deploy command
# To reproduce
1. create an `app.yaml` with the `name` property set to an app that doesn't exist
2. create some throwaway functions in the same directory
3. run `fn deploy --all`
4. check the application listing with `curl http://localhost:8080/v1/apps`
5. observe that the `created_at` property is set to `0001-01-01T00:00:00.000Z`

This problem does not happen if an app is created via the CLI prior to `fn deploy` being run

rdallman:
push validate/defaults into datastore

we weren't setting a timestamp in route insert when we needed to create an app
there. that whole thing isn't atomic, but this fixes the timestamp issue.

closes #738

seems like we should do similar with the FireBeforeX stuff too.

also:

* app name validation was buggy, an upper cased letter failed. now it doesn't.
uses unicode now.
* removes duplicate errors for datastore and models validation that were used
interchangably but weren't.

### DIFF
--- a/api/datastore/internal/datastoretest/test.go
+++ b/api/datastore/internal/datastoretest/test.go
@@ -280,8 +280,8 @@ func Test(t *testing.T, dsf func(t *testing.T) models.Datastore) {
 		}
 
 		_, err = ds.InsertApp(ctx, &models.App{})
-		if err != models.ErrDatastoreEmptyAppName {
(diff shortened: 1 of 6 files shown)


rdallman:
i checked and we'd have to do an api call to get the app after doing a post route. it kind of sucks (our tests suck, really, 2/3 of them)

PR #742, opened 2018-02-05
push down app listeners to a datastore

rdallman:
fnext.NewDatastore returns a datastore that wraps the appropriate methods for
AppListener in a Datastore implementation. this is more future proof than
needing to wrap every call of GetApp/UpdateApp/etc with the listeners, there
are a few places where this can happen and it seems like the AppListener
behavior is supposed to wrap the datastore, not just the front end methods
surrounding CRUD ops on an app. the hairy case that came up was when fiddling
with the create/update route business.

this changes the FireBeforeApp* ops to be an AppListener implementation itself
rather than having the Server itself expose certain methods to fire off the
app listeners, now they're on the datastore itself, which the server can
return the instance of.

small change to BeforeAppDelete/AfterAppDelete -- we were passing in a half
baked struct with only the name filled in and not filling in the fields
anywhere. this is mostly just misleading, we could fill in the app, but we
weren't and don't really want to, it's more to notify of an app deletion event
so that an extension can behave accordingly instead of letting a user inspect
the app. i know of 3 extensions and the changes required to update are very
small.

cleans up all the front end implementations FireBefore/FireAfter.

this seems potentially less flexible than previous version if we do want to
allow users some way to call the database methods without using the
extensions, but that's exactly the trade off, as far as the AppListener's are
described it seems heavily implied that this should be the case.

mostly a feeler, for the above reasons, but this was kind of odorous so just
went for it. we do need to lock in the extension api stuff.

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -76,7 +76,7 @@ type Server struct {
 	mq              models.MessageQueue
 	logstore        models.LogStore
 	nodeType        ServerNodeType
-	appListeners    []fnext.AppListener
+	appListeners    *appListeners
 	rootMiddlewares []fnext.Middleware
 	apiMiddlewares  []fnext.Middleware
 }
@@ -279,6 +279,10 @@ func New(ctx context.Context, opts ...ServerOption) *Server {
 	s.Router.Use(loggerWrap, traceWrap, panicWrap) // TODO should be opts
 	optionalCorsWrap(s.Router)                     // TODO should be an opt
 	s.bindHandlers(ctx)
+
+	s.appListeners = new(appListeners)
+	s.datastore = fnext.NewDatastore(s.datastore, s.appListeners)
+
 	return s
 }
 
--- a/fnext/listeners.go
+++ b/fnext/listeners.go
@@ -17,9 +17,9 @@ type AppListener interface {
 	// AfterAppUpdate called after updating App in the database
 	AfterAppUpdate(ctx context.Context, app *models.App) error
 	// BeforeAppDelete called right before deleting App in the database
-	BeforeAppDelete(ctx context.Context, appName string) error
+	BeforeAppDelete(ctx context.Context, app *models.App) error
 	// AfterAppDelete called after deleting App in the database
-	AfterAppDelete(ctx context.Context, appName string) error
+	AfterAppDelete(ctx context.Context, app *models.App) error
 	// BeforeAppGet called right before getting an app
 	BeforeAppGet(ctx context.Context, appName string) error
 	// AfterAppGet called after getting app from database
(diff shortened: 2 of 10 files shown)


api/server/server.go
```diff
@@ -279,6 +279,10 @@ func New(ctx context.Context, opts ...ServerOption) *Server {
 	s.Router.Use(loggerWrap, traceWrap, panicWrap) // TODO should be opts
 	optionalCorsWrap(s.Router)                     // TODO should be an opt
 	s.bindHandlers(ctx)
+
+	s.appListeners = new(appListeners)
+	s.datastore = fnext.NewDatastore(s.datastore, s.appListeners)
```
rdallman: could do this lazily too on `AddAppListener`, flirted with it... lmk

rdallman:
awaiting @treeder thoughts before big green button

fnext/listeners.go
```diff
@@ -17,9 +17,9 @@ type AppListener interface {
 	// AfterAppUpdate called after updating App in the database
 	AfterAppUpdate(ctx context.Context, app *models.App) error
 	// BeforeAppDelete called right before deleting App in the database
-	BeforeAppDelete(ctx context.Context, app *models.App) error
+	BeforeAppDelete(ctx context.Context, appName string) error
 	// AfterAppDelete called after deleting App in the database
-	AfterAppDelete(ctx context.Context, app *models.App) error
+	AfterAppDelete(ctx context.Context, appName string) error
```
treeder: I think it would be useful for to have all of the application info here. Why the change?

rdallman: well, it means we have to go fill it in instead of just deleting it (which we weren't / aren't, which is misleading at least), the db call itself to delete doesn't fetch and this turns into a plumbing thing. I can't think of other fields in the app itself that would cause any practical use cases to prevent deletion really... 'don't delete apps if they have a config value with key=AZEROTH' ?

rdallman:
> Looks good for the most part. From an extension developers perspective, it doesn't change anything ya?

if they use the datastore directly, it will use their listeners now, not just the front end methods that use these. this seems to be the intent but was mostly asking your opinion

fnext/listeners.go
rdallman: > small change to BeforeAppDelete/AfterAppDelete -- we were passing in a half
baked struct with only the name filled in and not filling in the fields
anywhere. this is mostly just misleading, we could fill in the app, but we
weren't and don't really want to, it's more to notify of an app deletion event
so that an extension can behave accordingly instead of letting a user inspect
the app. i know of 3 extensions and the changes required to update are very
small.

(from PR comment)

treeder:
Besides the app vs app name thing, LGTM

rdallman:
> Besides the app vs app name thing

do you propose we leave it as a not-filled-in app then or what do you want?

rdallman:
I've acquiesced, one day we'll get app ids in and can re-open the worm can

PR #744, opened 2018-02-06
fn: docker version checks

skinowski:
1) now required docker version is 17.06
2) enable circle ci latest docker install
3) docker driver & agent check minimum version before start


skinowski:
required by: #733

skinowski:
uses semver, '-ce' should be ok.

Issue #714, opened 2018-01-23
Version 1.0 - Finalize all user facing API's, commands and contracts


treeder:
After this point, we'll lock things in and will not introduce any breaking changes until version 2.0.

This includes:

* API 
* Container I/O formats
* CLI commands
* Extension interfaces

Related issues:

* [ ] app_id #607 
* [ ] move some interfaces to `github.com/fnproject/fn` namespace #556 

Suggested date: February 16, 2018.

treeder:
cc @zootalures , per your comments earlier today.

zootalures:
I'd like to see #662 in here too

zootalures:
Also perhaps consider ditching http if we can come up with a streamable JSON framing.

zootalures:
Per comments from @denismakogan would be good to have a version flag to allow/detect future breaking changes

denismakogon:
Protocol version is more of just a flag. Protocol version should be taken into consideration in FDK implementations. Since we develop protocol handlers those handlers should be versioned with respect to Fn protocol versions, or as an alternative: FDKs should be aware of Fn version they work with (this would an implicit Fn-FDK compatibility characteristic).

Also, we need to develop a mechanism that will open issues for all officially supported FDKs when we see changes appear in `agent/protocol` files.

treeder:
The idea of this is that we don't make the protocols backwards incompatible, so there shouldn't need to be versions. We can always add things, but should almost never remove something. If we do end up breaking a protocol, for whatever reason, we should probably make it a new format, eg: json2 or whatever.

denismakogon:
Vote +1 for getting rid of default format.

denismakogon:
@treeder i'd like to see protocols being tested against actual functions as part of Fn CI, only this type of verification is required for saying that we're not breaking protocols. FDK regression framework would fit here perfectly.

treeder:
I created #819 about namespacing, as it's a big breaking change. There's also #820 which is another breaking change.

treeder:
Can we all agree on the input formats as being locked down now, in particular default and JSON?  HTTP could be dropped or deprecated (#785). Thumb up or down on this comment to vote.

zootalures:
Can we at least add base64 to JSON -  a lot of strange stuff happens with binary input.

treeder:
Added something else to consider, error handling: #870.

denismakogon:
What about protocol versioning? All things said below related to JSON protocol only.
[...]

Issue #137, opened 2017-07-27
discussion: default to being hot


rdallman:
In gitlab by @rdallman on Jul 18, 2017, 13:12

Like Tom Brady's children? ![9XHl18c](/uploads/a7377537f0094f516a60c58fc6b6439f/9XHl18c.gif)

We've discussed this many times but there doesn't seem to be a card for it (I looked). It seems like it would be possible to default to 'hot' regardless of format. Bear with me, but poke holes if there are blatant issues. This takes care of 2 'issues', one being having to use the http format (yes, even with easy to use wrappers) and the other is internally having to treat hot and cold differently (in our server code). Defaulting to 'hot' will still basically entail the user having their code a certain way (ideally, using our wrapper), but we don't really have to treat the two that much differently.

Ultimately, we _can_ run every container as 'hot', it's just that if not coded as such the container will so happen to exit (i.e. if they don't have a for loop / our wrapper). what this means is that as long as we don't queue up work against any container, expecting it to run more things, then that's ok and the container can exit (how it is now), it's just not optimal -- but if the code isn't there to do it, then whatever, not much we can do. It would be possible for us to use docker exec to change the env vars between executions with the `default` format, and it would be possible to frame STDIN with EOFs, are there other concerns here for `default` format?

Would this give us a lot of sugar? I think this would give us two things really:

* not having to make the distinction whether a function is 'hot' or 'cold' on the server. that is, not having to change the configuration for a route at all (where 'http' => 'hot'). if the code happens to loop, great. if not, nobody cares but things are a little slower. this would make it so if the user just updates their code they don't have to change 'format' on their route.
* formats are less important as a distinction for hot/cold. as long as we say we frame 'default' with EOF then we can use it for 'hot'. obviously with wrappers this isn't such a big deal, but it's a little bit 'magic' to have to set your format to 'http' even though your code isn't explicitly doing http parsing (the wrapper hides it) and it's possible right now even to have a "cold" function that talks the 'http' format (just exit instead of looping, we can't prevent this). this also means in the server code we can hit the same execution path for any function invocation (instead of switching on format "type").

possible cons:

* it will be hard to tell whether a function is 'hot' or 'cold' based on format alone, which the argument could be made is already the case with 'http' anyway, since the user can just exit (obviously, the question is, why would they? but ignore that, it's only important that it's _possible_ is the point). 
* we're not _really_ defaulting to 'hot' since the code has to be a certain way, maybe that could be confusing

[...]

PR #745, opened 2018-02-06
[WIP] Fn can do hot only, no more cold

denismakogon:
Related to #714


denismakogon:
Tests are broken so far, mostly because they use the functions that were not designed to be hot.

treeder:
This is obviously something that needs to be discussed bigly before implementing.

denismakogon:
@treeder as far as i understand, there's nothing much to discuss regarding default protocol. The only concern is the number of FDKs we support, so having both HTTP and JSON protocol is more than enough even if you start developing functions with the programming language that is not backed up with FDK we (or someone else) implemented. There would be the certain impact on Fn users, but pre-V1 is the exact time to do so.

rdallman:
yea, wouldn't spend a lot of time on this yet

rdallman:
move discussion to #137 and close this ?

PR #746, opened 2018-02-07
fn: memory limit adjustments

skinowski:
1) limit kernel memory which was previously unlimited, using
   same limits as user memory for a unified approach.
2) disable swap memory for containers


rdallman:
this means a container could potentially use e.g. 128MB user space + 128MB kernel space? not clear how to rectify if that's the case?

skinowski:
@rdallman  According to kernel docs, kernel usage is reported under user accounting (but only when kernel limits are enabled), so this means, setting both to the same limit is supposed to clamp *all* usage to 128MB following your example.

PR #747, opened 2018-02-08
Add options to impose server-side mandated limits on container resources

hhexo:
This change introduces configurable server-side mandated limits on CPU, memory and disk usage. It required a little bit of refactoring in the agent package.

This is the start of a broader work on managing resource usage on Fn nodes and preventing one function from causing disruption to other functions by overusing resources. It's a complex issue and this is just a part of it.
The goal is not just to defend from malicious agents, but also accidental mistakes (e.g. rogue loop writing terabytes to disk).

Functions already have CPU and memory limits set as part of their configuration, but this is more about introducing service-mandated limits - so, the actual limits to function resources will be the minimum between the service-mandated limits and the limits requested by the function developer.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -146,6 +146,13 @@ func New(da DataAccess, fl FunctionLimits) Agent {
 		ServerVersion: "17.06.0",
 	})
 
+	if fl.MaxFilesystemSize() != 0 {
+		supported := driver.SupportsLimit(nil, drivers.LimitFilesystem)
+		if !supported {
+			logrus.Fatal("Can't set FN_MAX_FUNC_FILESYSTEM_SIZE, storage driver does not support quotas")
+		}
+	}
+
 	freezeIdleMsecs, err := getEnvMsecs("FN_FREEZE_IDLE_MSECS", 50*time.Millisecond)
 	if err != nil {
 		logrus.WithError(err).Fatal("error initializing freeze idle delay")
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -627,12 +627,16 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	}
 }
 
-// Golang doesn't have an integer min...
-func min(a uint64, b uint64) uint64 {
-	if a < b {
-		return a
+// nonZeroMin returns zero only if both parameters are zero, otherwise it
+// returns the minimum non-zero number of the two.
+func nonZeroMin(a uint64, b uint64) uint64 {
+	if a != 0 && b != 0 {
+		if a < b {
+			return a
+		}
+		return b
 	}
-	return b
+	return a ^ b // We know one of them is zero, bitwise xor will return the other one
 }
 
 func (a *agent) prepCold(ctx context.Context, call *call, tok ResourceToken, ch chan Slot) {
@@ -653,18 +657,9 @@ func (a *agent) prepCold(ctx context.Context, call *call, tok ResourceToken, ch 
 		call.Config[k] = strings.Join(v, ", ")
 	}
 
-	memory := call.Memory
-	if a.limits.MaxMemory() != 0 {
-		memory = min(memory, a.limits.MaxMemory())
-	}
-	cpus := uint64(call.CPUs)
-	if a.limits.MaxCPUs() != 0 {
-		cpus = min(cpus, a.limits.MaxCPUs())
-	}
-	filesystem := uint64(call.FilesystemSize)
-	if a.limits.MaxFilesystemSize() != 0 {
-		filesystem = min(filesystem, a.limits.MaxFilesystemSize())
-	}
+	memory := nonZeroMin(uint64(call.Memory), a.limits.MaxMemory())
+	cpus := nonZeroMin(uint64(call.CPUs), a.limits.MaxCPUs())
+	filesystem := nonZeroMin(uint64(call.FilesystemSize), a.limits.MaxFilesystemSize())
 	container := &container{
 		id:         id.New().String(), // XXX we could just let docker generate ids...
 		image:      call.Image,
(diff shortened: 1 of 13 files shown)


api/agent/call.go
```diff
@@ -94,16 +94,17 @@ func FromRequest(appName, path string, req *http.Request) CallOpt {
...
-			Headers:     req.Header,
-			CreatedAt:   strfmt.DateTime(time.Now()),
-			URL:         reqURL(req),
-			Method:      req.Method,
+			Priority:       new(int32), // TODO this is crucial, apparently
+			Timeout:        route.Timeout,
+			IdleTimeout:    route.IdleTimeout,
+			Memory:         route.Memory,
+			CPUs:           route.CPUs,
+			FilesystemSize: 0, // TODO make it configurable per route as well
```
hhexo: Need to implement this TODO before merging, at the moment the min returns 0 because of this.

api/agent/agent.go
```diff
@@ -602,6 +618,14 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
...
 }
 
+// Golang doesn't have an integer min...
+func min(a uint64, b uint64) uint64 {
+	if a < b {
+		return a
+	}
+	return b
+}
+
```
skinowski: some of these functions are in resource.go (we might want to move them to a common.go or utils.go or something like that.) Not important. :-)

rdallman:
only did a quick pass, looks pretty good but will let you finish. one note is that the storage-size option for the docker fs driver is fs driver specific afaik and this appears to rely on the args for overlay[2], which is fine for prod that's what we'll use but I can see other users running into potential issues if they try to play with different fs drivers or if we do at some point for whatever reason -- not sure of the best policy, we could implement the various args based on what docker fs driver we get a handle on in fn, or we could just warn or shut down if it's not overlay, maybe something else?

api/agent/agent.go
rdallman: `int(math.Min(float64(a), float64(b)))` _usually_ works, float precision can get dicey but I doubt it for the numbers we're inputting. bit gross

api/agent/call.go
skinowski: I think a system (fleet) wide setting is fine for this, no? To my understanding, this doesn't have to be large due to overlay layers, right?

skinowski:
Why not reject the requests if they exceed these limits? The execution is misleading to users. We execute with clamped limits. We should not execute at all and return 400 with proper error message.

Also, the clamped values completely bypass resource manager. Say max mem configured is 2GB, and we execute a function req with 4GB. With this PR, resource mgr thinks it is running a 4GB job where really container is running with 2GB...

skinowski:
does storage option "size" produce error when running on an unsupported combo: eg. ext4 + overlay ?

skinowski:
I'm specifically thinking about "local" anon volumes (which get removed) when container is removed. but let me check...

skinowski:
@rdallman @hhexo Just double checked the volume issue, Volume definitions in DockerFile do not cause mounts at run time.

api/agent/call.go
hhexo: Yeah, now I fixed the logic to determine the minimum limits later on, so we can leave this at 0 for now and defer the problem of whether to support per-route disk space limits.

hhexo:
@rdallman I have added a check that the docker storage driver supports filesystem quotas. Now if you try to set FN_MAX_FUNC_FILESYSTEM_SIZE on a system that doesn't support filesystem quotas you get a fatal error starting Fn.

Also the erroneous calculation of the minimum "clamp" is fixed.

@skinowski With regards to actually refusing requests which want to have more disk space than the server-side maximum, I'm thinking we should refuse to even _deploy_ such functions in the first place ("sorry, you can't gobble up all our disk space, be nice to your neighbours"). But, how do we know on an API node that the runners have a size limit? We'd have to assume that they are all started with the same configuration... is it a fair assumption? Could Fn deployments have a heterogeneous set of runners? (there be dragons, I think!)

skinowski:
@hhexo if we assume heterogeneous runners, then we can probably just refuse the call at submit . But if we decide on executing these request with clamped values, then please also pass in clamped cpu & mem values to a.resources.GetResourceToken() to the resource mgr (agent.go), which should keep things in sync.

rdallman:
think we have to check if the call is unable to be satisfied at runtime (submit) for now. figuring out how to schedule those into a heterogeneous cluster so that they won't fail is a whole thing we're not quite ready for

api/agent/agent.go
```diff
@@ -609,6 +636,18 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	}
 }
 
+// nonZeroMin returns zero only if both parameters are zero, otherwise it
+// returns the minimum non-zero number of the two.
```
skinowski: For CPU, zero value in call means no limit & skip. In this case, nonZeroMin() should not pick up MaxCPU limit, since it'll accidentally set max CPU to no-limit calls.

api/agent/drivers/docker/docker.go
```diff
@@ -203,6 +210,25 @@ func (drv *DockerDriver) Prepare(ctx context.Context, task drivers.ContainerTask
 	return &cookie{id: task.Id(), task: task, drv: drv}, nil
 }
 
+func (drv *DockerDriver) SupportsLimit(ctx context.Context, l drivers.Limit) bool {
```
skinowski: We can improve later, but just noting it here: we do fetch docker info in checkDockerVersion() in docker.go. We should probably cache the results for referencing it below.

skinowski:
If a user is asking a specific cpu & mem & fs_size, clamping these down may cause errors/issues during the execution and could be confusing to troubleshoot. (eg. agent.runHot() logs actual requested cpu & mem. How would a user or operator find out to see if a request was clamped down?)

On the other hand, the resource manager behaves much differently. If a call comes in that requests for cpu or mem that is higher than total cpu or mem on the server, we reject the call immediately as "a request that can never be executed" And all other requests with cpu & mem requirements has to wait until these resources are available or they do get a 503. If we can't honor user contract for cpu & mem requested, we don't execute them.

Our approach could be OK here, but I'm just curious why we aren't just refusing these calls that are over limits. It's also much simpler than surgically injecting clamping code in various parts of the code.

skinowski:
related: #853

skinowski:
@hhexo Given #853 and #860 (which I've copied some of the impl from here), do you think we can close this?

hhexo:
Yes, I think we can close this one.

PR #748, opened 2018-02-09
limit log size in containers

### LINKED ISSUE #317: hot container logs grow unbounded on disk
while we are tracking logs for each invocation of a call ran in a hot container in memory, these are still getting written to disk via docker's json log file interface. these logs will just keep getting written to for each container and in the pathological case where somebody runs 1 task every 1 second in a hot container or so, they can manage to keep a container alive for theoretically infinite time (good luck). this isn't the end of the world, but the logs will eventually run the host out of space...

we probably need to kill a hot container after some amount of time anyway due to memory accretion and stuff (idk, like 10 minutes or something seems reasonable). we're not really accounting for this and a container could have memory from prior tasks hanging out and be penalized for that (java's GC is notoriously delaying gc's...). 

I've tried turning the logging drivers off as outlined here: https://docs.docker.com/engine/admin/logging/overview/#supported-logging-drivers with no real luck. doing so meant that the container emitted no logs when attaching, either. maybe this behavior has changed, it would be worth testing, as we don't really need the json logs. otherwise, it seems like the best option may be to just add some log-opts to the json-file driver to limit the amount of logs we write in (since we aren't using them, something really small seems fine, as we don't want to constantly be deleting either thanks to i/o overhead). details are here https://docs.docker.com/engine/admin/logging/json-file/#examples

on a related note, we aren't really limiting the disk used by a container either. at some point...

rdallman:
closes #317

we could fiddle with this, but we need to at least bound these. this
accomplishes that. 1m is picked since that's our default max log size for the
time being per call, it also takes a little time to generate that many bytes
through logs, typically (i.e. without trying to). I tested with 0, which
spiked the i/o rate on my machine because it's constantly deleting the json
log file. I also tested with 1k and it was similar (for a task that generated
about 1k in logs quickly) -- in testing, this halved my throughput, whereas
using 1m did not change the throughput at all. trying the 'none' driver and
'syslog' driver weren't great, 'none' turns off all stderr and 'syslog' blocks
every log line (boo). anyway, this option seems to have no effect on the
output we get in 'attach', which is what we really care about (i.e. docker is
not logically capping this, just swapping out the log file).

using 1m for this, e.g. if we have 500 hot containers on a machine we have
potentially half a gig of worthless logs laying around. we don't need the
docker logs laying around at all really, but short of writing a storage driver
ourselves there don't seem to be too many better options. open to idears, but
this is likely to hold us over for some time.

### DIFF
--- a/api/agent/drivers/docker/docker.go
+++ b/api/agent/drivers/docker/docker.go
@@ -147,8 +147,16 @@ func (drv *DockerDriver) Prepare(ctx context.Context, task drivers.ContainerTask
 			AttachStdin:  true,
 			StdinOnce:    true,
 		},
-		HostConfig: &docker.HostConfig{},
-		Context:    ctx,
+		// turn off logs since we're collecting them from attach
+		HostConfig: &docker.HostConfig{
+			LogConfig: docker.LogConfig{
+				Type: "json-file", // o/w attach does not work
+				Config: map[string]string{
+					"max-size": "1m",
+				},
+			},
+		},
+		Context: ctx,
 	}
 
 	// Translate milli cpus into CPUQuota & CPUPeriod (see Linux cGroups CFS cgroup v1 documentation)


rdallman:
thanks!

PR #749, opened 2018-02-09
Trivial fix to get 'make test' working on Mac

ostrain:
Apparently GNU find lets you omit the directory but BSD find does not.


rdallman:
thanks! if you can make https://github.com/oracle public on your user then the cla_bot will pass... i'm not sure if we'll have to kick it somehow after doing that, maybe close this one and open another after doing that or push a blank commit maybe?

PR #751, opened 2018-02-09
fn: hot container timer improvements

skinowski:
With this change, now we are allocating the timers
when the container starts and managing them via
stop/clear as needed, which should not only be more
efficient, but also easier to follow.

For example, previously, if eject time out was
set to 10 secs, this could have delayed idle timeout
up to 10 secs as well. It is also not necessary to do
any math for elapsed time.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -140,6 +140,9 @@ func New(da DataAccess) Agent {
 	if err != nil {
 		logrus.WithError(err).Fatal("error initializing eject idle delay")
 	}
+	if ejectIdleMsecs == time.Duration(0) {
+		logrus.Fatal("eject idle delay cannot be zero")
+	}
 
 	logrus.WithFields(logrus.Fields{"eject_msec": ejectIdleMsecs, "free_msec": freezeIdleMsecs}).Info("agent starting")
 
@@ -555,9 +558,9 @@ func (s *coldSlot) Close(ctx context.Context) error {
 
 // implements Slot
 type hotSlot struct {
-	done      chan<- struct{} // signal we are done with slot
-	errC      <-chan error    // container error
-	container *container      // TODO mask this
+	done      chan struct{} // signal we are done with slot
+	errC      <-chan error  // container error
+	container *container    // TODO mask this
 	err       error
 }
 
@@ -651,23 +654,6 @@ func (a *agent) prepCold(ctx context.Context, call *call, tok ResourceToken, ch 
 	}
 }
 
-func stopTimer(timer *time.Timer) {
-	if timer.Stop() {
-		return
-	}
-	select {
-	case <-timer.C:
-	default:
-	}
-}
-
-func reloadTimer(timer *time.Timer, dur time.Duration) {
-	if !timer.Stop() {
-		<-timer.C
-	}
-	timer.Reset(dur)
-}
-
 func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state ContainerState) {
 	// IMPORTANT: get a context that has a child span / logger but NO timeout
 	// TODO this is a 'FollowsFrom'
@@ -729,16 +715,6 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
 	go func() {
 		defer shutdownContainer() // also close if we get an agent shutdown / idle timeout
 
-		ejectTimer := time.NewTimer(time.Duration(0))
-		freezeTimer := time.NewTimer(time.Duration(0))
-		idleTimer := time.NewTimer(time.Duration(0))
-
-		defer func() {
-			stopTimer(ejectTimer)
-			stopTimer(freezeTimer)
-			stopTimer(idleTimer)
-		}()
-
 		for {
 			select { // make sure everything is up before trying to send slot
 			case <-ctx.Done(): // container shutdown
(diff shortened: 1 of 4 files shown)


api/agent/agent.go
```diff
@@ -718,49 +740,49 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
...
-			freezerTicker := a.freezeIdleMsecs
-			idleTimeout := time.Duration(call.IdleTimeout) * time.Second
 
 			done := make(chan struct{})
 			state.UpdateState(ctx, ContainerStateIdle, call.slots)
 			s := call.slots.queueSlot(&hotSlot{done, errC, container, nil})
 
+			stopTimer(freezeTimer)
+			stopTimer(idleTimer)
+			stopTimer(ejectTimer)
```
rdallman: since there's a select with a default i'm not sure we'll properly drain any of the timer.C channels, per the docs this will fail spectacularly if we call reset afterwards. i think as it is the select statement should pick the timer.C receive case but there may be a runtime scheduling race in doing that since if timer.C is not ready to send the default case will be selected (the docs just do `if timer.Stop() { <-timer.C }` which may wait an implied short amount of time, as i'm sure you know). idk for sure ?

skinowski: Changing this...

api/agent/agent.go
```diff
@@ -718,49 +737,49 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
...
-								return
-							}
-							isFrozen = true
+				case <-idleTimer.C:
+				case <-freezeTimer.C:
+					if !isFrozen {
+						err := cookie.Freeze(ctx)
+						if err != nil {
+							logger.WithError(err).Error("freeze error")
+							return
```
rdallman: only issue is I think since we took the freezeTimer.C cookie that the deferred stop of that timer will wait for the cookie since it didn't stop the timer, but nobody will ever send a cookie (we'll get blocked in defer forever and leak this goroutine)

skinowski:
I need to test this a bit more... Let's hold off on merging...

api/agent/agent.go
```diff
@@ -807,6 +742,107 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
...
+		isFrozen = true
+	}
+
+	state.UpdateState(ctx, ContainerStateIdle, call.slots)
+	s := call.slots.queueSlot(slot)
+
+	for {
+		select {
+		case <-s.trigger: // slot already consumed, unfreeze ASAP
+			if !isFrozen {
```
rdallman: we have to freeze it just so that we can unfreeze it? seems a bit :(

api/agent/agent.go
```diff
@@ -807,6 +742,107 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
...
+	state.UpdateState(ctx, ContainerStateBusy, call.slots)
+
+	// request processing may take a long time, clean up timers now
+	ejectTicker.Stop()
+	freezeTimer.Stop()
+	idleTimer.Stop()
+
+	// wait for this call to finish
+	// NOTE do NOT select with shutdown / other channels. slot handles this.
+	<-slot.done
```
rdallman: one option to do this (`<-slot.done`) in the caller, which would remove the double ticker/timer stops

api/agent/agent.go
```diff
@@ -807,6 +742,107 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
...
+// the slot is consumed. A return value of false means, the container should shutdown and no subsequent
+// calls should be made to this function.
+func (a *agent) runHotReq(ctx context.Context, call *call, state ContainerState, logger logrus.FieldLogger, cookie drivers.Cookie, slot *hotSlot) bool {
+
+	var err error
+	isFrozen := false
+
+	freezeTimer := time.NewTimer(a.freezeIdleMsecs)
+	idleTimer := time.NewTimer(time.Duration(call.IdleTimeout) * time.Second)
+	ejectTicker := time.NewTicker(a.ejectIdleMsecs)
```
rdallman: i gave this a solid 30 seconds of thought, but I can't come up with why this one is a ticker and the other 2 are timers. tickers may drop ticks if the receiver is slow, which isn't something we want exactly I don't think (say, freeze latency puts us past the eject timeout, do we want to skip ejection due to this?). a ticker also will send > 1 time, whereas this function pertains to one instance of acquiring a slot or $alternative (alternative being: shutdown, eject, idle). anyway... ?

api/agent/slots.go
```diff
@@ -70,33 +70,24 @@ func NewSlotQueue(key string) *slotQueue {
...
 	if !atomic.CompareAndSwapUint32(&s.isBusy, 0, 1) {
 		return false
 	}
 
 	a.cond.L.Lock()
-	for i := 0; i < len(a.slots); i++ {
-		if a.slots[i].id == s.id {
-			a.slots = append(a.slots[:i], a.slots[i+1:]...)
-			break
+	if len(a.slots) > 0 {
```
rdallman: the loop works without this branch fwiw

api/agent/agent.go
skinowski: No, only if frozen. An optimization only, if s.trigger is triggered, we can proceed to unfreeze before attempting acquire. I'm going to remove this as it's hard to understand and perhaps premature.

api/agent/slots.go
```diff
@@ -131,19 +122,13 @@ func (a *slotQueue) startDequeuer(ctx context.Context) chan *slotToken {
 				return
 			}
 
-			// pop
 			item := a.slots[len(a.slots)-1]
-			a.slots = a.slots[:len(a.slots)-1]
```
rdallman: i'm finding it challenging to audit the correctness of the process of acquiring a slot now. it now seems like we are relying on acquireSlot to do this, however after the loop in runHotReq we appear to do a slot.Close if we acquire the slot, which now appears to be something we have to do regardless of whether we are trying to eject -- so I'm confused how hot functions are running at all since we seem to be closing every slot we acquire?

api/agent/agent.go
rdallman: this code appears to do the opposite (it freezes if not frozen)

api/agent/slots.go
skinowski: in runHot(), if acquire succeeds, then it means consumer did not acquire it, then we close and exit processing (since it's timeout, shutdown, etc.)

on the consumer side, if acquire fails, it means the token was ejected and consumer must get another one.

api/agent/agent.go
skinowski: I think missing ticks due to freeze latency is OK. It's not really an idleness due to no processing. It is a ticker because we want to check if we can eject it every eg 1 seconds if it was idle.

rdallman: ah, that makes sense.

api/agent/agent.go
skinowski: arrghh, anyway, it's gone. :-)

api/agent/slots.go
rdallman: that's right. so we just moved slot close up to the caller, that makes sense

skinowski:
Adjusted the explanation of this idle eject settings.. The fact that it is a periodic check is not really clear. We can wake up containers when cpu & mem is starved, but it's going to trigger complicated channel/cond.var plumbing, which I'm not sure worth it.... (Perhaps in future..) For now, I think a periodic check of 1 sec is simple/enough... (Although potentially add 1 sec latency to blocked requests.)

PR #752, opened 2018-02-09
fn: more functionality into fn-test-utils

skinowski:
read/write a file on disk, alloc/leak memory.


skinowski:
TODO:
1) oom test (I'll add this soon)
2) xfs - size test for supported platforms
3) leak tests could be nice to see how a bad behaving hot container impacts series of requests.
4) include fdk-go utils to simulate corrupt json/http, slow I/O, close pipe, infinite data pump, etc.

PR #756, opened 2018-02-12
fn: fn-test-utils: partial output and invalid http or json

skinowski:
Simulate partial output or invalid json/html in fn-test-utils.


rdallman:
seems like extracting 'util' didn't buy too much sugar, just the ctx stuff (useful, we updated that a good bit)

PR #760, opened 2018-02-12
Include version number in Docker images

### LINKED ISSUE #758: Docker images should be built with version numbers built into image metadata
I used `fn-helm` to run fnserver inside of a minikube. I'm trying to diagnose an issue with the fnserver, but the helm chart specifies the image as just `fnproject/fnserver:latest`. The docker image itself doesn't have the project version or tag numbers built into it so it's hard to tell what commit the image was built from. 

It's possible to pull other tagged version from Docker hub and compare the image hashes, but it would be much easier if the the Docker image itself had the commit or tag version number built in to it so that it could be found via `docker inspect` or `docker history`.

kojustin:
When building the fnserver docker image, stamp the version number into
the image. This will help with tracking down specific images to specific
version numbers. This will especially be useful with ":latest" images.

Fixes #758.

### DIFF
--- a/api/version/version.go
+++ b/api/version/version.go
@@ -1,4 +1,4 @@
 package version
 
 // Version of Functions
-var Version = "0.3.395"
+var Version = "0.3.396"


Makefile
```diff
@@ -83,7 +83,7 @@ run: build
 	GIN_MODE=debug ./fnserver
 
 docker-build:
-	docker build --build-arg HTTPS_PROXY --build-arg HTTP_PROXY -t fnproject/fnserver:latest .
+	docker build --build-arg HTTPS_PROXY --build-arg HTTP_PROXY --build-arg BUILD_ARG_FN_VERSION -t fnproject/fnserver:latest .
```
rdallman: `--label=$version` ?

PR #761, opened 2018-02-13
fn: favor fn-test-utils over hello (to be decommissioned)

skinowski:
TODO: clean up fn-api-tests to use fn-test-utils.


skinowski:
Part of #656

PR #762, opened 2018-02-13
cap docker retries to fixed number

### LINKED ISSUE #727: return error to user when registry cannot be accessed
seems like multiple users have run into not being able to address their local/private registry when running fn due to container misconfiguration (ECONNREFUSED)

right now this is seemingly tricky to debug since the error shows up only in fn logs, seems like another error we should white list to return in the http response, possibly with a hint/warning about `localhost` ;)

rdallman:
previously we would retry infinitely up to the context with some backoff in
between. for hot functions, since we don't set any dead line on pulling or
creating the container, this means it would retry forever without making any
progress if e.g. the registry is inaccessable or any other temporary error
that isn't actually temporary.  this adds a hard cap of 10 retries, which
gives approximately 13s if the ops take no time, still respecting the context
deadline enclosed.

the case where this was coming up is now tested for and was otherwise
confusing for users to debug, now it spits out an ECONNREFUSED with the
address of the registry, which should help users debug without having to poke
around fn logs (though I don't like this as an excuse, not all users will be
operators at some point in the near future, and this one makes sense)

closes #727

### DIFF
--- a/api/agent/drivers/docker/docker_client.go
+++ b/api/agent/drivers/docker/docker_client.go
@@ -100,12 +100,14 @@ type dockerWrap struct {
 
 func (d *dockerWrap) retry(ctx context.Context, f func() error) error {
 	var i int
+	var err error
 	span := opentracing.SpanFromContext(ctx)
 	defer func() { span.LogFields(log.Int("docker_call_retries", i)) }()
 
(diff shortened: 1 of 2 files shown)


rdallman:
thanks!

PR #765, opened 2018-02-13
support deeper / nesting of image names

### LINKED ISSUE #764: Private repos do not support more than 3 subfolders
/kind bug

If you use more than 3 subfolders in a private repo (with autentication) it silently fails with "no authentication". I traced the root of the problem down to this function:

https://github.com/fnproject/fn/blob/f27d47f2dd9520647f8799043bfcb3d121709958/api/agent/drivers/driver.go#L283-L295

There should be a default case for deeper nestings. If you do not want to support cases like this, at least an error log is necessary.

Workaround is to use shorter nestings. However, this can be problematic if e.g. your company forces you to use prefix folders (e.g. per team and/or parent product) that can then compose to longer names.

rdallman:
closes #764

### DIFF
--- a/api/agent/drivers/driver.go
+++ b/api/agent/drivers/driver.go
@@ -179,6 +179,8 @@ const (
 	StatusTimeout   = "timeout"
 	StatusKilled    = "killed"
 	StatusCancelled = "cancelled"
+
+	defaultDomain = "docker.io"
 )
 
 type Config struct {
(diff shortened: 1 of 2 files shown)


api/agent/drivers/driver.go
```diff
@@ -179,6 +179,8 @@ const (
 	StatusTimeout   = "timeout"
 	StatusKilled    = "killed"
 	StatusCancelled = "cancelled"
+
+	defaultDomain = "docker.io"
```
skinowski: unused?

PR #772, opened 2018-02-14
fixup json,http protocols

### LINKED ISSUE #326: Issue with new streaming responses

rdallman:
* http now buffers the entire request body from the container before copying
it to the response writer (and sets content length). this is a level of sad i
don't feel comfortable talking about but it is what it is.
* json protocol was buffering the entire body so there wasn't any reason for
us to try to write this directly to the container stdin manually, we needed to
add a bufio.Writer around it anyway it was making too many write(fd) syscalls
with the way it was. this is just easier overall and has the same performance
as http now in my tests, whereas previously this was 50% slower [than http].
* add buffer pool for http & json to share/use. json doesn't create a new
buffer every stinkin request. we need to plumb down content length so that we
can properly size the buffer for json, have to add header size and everything
together but it's probably faster than malloc(); punting on properly sizing.
* json now sets content type to the length of the body from the returned json
blurb from the container

this does not handle imposing a maximum size of the response returned from a
container, which we need to add, but this has been open for some time
(specifically, on json). we can impose this by wrapping the pipes, but there's
some discussion to be had for json specifically we won't be able to just cut
off the output stream and use that (http we can do this). anyway, filing a
ticket...

closes #326 :(((((((

### DIFF
--- a/api/agent/protocol/http.go
+++ b/api/agent/protocol/http.go
@@ -2,13 +2,21 @@ package protocol
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net/http"
+	"strconv"
+	"sync"
 
 	"github.com/fnproject/fn/api/models"
 	opentracing "github.com/opentracing/opentracing-go"
+)
+
+var (
+	bufPool = &sync.Pool{New: func() interface{} { return new(bytes.Buffer) }}
 )
 
 // HTTPProtocol converts stdin/stdout streams into HTTP/1.1 compliant
@@ -50,10 +58,19 @@ func (h *HTTPProtocol) Dispatch(ctx context.Context, ci CallInfo, w io.Writer) e
 	if err != nil {
 		return models.NewAPIError(http.StatusBadGateway, fmt.Errorf("invalid http response from function err: %v", err))
 	}
-	defer resp.Body.Close()
 
 	span, _ = opentracing.StartSpanFromContext(ctx, "dispatch_http_write_response")
 	defer span.Finish()
+
+	buf := bufPool.Get().(*bytes.Buffer)
+	buf.Reset()
+	defer bufPool.Put(buf)
+
+	// copy the response body into a buffer so that we read the whole thing. then set the content length.
+	io.Copy(buf, resp.Body)
+	resp.Body.Close()
+	resp.Body = ioutil.NopCloser(buf)
+	resp.Header.Set("Content-Length", strconv.Itoa(buf.Len()))
 
 	rw, ok := w.(http.ResponseWriter)
 	if !ok {
--- a/api/agent/protocol/json.go
+++ b/api/agent/protocol/json.go
@@ -7,21 +7,14 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strconv"
 
 	"github.com/fnproject/fn/api/models"
 	opentracing "github.com/opentracing/opentracing-go"
 )
 
-// This is sent into the function
-// All HTTP request headers should be set in env
-type jsonio struct {
-	Body        string `json:"body"`
-	ContentType string `json:"content_type"`
-}
-
 // CallRequestHTTP for the protocol that was used by the end user to call this function. We only have HTTP right now.
 type CallRequestHTTP struct {
-	// TODO request method ?
 	Type       string      `json:"type"`
 	Method     string      `json:"method"`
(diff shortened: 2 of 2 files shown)


api/agent/protocol/json.go
```diff
@@ -36,19 +29,19 @@ type CallResponseHTTP struct {
...
 // jsonIn We're not using this since we're writing JSON directly right now, but trying to keep it current anyways, much easier to read/follow
 type jsonIn struct {
-	jsonio
-	CallID      string           `json:"call_id"`
-	ContentType string           `json:"content_type"`
-	Type        string           `json:"type"`
-	Deadline    string           `json:"deadline"`
-	Body        string           `json:"body"`
-	Protocol    *CallRequestHTTP `json:"protocol"`
+	CallID      string          `json:"call_id"`
```
skinowski: :clap:

api/agent/protocol/json.go
```diff
@@ -36,19 +29,19 @@ type CallResponseHTTP struct {
...
+	Type        string          `json:"type"`
+	Deadline    string          `json:"deadline"`
+	Body        string          `json:"body"`
+	ContentType string          `json:"content_type"`
```
skinowski: :clap:

api/agent/protocol/json.go
```diff
@@ -36,19 +29,19 @@ type CallResponseHTTP struct {
...
+	Protocol    CallRequestHTTP `json:"protocol"`
 }
 
 // jsonOut the expected response from the function container
 type jsonOut struct {
-	jsonio
-	Protocol *CallResponseHTTP `json:"protocol,omitempty"`
+	Body        string            `json:"body"`
+	ContentType string            `json:"content_type"`
```
skinowski: :clap:

api/agent/protocol/http.go
```diff
@@ -2,15 +2,23 @@ package protocol
...
+	"sync"
 
 	"github.com/fnproject/fn/api/models"
 	opentracing "github.com/opentracing/opentracing-go"
 )
 
+var (
+	bufPool = &sync.Pool{New: func() interface{} { return new(bytes.Buffer) }}
+)
+
```
skinowski: maybe factory.go is a better place? not a big deal...

api/agent/protocol/json.go
```diff
@@ -62,131 +55,33 @@ func (p *JSONProtocol) IsStreamable() bool {
...
-			return err
-		}
-		err = stdin.Encode(ci.Headers())
-	}
-	err = writeString(err, h.in, "}")
-
-	// close
-	err = writeString(err, h.in, "\n}\n\n")
-	return err
+	return json.NewEncoder(h.in).Encode(in)
```
skinowski: :clap:

rdallman:
thanks. wrt the buffer pool... we have 2 in func_logger as it is and this one as well. it seems like it would be a useful thing to have a general purpose buffer pool that can dispense based on some expected size. for this one, it's not perfect since http is using it for container output and json for request input, but in a heterogeneous workload we can't expect that those 2 differ a whole lot (one of the func_logger ones is just for lines, which should be short generally).

skinowski:
So, with this, we are always buffering. (no streaming mode), which is fine in my opinion. Only to-do here is to set a system wide limit and if resp exceeds, cut off response, get rid of buffered data (clean pipes), then return an error (eg. data too large), right?

rdallman:
something like that. opened #773 - default needs to be handled, as well. we could either return an error or return partial output and somehow try to indicate that their output was too long. not sure how worth it / challenging the partial output thing is (we're truncating logs, but stdout is a little different).

PR #776, opened 2018-02-15
fn: ui no longer uses /stats

skinowski:
Decommission /stats related code.


skinowski:
This could be a bit early and requires latest fn-ui, but /stats is no longer used in UI.

rdallman:
sweet, just did this too (easy rebase, did exact same thing). let's freeze this wart off

PR #778, opened 2018-02-16
reduce allocs in getSlotQueueKey

rdallman:
this somewhat minimally comes up in profiling, but it was an itch i needed to
scratch. this does 10x less allocations and is 3x faster (with 3x less bytes),
and they're the small painful kind of allocation. we're only reading these
strings so the uses of unsafe are fine (I think audit me).  the byte array
we're casting to a string at the end is also heap allocated and does 
escape. I only count 2 allocations, but there's 3 (`hash.Sum` and
`make([]string)`), using a pool of sha1 hash.Hash shaves 120 byte and an alloc
off so seems worth it (it's minimal). if we set a max size of config vals with
a constant we could avoid that allocation and we could probably find a
checksum package that doesn't use the `hash.Hash` that would speed things up a
little (no dynamic dispatch, doesn't allocate in Sum) but there's not one I
know of in stdlib.

master:
```
✗: go test -run=yodawg -bench . -benchmem -benchtime 1s -cpuprofile cpu.out
goos: linux
goarch: amd64
pkg: github.com/fnproject/fn/api/agent
BenchmarkSlotKey          200000              6068 ns/op             696 B/op         31 allocs/op
PASS
ok      github.com/fnproject/fn/api/agent       1.454s
```

now:
```
✗: go test -run=yodawg -bench . -benchmem -benchtime 1s -cpuprofile cpu.out
goos: linux
goarch: amd64
pkg: github.com/fnproject/fn/api/agent
BenchmarkSlotKey         1000000              1901 ns/op             168 B/op          3 allocs/op
PASS
ok      github.com/fnproject/fn/api/agent       2.092s
```

once we have versioned apps/routes we don't need to build a sha or sort
configs so this will get a lot faster.

anyway, mostly funsies here... my life is that sad now.

### DIFF
--- a/api/agent/slots.go
+++ b/api/agent/slots.go
@@ -3,10 +3,12 @@ package agent
 import (
 	"context"
 	"crypto/sha1"
-	"fmt"
+	"encoding/binary"
+	"hash"
 	"sort"
 	"sync"
 	"sync/atomic"
+	"unsafe"
 )
 
 //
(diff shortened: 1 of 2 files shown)


api/agent/slots.go
```diff
@@ -270,31 +272,70 @@ func (a *slotQueueMgr) deleteSlotQueue(slots *slotQueue) bool {
...
+	binary.LittleEndian.PutUint64(byt[:], uint64(call.CPUs))
+	hash.Write(byt[:])
+	hash.Write(unsafeBytes("\x00"))
+
+	// we have to sort these before printing, yay.
+	// TODO if we had a max size for config const we could avoid this!
 	keys := make([]string, 0, len(call.Config))
 	for k := range call.Config {
-		keys = append(keys, k)
+		i := sort.SearchStrings(keys, k)
```
skinowski: avoids recursive sort and hence faster?

skinowski: oh I see, memmove + bin search insertion sort.

rdallman: this reduces the number of times we do linear operations to 2. previously we ran over call.Config to append keys, then sort, then add the strings. we still have the same asymptotic bounds as the sort (nlogn) [I think] since we're doing logn search n times, but we save one extra n by doing it in this loop we were already taking. it's not much, but it's something (the nature of this patch).

rdallman:
thanks

PR #784, opened 2018-02-20
use sync call for some api tests

rdallman:
a lot of the logging tests use async calls which was flaky with timings. there
are a couple left that use async, that we need to test, but this reduces the
probability of the flaky tests flaking to 2/5 instead of 5/5.

the behaviors we are testing here we should not be testing through the api,
for this exact reason, we can test a ton of this in the agent itself and not
have to rely on timings. not taking that here, but this fixes circle ci
flaking out for a few days.


skinowski:
:clap:  merged this (since I really needed it)

Issue #773, opened 2018-02-14
option to limit container output size as a server setting


rdallman:
services will need to impose limits on this or users will easily be able to OOM `fn` servers by writing large outputs from their container. we already have the ability to limit the request size of the input with https://github.com/fnproject/fn/blob/master/api/server/server_options.go#L20 for symmetry.

this needs to be imposed for all formats and is mostly a plumbing issue.

PR #786, opened 2018-02-20
Response size clamp

skinowski:
*) Limit response http body or json response size to FN_MAX_RESPONSE_SIZE (default unlimited)
*) If limits are exceeded 502 is returned with 'body too large' in the error message

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -505,10 +505,17 @@ func (s *coldSlot) exec(ctx context.Context, call *call) error {
 		return models.NewAPIError(http.StatusBadGateway, fmt.Errorf("invalid response from function err: %v", s.clamper.E))
 	}
 
-	// yes, we leak this and retain the buf until client times out.
-	buf := s.clamper.W.(*bytes.Buffer)
 	errApp := make(chan error, 1)
+
+	// take ownership of this buffer to avoid cold token.Close() to touch it.
+	buf := s.clamper.W.(*bytes.Buffer)
+	s.clamper = nil
+	// yes, we leak this and retain the buf until client times out. But this timeout is
+	// closely bounded by our own ctx timeout below. Once we return err, we expect gin
+	// to close/finalize client request/connection.
 	go func() {
+		// IMPORTANT: keep the buffer until client is done.
+		defer protocol.BufPool.Put(buf)
 		_, err := io.Copy(call.w, ioutil.NopCloser(buf))
 		errApp <- err
 	}()
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -503,11 +503,11 @@ func (s *coldSlot) exec(ctx context.Context, call *call) error {
 		return res.Error()
 	}
 
-	if s.stdout == nil {
+	// This means default IO with no limits (no clamper)
+	clamper, ok := s.stdout.(*protocol.ClampWriter)
+	if !ok {
 		return ctx.Err()
 	}
-
-	clamper := s.stdout.(*protocol.ClampWriter)
 	if clamper.E {
 		return models.NewAPIError(http.StatusBadGateway, fmt.Errorf("invalid response from function err: body too large"))
 	}
@@ -659,13 +659,7 @@ func (a *agent) prepCold(ctx context.Context, call *call, tok ResourceToken, ch 
 	buf := protocol.BufPool.Get().(*bytes.Buffer)
 	buf.Reset()
 
-	var stdout io.Writer
-	if protocol.BufPoolChunkSize == 0 {
-		stdout = buf
-	} else {
-		stdout = &protocol.ClampWriter{W: buf, N: protocol.BufPoolChunkSize}
-	}
-
+	stdout := protocol.NewClampWriter(buf)
 	container := &container{
 		id:      id.New().String(), // XXX we could just let docker generate ids...
 		image:   call.Image,
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -518,10 +518,11 @@ func (s *coldSlot) Close(ctx context.Context) error {
 
 // implements Slot
 type hotSlot struct {
-	done      chan struct{} // signal we are done with slot
-	errC      <-chan error  // container error
-	container *container    // TODO mask this
-	err       error
+	done        chan struct{} // signal we are done with slot
+	errC        <-chan error  // container error
+	container   *container    // TODO mask this
+	maxRespSize uint64        // TODO boo.
+	err         error
 }
 
 func (s *hotSlot) Close(ctx context.Context) error {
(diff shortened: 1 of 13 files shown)


api/agent/agent.go
```diff
@@ -112,72 +113,42 @@ type agent struct {
...
-		ServerVersion: "17.06.0-ce",
-	})
-
-	freezeIdleMsecs, err := getEnvMsecs("FN_FREEZE_IDLE_MSECS", 50*time.Millisecond)
-	if err != nil {
-		logrus.WithError(err).Fatal("error initializing freeze idle delay")
-	}
 
-	ejectIdleMsecs, err := getEnvMsecs("FN_EJECT_IDLE_MSECS", 1000*time.Millisecond)
+	cfg, err := NewAgentConfig()
```
rdallman: thanks for adding this.

api/agent/protocol/http.go
```diff
@@ -62,12 +63,18 @@ func (h *HTTPProtocol) Dispatch(ctx context.Context, ci CallInfo, w io.Writer) e
 	span, _ = opentracing.StartSpanFromContext(ctx, "dispatch_http_write_response")
 	defer span.Finish()
 
-	buf := bufPool.Get().(*bytes.Buffer)
+	buf := BufPool.Get().(*bytes.Buffer)
 	buf.Reset()
-	defer bufPool.Put(buf)
+	defer BufPool.Put(buf)
+
+	reader := NewClampReader(resp.Body)
```
rdallman: i'm worried about doubly imposing this limit with https://github.com/fnproject/fn/blob/master/api/server/server_options.go#L20 -- that guy has the benefit of limiting the request body for every handler, but plumbing wise checking for that error from in here sucks. some options I guess. at least, there's some dupe code for limit reader stuff it looks like.

api/agent/agent.go
```diff
@@ -112,72 +113,42 @@ type agent struct {
...
+		cfg:         *cfg,
+		da:          da,
+		driver:      driver,
+		slotMgr:     NewSlotQueueMgr(),
+		resources:   NewResourceTracker(),
+		shutdown:    make(chan struct{}),
+		promHandler: promhttp.Handler(),
 	}
 
+	protocol.BufPoolChunkSize = int64(cfg.MaxResponseSize)
```
rdallman: this one kinda smells. I see a lot of ways to skin the cat on this one. was this preferable to wrapping the reader or writer being passed off to dispatch [as stdin/stdout] in the same manner we're doing for stderr (`func_logger.go`)?

api/agent/protocol/factory.go
```diff
@@ -172,3 +172,57 @@ func New(p Protocol, in io.Writer, out io.Reader) ContainerIO {
 // IsStreamable says whether the given protocol can be used for streaming into
 // hot functions.
 func IsStreamable(p Protocol) bool { return New(p, nil, nil).IsStreamable() }
+
+type ClampWriter struct {
```
rdallman: `func_logger.go` has this guy too.

rdallman: we used to have `common/writers.go` that had various io.Writer utilities, we could do that again

api/agent/agent.go
```diff
@@ -532,7 +504,37 @@ func (s *coldSlot) exec(ctx context.Context, call *call) error {
...
+	s.stdout = nil
+	// yes, we leak this and retain the buf until client times out. But this timeout is
+	// closely bounded by our own ctx timeout below. Once we return err, we expect gin
+	// to close/finalize client request/connection.
+	go func() {
+		// IMPORTANT: keep the buffer until client is done.
+		defer protocol.BufPool.Put(buf)
+		_, err := io.Copy(call.w, ioutil.NopCloser(buf))
+		errApp <- err
+	}()
```
rdallman: Dispatch and cold have never played well together, the back story on that is that if you start the container and it ignores stdin then it can exit before dispatch is ready to see what's going on or we've handed out the slot to actually receive the error from the container. it sucks. i wish we could thread cold through Dispatch. maybe someone with a bigger brain than I can solve the puzzle there.

all that's to say, this seems really hairy. this appears to add buffering of default format responses before shipping them out. dispatch was a good spot to put this (for response headers crap, mostly) and i'm sad we can't put this there and I don't think we can thread a buffer as the writer in front of all calls due to the headers crap (that has to be done in dispatch on a ResponseWriter). am trying to think but all jolted on cheap burnt conference coffee and need to eat. i'll leave that to marinate on and be back later... in any case, it seems like we should find a unified place to either tuck this and return a `models.APIError` from the `io.Copy` calls or from somewhere else (there are challenges here with pipes, i'm aware).

api/agent/agent.go
```diff
@@ -583,24 +592,54 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
...
+	// more data after) but this does clear naive (most common) cases out
+	// such as container spitting out too much (or faulty) data at once.
+	// If we fail to clear this pipe (stdoutRead), then docker IO demultiplex
+	// stdCopy will notice that we have not read all the data from it and will
+	// shutdown the container.
+	select {
+	case finalError = <-s.errC: // give precedence to error from container
+	case <-time.After(time.Duration(500) * time.Millisecond):
+	case <-ctx.Done(): // call timeout
+	}
```
rdallman: closing the pipes takes care of this and this is redundant I think? the docker stdcopy read will be stuck on an old pipe reader in io.Read? see: conf coffee issue. bbl.

api/agent/agent.go
skinowski: yeah, I'll take a look at this one more time. I kind of started with global var like this because this was tied to system wide env var, but now I moved it under the agent, who sets it, so it surely looks a bit unsavory.

api/agent/protocol/factory.go
skinowski: I should move these to common/writers.go. But the func_logger one is a bit specialized for log writing it seems.

api/agent/agent.go
skinowski: So, this has turned out to be tricky. If we close the pipe here, then docker stdcopy doesn't like it. Partial read causes docker to exit the container. That exit 2 event happens after we return here, and sometimes causes next request to fail. Adding 500msec latency to these failures is something I don't like (and it also doesn't 100% cover all cases), but the alternative is to kill the hot container. (which I actually like. :-) )

api/agent/protocol/http.go
skinowski: Doesn't that pertain to gin contexts (specifically request.Body) though? This is the response, and it is a brand new http response we synthesize here.

api/agent/agent.go
```diff
@@ -583,24 +592,54 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
...
 	go func() {
 		ci := protocol.NewCallInfo(call.Call, call.req)
-		errApp <- proto.Dispatch(ctx, ci, call.w)
+		err := proto.Dispatch(ctx, ci, call.w)
+
+		// drain the container stdout if bad gw (aka container communication failed)
+		if err != nil && models.GetAPIErrorCode(err) == http.StatusBadGateway {
+			go func() {
+				// stdoutWrite.Close() will unblock this with EOF incoming from stdoutRead
+				io.Copy(ioutil.Discard, stdoutRead)
```
rdallman: idear: can we do this in the reader when we exceed the max instead of the contract requiring the caller to do this (see this at least 2 places)

api/agent/agent.go
rdallman: hmm. would this add a condition to our going terminator on containers?

* if container i/o exceeds time bounds
* if container i/o doesn't respect our imposed bounds?

this one seems recoverable, but maybe no

api/agent/agent.go
```diff
@@ -846,14 +888,10 @@ type container struct {
 }
 
 func (c *container) swap(stdin io.Reader, stdout, stderr io.Writer, cs *drivers.Stats) func() {
-	ostdin := c.stdin.(*ghostReader).inner
-	ostdout := c.stdout.(*ghostWriter).inner
-	ostderr := c.stderr.(*ghostWriter).inner
```
rdallman: d'oh... thanks.

api/agent/agent.go
skinowski: I could try to move this code into Dispatch(), as default protocol. There's already some dead code in there for cold-dispatch (eg. default.go)

skinowski: I looked at this one more time, but trying to move this into Dispatch() doesn't gain us much, the exec path is significantly different for cold.

Also I think we do want buffering of responses in all cases anyway, right?

skinowski:
Fixes #773

skinowski:
@rdallman this is rebased/cleanup now, which is essentially your pipe/clamp changes.

PR #641, opened 2018-01-05
App ID

denismakogon:
Closes: #607

PR #787, opened 2018-02-21
Switch migrations framework

denismakogon:
Why?
-----
Only SQL statements can solve problems partially, no way to work with the external data source.

What's used?
-------------
Goose, but it's still bad, but better than what we have at this moment. With some hacking, we can use go migrations which is simple and more powerful than just SQL statement because gives us the chance to hook into datastore modifications within the migration with help of `sql.Tx`.

What's changed?
-----------------
The way we write and consume migrations, no more code generation.
Go migrations instead of raw SQL migrations

Related to #641


denismakogon:
Okay, done. Migrations work as suppose to. No downsides observed. Goose uses the different table to store migration versions, but it doesn't break existing models, however, have to adjust table definitions to make migrations less harmful, meaning that any migration should not be manually reflected in the table definition (see the code where I dropped fields from the tables that are present at the migrations, this is how should actual migrations look like AFAIK). So, the migrations itself were written with the assumption that there is the datastore that Fn uses already, meaning that migration wouldn't fail with `duplicate column` error.

rdallman:
> So, the migrations itself were written with the assumption that there is the datastore that Fn uses already, meaning that migration wouldn't fail with duplicate column error.

this seems pretty dangerous. if a migration fails we need to bail and let the user manually recover the db itself. migrations can depend on previous migrations and we're in a weird state where the migration version is the latest, but a migration didn't work, so it's not clear which migration broke and which migrations need to be ran again after resolving the issue. basically, we need to check the errors from up migrations (they are ignored).

since we're migrating migration libraries, we need to read the version from the old migration table (we don't necessarily need to keep the whole import, we can do a quick sql query) and use it to start the migrations from there using the new migration code. after we do that, we can drop the old migration table (this is a migration in itself?) to indicate fn/db has upgraded (version is sufficient here). this is a little bit of a pain, but doesn't seem too bad.

api/datastore/sql/sql_test.go
```diff
@@ -22,12 +24,7 @@ func newWithMigrations(ctx context.Context, url *url.URL) (*sqlStore, error) {
 		return nil, err
 	}
 
-	m, err := migrator(url.String())
-	if err != nil {
-		return nil, err
-	}
-
-	err = m.Down()
+	err = migrations.DownByOne(url.Scheme, ds.db.DB)
```
rdallman: `m.Down()` went all the way down, which then let the subsequent opening of the db go all the way back up. this ensures all migrations work (against all sql stores, except sqlite3 which doesn't like `DROP COLUMN`).

api/datastore/sql/sql.go
```diff
@@ -43,50 +40,6 @@ import (
...
-	PRIMARY KEY (id)
-);`,
-
-	`CREATE TABLE IF NOT EXISTS logs (
-	id varchar(256) NOT NULL PRIMARY KEY,
-	app_name varchar(256) NOT NULL,
-	log text NOT NULL
-);`,
-}
-
```
rdallman: i don't see any code to replace these... we still need to handle fresh dbs.

api/datastore/sql/sql.go
```diff
@@ -178,77 +125,6 @@ func pingWithRetry(ctx context.Context, attempts int, sleep time.Duration, db *s
...
-		return m.Force(latestVersion(migrations.AssetNames()))
-	}
-
-	// run any migrations needed to get to latest, if any
-	err = m.Up()
-	if err == migrate.ErrNoChange { // we don't care, but want other errors
-		err = nil
-	}
-	return err
-}
```
rdallman: i think we need to keep this logic so that fresh dbs skip all migrations still? this logic seems mostly drop in here.

api/datastore/sql/sql_test.go
denismakogon: Okay, can adjust that.

api/datastore/sql/sql.go
denismakogon: That is handled inside the migration. Fresh DB means that it doesn't have the migrations applied, so the migration DB check will return `-1` as the indicator of a fresh new datastore.

rdallman: right, but I don't see the `CREATE TABLE` replacements ? maybe github didn't expand the new file they are in and my search is failing?

api/datastore/sql/sql.go
denismakogon: i disagree, even fresh datastore should be migrated since the tables were changed, so right after we create necessary tables we need to apply all migrations we have to make datastore consistent with Fn models.

rdallman: then we have to reconstruct the tables themselves from the deltas of all migrations in our head to know what a table looks like. cognitively, this is challenging with a non-trivial number of migrations.

api/datastore/sql/sql.go
denismakogon: it's there, at `migration.so`, no worries.

api/datastore/sql/sql.go
denismakogon: That's the point of the goose migrations, every change to the table (or a new table creation) should lay in migration, but not somewhere in the code. This is what we did a bit incorrectly at the very beginning. But this is what we have to do at this moment with goose, we **have** to reconstruct the tables by applying any number of the migrations we have to make datastore consistent.

rdallman: cat skinning...

if we kept schema files with the current state somewhere in our codebase it would be beneficial. as a dev, i don't want to go run sql queries against the db (nor is this plausible always) just to see what the db looks like before i go modify some fields in a patch.

denismakogon: well, this is some kind of limitation that we have to live with, just as we lived with SQL only migrations, gonna have to get used to.

rdallman: we don't have to do anything, such is life :)

since we're going to have crippled CREATE TABLE to reference now can we find some way to have full schema of each table with the latest db and ensure in our tests that they actually work? right now we can reference `INSERT INTO` for a given table since fields are generally enumerated in that query, and this is tested in tests [generally], but this does not contain the type itself. one way to do this would be to in tests seed the db with the full schema to CREATE TABLE for each of our tables and then run all of our tests against the db, if this blows up then the schemas need to be updated (this would only happen in testing/CI). it's significantly less error prone and creates a significantly better user experience (when we have 100s of migrations to run on first boot) to just do this and skip migrations for fresh dbs...

api/datastore/sql/migrations/migrations.go
```diff
@@ -1,534 +1,169 @@
...
-		return RestoreAsset(dir, name)
+		if err != goose.ErrNoNextVersion {
+			return err
+		}
 	}
-	// Dir
-	for _, child := range children {
-		err = RestoreAssets(dir, filepath.Join(name, child))
-		if err != nil {
+	// datastore is fresh new
```
rdallman: we also need to check if they are coming from the previous migration package? If we special case errors coming out of there then we can end up with a db that's in a funky state (say some DBA went around changing the tables..)

api/datastore/sql/sql.go
denismakogon: > it's significantly less error prone and creates a significantly better user experience (when we have 100s of migrations to run on first boot) to just do this and skip migrations for fresh dbs...

well, from my previous experience with other migrations frameworks (like one in Django, oslo.db, alembic), the number of migration to apply at the first boot guaranties that datastore 100% consistent, it happens once, so having migrations run once is totally fine, even in cluster, only the first Fn instance suppose to run all that migrations at once, any other instances will identify the migrations version and will run only the delta (from the particular version to the last one).

> since we're going to have crippled CREATE TABLE to reference now can we find some way to have full schema of each table with the latest db and ensure in our tests that they actually work?

This is what we had to worry about when we decided to use the migrations. At this moment we have no versioned objects or updated tests. So we don't know if the upgrade actually works until we test that manually. This is our fault that we have to fix.

Despite all of that, every time Fn will call datastore.New it will get the consistent datastore with latest migrations, so if the developer forgot to include the migration to the chain, it's his fault. This type of problems we already can catch with datastore tests as long as we test the migrations.

api/datastore/sql/migrations/migrations.go
denismakogon: I'm not sure we actually have to worry about that because the migrations we have at this moment check SQL error to make sure that it's a duplicated column error (specific to each datastore we support). So we don't have to worry about previous version table, according to the process of migrations applying (in this patch) there's no point to check that version because we are safe to apply the migrations.

api/datastore/sql/sql.go
rdallman: > This type of problems we already can catch with datastore tests as long as we test the migrations.

slightly different problem: 

* me, developer, wants to change the type of a column
* how do i determine the type of that column? I must sift through all migrations or go ask a current db for the type of that column?

api/datastore/sql/migrations/migrations.go
denismakogon: regarding DBA, we aren't exposing datastore API to change the datastore schemas, if someone did that - it's his fault.

api/datastore/sql/sql.go
denismakogon: > how do i determine the type of that column? I must sift through all migrations or go ask a current db for the type of that column?

Exactly, consistent datastore or the migrations are the sources of truth if you want to identify the column type.

rdallman: the last time i worked with a sql db was mostly rails, which active record very nicely handles maintaining a schema which is generated after migrations are run and this is dumped and checked into VCS and used to create new databases as well as for tests (what I'm hinting at). if we could automate that, great. maybe someone else can articulate what I mean: http://guides.rubyonrails.org/v3.2.8/migrations.html#schema-dumping-and-you -- we want/need an authoritative reference for each table, and there are some side benefits we can take advantage if we aren't religiously opposed to certain things.

api/datastore/sql/migrations/migrations.go
rdallman: we're running migrations twice, effectively? it kind of smells, I could see that causing issues. we're effectively migrating a database that is up to date - there seems to be a reason to skip these, what with the versioning of each migration and deliberately skipping over migrations that have already run.

denismakogon: We're not running migrations twice. If datastore has the latest version we skip the migrations.

denismakogon: see `migrations.go`, lines 158 to 166

rdallman: this is only if they were coming from this migrations library? there was a note somewhere that the table that keeps the migration version is different for goose than for the library we were using.

denismakogon: that's true, indeed, tables are different, but as noted as well, reading another migrations table is redundant, it brings yet another overhead to the migrations. Basically, nothing would be changed inside Fn tables, but a new table for the migrations would be created.

So the migration from `migrate` framework to `goose` ends up in a new table for the migration versions and zero diffs for the Fn tables.

rdallman: right. we could just read the version from old migrate table, drop that table, and then proceed like normal? this would fix the error handling stuff, and at least doesn't seem too challenging to do but maybe it is?

denismakogon: Done, new commit adds corresponding check, see 68a523d.

api/datastore/sql/sql.go
denismakogon: this kind of migrations works only for that particular framework, goose would fail every migration if the table would already contain exactly the same column, so it's bad behaviour in our particular case. So, having initial tables plus bunch of migrations (1-7) is definitely our path at this moment as long we use goose.

rdallman:
will get back to this tomorrow, sorry for the delay, took longer to get stuff done than i wanted the last 2 days

api/datastore/sql/migrations/migrations.go
```diff
@@ -55,6 +55,35 @@ var initialTables = [...]string{`CREATE TABLE IF NOT EXISTS routes (
 );`,
 }
 
+func checkOldMigrationTableVersionIfExists(db *sql.DB) (version int64, dirty bool, err error) {
+	migrationsTable := "schema_migrations"
```
rdallman: can you document this a little better please? `// "schema_migrations" is the table used by github.com/mattes/migrate to maintain migration versions in an old version of fn` -- in 3-6 months we can probably do away with this code, it's nice to leave a note so we don't have to sift through our faulty brains / github.

api/datastore/sql/migrations/migrations.go
```diff
@@ -55,6 +55,35 @@ var initialTables = [...]string{`CREATE TABLE IF NOT EXISTS routes (
...
+	ctx := context.Background()
+
+	// deleting old table after all
+	defer func() {
+		_, _ = db.ExecContext(ctx, "DROP TABLE IF EXISTS "+migrationsTable)
```
rdallman: this needs to happen after we set the version in the new migrations table only? otherwise if migrations fail, we risk skipping them.

api/datastore/sql/migrations/migrations.go
```diff
@@ -55,6 +55,35 @@ var initialTables = [...]string{`CREATE TABLE IF NOT EXISTS routes (
...
+
+	err = db.QueryRowContext(
+		ctx, "SELECT version, dirty FROM "+migrationsTable+" LIMIT 1").
+		Scan(&version, &dirty)
+	switch {
+	case err == sql.ErrNoRows:
+		return -1, false, nil
+
+	case err != nil:
+		if e, ok := err.(*mysql.MySQLError); ok {
```
rdallman: sqlite3 is probably the most popular option (also pg, less popular)? not sure if required, but it seems like as long as we check the `dirty` bit, we can skip this error type checking altogether and just remove this, but i didn't test.

api/datastore/sql/migrations/migrations.go
```diff
@@ -126,6 +155,11 @@ func DownAll(driver string, db *sql.DB) error {
 func ApplyMigrations(ctx context.Context, driver string, db *sql.DB) error {
 	log := common.Logger(ctx)
 
+	// no point to handle error from here, basically err
+	// indicates about any problem with old migrations table
+	current, _, err := checkOldMigrationTableVersionIfExists(db)
```
rdallman: if the dirty bit is set, we need to exit the program and suggest the user to fix their db manually (I don't expect this to happen, but i'd rather have this than not and try to debug it)

api/datastore/sql/migrations/migrations.go
rdallman: some confirmation that this was tested? it's not worth having automated tests for, imo

api/datastore/sql/sql.go
rdallman: I don't understand the concrete arguments for doing it this way other than that 'it's easier'. now that we at least try to skip the old migrations, I still do not understand the benefits of us running every migration on first boot and not maintaining a schema file where we have to figure out deltas as devs aside from the benefit of being able to skip the migrations. this seems to have nothing to do with the framework we're using, goose is very much just a library to do migrations, we can use it however we want. i see 0 blockers on using the same behavior as we had previously with migrate using goose, and 0 good reasons to switch to this newly proposed behavior. the old migrate library didn't maintain schema files either, we elected to keep an up to date schema in the CREATE TABLE blocks in our code and skip migrations for fresh boots. this new way proposed needs justification.

rdallman:
some comments, please take your time and make sure this works (test all dbs, etc), breaking user dbs is no fun. we're not in a big hurry

api/datastore/sql/migrations/1_add_route_created_at.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"database/sql"
+	"github.com/pressly/goose"
+)
+
+// we skip the error here to make previous datastore tables work fine
+func up1(tx *sql.Tx) error {
+	_, err := tx.Exec("ALTER TABLE routes ADD created_at text;")
+	return checkMigrationsUpError(err)
```
rdallman: we shouldn't be doing this in any of our migrations, the migration should be written differently and not rely on this behavior

rdallman: this being 'checkMigrationsUpError' - delete this

api/datastore/sql/migrations/migrations.go
denismakogon: It was tested, i used the following steps:
1. checkout `master` and build the binary
2. start Fn server with corresponding datastore(sqlite, mysql, pg), create an app, shut down the Fn server.
3. checkout to this branch and build the binary.
4. start the Fn server
this is what you're going to get:
```
INFO[2018-02-28T22:35:37+02:00] datastore dialed                              datastore=sqlite3 max_idle_connections=256
INFO[2018-02-28T22:35:37+02:00] goose: current datastore version: 0          
DEBU[2018-02-28T22:35:37+02:00] old migration table version was: 7           
OK    1_add_route_created_at.go
OK    2_add_call_stats.go
OK    3_add_call_error.go
OK    4_add_route_updated_at.go
OK    5_add_app_created_at.go
OK    6_add_app_updated_at.go
OK    7_add_route_cpus.go
DEBU[2018-02-28T22:35:37+02:00] Migrations left to apply: 0                  
```
It would be the same for mysql and postgres.

api/datastore/sql/migrations/migrations.go
denismakogon: fixed

api/datastore/sql/migrations/migrations.go
denismakogon: fixed

api/datastore/sql/migrations/migrations.go
denismakogon: fixed

api/datastore/sql/migrations/migrations.go
denismakogon: fixed, failing if dirty

api/datastore/sql/migrations/1_add_route_created_at.go
denismakogon: this was necessary in that case if we want to keep table definitions as they were (with new tables from the migrations), since tables where adjusted, there's no point to hide dup. column error.

denismakogon:
I reverted dup. column check because it doesn't let migrations to be applied, we need this error check only for old migrations, there is no way to make it more clean and nice expect that odd thing - keeping table definitions update with every new migration which is 100% not developer friendly, this is one more reason to play the blame-game with an old migrations framework.

rdallman:
> I reverted dup. column check because it doesn't let migrations to be applied, we need this error check only for old migrations,

we should skip the old migrations though? not following here. we grab the version we were at from the old migrations, set the new migrations version to that, then run new migrations from that point? what is the current behavior?

denismakogon:
> we should skip the old migrations though?

I don't see the point in skipping the migrations, it will make us do a lot of plumbing around goose because the version is not just an int somewhere in the database it is the whole table, so we need to go through the goose to make do version increment of us, that's why we need to run migrations.

> we grab the version we were at from the old migrations, set the new migrations version to that, then run new migrations from that point? what is the current behaviour?

Yes, we grab the version and that's it. To make good bump its version we need to run migrations, we can't just about across that, if we want to bump that manually we need to implement our own migrations framework.

rdallman:
> I don't see the point in skipping the migrations, it will make us do a lot of plumbing around goose because the version is not just an int somewhere in the database it is the whole table, so we need to go through the goose to make do version increment of us, that's why we need to run migrations.

a quick glance and it really is https://github.com/pressly/goose/blob/master/dialect.go#L65 - they just don't have a way to set this version it seems from your later comment? we don't need to make our own migrations framework, no need for hyperbole, it really doesn't seem all that challenging to add this functionality to the goose library upstream. 

the point in skipping them is that it's pretty hairy to run migrations that have already been ran, and then try to ignore errors on this basis. this can leave us with a database in an inconsistent state.

unrelated, another thing I see we'll need to add to goose is oracle db support. this won't require us to make our own migrations framework, either; at worst, we fork goose, at best we push this upstream.

rdallman:
there are likely other options there, too. sad ones, but options. we could keep migrate to run all the old migrations and then only cut over new migrations to use goose. later on, we can drop the old migrations. this is kind of sad since it seems easy to avoid. there's possibly other options than this

rdallman:
really the whole migrations thing originally we could have implemented our own fairly trivially, it was just something of a convenience to use an OSS one but the functionality is not hard to implement at all (version sql queries and run them in order). if this remains contentious, can go off and do this, but goose seems like about what we need short of a few easy to add things.

denismakogon:
The latest commit does one tricky thing, it keeps the old table around and applies only newer migrations that will be reflected in new goose table. The code is fairly simple.
But i do feel like we need to create a function that will sync two tables and then run the migrations. Unfortunately, it will bring a lot of hairy code, but hopefully, it will go away in a half a year, maximum.

rdallman:
>  But i do feel like we need to create a function that will sync two tables and then run the migrations. Unfortunately, it will bring a lot of hairy code, but hopefully, it will go away in a half a year, maximum.

i still don't think we're on the same wave length here. I'll try to enumerate all the cases I can think up, none of which don't seem like they would require complex logic and there are multiple ways to interpret the RHS:

* fresh db -> skip all migrations, create tables from schema
* db up to date with mattes/migrate migrations -> run any goose migrations (optional: drop mattes table)
* db behind on mattes/migrate -> run mattes/migrate migrations, run goose migrations
* db up to date on goose migrations -> do nothing

there are some ways to skin the cat there, but I don't see any need to sync the two tables. our code needs to make sure first that the db is up to date with the old migrations, and then with the goose migrations; that's really it. at present it looks like we're skipping altogether based on version, and there was another attempt to just run the goose migrations over whatever, neither of these seem like feasible options. i do think there are a couple ways to actually go about running the required migrations (one model, where they all run in goose and we start goose at the right spot. another model, where we just use the mattes lib like we have and then start the goose ones fresh).

denismakogon:
> fresh db -> skip all migrations, create tables from schema (set version in goose?)

Disagree, in case of fresh DB we still need to apply all the migrations we have.

> db up to date with mattes/migrate migrations -> run any goose migrations (optional: drop mattes table)

Disagree, in case of up-to-date (or version(s) behind) mattes/migrate table we need to sync that into a new goose table (which is the very easy thing to do), then see if we still need to apply more migrations starting from the point where mattes/migrate ended.

> db behind on mattes/migrate -> run mattes/migrate migrations, run goose migrations

The worst option, I've spent time reading mattes/migrate and it appeared that we violate its rules because it says (and shows in the examples) that the table creation **IS** the migration, but we started our migrations from next step, so we had to make tables creation our migrations version 1. This why we have so much hard time reaching the consensus here. I am strongly against any attempt to leave mattes/migrate with us.

> db up to date on goose migrations -> do nothing

this implies to your 2nd point - we need to merge old mattes table to goose table (delete mattes table after all) and then based on the latest version from mattes table we apply the rest of migrations we need.
Basically, all Fn versions right now have migration version 7, so we wouldn't need to run any migrations but we need to merge mattes table to goose table (shift and lift).

So, based on our bad decisions, i would choose the following:
1. Get rid of mattes/migrate forever.
2. Merge mattes/migrate to goose table.
3. Apply the rest of migrations we need starting from the latest version (which is 7).
4. We're done, users are happy.

rdallman:
it seems phrasing was contentious, does the latest commit accomplish this?

* fresh db -> skip all migrations, create tables from schema (, set version in goose?)
* db up to date with old migrations -> run any new migrations (optional: drop old migrate table)
* db behind on old migrations -> run old migrations, run new migrations (optional: drop old migrate table)
* db up to date on new migrations -> do nothing

the only part of the above that seemed to be contended was skipping migrations on fresh boot. maintaining the schema is entirely possible. is setting the version possible? are there religious reasons that should be tax exempt for us to not set the version? are there any benefits to us maintaining a half baked schema and trying to work out the current db schema by inspecting migrations? are there any benefits to subjecting users to the possibility of migrations blowing up on their first boot and having to eat the latency of running them?

rdallman:
> Apply the rest of migrations we need starting from the latest version (which is 7).

this one is the one that sticks out to me as what was weird. we need to apply the old migrations, but only any old migrations we have not run. in no scenario do we want to run migrations that have already been run, as was the case at least earlier in this patch.

awaiting some verification of assumptions before sinking time reviewing

denismakogon:
> fresh db -> skip all migrations, create tables from schema (, set version in goose?)

The code would never do that, because it’s a mistake we did before, and I disagree to do this again, table is a migration in terms of any migrations framework.

Despite 1st statement, everything else is works exactly as described.

The code runs only migrations that haven’t applied yet with respect to the old migrations table, we wouldn’t run any migration twice, especially while syncing old migrations table to a new one. That’s why I decided to move records from the the old table to a new one and run the rest of migrations starting from the latest version found in old migration table, it looks more clean and doesn’t look inconsistent to my initial though to ignore dup. column errors.

rdallman:
> The code would never do that, because it’s a mistake we did before, and I disagree to do this again, table creation is a migration in terms of any migrations framework.

there's still no basis in this being a mistake other than 1 person's opinion. I can find numerous counter arguments. please, indulge me.

rdallman:
as long as we have schemas somewhere to reference, whatever, we can keep all the bad parts of not using them so that the internet can have one less battle over useless shit.

denismakogon:
> there's still no basis for this being a mistake other than 1 person's opinion. I can find numerous counter-arguments. please, indulge me.

I see where this discussion goes and don't see any value in continuing. If you see the benefits of manually updating schemas with each new migrations - it up to you. I still will have my own thoughts.

However, the latest commit brings complete schemas back and does was you've asked for - fresh DB -skip all migrations, create tables from the schema and set the version in the goose.

rdallman:
maybe, there's 40 some odd people working on this now, we have no shortage of opinions (it is easy to get a 3rd or 13th opinion on anything when needed). in general best to justify based on concrete correctness / widely agreed upon best practices.

thanks, the schemas will be critical when we have 70 or 700 migrations and not just 7. there are a couple ways to go about maintaining these, using them for fresh creation and keeping them in Go is one. maintaining separate schema files is another thing we could do and that would allow us to keep the original CREATE TABLE statements as was attempted here.

api/datastore/sql/migrations/1_add_route_created_at.go
```diff
@@ -0,0 +1,27 @@
+package migrations
+
+import (
+	"database/sql"
+	"github.com/pressly/goose"
+)
+
+// we skip the error here to make previous datastore tables work fine
```
rdallman: this comment line is in every one of these migration files, as well as the README here. a simple sed should get rid of them all.

api/datastore/sql/migrations/old.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func checkOldMigration(ctx context.Context, db *sqlx.DB) (int64, error) {
+	log := common.Logger(ctx)
+	current, dirty, err := checkOldMigrationTableVersionIfExists(db)
+	if err != nil {
+		return -1, err
+	}
+	if dirty {
+		log.Fatal("database corrupted")
```
rdallman: "database corrupted, please return the database to a stable state under this migration version (run the down migration at the version indicated?) in table `schema_migrations` and unset the dirty bit in that table to restore the database to a working condition"

"database corrupted" is pretty jarring, and this _should_ be recoverable. it just means a migration was running and the db exited, should just have to run down migration and then unset the dirty bit.

api/datastore/sql/migrations/migrations.go
```diff
@@ -1,534 +1,136 @@
...
-	"io"
-	"io/ioutil"
-	"os"
-	"path/filepath"
-	"strings"
-	"time"
+	"github.com/fnproject/fn/api/common"
+	"github.com/jmoiron/sqlx"
+	"github.com/pressly/goose"
+	"sort"
```
rdallman: run `goimports` on this file please

api/datastore/sql/migrations/migrations.go
```diff
@@ -1,534 +1,136 @@
...
-// could not be loaded.
-func AssetInfo(name string) (os.FileInfo, error) {
-	cannonicalName := strings.Replace(name, "\\", "/", -1)
-	if f, ok := _bindata[cannonicalName]; ok {
-		a, err := f()
+		current, err := migrations.Current(currentVersion)
 		if err != nil {
-			return nil, fmt.Errorf("AssetInfo %s can't read by error: %v", name, err)
+			fmt.Printf("goose: no migrations to run. current version: %d\n", currentVersion)
+			return nil
```
rdallman: should we check for a specific error here? should we return this error? unclear if `return nil` is in err

api/datastore/sql/migrations/old.go
```diff
@@ -0,0 +1,112 @@
+package migrations
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"github.com/fnproject/fn/api/common"
+	"github.com/jmoiron/sqlx"
+	"github.com/pressly/goose"
```
rdallman: `goimports` again please (all files not under `vendor/` probably)

api/datastore/sql/migrations/old.go
```diff
@@ -0,0 +1,112 @@
...
+					return err
+				}
+				log.Debug("inserting new goose version: ", version)
+			}
+			return nil
+		})
+		return goose.Migrations{}, err
+	}
+
+	// in case of a new datastore this will return a new slice [0:], with all the migrations we have
```
rdallman: not sure if this one is still true, the previous block for a new datastore returns a slice with length 0

api/datastore/sql/migrations/old.go
```diff
@@ -0,0 +1,112 @@
...
+
+	// will return the latest version
+	migrateCurrent, err := checkOldMigration(ctx, db)
+	if err != nil {
+		return nil, err
+	}
+	log.Debug("current mattes migrations version: ", migrateCurrent)
+
+	sortedMigrations := sortAndConnectMigrations(migrations)
+	newMigrationsQuery := fmt.Sprintf("INSERT INTO goose_db_version (version_id, is_applied) VALUES (?, ?);")
```
rdallman: postgres uses `$` syntax and oracle db will use `:named` syntax for these queries

rdallman:
the code now appears to handle all possible cases and have schemas, thank you. I have some concerns:

* goose doesn't support oracle db, we'll have to add this
* goose doesn't use any LOCK / UNLOCK behavior, we will have to add this to all of our migrations ourselves since we will be running multiple FN nodes against the same db
* we are muddling around in goose's tables (as was pointed out many times), which appears to work ok, but feels wrong. this is slightly brittle as we have to handle the various dialects ourselves since we aren't using goose.
* tested against pg? mysql? sqlite3? all 4 highlighted cases above, ideally (this sucks, sorry, been here). our automated tests should test fresh db and full down/up for pg & mysql but not 'stuck on old migration, move to latest'. the queries in the code don't seem like they would work for pg, so am not sure how this got through CI. 

I have a question. I am thinking about writing a migration tool that just uses `sqlx.Tx` that does proper locking, which would cover the first 3 of the above cases since we can basically do whatever we want API-wise (cater to our needs). as a bonus, we could own development of this and wouldn't have to deal with the trials of getting patches upstream. It seems like go is lacking a good, maintained migrate library, if not for us it seems widely useful. I think it will take 2-3 days to have something reasonably well tested and usable (accounting for the amount of time I spend staring at github lately). It is then up to us if we want to maintain it as a separate OSS project, but at least it seems like we need one that fits our needs and none exist off the shelf that work :( what do you think?

rdallman:
https://github.com/github/gh-ost looks really interesting for more intense migrations. none of ours currently are very challenging, and I guess it'd be ideal to never need something like this, but it's interesting none-the-less while we're on the topic. definitely not looking to go that route for `sqlx.Tx` migrator, want to keep it pretty simple.

api/datastore/sql/migrations/old.go
denismakogon: all these statements going to rebind, so it works fine for all our datastores.

api/datastore/sql/migrations/migrations.go
denismakogon: fixed

api/datastore/sql/migrations/1_add_route_created_at.go
denismakogon: fixed

api/datastore/sql/migrations/old.go
denismakogon: fixed

api/datastore/sql/migrations/migrations.go
denismakogon: fixex

denismakogon:
> tested against pg? mysql? sqlite3?

Tested on all supported datastores, all possible cases.

> goose doesn't support oracle db, we'll have to add this

meh, we don't know if Oracle DB works fine with the Fn.

> goose doesn't use any LOCK / UNLOCK behavior, we will have to add this to all of our migrations ourselves since we will be running multiple FN nodes against the same db

yeah, would be nice to have something in-between of mattes/migrate and goose.

> I have a question. I am thinking about writing a migration tool that just uses sqlx.Tx that does proper locking, which would cover the first 3 of the above cases since we can basically do whatever we want API-wise (cater to our needs). as a bonus, we could own development of this and wouldn't have to deal with the trials of getting patches upstream

maybe if mattes/migrate can support migrations like we want then that would be awesome, so, I don't think that starting from the very beginning is really the path we would like to go, but improve mattes/migrate or its fork to do the job we need.

rdallman:
> Tested on all supported datastores, all possible cases.

thanks.

> meh, we don't know if Oracle DB works fine with the Fn.

allegedly, it works with sqlx. to be as opaque as possible, this is going to be a requirement shortly.

> maybe if mattes/migrate can support migrations like we want then that would be awesome, so, I don't think that starting from the very beginning is really the path we would like to go, but improve mattes/migrate or its fork to do the job we need.

seems wise. sqlx itself have dug around in mattes and it will require a good bit of munging the code around to get it to work: mattes does not expose tx through its interface, and mattes supports a _ton_ of different backends that aren't sql, which we aren't using. have a feeling it will be faster to work from 0 with knowledge of how mattes/goose work now, and can completely toss out the part about a CLI to generate and run migrations. may give it a go and see how far I get in a few hours, thanks for feedback.

the LOCK thing is a blocker imo for goose, and seeing how painful this was, would hate to migrate migrators _again_

denismakogon:
> may give it a go and see how far I get in a few hours, 
even if mattes don't expose tx we can make a new API (similar to `Run(m io.Reader)` from mattes) it accept a list of functions that accept the Tx and then get the Tx from the inside of mattes somehow.

denismakogon:
closing in favour of #825

PR #788, opened 2018-02-22
fix up response headers

### LINKED ISSUE #782: JSON proto ignores top-level `content_type` on return doc
In a JSON function the output event includes a content-type, but this is not sent to the HTTP response of fn 

E.g: 

```
{"body":"OK","content_type":"text/plain","protocol":{"status_code":200,"headers":{}}}
```

Fn returns: 
```< HTTP/1.1 200 OK
< Content-Length: 2
< Content-Type: application/json
< Fn_call_id: 01C6Q7YY8B47WMT00000000000
< Xxx-Fxlb-Wait: 551.4783ms
< Date: Mon, 19 Feb 2018 14:37:26 GMT
<
* Connection #0 to host localhost left intact
OK%```
```

Expect http content type to reflect the message content type  

would need to resolve conflicting values from headers somewhere.

rdallman:
* stops defaulting to application/json. this was something awful, go stdlib has
a func to detect content type. sadly, it doesn't contain json, but we can do a
pretty good job by checking for an opening '{'... there are other fish in the
sea, and now we handle them nicely instead of saying it's a json [when it's
not]. a test confirms this, there should be no breakage for any routes
returning a json blob that were relying on us defaulting to this format
(granted that they start with a '{').
* buffers output now to a buffer for all protocol types (default is no longer
left out in the cold). use a little response writer so that we can still let
users write headers from their functions. this is useful for content type
detection instead of having to do it in multiple places.
* plumbs the little content type bit into fn-test-util just so we can test it,
we don't want to put this in the fdk since it's redundant.

I am totally in favor of getting rid of content type from the top level json
blurb. it's redundant, at best, and can have confusing behaviors if a user
uses both the headers and the content_type field (we override with the latter,
now). it's client protocol specific to http to a certain degree, other
protocols may use this concept but have their own way to set it (like http
does in headers..). I realize that it mostly exists because it's somewhat gross
to have to index a list from the headers in certain languages more than
others, but with the ^ behavior, is it really worth it?

closes #782


rdallman:
@skinowski i know there's work to do the buffering in #786 -- I am happy to line up behind that, I do not know if this makes your life easier in 786 or not (sorry if so).

treeder:
Why wouldn't content-type be top level?  Doesn't matter what protocol it is, that's a must have.

rdallman:
> Why wouldn't content-type be top level? Doesn't matter what protocol it is, that's a must have.

each protocol that has this concept seems likely to have their own concept of how to expose it. maybe it has a capital C and a capital L, maybe there's an underscore in between the two, maybe it's a dash, maybe it's an ascii shrug.

rdallman:
grpc for example, only has `application/grpc`, so contentType is a little bit out of place.

rdallman:
but the fact that it did not even work until this patch means that it's probably pretty safe to remove it, from output at least. the input one we could leave since it appears at least the ruby fdk is using it, but but symmetry... it is duplicated in headers on input, as well.

treeder:
What do you mean it didn't work?  It's fed into the function, that's all the "work" it has to do.

rdallman:
> What do you mean it didn't work? It's fed into the function, that's all the "work" it has to do.

it didn't work: https://github.com/fnproject/fn/issues/782 -- this patch also is the first to hook it up... it didn't work.

treeder:
Oh sorry, was thinking about the content_type for input (which did work). Although content_type is still required for output. Anyways. LGTM.

treeder:
@rdallman want to fix those conflicts then merge?

skinowski:
@rdallman we can merge this, I'll rework the I/O limit changes on top of this. I need to work/massage those changes anyway.

api/server/runner.go
```diff
@@ -102,16 +114,46 @@ func (s *Server) serve(c *gin.Context, appName, path string) error {
...
+		c.Writer.WriteHeader(writer.status)
+	}
+	io.Copy(c.Writer, &writer)
 
 	return nil
 }
+
+var jsonPrefix = []byte("{")
+
+var _ http.ResponseWriter = new(syncResponseWriter)
```
skinowski: do we need this line?

api/server/runner_test.go
```diff
@@ -138,15 +138,15 @@ func TestRouteRunnerExecution(t *testing.T) {
...
-			{Path: "/myerror", AppName: "myapp", Image: "fnproject/fn-test-utils", Type: "sync", Memory: 128, Timeout: 30, IdleTimeout: 30, Headers: map[string][]string{"X-Function": {"Test"}}},
-			{Path: "/mydne", AppName: "myapp", Image: "fnproject/imagethatdoesnotexist", Type: "sync", Memory: 128, Timeout: 30, IdleTimeout: 30},
-			{Path: "/mydnehot", AppName: "myapp", Image: "fnproject/imagethatdoesnotexist", Type: "sync", Format: "http", Memory: 128, Timeout: 30, IdleTimeout: 30},
-			{Path: "/mydneregistry", AppName: "myapp", Image: "localhost:5000/fnproject/imagethatdoesnotexist", Type: "sync", Format: "http", Memory: 128, Timeout: 30, IdleTimeout: 30},
-			{Path: "/myoom", AppName: "myapp", Image: "fnproject/fn-test-utils", Type: "sync", Memory: 8, Timeout: 30, IdleTimeout: 30},
+			{Path: "/", AppName: "myapp", Image: "fnproject/fn-test-utils", Type: "sync", Memory: 128, Timeout: 3, IdleTimeout: 30, Headers: map[string][]string{"X-Function": {"Test"}}},
+			{Path: "/myhot", AppName: "myapp", Image: "fnproject/fn-test-utils", Type: "sync", Format: "http", Memory: 128, Timeout: 30, IdleTimeout: 3, Headers: map[string][]string{"X-Function": {"Test"}}},
```
skinowski: I wouldn't change idle timeout this much. Two tests depend on hot container recovery in I/O, so we need to make sure it's alive.

api/server/runner.go
rdallman: this is something of an idiom so that the compiler will generate a nice message that says we don't fulfill that interface directly here instead of generating error messages from the call sites where it is used. as a plus it nicely documents its use

api/server/runner_test.go
rdallman: i'll change these back. i was having test issues and that helped the feedback loop a lot, thanks

rdallman:
thanks @skinowski - updated.

rdallman:
nobody caught this one:

> (default is no longer left out in the cold)

boo

skinowski:
hhmm, why? did you mean we missed that? But resp writer thing in call/runner gets all protocols no?

rdallman:
missed, as in missed a joke :(

skinowski:
It's a good one too. :-)

PR #792, opened 2018-02-26
 Moved grafana from fn/examples to tutorials repo

sachin-pikle:


examples/grafana/prometheus.yml
```diff
@@ -1,20 +0,0 @@
-global:
```
denismakogon: [compose depends on this one](https://github.com/fnproject/fn/blob/master/docker-compose.yml#L69), so i do like to keep this in place

sachin-pikle: Does the compose file depend upon just prometheus.yml or even the dashboard, etc.? And just a wider question - is the compose file like a core Fn feature? If yes, why would it depend on an "example" prometheus file? Also, is the prometheus file a core feature? If yes why is it under examples? Trying to understand this better. Maybe we should have a separate 1:1 chat. And a related question is what if we duplicate the Prometheus file under tutorials what does that mean? Does it mean changing the example prometheus file also means we need to duplicate the change under tutorials? Or are they two separate independent files that can change independently?

rdallman: >  is the compose file like a core Fn feature?

it has seeped in, we need to move it as the intention wasn't to officially support it (whereas the helm chart the intention is to officially support it)

denismakogon: > Does the compose file depend upon just prometheus.yml or even the dashboard, etc.? 

Only yml file. Dashboard doesn't need that.

> is the compose file like a core Fn feature
It used as quick Fn deployment tool for development and local testing.

> if yes, why would it depend on an "example" prometheus file?

Because example file was written that was to reuse example file, because compose delivers some sort of sandbox deployment, that is why it relays on the example file because it works as is with docker compose.

> Also, is the prometheus file a core feature?

Fn doesn't have any "core" vs "non-core" definitions. So, prometheus for the Fn is a storage fo the OpenTracing metrics as well as zipkin AFAIK.

> Does it mean changing the example prometheus file also means we need to duplicate the change in tutorials?

the adjustments should be reflected only in one place, the one that is prometheus config example config for compose.

> Or are they two separate independent files that can change independently?

Yes, but you have to be sure that compose still works as expected, we do not gate that, so have to ensure manually.

sachin-pikle: From what you are saying Prometheus.yml file "belongs" to fn repo. There should be only one copy of this file. Today it is used by compose for standing up the local dev environment with prometheus monitoring enabled. Tomorrow it may be used by Helm (or other similar tools) to stand up prod clusters with prometheus monitoring enabled.

And hence the tutorial should simply refer to this file instead of duplicating it in the tutorials folder, correct?

skinowski:
Dashboards could also be referenced from the tutorials. We might want to keep them here and update them in fn repo since they are tied to stats. And I want to have much fancier dashboards going forward. But either way is fine.

sachin-pikle:
Ok. Thank you for the review comments. The more I think about your comments and the way it should work is the prometheus yaml, and the dashboard json files "belong" in the fn project. They are an integral part of fn's monitoring story i.e. batteries included but replaceable (or extendable). Anyone compose or helm or tutorials who needs an integrated set up of Fn + Prometheus + Grafana (with our 3 dashboards) should refer/point to these three files.

sachin-pikle:
Closing this Pull Request. Will open a new one if needed.

PR #797, opened 2018-02-27
fn: introduce agent config and minor ghostreader tweak

skinowski:
TODO: move all constants/tweaks in agent to agent config.

### DIFF
--- a/api/agent/config.go
+++ b/api/agent/config.go
@@ -9,9 +9,9 @@ import (
 )
 
 type AgentConfig struct {
-	MinDockerVersion string        `json:"minDockerVersion"`
-	FreezeIdleMsecs  time.Duration `json:"freezeIdleMsecs"`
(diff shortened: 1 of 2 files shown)


api/agent/config.go
```diff
@@ -0,0 +1,60 @@
...
+	"math"
+	"os"
+	"strconv"
+	"time"
+)
+
+type AgentConfig struct {
+	MinDockerVersion string        `json:"minDockerVersion"`
+	FreezeIdleMsecs  time.Duration `json:"freezeIdleMsecs"`
+	EjectIdleMsecs   time.Duration `json:"ejectIdleMsecs"`
```
rdallman: while we're here, by convention these should be `min_docker_version`, etc. -- if we ever have it so simple as being able to use a json config

rdallman: lost specificity: `json:"min_docker_version"` - the go fields themselves being camel cased is appropriate

PR #799, opened 2018-02-27
fn: remove cold implementation

skinowski:
Cold requests/containers now become execute-once
hot containers. There's no visible change to users
and/or operators.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -3,8 +3,8 @@ package agent
 import (
 	"context"
 	"io"
+	"io/ioutil"
 	"net/http"
-	"strings"
 	"sync"
 	"time"
 
@@ -15,7 +15,6 @@ import (
 	"github.com/fnproject/fn/api/id"
 	"github.com/fnproject/fn/api/models"
 	"github.com/fnproject/fn/fnext"
-	"github.com/go-openapi/strfmt"
 	"github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sirupsen/logrus"
@@ -30,8 +29,6 @@ import (
 // to be much more robust. now we're at least running it if we delete the msg,
 // but we may never store info about that execution so still broked (if fn
 // dies). need coordination w/ db.
-// TODO if a cold call times out but container is created but hasn't replied, could
-// end up that the client doesn't get a reply until long after the timeout (b/c of container removal, async it?)
 // TODO if async would store requests (or interchange format) it would be slick, but
 // if we're going to store full calls in db maybe we should only queue pointers to ids?
 // TODO examine cases where hot can't start a container and the user would never see an error
@@ -182,23 +179,11 @@ func (a *agent) Submit(callI Call) error {
 }
 
 func (a *agent) startStateTrackers(ctx context.Context, call *call) {
-
-	if !protocol.IsStreamable(protocol.Protocol(call.Format)) {
-		// For cold containers, we track the container state in call
-		call.containerState = NewContainerState()
-	}
-
 	call.requestState = NewRequestState()
 }
 
 func (a *agent) endStateTrackers(ctx context.Context, call *call) {
-
 	call.requestState.UpdateState(ctx, RequestStateDone, call.slots)
-
-	// For cold containers, we are done with the container.
-	if call.containerState != nil {
-		call.containerState.UpdateState(ctx, ContainerStateDone, call.slots)
-	}
 }
 
 func (a *agent) submit(ctx context.Context, call *call) error {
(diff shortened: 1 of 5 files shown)


api/agent/agent.go
```diff
@@ -556,7 +460,29 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
...
+			go func() {
+				io.CopyN(ioutil.Discard, stdoutRead, 100*1024*1024)
+				close(done)
+			}()
+
+			// let drain, 50 msecs or 100MB, whichever comes first
+			select {
+			case <-done:
+			case <-time.After(time.Duration(50) * time.Millisecond):
+			}
```
rdallman: we seem to have hit 3 cases where it makes a lot more sense to just shut the container down than try to recover. i give up, let's do that. killing off the ghost thing might be nice, but it means we end up with a weird stderr pipe too, anyway, an option

skinowski: yup, I agree... (Also I know this is not going to interact nice with I'm going to merge in the next few days: I/O size clamping)

api/agent/agent.go
```diff
@@ -795,12 +703,15 @@ func (a *agent) runHotReq(ctx context.Context, call *call, state ContainerState,
...
-	timeout time.Duration // cold only (superfluous, but in case)
+	id     string // contrived
+	image  string
+	env    map[string]string
+	memory uint64
+	cpus   uint64
+
+	// limit total requests that can be run on this container
+	// default: zero - unlimited
+	maxRequests uint64
```
rdallman: conceptually, this doesn't seem like a 'container' construct wrt the docker api itself (this `*container`) as the `fsouza.DockerContainer` doesn't have any idea of this. can we push it up into `runHot` and build from the call state (it seems like a 'route' construct potentially, which we could hang off the call)

rdallman:
without re-plumbing the context for launch hot i see some nascient issues:

* containers will go launch when convenient even after the request that spawned them is long gone on a loaded system, we can bail outta this way before (the hot launcher loop will be going)
* running a bunch of calls and doing async removal is going to backlog docker for removals, this just means we end up queued in a different place than before, as long as we hold the resource token over docker removal we won't over provision the system at the same rate as now.

it seems like a slightly better cut point may be to skip the slot queue? but i'm not sure of issues there (intuitively, seems ok). then we at least get the same execution path, we just lop off the launcher.

skinowski:
@rdallman  plumbed up the slot/orig contexts now. It looks a bit messy in there though.

skinowski:
Friendly ping/reminder. :-)

api/agent/agent.go
```diff
@@ -563,143 +483,133 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	case err := <-s.errC: // error from container
 		return err
 	case err := <-errApp: // from dispatch
+		if err != nil && models.IsAPIError(err) && s.fatalErr != nil {
+			// TODO: narrow this down to StatusBadGateway, we do not want to accidentally
+			// shutdown container due to client side errors. Alternatively Dispatch()
+			// should tell us if the error is from the container or not.
+			// TODO: we can skip this for Cold containers, but let's not optimize Cold path
+			s.fatalErr = err
+		}
```
rdallman: this ends up being a pretty big behavior change, didn't see a note about it so just making sure this is intended

skinowski: We briefly discussed this, we failed to parse the protocol. Before this commit, the logic here tried to recover by draining the pipes. (50 msec or 100MB read) But we said, this is too difficult and not certain to fix the pipes and adds latency. It's best to kill the container when we go into such inconsistent state. (But I might have misunderstood the comm.)

skinowski: commit, I am referring to previous commit in this PR.

rdallman: when I made the comment, it seemed to also encompass timeout / cancel and 'oversize output' but neither of those cases are handled here (the latter not being upstream, the former not a 'models.APIError') -- though we probably should handle them similarly, this is ok

skinowski: so, ok to shutdown in case of protocol errors, right? I think dispatch() probably should return two different errors, one for errors that we want to log or attempt to relay to client, etc. The other for errors related to container, which warrants a shutdown. The distinction is not that clear here. But regardless, I expect some post-cleanup work (probably outside of scope of this PR)

skinowski:
Closing this for now, as the result ended up being equally complicated and also the main line has evolved significantly.

One of the biggest gains out of this was the side effect of making all cold tests go through hot to increase coverage, but this is also debatable with code littered with 'if-cold' branches.

We can re-attempt this going forward, but I'm closing this for now.

Issue #796, opened 2018-02-27
Flaky test: TestMultiLog


rdallman:
```
--- FAIL: TestMultiLog (6.15s)
	exec_test.go:55: Sync execution call ID: 01C7C6BA67T7WG600000000000
	exec_test.go:293: Log entry must contain `First line` string, but got: Second line
```

this is in the api tests i think and is habitually a flaky test offender for one reason or another. aside from not testing tricky behaviors in the api tests at all, we should fix this situation. the api tests here simply can check that the log endpoint returns a log, that's all we need to do, and there are multiple tests for that.

skinowski:
Seems like 3 second sleep makes it easier to trigger this.. New tests in #801 with 1 sec sleep pass.

skinowski:
Note: #801 does not fix the issue.

gviedma:
Note: possibly related to new flaky test issue #822

gviedma:
I have confirmed this is still an issue after #809 
```
--- FAIL: TestRouteRunnerExecution (22.07s)
	runner_test.go:496: Test 14: GetLog read IO call_id:01C82PNVZP27WJM00000000000 cannot find: EndOfLogs in logs: 2018/03/08 11:41:57 BeginOfLogs
		2018/03/08 11:41:57 Received format http
		2018/03/08 11:41:57 Received request main.AppRequest{SleepTime:1000, ResponseCode:0, ResponseContentType:"", JasonContentType:"", EchoContent:"", IsDebug:true, IsCrash:false, ReadFile:"", ReadFileSize:0, CreateFile:"", CreateFileSize:0, AllocateMemory:0, LeakMemory:0, TrailerRepeat:0, InvalidResponse:false, PostSleepTime:0, PostOutGarbage:"", PostErrGarbage:""}
		2018/03/08 11:41:57 Received headers map[User-Agent:[Go-http-client/1.1] Content-Length:[36] Fn_call_id:[01C82PNVZP27WJM00000000000] Fn_deadline:[2018-03-08T11:42:27.110Z] Fn_method:[GET] Fn_request_url:[http://127.0.0.1:8080/r/myapp/myhot]]
[...]

PR #801, opened 2018-02-28
fn: clean api tests: removed multi log

skinowski:
*) fn-test-utils covers multi-log case, with sleep in between.
*) added trailer data to fn-test-utils (going to be useful for large response tests)
*) added logging to panic/error cases in fn-test-utils.


skinowski:
Fixes: #796

rdallman:
a quick glance at this has me curious whether our turning off of the docker logs and using async attach is the race, and not a poorly built multi log image (though i look forward to removing this image, i can't imagine it's doing anything that crazy). to be clear: want to move forward with this, and thanks for doing it; just got me thinking about the root of the issue since when it happens it seems to skip the first line and it seems pursuant to a log race.

skinowski:
Another reason why I really want to remove the cold-impl... Hot has hostWriter in stderr, but cold doesn't.. The test case only tests cold...

rdallman:
i think the race exists but is more far-fetched for hot. if attach takes longer than start + hand out of slot, then we can miss logs. attach is done in parallel to that, so it could happen. previously, we streamed all logs up to attach time, which prevented this bug. it seems like we just need to add synchronous attachment to the container and add our own goroutine that we can manage so that we can properly manage the logs (it's very much worth turning off the json logs).

rdallman:
in any event, we can do that separately (can file another issue)

skinowski:
Raising it to 3 secs triggered it now, but only for hot, cold passed. Arrghh. I'll merge this and we can work on the runner test instead of api tests.

skinowski:
Nevermind, it's off/on, seems random...

rdallman:
yea, i think it's just racy, multi log test waffled back and forth

skinowski:
I'll take a closer look tomorrow at this...

PR #802, opened 2018-02-28
opentracing -> opencensus

### LINKED ISSUE #695: research: http://opencensus.io/
this looks like a promising project (from Google) largely based on opentracing's foundations, which has additional interfaces for normal things like gauges and counters, etc. at present we have our own little special way of doing these kinds of metrics in opentracing, but it seems like opencensus has all the sugar that opentracing has and more (clients for every lang, zipkin/jaeger support, standard interface with pluggable backends). out of the box they support a prometheus backend and a zipkin backend for the go one: https://github.com/census-instrumentation/opencensus-go

i think it would be pretty easy to swap out for with opentracing, from the looks of it. it seems like this would solve some of our headaches munging data from opentracing/openzipkin into prometheus.

rdallman:
closes #695 

this is gonna be long, grab a beverage. i'll try to keep it digestible.

## why? 

https://opencensus.io/ is a spiritual successor to opentracing, the interface we are currently using to extract span and stats telemetry in our code. opentracing and opencensus both provide exportation of span data to jaeger/zipkin, the differentiator being that opencensus defines its own wire and code formats for these, which is beneficial as one sticking point we had using opentracing was that we were tied to the zipkin's backend span format and it wasn't easy to switch to jaeger due to this. the other benefit of opencensus is that it provides open telemetry interfaces in the same vein as the tracing api, but for gauges / histograms / counters -- the benefit here being we do not have our own proprietary telemetry code API, with our own limited set of exporters to various backends. since the tides change frequently for telemetry backends, it will be nice to have opencensus for not only us but anybody building extensions with us can simply use the opencensus apis and exporters instead of us having to build / maintain / use our own -- the dream obviously being that another user already using opencensus to instrument their code can seamlessly build fn against their code (say, some extension) and get stats/traces out in a familiar way.

conceptually, the status quo for us was to toss metrics into tracing log events, using different key names to turn them into the correct stat type (gauge, etc.), and then collect these in span finish. while this works, and let us use open apis (the dream), there were performance issues with doing this and it wasn't perfectly hygienic; for one, we had to sample at 100%, create a new span per non-trace metric, and our mechanisms for scraping these were something of a bottleneck performance wise (I get a 20% speedup with this, even with 100% sampling). nonetheless, it was a great first effort and allowed us to get to this point with our telemetry stuff, markedly better than we were beforehand. code wise for the traces themselves, the api is strikingly similar, whereas for the statistics we had already begun to define our own set of stats apis and now we can use the more expressive opencensus code api.

tiny note, it is early for opencensus, but we can vendor, it's backed by google with some pretty big horsepower behind it people-wise, and it seems wise to pick a stats interface as early as possible, as changing these is historically not very fun.

## what works

* https://github.com/fnproject/fn/tree/05ec91e31033c791a0c34541c5e4bb8b0ae04d08/examples/grafana -- dashboard 1 & 3 [_should_] still work
* `/metrics` still works for prometheus, this includes app name and path. in general, metrics from dashboard 1&3 above are all the same and untouched (in `/metrics`)
* UI still [appears to] work, i gave this a brief test. it uses `/metrics` for metrics in dash 1 mostly
* zipkin spans still work, same configuration. currently samples at 100% if on, i have a TODO to add more config for this, but this patch is rather noisy as it is.
* hybrid servers still send spans back and forth (continuations), and we're still starting a span based on the sampling ratio from the root http handler (opencensus made this easy with their handler)

## what broke

* `agent_submit_global`, `agent_submit_app`, `agent_submit` and every stat that was under `fn_span_` is now elided from `/metrics`, i.e. all metrics in dashboard 2 from https://github.com/fnproject/fn/tree/05ec91e31033c791a0c34541c5e4bb8b0ae04d08/examples/grafana. these are still available to tracers (zipkin, et al). it seems to me that this wasn't advisable to do from the get go, as reluctantly now we have people creating dashboards that rely on these very internal and very whimsical metrics that may come and go at any time. zipkin and jaeger are much less rigid for their telemetry viewing, they basically populate whatever is shoved in there and we don't have to declare any fields up front at all (woo). if we do want some of these that we officially support (as tears roll down my cheek), we can go in and add opencensus histogram measurements on the spot; it doesn't seem wise for us to make a prometheus tracing exporter and export all spans, as we'd want to sample the tracer at 100% and we don't really want to do that for zipkin (for performance, mostly, but also we don't need that data -- we just need a small sample size and any forced traces on the spot with a really low retention). I'm open to ideas here, i spent a lot of time scratching my chin around this...
* the lb will need to switch to opencensus in order for us to continue its spans (relatively easy, we didn't have all the prom stuff over there)
* if i missed something (i.e. you don't see it in 'what works' nor 'what broke') please notify, that's everything I can think of.

## miscellaneous deltas

* removes opentracing / zipkin-go-opentracing and associated imports (vast) & adds opencensus
* moves setup of prometheus http handler to server (from agent, plumbing bugaboo)
* no longer do we have 'inconsistent label cardinality' in prometheus stats, took some scrubbing here; i'm a little angered by the whole situation but overall it enforces us to do the right thing I think. this does mean that most every statistic has app name & path and to get global stats requires `*` in grafana.
* no longer relying on `openzipkin.Trace` structs, which means we can trivially try out jaeger (there's some artificial excitement around this idk why, zipkin works fine really for our purposes)
* we no longer have 'baggage items' as they were deemed in opentracing, we now have 'tags' shoved onto each request's context, and we're also inserting them in the async agent for any there. this required extraction and re-insertion in some of the contexts to propagate correctly. this works well, sadly there's no way to get a context with all the values sans any deadlines/cancellations without doing it explicitly yourself, but the compromise isn't bad (considering that we have a ~unique use case).
* in general, this is a lot of code that was replaced simply using sed (all traces, imports) and is pretty straight-forward, most of the patch is deleting
* deletes `api/common/metrics.go`, `api/server/fntracer.go`, `api/server/prom_zip_collector.go` - big win that somebody else wrote the prom collector / stats interface for us now, woo hoo.

## future / discussion points

* it seems the intention of opencensus to have views and measurements more separate, spent some time trying to figure out where we should declare all of these, at present they are co-located. this works, but for some reason I think we can do better here (I was going to discuss with the opencensus team after we had something). would be grateful for ideas around this.
* we can change all the request state and container state stuff to grab the opencensus view and use the numbers from there, as right now we're maintaining our own state counters separately that are identical.
* am somewhat concerned that client-side aggregation, which we were already doing for prometheus exporter but is now done in opencensus itself is gonna muddle stats too much for them to be of any use. we can get traces for more in depth looks but still, eesh. the industry seems to be moving this direction though, so going with the flow.
* need to allow configuration of tracing sampling rate and have some header that will trigger a span to be sampled for that request (could not find the latter in their docs, but easy to do)
* add span links for our hot requests to link the request span with the hot container span (opencensus seems to support a 'multiple parent' model -- this will be awesome for debugging hot, things like freeze / start / etc -- couldn't figure out how to do it with opentracing).

this seems like an improvement over what we had for performance reasons (>20% speed up) as well as things like having an open telemetry api and not having our own prometheus collector, and would like to move forward here on that basis if we get some consensus on that -- we can further massage the measurement / view / tag stuff to our liking.


rdallman:
p.s. not sure if @nigeldeakin switched away yet, but if not, you the man, would love your thoughts here.

api/agent/drivers/docker/docker_client.go
```diff
@@ -176,22 +282,17 @@ func filterNoSuchContainer(ctx context.Context, err error) error {
...
-		info, err = d.docker.Info()
-		return err
-	})
-	return info, err
+	// NOTE: we're not very responsible and prometheus wasn't loved as a child, this
+	// threads through directly down to the docker call, skipping retires, so that we
+	// don't have to add tags / tracing / logger to the bare context handed to the one
+	// place this is called in initialization that has no context to report consistent
+	// stats like everything else in here. tl;dr this works, just don't use it for anything else.
+	return d.docker.Info()
```
skinowski: :+1:

api/agent/stats.go
```diff
@@ -2,58 +2,256 @@ package agent
...
-	common.IncrementCounter(ctx, serverBusyMetricName)
+func statsTooBusy(ctx context.Context) {
+	stats.Record(ctx, serverBusyMeasure.M(1))
 }
 
 const (
+	// TODO we should probably prefix these with calls_ ?
 	queuedMetricName     = "queued"
-	callsMetricName      = "calls"
+	callsMetricName      = "calls" // TODO this is a dupe of sum {complete,failed} ?
```
skinowski: no, it's more like complete + failed + server_busy. The notion that 'failed' does not include server_busy I think it confusing, but preserved this for some time for UI consistency... (AKA UI won't show server too busy errors in failed stats. I think it should.. But anyways.)

skinowski: Unrelated with this PR, but I think the stats.go functions (especially after recent code reductions) do not provide much value anymore. They even make the code slightly harder to follow. I kind of want to break these "two-metric" functions here into "operate-on-one-metric" only functions or just get rid of them...

api/common/metrics.go
```diff
@@ -1,140 +0,0 @@
-package common
```
skinowski: :clap:

api/agent/stats.go
rdallman: that makes sense. `calls` stills seems superfluous then

api/agent/agent.go
```diff
@@ -847,24 +821,56 @@ func (c *container) CPUs() uint64                   { return c.cpus }
...
 // WriteStat publishes each metric in the specified Stats structure as a histogram metric
 func (c *container) WriteStat(ctx context.Context, stat drivers.Stat) {
-
-	// Convert each metric value from uint64 to float64
-	// and, for backward compatibility reasons, prepend each metric name with "docker_stats_fn_"
-	// (if we don't care about compatibility then we can remove that)
-	var metrics = make(map[string]float64)
 	for key, value := range stat.Metrics {
-		metrics["docker_stats_fn_"+key] = float64(value)
+		stats.Record(ctx, stats.FindMeasure("docker_stats_"+key).(*stats.Int64Measure).M(int64(value)))
```
skinowski: We probably don't need 'docker_stats' prefix and any append here... But we can tweak later... The driver could simply return an array (and perhaps prefix them with 'container_'  or 'docker_stats' already) instead of this map too... Lots of unnecessary translations we do for stats. TODO & not important...

rdallman: dashboard number 3 in our examples relies on this format, was trying to minimize name breakages in this PR for digestibility's sake - we need to define / finalize probably every stat we're exporting and decide on them at some point.

api/agent/drivers/docker/docker_client.go
```diff
@@ -98,18 +100,123 @@ type dockerWrap struct {
...
+	{
+		dockerOOMMeasure, err = stats.Int64("docker_oom", "docker oom", "")
+		if err != nil {
+			logrus.Fatal(err)
+		}
+		v, err := view.New(
+			"docker_oom_count",
+			"number of docker container oom",
+			[]tag.Key{appKey, pathKey},
+			dockerErrorMeasure,
```
skinowski: dockerOOMMeasure? typo here... maybe move this repeated blocks to a func? we can improve later...

api/agent/drivers/docker/docker_client.go
```diff
@@ -119,11 +226,10 @@ func (d *dockerWrap) retry(ctx context.Context, logger logrus.FieldLogger, f fun
 		if common.IsTemporary(err) || isDocker50x(err) {
 			logger.WithError(err).Warn("docker temporary error, retrying")
 			b.Sleep(ctx)
-			span.LogFields(log.String("task", "tmperror.docker"))
```
skinowski: would be good to still track these? but ok...

api/agent/drivers/docker/docker.go
```diff
@@ -576,7 +577,7 @@ func (w *waitResult) wait(ctx context.Context) (status string, err error) {
 	case 0:
```
skinowski: Note to self... Should track abnormal container exit codes perhaps...

api/agent/drivers/docker/docker_client.go
```diff
@@ -262,13 +363,15 @@ func (d *dockerWrap) PullImage(opts docker.PullImageOptions, auth docker.AuthCon
...
-	defer span.Finish()
-	ctx := opentracing.ContextWithSpan(context.Background(), span)
+	ctx := common.BackgroundContext(opts.Context)
+	ctx, span := trace.StartSpan(ctx, "docker_remove_container")
+	defer span.End()
 
 	ctx, cancel := context.WithTimeout(ctx, retryTimeout)
 	defer cancel()
 
+	opts.Context = ctx
```
skinowski: none of the callers of this function set a timeout in agent.. Now, we start to impose this and removals can timeout.
I don't think we should set opts.Context with the new ctx.

I think we can revisit these retry timeouts in general in this package in the future. In some functions, ctx arg is used only for logging, in some it's used to enforce a timeout, sometimes retry timeout can hold up the request longer, could be using docker or dockeNoTimeout client, etc.

api/agent/drivers/docker/docker_client.go
rdallman: i think they are duplicates of the retries count sum. the terminal cases (non-temporary error, timeout) have explicit stats and any non-zero retries count would sum to the same as this previous `tmperror`. can add back tho?

api/agent/drivers/docker/docker_client.go
rdallman: thanks, this was a bug

api/logs/s3/s3.go
```diff
@@ -144,13 +146,13 @@ func (s *store) InsertLog(ctx context.Context, appName, callID string, callLog i
 		return fmt.Errorf("failed to write log, %v", err)
 	}
 
-	common.PublishHistogramToSpan(span, "s3_log_upload_size", float64(cr.count))
+	stats.Record(ctx, downloadSizeMeasure.M(int64(cr.count)))
```
skinowski: uploadSizeMeasure?

api/logs/s3/s3.go
```diff
@@ -169,6 +171,65 @@ func (s *store) GetLog(ctx context.Context, appName, callID string) (io.Reader,
...
+	{
+		downloadSizeMeasure, err = stats.Int64("s3_log_download_size", "downloaded log size", "byte")
+		if err != nil {
+			logrus.Fatal(err)
+		}
+		v, err := view.New(
+			"s3_log_download_size",
+			"downloaded log size",
+			[]tag.Key{appKey, pathKey},
+			uploadSizeMeasure,
```
skinowski: downloadSizeMeasure?

api/server/server.go
```diff
@@ -434,8 +425,9 @@ func (s *Server) bindHandlers(ctx context.Context) {
 
 	engine.GET("/", handlePing)
 	engine.GET("/version", handleVersion)
-	// TODO: move the following under v1
-	engine.GET("/metrics", s.handlePrometheusMetrics)
+
+	// TODO: move under v1 ?
+	engine.GET("/metrics", gin.WrapH(s.promExporter))
```
skinowski: s.promExporter could be nil, right?

api/agent/drivers/docker/docker_client.go
skinowski: I don't know, if docker starts to spits 50x, etc. it would be good to see, but we can tweak later too...

rdallman: i am positive that it's encompassed in the retries sum a little higher up (this was also a sum)

rdallman:
additional future stuff:

* figure out histogram thing, we probably want these but defining bounds for each is a pain. currently we can get average / count / sum which is OK (same deal on master, most histograms aren't very useful since bounds aren't well defined per stat)
* get the go proc / gc / cpu stats back into the prom graph, opencensus doesn't seem to have these

thank you, Tolga!

rdallman:
gonna await monday on big green button in case somebody gets overzealous with the prometheus tutorials over the weekend..

rdallman:
https://github.com/fnproject/tutorials/pull/37 since we now have grafana docs in 2 places

updated docs here. bandaid rip time *gulp*

PR #804, opened 2018-02-28
Remove replicated expvar handler

ezbercih:
expvar package exports Handler which can be directly used instead of copying the expvarHandler function.


ezbercih:
Closing this pull request to raise a new one to fix a typo in commit comment.

PR #805, opened 2018-02-28
Remove replicated expvar handler

ezbercih:
expvar package exports Handler which can be directly used instead of copying the expvarHandler function.


rdallman:
thanks, I don't think there's a lot of CLA conflict here since this is stdlib code, our CLA process sucks (sorry) so let's do it

ezbercih:
Thanks!

PR #806, opened 2018-03-01
fix undefined string slot key

rdallman:
while escape analysis didn't lie that the bytes underlying this string escaped
to the heap, the reference to them died and led to us getting an undefined
byte array underlying the string.

sadly, this makes 4 allocs here (still down from 31), but only adds 100ns per
op. I still don't get why 'buf' and 'byts' escape to the heap, blaming faulty
escape analysis code. we could reduce one of them i guess, the hash sum
though will still escape.

this one is kind of impossible to write a test for.

found this from doing benchmarking stuff and was getting weird behavior at the
end of runs where calls didn't find a slot, ran bisect on a known-good commit
from a couple weeks ago and found that it was this. voila. this could explain
the variance from the slack dude's benchmarks, too. anyway, confirmed that
this fixes the issue.

### DIFF
--- a/api/agent/slots.go
+++ b/api/agent/slots.go
@@ -326,16 +326,11 @@ func getSlotQueueKey(call *call) string {
 	}
 
 	var buf [sha1.Size]byte
-	byts := hash.Sum(buf[:0])
-	return unsafeString(byts)
+	hash.Sum(buf[:0])
+	return string(buf[:])
 }
 
 // WARN: this is read only
 func unsafeBytes(a string) []byte {
 	return *(*[]byte)(unsafe.Pointer(&a))
-}
-
-// WARN: this is read only
-func unsafeString(b []byte) string {
-	return *(*string)(unsafe.Pointer(&b))
 }


api/agent/slots.go
```diff
@@ -326,16 +326,11 @@ func getSlotQueueKey(call *call) string {
 	}
 
 	var buf [sha1.Size]byte
-	byts := hash.Sum(buf[:0])
-	return unsafeString(byts)
+	hash.Sum(buf[:0])
```
skinowski: shouldn't this be something like this?

```
return string(hash.Sum(nil))
```

rdallman: that is effectively the same, since go's escape analysis stinks. the contract of  `hash.Hash.Sum([]byte)` is that it will append the result to the provided byte slice, returning the resulting slice. in theory, providing your own perfectly sized byte slice should avoid an allocation, in practice it seems to simply allow one to re-use a byte slice to do repeated sums instead of allocating a new one each sum.

rdallman: example: https://play.golang.org/p/Ry4G5N2KDwP

skinowski: OK, got confused by sha1 docs, different function:
https://golang.org/pkg/crypto/sha1/#Sum

rdallman: we should prob use another hash function like siphash / xxhash but it's an extra import. those tend to provide a more allocation friendly api than `hash.Hash` and we don't need a cryptographic hash function for this... once we add versioning of routes / apps this dies tho.

rdallman:
thanks

PR #809, opened 2018-03-02
fn: I/O related improvements

skinowski:
*) I/O protocol parse issues should shutdown the container as the container
goes to inconsistent state between calls. (eg. next call may receive previous
calls left overs.)
*) Move ghost read/write code into io_utils in common.
*) Clean unused error from docker Wait()
*) We can catch one case in JSON, if there's remaining unparsed data in
decoder buffer, we can shut the container
*) stdout/stderr when container is not handling a request are now blocked if freezer is also enabled.
*) if a fatal err is set for slot, we do not requeue it and proceed to shutdown
*) added a test function for a few cases with freezer strict behavior

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -3,7 +3,6 @@ package agent
 import (
 	"context"
 	"io"
-	"io/ioutil"
 	"strings"
 	"sync"
 	"time"
@@ -787,9 +786,6 @@ func NewHotContainer(call *call) (*container, func()) {
 	stdin := common.NewGhostReader()
 	stderr := common.NewGhostWriter()
 	stdout := common.NewGhostWriter()
-
-	// any write between calls should be discarded
-	stdout.Swap(ioutil.Discard)
 
 	// direct stderr to log writer between calls
 	stderr.Swap(newLineWriter(&logWriter{
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -464,10 +464,8 @@ func (s *coldSlot) exec(ctx context.Context, call *call) error {
 		return err
 	}
 
-	res, err := waiter.Wait(ctx)
-	if err != nil {
-		return err
-	} else if res.Error() != nil {
+	res := waiter.Wait(ctx)
+	if res.Error() != nil {
 		// check for call error (oom/exit) and beam it up
 		return res.Error()
 	}
@@ -496,16 +494,20 @@ type hotSlot struct {
 	errC        <-chan error  // container error
 	container   *container    // TODO mask this
 	maxRespSize uint64        // TODO boo.
-	err         error
+	shutFun     func()        // shutdown container
+	fatalErr    error
 }
 
 func (s *hotSlot) Close(ctx context.Context) error {
+	if s.fatalErr != nil && s.shutFun != nil {
+		s.shutFun()
+	}
 	close(s.done)
 	return nil
 }
 
 func (s *hotSlot) Error() error {
-	return s.err
+	return s.fatalErr
 }
 
 func (s *hotSlot) exec(ctx context.Context, call *call) error {
@@ -542,6 +544,9 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	case err := <-s.errC: // error from container
 		return err
 	case err := <-errApp: // from dispatch
+		if err != nil && models.IsAPIError(err) && s.fatalErr == nil {
+			s.fatalErr = err
+		}
 		return err
 	case <-ctx.Done(): // call timeout
 		return ctx.Err()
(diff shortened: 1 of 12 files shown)


api/agent/agent.go
```diff
@@ -630,42 +635,38 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
 	state.UpdateState(ctx, ContainerStateStart, call.slots)
 	defer state.UpdateState(ctx, ContainerStateDone, call.slots)
 
-	cid := id.New().String()
-
-	// set up the stderr to capture any logs before the slot is executed and
-	// between hot functions
-	stderr := newLineWriter(&logWriter{
-		logrus.WithFields(logrus.Fields{"between_log": true, "app_name": call.AppName, "path": call.Path, "image": call.Image, "container_id": cid}),
-	})
```
rdallman: removing this does mean we'll lose logs between calls, where this will put them in the server logs for users at least. a quick `stderr.swap(betweenLogger)` below would do.

skinowski: this one is a policy discussion for sure.. so, here, we actually block stderr & stdout, so I don't think we lose logs. (I do need to verify this claim.) A bit related to this is the fact that we freeze the containers in between calls.... (However, freeze behavior is an operator option, which could be disabled.) Given this:

1) is it OK to block stderr? Probably, since we do block stdout....
2) or keep it open, but incur I/O be noisy and log it? We could clamp this with some max, etc. too.

I'm ok with either...

rdallman: that's true, freeze reduces the surface area. one case where we would lose logs I think is if the container starts, logs and exits before the container's first slot is acquired. 

otherwise, it does seem odd to get logs from before a call was started, as a user in certain cases I could see how it would be useful (why did this fail? oh, that's why) but also confusing (why is this log line 3 seconds before the call was started? I can ignore it I guess since everything was okay? is time a circle where all roads lead back to this function call?).

am not sure what the other FaaS providers are doing here.

for stdout, it seems like we should be discarding between functions possibly, or maybe just logging it too? otherwise we can have a container that for 1 input (or 0 input) produces an additional output -- in the 1->2 case, we'd get succeeding functions that reply without accounting for the input (smells funny). it seems in general like a case of 'containers behaving badly', not a clear best way to defend here I can see. we could add it to our shut down shit list, too.

api/agent/agent.go
```diff
@@ -670,18 +660,20 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
 			// wait for this call to finish
 			// NOTE do NOT select with shutdown / other channels. slot handles this.
 			<-slot.done
+
+			if slot.fatalErr != nil {
+				logger.WithError(slot.fatalErr).Info("hot function terminating")
+				return
+			}
```
rdallman: 🥇

skinowski: lol. :-)

api/agent/protocol/json.go
```diff
@@ -141,5 +143,16 @@ func (h *JSONProtocol) Dispatch(ctx context.Context, ci CallInfo, w io.Writer) e
...
+}
+
+func (h *JSONProtocol) isExcessData(err error, decoder *json.Decoder) error {
+	if err == nil {
+		// Now check for excess output, if this is the case, we can be certain that the next request will fail.
+		tmp, ok := decoder.Buffered().(*bytes.Reader)
+		if ok && tmp.Len() > 0 {
+			return ErrExcessData
+		}
+	}
```
rdallman: i guess this has been consumed from the stdout pipe but yet chewed off in the decoder? blimey. it would be nice to do this for all protos (no need here, the mystical 'one day')

skinowski: tried similar things on http, but with no luck... yet.. TODO... 

I'm not sure if this is 100% reliable since it's based on that buf, but hey, if we can catch it, it's a gain...

skinowski: Seems like buffered unread data is lost if we ignore it. Very likely to corrupt next request...

api/agent/agent.go
```diff
@@ -542,6 +540,15 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
...
 	case err := <-errApp: // from dispatch
+		if s.fatalErr == nil && err != nil {
+			if models.IsAPIError(err) {
+				s.fatalErr = err
+			} else if err == protocol.ErrExcessData {
+				s.fatalErr = err
+				// suppress excess data error, but do shutdown the container
+				return nil
+			}
+		}
```
rdallman: other future idea: the errors we expect we could make an error interface along the lines of `interface{ Fatal() bool }` and then a little `func IsFatal(err) bool { f, ok := err.(FatalErr); return ok && f.Fatal() }` -- this would allow us to limit to certain errors we expect like too much output, protocol error, size error, etc. more tersely. though from dispatch probably any error we get is fatal?

skinowski: yes, that sounds good. Dispatch() results are tricky... And this keeps changing based on policy discussions.. So far, three types we have encountered;

1) client messed up, timeout, etc, container goes on
2) protocol corrupt, client gets error, container terminates
3) protocol corrupt, but we could service *this* request, client gets OK, container terminates
4) protocol corrupt, client gets error, container goes on (we honor '2' instead)

I think we still have some to-do on case '1'.

PR #810, opened 2018-03-02
fn: wait for async attach with success channel

skinowski:


api/agent/drivers/docker/docker.go
```diff
@@ -344,30 +344,52 @@ func dockerMsg(derr *docker.Error) string {
...
+			go drv.collectStats(ctx, stopSignal, container, task)
+
+			err = drv.startTask(ctx, container)
+			if err != nil && ctx.Err() == nil {
+				// if there's just a timeout making the docker calls, drv.wait below will rewrite it to timeout
+				startWait <- err
+			} else {
+				startWait <- nil
+			}
+		}()
```
rdallman: is there a reason to change the behavior here to return a container that may yet be started? to fix the attach bug the only code change we need appears to be (with error handling skipped):

```go
... : = drv.AttachToContainerNonBlocking(opts)

// wait to attach to container before we start it so that we do not miss output!
select {
case ctx.Done():
  return nil, ctx.Err()
case <-attachSuccess:
  attachSuccess <- struct{}{}
}
go drv.collectStats(...)

err = drv.startTask(...)
// handle err ...

return &waitResult{...}, nil
```

the above code ensures that we have attached to the container before it has started so that we do not miss any logs (the bug, as I understand it?) but doesn't require threading through start bugs all the way to the waiter, since `startTask` will give us a thumbs up or down on whether it started before returning.

rdallman: alternatively it seems like we could just use `AttachToContainer`? I think originally (or maybe b/c it is so) we had to use non blocking attach since attach would fail until the container was up, but my memory is foggy and the code here appears to do that and the tests pass.

skinowski:
@rdallman This version is simpler, with a gate before startTask.

api/agent/drivers/docker/docker.go
```diff
@@ -345,17 +345,36 @@ func dockerMsg(derr *docker.Error) string {
...
 		return nil, err
 	}
 
+	// Sync up with NB Attacher above before starting the task
+	if err == nil {
+		// WARNING: the I/O below requires docker hijack function to honor
+		// the contract below, specifically if an error is not returned
+		// from AttachToContainerNonBlocking, then max blocking time
+		// here should be what drv.docker dialer/client config was set to.
+		<-attachSuccess
```
rdallman: to be nice we should probably select here with `ctx.Done()`, for cold at least it could happen

api/agent/drivers/docker/docker_client.go
```diff
@@ -294,6 +294,9 @@ func (d *dockerWrap) AttachToContainerNonBlocking(ctx context.Context, opts dock
 	ctx, span := trace.StartSpan(ctx, "docker_attach_container")
 	defer span.End()
 
+	// WARNING: fragile logic below, especially if Success channel is set in opts.
+	// Current impl in d.docker seems safe to repeat the attach even with
+	// Success channel set.
```
rdallman: the `Stats` has a similar issue. honestly we should just skip retry for this one and let it fail. in general, `create container` is the first one that screws up if docker goes down (or pull). it sounds better to have a tiny % fail here than try to reason about this behavior if we ever hit it (task without logs?)

api/agent/drivers/docker/docker.go
skinowski: seems super risky, I went back and forth about this, we risk a go routine leak if we skip this...

rdallman: i see, if nobody receives this it hangs out forever. well, shit, I completely missed that part.

PR #814, opened 2018-03-05
Support load-balanced runner groups for multitenant compute isolation

gviedma:
## Use-Case

In order to support multitenancy, operators should be able to define independent pools of compute resources available to functions in the platform. By associating a function with a particular _load balancing group_, its invocations are guaranteed to execute on the compute resources assigned to that specific group. A pluggable _node pool manager_ abstraction can provide a mechanism to scale compute resources dynamically, based on capacity requirements advertised by the load-balancing layer. Together with load balancer groups, it allows independent capacity and scaling policies for different sets of users or tenants to be implemented.

## Implementation

This adds two implementations of the Fn Agent: one, a "pure runner" that runs only what it is asked to and which responds to GRPC; secondly, an implementation that delegates to an external set of Runners to actually place function invocations. The runner is intended to receive its workload through a push-only mechanism.

Unlike the current FNLB implementation, calls are queued on the "LB" side as late as possible; the idea being that a runner should accept a call if it has the capacity to handle the request immediately to hand. In the long-run, the objective is that this LB implementation will replace the existing FNLB.

(Note: that in itself does not preclude the runner also taking on an async workload.)

The lb_agent introduces some new abstractions: in particular, the NodePool, which plays an analogous role to the DBStore in FNLB; the Runner, which handles the go/no-go and interaction between the LB and the pure_runner; and the Placer, which encapsulates the placement logic.

Additionally, the lb_agent keeps track of the capacity requirements necessary to satisfy the outstanding requests it has. These can, in principle, be used to drive runner scaling decisions. (There is ongoing work in this area.)


skinowski:
One small thing (maybe with another PR later), it would be perhaps better if we can generate the test certs instead of checking them in..

system_test.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+set -exo pipefail
+
+source ./helpers.sh
+
+remove_containers
+
+case "$1" in
+    "sqlite3" )
+    rm -fr /tmp/fn_integration_tests.db
```
skinowski: I'd pick a different directory for this (this is same as in test.sh, which may collide in parallel runs.) (We probably should convert these to use mktemp in the long run.)

system_test.sh
```diff
@@ -0,0 +1,38 @@
...
+    touch /tmp/fn_integration_tests.db
+    export FN_DB_URL="sqlite3:///tmp/fn_integration_tests.db"
+    ;;
+
+    "mysql" )
+    DB_CONTAINER="func-mysql-test"
```
skinowski: again, the container names and ports here collide with api_test.sh, it's possible to get hit by errors here if api_test.sh is already running. For the scope of this PR, we can make sure this doesn't collide with api_test.sh, but in the long run I think we might move some of this logic to the Makefile itself where we spawn these at the beginning shared by both.

system_test.sh
skinowski: I meant api_test.sh

skinowski:
One to-do (probably the conflict cause), is to rebase this on opencensus changes make relevant changes in spans/traces.

api/agent/call.go
```diff
@@ -182,6 +198,47 @@ func WithContext(ctx context.Context) CallOpt {
...
+		a.startStateTrackers(ctx, c)
+		// The call to a.endStateTrackers() will then be done by Submit
+
+		// If we're provided a slot we use it, otherwise we go through the
+		// motions of allocating it.
+		if slot == nil {
+			// First check if we have capacity. We exit early if we don't, because
+			// reserving a slot must be a quick operation, as in, "can I reserve a
+			// slot RIGHT NOW please?".
+			if !a.resources.IsResourcePossible(c.Memory, uint64(c.CPUs), c.Type == models.TypeAsync) {
```
skinowski: I apologize, IsResourcePossible function name is not 100% clear here, but IsResourcePossible() checks if this request can ever be run. For example, if total RAM is 8GB, you can never ever run a 32GB job. So, here this logic seems flawed, as getSlot() will block for extended period even if IsResourcePossible() returned true.

An example:

8GB total ram, a request comes in for 1GB, IsResourcePossible() will return *always* True. And if there are other requests active (eg. 7.5 GB already in use), then getSlot() will block and wait until a free slot deadline available or context expires.

What you probably wanted was;

1) check IsResourcePossible() to fail impossibly big requests quickly (or skip this step and let '2' fail below)
2) run getSlot() with a quick/short slot deadline. Instead of Call.Timeout / 2.

api/server/server.go
```diff
@@ -402,6 +610,18 @@ func (s *Server) startGears(ctx context.Context, cancel context.CancelFunc) {
...
 
+	if s.nodeType == ServerTypePureRunner {
+		// Run grpc too
+		grpcAddr := fmt.Sprintf("%s:%d", whoAmI().String(), s.grpcListenPort)
+		pr, err := agent.CreatePureRunner(grpcAddr, s.agent, s.cert, s.certKey, s.certAuthority)
+		if err != nil {
+			logrus.WithError(err).Fatal("Pure runner server creation error")
+		}
+		go func() {
+			pr.Start()
```
skinowski: For the shutdown case, (interrupt + sigterm), we should log the error returned from Start(). See the go-routine a few lines below. There's also a Serve & Listen error case handling in there (see cancel()).

skinowski: https://github.com/fnproject/fn/pull/814/files#diff-44a527ea72ea9572e1cf71c7f1e2a0f0R632

skinowski: GracefulStop case needs handling here.. See:
https://github.com/fnproject/fn/pull/814/files#diff-44a527ea72ea9572e1cf71c7f1e2a0f0R645

api/server/server.go
```diff
@@ -402,6 +610,18 @@ func (s *Server) startGears(ctx context.Context, cancel context.CancelFunc) {
 
 	installChildReaper()
 
+	if s.nodeType == ServerTypePureRunner {
+		// Run grpc too
+		grpcAddr := fmt.Sprintf("%s:%d", whoAmI().String(), s.grpcListenPort)
```
skinowski: is it possible to skip specifying local IP part? I think server (http.Server) listen and gRPC listen should be in sync with each other regarding this.. If we decide to bind to a specific local IP, then both servers should do the same.

api/agent/lb_hash.go
```diff
@@ -0,0 +1,74 @@
...
+	sum64 := siphash.Hash(0, 0x4c617279426f6174, []byte(key))
+
+	for {
+		if time.Now().After(deadline) {
+			return models.ErrCallTimeoutServerBusy
+		}
+
+		runners := np.Runners(lbGroupID)
+		if len(runners) < 1 {
+			time.Sleep(retryWaitInterval)
```
skinowski: should be sleep(min(retryWaitInterval , delta(now, deadline)))

api/agent/lb_hash.go
```diff
@@ -0,0 +1,74 @@
...
+				logrus.WithError(err).Error("Failed during call placement")
+			}
+			if placed {
+				return err
+			}
+
+			i = (i + 1) % len(runners)
+		}
+
+		time.Sleep(retryWaitInterval)
```
skinowski: should be sleep(min(retryWaitInterval , delta(now, deadline)))

api/agent/lb_hash.go
skinowski: But better to select on both, since agent Submit could cancel this (eg. shutdown)

api/agent/lb_agent.go
```diff
@@ -0,0 +1,182 @@
...
+			if err != nil {
+				logrus.WithError(err).Error("Failed during call placement")
+			}
+			if placed {
+				return err
+			}
+
+		}
+
+		time.Sleep(retryWaitInterval)
```
skinowski: same issue as lb_hash side, we should select on both ctx and retryWaitInterval to handle ctx timing out or ctx being cancelled.

api/agent/lb_agent.go
```diff
@@ -0,0 +1,182 @@
...
+// a remoteSlot which will implement the actual running functionality.
+func (a *lbAgent) GetCall(opts ...CallOpt) (Call, error) {
+	slot := &remoteSlot{lb: a}
+	opts = append(opts, WithReservedSlot(context.Background(), slot))
+	return a.delegatedAgent.GetCall(opts...)
+}
+
+func (a *lbAgent) Close() error {
+	a.np.Shutdown()
+	return nil
```
skinowski: delegateAgent shutdown forgotten?

api/agent/pure_runner.go
```diff
@@ -0,0 +1,491 @@
...
+
+// callHandle represents the state of the call as handled by the pure runner, and additionally it implements the
+// interface of http.ResponseWriter so that it can be used for streaming the output back.
+type callHandle struct {
+	engagement runner.RunnerProtocol_EngageServer
+	c          *call // the agent's version of call
+	input      io.WriteCloser
+	started    bool
+	// As the state can be set and checked by both goroutines handling this state, we need a mutex.
+	stateMutex sync.Mutex
```
skinowski: There are many network calls while holding this mutex locked. Is that intended? If yes, then we do have upper bounds on timeouts on all of those rpc calls?

api/agent/pure_runner.go
```diff
@@ -0,0 +1,491 @@
...
+	return ch.outHeaders
+}
+
+func (ch *callHandle) WriteHeader(status int) {
+	ch.outStatus = status
+	ch.commitHeaders()
+}
+
+func (ch *callHandle) commitHeaders() error {
+	if ch.headerWritten {
```
skinowski: this doesn't need the mutex?

api/agent/nodepool/grpc/grpc_pool.go
```diff
@@ -0,0 +1,414 @@
...
+
+	// Get app and route information
+	// Construct model.Call with CONFIG in it already
+	modelJSON, err := json.Marshal(call.Model())
+	if err != nil {
+		logrus.WithError(err).Error("Failed to encode model as JSON")
+		// If we can't encode the model, no runner will ever be able to run this. Give up.
+		return true, err
+	}
+	runnerConnection, err := r.client.Engage(context.Background())
```
skinowski: Should the ctx here be the original ctx (which had execDeadline)?

skinowski:
IMHO, lb agent is too tightly coupled with the runner agent and this may slow us down in the long run.

LB agent does not seem to have much common with runner agent as seen from remoteSlot, and various if-else blocks in agent.submit(), call.GetCall(), etc.

Container/Request states in runners do not really correspond to the same states in LB.

For example, is call.Start() is applicable in this case? Since getSlot is skipped, all of this lose meaning including a XXX-FXLB-WAIT header in responses (was this intentional?) or stats counters such as enqueued.

remoteSlot is no longer really a slot, but a pointer back to lb agent, or exec() is not really exec but an LB forward. Timeouts, etc. are a bit more hard to follow. (They were already very confusing in current agent.)

I wonder if this would be much easier to follow and maintain if we just code the lb agent flow separately without trying to reuse any runner agent code.

treeder:
I haven't dug in too much to this, but any chance this could be implemented as an extension?

api/agent/call.go
hhexo: But getSlot() could also perform the download of the image, the container startup, etc, which will take time and fail the short deadline even if the node has all the capacity in the world.

What we need is just a capacity check. I guess we could use isResourceAvailableLocked() somehow.

hhexo:
I don't think this can work as an extension, because there is so much private stuff in the agent package that an extension can't possibly hope to access everything it needs (e.g. lowercase "call"). We even wanted to write the new stuff in a different package and it turned out to be too hard. In order to do this in anything other than the agent package, we'd need to open up a lot of the agent abstractions.

This is partly why we have reused and repurposed the existing abstractions to implement the LB Agent and the pure runner. The trick with remoteSlot is a bit odd, but the alternative is to rewrite a lot of the agent code (that we still need) in a different agent implementation. As far as I could see the Slot interface represents "the object you can use to actually run and terminate this thing" (because the interface only has exec() and Close() methods) and it has different implementations for hot and cold - so we thought of adding a "remote" implementation.

api/agent/pure_runner.go
hhexo: Each call (as in, the call of an Fn function) has a different handle and therefore a different mutex - it's not a sequential chokepoint.

Within a single Fn function call, there are only two goroutines that access the shared state: the receive loop and the code that runs Submit and sends back the output. The receive loop is mostly idle once the function has received the input, so there shouldn't be much contention on the mutex.

api/agent/pure_runner.go
hhexo: Only one goroutine reads/write that flag.

api/agent/call.go
hhexo: Actually the capacity check has to be more clever than that, because it has to recognize hot slots that can be reused... my bad, scratch my previous comment.

GetSlot() with a short deadline looks like a good compromise for now, except that it will probably fail if the image is not already on the machine - so in the 0-to-1 case nothing will ever run unless the image is preloaded and GetSlot() is fast.

I guess we are fighting the fact that the abstraction does not have the concept of a slot reservation (i.e. a ReserveSlot())... maybe we should just add that.

hhexo: I'm thinking about this and I think that the main problem here is that the slot creation and allocation mechanisms are based on a producer/consumer model which does not allow us to explicitly run queries like "give me a slot if you can, otherwise tell me why you can't".

The only thing the consumer (the call) can do is wait and hope that someday the producer (slot manager) will give them a slot. The consumer has no visibility of the current slot situation and can make no decisions to fail fast.

I'm open to suggestions, but I believe the current slot handling is unsuited to the "push" flow that we want to implement. Maybe we should have an alternative manager for handling slots...

rdallman: what are the benefits of reserving a slot in the LB if we have topological saturation knowledge? the latter is something of a requirement in some capacity or we risk O(n) placement. are there high % traffic cases where we would ask for a slot but not receive one (it seems low %)?

rdallman:
> I don't think this can work as an extension,

I agree in its current aims it will be rather painful to move _all_ of this into some abstraction. I've read through the patch and can think of some viable working product that uses current 'extensions', though these are free to modify to our needs of course. In no way is this meant to be entirely prescriptive, this is mostly a guess based on my current understanding of the machinations here & subsequent attempt to start the extension convo:

* extend FNLB's proxy layer to create (also ingest?) a grpc request with a fully constructed call object to send to FN
* extend FN to have a grpc endpoint that receives a full call, and throws it against agent.Submit (eliding pre-emptive slot logic, for now) and returns an object with the call execution info (started_at, status, etc) [and log (?)]. this can be added to server.Router ?
* create a new 'grouper' passed to the LB that returns a different set of nodes based on the given key (as opposed to 'allgrouper' which ignores the key - the health crap thing needs to be moved somewhere more re-usable, but perhaps copying for now is ok). this seems mostly a lift and drop of the pool manager?
* somewhere FNLB needs to construct a call from CP information, this seems the hardest plumbing issue of 'rev 1' (sans external slots) ? it seems the goal is to marry an API node and the LB, this seems achievable / sensible. we can probably hook fn up to import the LB and use it for (equiv) `/r/` routing and configure an API node & its routes alongside? if the LB simply needs to interface with some CP API (i've lost track) just ignore that last bit.
* [done] make a 'nil' `agent.DataAccess` layer that does not upload call / log
* [done] disable the async poller in agent

are there fatal flaws with future work in skipping the slot plumbing out for now? the above seems a plenty large chunk of forward progress to achieve. Also general question of best way to marry an API node and an LB if I'm understanding correctly (?). Now, what did I miss? :)

api/agent/call.go
skinowski: Also reservation needs to take into account of some cases, such as idle and eject timers. Due to these, the "actual" agent can eject/kill and timeout containers (aka slots.) We've built into some logic into agent that can detect sleeping containers and kill them. The reservation must not be delayed enough for these to kick in, or some of these mechanisms need to be disabled.

An example;

1) For RouteA, LB_1 reserves a slot and hence a slot is waiting for the actual request.
2) RouteB arrives for another reservation (perhaps from LB_2), but system seems busy and this triggers eject logic and kills RouteA container.
3) LB_1 forwards RouteA actual exec request, but reservation is gone.

So, LB already needs to handle reservations that are no longer valid. And grpc calls/responses needs to be tuned to ensure they don't get hit by idle/eject timeouts too much.

LB seems to be too tied into Agent implementation with the reservation system. To make it work, we'll need tweaks in runner agent as well. Eg. Step 3 can in theory be retried, but today it can't without tweaks/fixes.

I think it would be good if we could make all this work without reserving slots.

api/agent/pure_runner.go
skinowski: WriteHeader() is public and this io.Writer is passed out to the actual agent... And proto.Dispatch() runs on it's own go-routine.

system_test.sh
```diff
@@ -0,0 +1,38 @@
...
+    touch /tmp/fn_system_tests.db
+    export FN_DB_URL="sqlite3:///tmp/fn_system_tests.db"
+    ;;
+
+    "mysql" )
+    DB_CONTAINER="func-mysql-test"
+    docker rm -fv ${DB_CONTAINER} || echo No prev mysql test db container
+    docker run --name ${DB_CONTAINER} -p 3306:3306 -e MYSQL_DATABASE=funcs -e MYSQL_ROOT_PASSWORD=root -d mysql
+    MYSQL_HOST=`host ${DB_CONTAINER}`
+    MYSQL_PORT=3306
```
skinowski: also the ports collide with api_test.sh, not a super important issue, but we do run the builds in parallel.

api/server/server.go
reclaro: We added a call to cancel() in case of error in creating the pure runner or error in start it.

hhexo:
The slot plumbing is solving (well, currently failing to solve) the problem of being able to support the semantics of "try to run this call, if you have capacity right now commit to it and do it, otherwise don't do anything and tell me you can't run it - do not queue". These semantics are required by how the new multi-tenant multi-runner dataplane is supposed to work. At the moment however Submit is only able to perform the semantics of "queue this call and run it when you can", which is why the concept of slot reservation came to mind as a possible solution.

Slot reservation may not be the right solution, but the problem is there. Another way to solve this could be to split Submit, basically changing the interface of the agent to change the "GetCall -> Submit" flow into "GetCall -> AllocateResources -> Submit", then we'd have more control on the behaviour of AllocateResources (the conditions in which we accept to queue the call, or fail fast - maybe we could have SlotAllocationOptions?). Or, as a narrower change, we could just add a flag to Submit to control whether to queue the call or fail fast.

Personally I perceive this problem as missing core functionality required to implement the semantics we need, and I would like to see this in the open source rather than some kind of private extension.

hhexo:
> Or, as a narrower change, we could just add a flag to Submit

Actually no, that doesn't work because we need to make the decision to accept/reject the call _before_ calling Submit (the last controllable step) on it.

system_test.sh
reclaro: good spot, we changed the port and the name of the containers

hhexo:
Ok...

We have given up trying to add functionality to the Agent in order to deal with slot reservation, capacity checking etc. A longer discussion is needed to figure out how to extend the abstractions of the slot allocation mechanism so that different slot allocations algorithms (e.g. "queue until you get one" vs "don't queue and fail fast") can be selected.

_As a tactical decision_ we have moved the initial capacity check to a higher level, in the Pure Runner. This probably duplicates / reimplements some logic, but it gets us closer to where we want to be and probably removes a blocker for this PR. We can and will iterate on that.

Having removed the slot reservation mechanism, we can't use the little trick we used to use in LB Agent so the LB Agent is now reimplementing `Submit()` and `submit()` up to the point where a slot is assigned, then a `remoteSlot` is used.

api/agent/pure_runner.go
hhexo: True... but I'm assuming that only one thread will ever write to the http.ResponseWriter, otherwise the http response would be garbled anyway.

What I mean is, if we get concurrency issues in the writes here, adding the mutex lock would not solve the problem - there would be some broken logic in the program.

skinowski:
A side note, seems the api-tests and system-tests both try to use port 8080 which causes port collusion.
Ideally tests should bind to dynamic unalloc ports (bind port 0), but to get things on-going/unblocking, you might just try to choose a different port.

PR #815, opened 2018-03-05
VirtualBox/Vagrant Backend for CP

faiq:
A local CP interface that uses VirtualBox as a backend. This is to be used in a testing environment


Gopkg.lock
```diff
@@ -378,6 +384,12 @@
   packages = ["."]
   revision = "316fb6d3f031ae8f4d457c6c5186b9e3ded70435"
 
+[[projects]]
+  branch = "master"
+  name = "github.com/google/uuid"
+  packages = ["."]
+  revision = "dec09d789f3dba190787f8b4454c7d3c936fed9e"
+
```
rdallman: we have a somewhat reasonable id generator https://github.com/fnproject/fn/blob/master/api/id/id.go that doesn't require any imports -- this pool manager appears mostly to be for testing purposes and the linked id format will suffice without need to set the machine id.

rdallman:
if you make the oracle org public on your GH you will auto pass the CLA bot

Gopkg.lock
faiq: 👍 will change

zootalures:
Hey - Can you (plan to) move the node-pool-manager out as a seperate component to Fn - it's an implementation detail of the distributed FnLB  implementation and I'd it was built and maintained seperately (with a Plugin/Extension interface compiled into the LB) rather than as a gestalt component.

rdallman:
+1 -- we don't have an 'official' extensions repository at the moment which makes it less clear where to put things. to date, they've been on personal repos (e.g. https://github.com/fnproject/ext-auth ) but you also should have access to creating a repo on `fnproject` and that would work for now, as well.

faiq:
@rdallman Thanks for the comment. Right now I'm writing this code based on the node pool and control plane implentations that the bristol team has done on #814, which currently does not have an extension system for the poolmanager. This pull request allows for local testing of this implementation of the pool manager, so that deveoplers can test out images/logic/etc for the runners.

PR #816, opened 2018-03-06
add benchmarks for ids

rdallman:
... I was curious, may as well keep them

not bad, mostly `time.Now()` and the atomic add on high core machines might make this worse, anyway it's pretty good:

```
✗: go test -bench . -benchmem
goos: linux
goarch: amd64
pkg: github.com/fnproject/fn/api/id
BenchmarkGen-2                  20000000                73.7 ns/op             0 B/op          0 allocs/op
BenchmarkMarshalText-2          30000000                47.5 ns/op            32 B/op          1 allocs/op
BenchmarkUnmarshalText-2        100000000               17.6 ns/op             0 B/op          0 allocs/op
PASS
```

### DIFF
--- a/api/id/id.go
+++ b/api/id/id.go
@@ -47,7 +47,7 @@ func SetMachineIdHost(addr net.IP, port uint16) {
 func New() Id {
 	var id Id
 	t := time.Now()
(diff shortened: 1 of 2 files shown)


rdallman:
thanks!

Issue #822, opened 2018-03-08
Flaky Test: TestRouteRunnerIOPipes


gviedma:
#809 introduces a flaky test _TestRouteRunnerIOPipes_, possibly related to the fixing of #796. I have removed it while the issue is resolved.

```
--- FAIL: TestRouteRunnerIOPipes (5.89s)
	runner_test.go:275: Test 0: dockerId: 91e1f0ac28101d2cb6ce26df1e553fa87a547d8a9891f64344597734b340d8e5
	runner_test.go:247: Test 1: Expected status code to be 200 but was 502. body: {"error":{"message":"invalid json response from function err: invalid character 'Y' looking for beginning of value"}}
	runner_test.go:275: Test 1: dockerId: N/A
	runner_test.go:247: Test 2: Expected status code to be 502 but was 200. body: {"request":{"isDebug":true},"header":{"Content-Length":["17"]},"config":{"ENABLE_FOOTER":"yes","ENABLE_HEADER":"yes","FN_APP_NAME":"zoo","FN_FORMAT":"json","FN_MEMORY":"64","FN_PATH":"/json","FN_TYPE":"sync","HOME":"/root","HOSTNAME":"default-c904ee0f-93fa-4cf7-907f-0193d6b3b045","PATH":"/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"},"data":{"DockerHostname":"default-c904ee0f-93fa-4cf7-907f-0193d6b3b045","DockerId":"bf59639a0e483ff6f4f76fccaf000dc9eabf113d5cdf307b15c45401add6ea1c"},"trailer":[]}
[...]

PR #823, opened 2018-03-08
Removes flaky runner test TestRouteRunnerIOPipes while #822 is resolved

gviedma:


zootalures:
@skinowski We are disabling this flakey test

gviedma:
@skinowski New flappy test is tracked in #822, old flappy test is tracked in #796

PR #825, opened 2018-03-08
phase 1 sqlx migrator

rdallman:
code is feature complete in the general sense, with minor TODO left.

this is just a patch with 'migratex' and does not use it for fn's migrations
yet, would like to get feedback prior to doing that.

presenting:

A migration library loosely based on pressly/goose and mattes/migrate design,
that does migrations across a smattering of sql databases by only accepting a
`*sqlx.DB`.

why?

* goose didn't support kindly allowing us to rebind transactions based on a
given db to various dialects or offer oracle support
* goose didn't support locking the db (maybe not needed with tx? it's late..
we may want to lock the whole db eventually?)
* goose requires us to do semi-complex migration to it from mattes/migrate
* mattes has stepped down as migrate maintainer and the project is in flux
* mattes/migrate did not allow us to define migrations in go and rebind to
different dialects, an issue since we need to insert ids in our own format and
can't define this in sql
* neither handled context plumbing and risked issues there for various
reasons (deadlock, etc).
* I think I'm forgetting 1 or 2

in the style of goose, this lets us define `*sqlx.Tx` up and down funcs in go
code, but uses mattes' migration table so we don't need to migrate that and
retains its lock behavior with added tx sugar and less errors. most
importantly, this code is terse, leveraging sqlx to support a lot of sql dbs
(unlike mattes) and we control this. there is one useful TODO to handle
migrations failing at startup more gracefully, in prod stuff like that will be
nice to have. open to discussion of putting in a separate library, the
landscape of go sql migrators is... really something.

TODO make test suite and test against sqlite3, pg, mysql [, oracledb] like we
have for our own unit tests. I'm thinking it's faster to wire up through
there and use our bevy of migrations?

### DIFF
--- a/api/datastore/sql/migratex/migrate.go
+++ b/api/datastore/sql/migratex/migrate.go
@@ -1,0 +1,337 @@
+package migratex
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"hash/crc32"
+	"sort"
+	"strings"
+
+	"github.com/go-sql-driver/mysql"
+	"github.com/jmoiron/sqlx"
+	"github.com/lib/pq"
+)
+
+var (
+	// use same migration table name as mattes/migrate, so that we don't have to
+	// migrate that.
+	// TODO doesn't have to be a glob
+	MigrationsTable = "schema_migrations"
+
+	ErrLocked     = errors.New("database is locked")
+	ErrDirty      = errors.New("database is dirty")
+	ErrOutOfOrder = errors.New("non-contiguous migration attempted")
+)
+
+const (
+	NilVersion = -1
+)
+
+type Migration interface {
+	Up(*sqlx.Tx) error
+	Down(*sqlx.Tx) error
+	Version() int64
+}
+
+type sorted []Migration
+
+func (s sorted) Len() int           { return len(s) }
+func (s sorted) Less(i, j int) bool { return s[i].Version() < s[j].Version() }
+func (s sorted) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
+
+var _ Migration = new(MigFields)
+
+// MigFields implements Migration and can be used for convenience.
+type MigFields struct {
+	UpFunc      func(*sqlx.Tx) error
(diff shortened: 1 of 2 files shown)


rdallman:
@denismakogon please let me know what you think of the general idea here + any functionality issues. ended up being pretty small but kind of dense (no generation stuff is nice).

api/datastore/sql/migratex/migrate.go
```diff
@@ -0,0 +1,337 @@
+package migratex
```
denismakogon: Looks straightforward, so the only thing that bothers me is migration collecting, for migrations we have I'd like to make this framework go and collect every migration it has and then apply them step by step, similar to what I did in my PR. This is what i felt when saw the test case for the migrations. At this moment, it looks quite complex.

So, we want people to create the migrations and then just append them to the global list of migrations and then let the framework sort them, and apply, we shouldn't expose much logic to users except the transaction object sqlx.Tx.

I see that we have necessary logic for sorting those, so there's nothing much left to do.

rdallman: > Looks straightforward, so the only thing that bothers me is migration collecting, for migrations we have I'd like to make this framework go and collect every migration it has and then apply them step by step, similar to what I did in my PR. This is what i felt when saw the test case for the migrations. At this moment, it looks quite complex.

could add a step-wise instead of all up or all down; just trying to keep the surface area small since we were only ever going all the way up or down (mattes has 5-6 ways to run migrations, it's all noise). this is what you mean, correct?

> So, we want people to create the migrations and then just append them to the global list of migrations and then let the framework sort them, and apply, we shouldn't expose much logic to users except the transaction object sqlx.Tx.

I suppose it's pretty flexible since we just take a `[]Migration`, the global list is one way to get them in there. does this comment imply some change you'd like to see? it's not clear.

thank you for reviewing.

rdallman:
> Okay, i'm cool with this. So, maybe we need update our current migrations?

yep, phase 2. thanks. we can mostly copy go migration files from #787 and the code in `database/sql/sql.go` is mostly drop in changing import. will try to get a PR up today

PR #830, opened 2018-03-09
fn: json excess data check should ignore whitespace

skinowski:


skinowski:
I don't know if we can avoid that io.Copy() there, but this fixes the excess json data issue.

api/agent/protocol/json.go
```diff
@@ -149,9 +150,23 @@ func (h *JSONProtocol) Dispatch(ctx context.Context, ci CallInfo, w io.Writer) e
...
+			defer bufPool.Put(buf)
+
+			// Let's check if extra data is whitespace, which is valid/ignored in json
+			io.CopyN(buf, reader, int64(reader.Len()))
+
+			isNonSpace := func(c rune) bool {
+				return !unicode.IsSpace(c)
+			}
+
+			if bytes.IndexFunc(buf.Bytes(), isNonSpace) != -1 {
```
rdallman: without the the closure or buffer; I think this is potentially an exploitable buffer:

```go
// rationale: we must not allocate any buffers/slices, this is potentially an attack vector (shove 8GB of whitespace between objects and we OOM)

for {
  r, err := reader.ReadRune()
  if err == io.EOF { // no more bytes
    break
  }
  if !unicode.IsSpace(r) {
    return ErrExcessData
  }
}
```

the IndexFunc implementation appears to go rune by rune and not do any optimized text search (don't think it can, with the closure matcher) so this would not be any slower from that angle. I don't think we need the type check `ok` since json's implementation would have to change for this to fail, and it would lead us to leaving stuff in there (back to where we were), so maybe we just eat the panic in tests if implementation changes?

complete aside:

we should get rid of whitespace in our json protocol definition. json has a perfectly good spec for delimiting objects, we don't need a white space in there at the protocol level (though ignoring them here doesn't hurt, we should do this too)

rdallman:
> Can we get test for this?

+1, I think we have tests already that spy on containers launching and we can just check after 2 json tasks that only 1 container launched (http is covered here I believe), but if not we can punt.

api/agent/protocol/json.go
skinowski: So here Buffered() is remaining data in buffer. It's already read in a slice to my understanding. But I do like that ReadRune() for loop as it avoids extra copy... I'll submit another commit.

skinowski:
@rdallman @denismakogon ping..

PR #832, opened 2018-03-09
Add a basic concurrency test for the dataplane system tests

hhexo:
Add a basic concurrency test for the dataplane system tests.

Also fix a minor issue with capacity accounting, discovered with tests, and remove some spurious logging.


api/agent/nodepool/grpc/grpc_pool.go
```diff
@@ -311,7 +311,7 @@ func (r *gRPCRunner) TryExec(ctx context.Context, call agent.Call) (bool, error)
 	switch body := msg.Body.(type) {
 	case *pb.RunnerMsg_Acknowledged:
 		if !body.Acknowledged.Committed {
-			logrus.Errorf("Runner didn't commit invocation request: %v", body.Acknowledged.Details)
+			logrus.Debugf("Runner didn't commit invocation request: %v", body.Acknowledged.Details)
```
hhexo: (this is genuinely not an error: it's expected that some runners can refuse calls if they're at capacity)

PR #833, opened 2018-03-09
Refactor controlplane into a go plugin

mattstep:
Wait to merge this until I add some docs for it. Just getting this up to start a review now.


poolmanager/server/controlplane/controlplane.go
```diff
@@ -0,0 +1,20 @@
+/*
+	Interface between the Node Pool Manager and the Control Plane
+*/
+
+package controlplane
+
+const CAPACITY_PER_RUNNER = 4096
```
rdallman: Go's idiom for constants is mixed casing for constants https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps (and they note this is odd, but the stdlib is this way)

poolmanager/server/controlplane/controlplane.go
```diff
@@ -0,0 +1,20 @@
...
+	Address string
+	// Other: certs etc here as managed and installed by CP
+	Capacity int64
+}
+
+type ControlPlane interface {
+	GetLBGRunners(lgbId string) ([]*Runner, error)
+	ProvisionRunners(lgbId string, n int) (int, error)
+	RemoveRunner(lbgId string, id string) error
+}
```
rdallman: 🥇

poolmanager/server/controlplane/plugin/noop.go
```diff
@@ -1,56 +1,48 @@
...
+	"github.com/fnproject/fn/poolmanager/server/controlplane"
+)
 
-type ControlPlane interface {
-	GetLBGRunners(lgbId string) ([]*Runner, error)
-	ProvisionRunners(lgbId string, n int) (int, error)
-	RemoveRunner(lbgId string, id string) error
-}
+const (
+	EnvFixedRunners = "FN_RUNNER_ADDRESSES"
```
rdallman: ditto

poolmanager/server/controlplane/plugin/noop.go
```diff
@@ -121,3 +113,13 @@ func (cp *controlPlane) RemoveRunner(lbgId string, id string) error {
...
+
+func getEnv(key string) string {
+	value, ok := os.LookupEnv(key)
+	if !ok {
+		log.Panicf("Missing config key: %v", key)
+	}
+	return value
+}
+
+var ControlPlane noopControlPlane
```
rdallman: not sure about this one, it has uninitialized maps and `NewControlPlane()` returns an instance of one of these to get a handle on, perhaps this was cruft from shifting stuff around?

poolmanager/server/controlplane/plugin/noop.go
```diff
@@ -1,56 +1,48 @@
-/*
-	Interface between the Node Pool Manager and the Control Plane
-*/
-
-package cp
+/**
+ * Dummy implementation for the controlplane that just adds delays
+ */
+package main
```
rdallman: package noop? package noopcp? I do not see a main in this package checked in, at least.

rdallman:
thanks for this, looks great. some minor comments

poolmanager/server/main.go
```diff
@@ -116,6 +115,25 @@ func createGrpcCreds(cert string, key string, ca string) (grpc.ServerOption, err
 	return grpc.Creds(creds), nil
 }
 
+func newPluggableControlPlane() controlplane.ControlPlane {
```
faiq: Do we know for sure that this instantionalizes the object for sure? I think @rdallman also commented on the bottom of the NoopControlPlane with a similar concern.

We could return a factory as an export, and call the function that was exported.

poolmanager/manager.go
```diff
@@ -8,7 +8,7 @@ import (
 	"sync"
 
 	model "github.com/fnproject/fn/poolmanager/grpc"
-	"github.com/fnproject/fn/poolmanager/server/cp"
```
faiq: HOW WAS EVERYTHING NAMED CP 👍  on the change

poolmanager/manager.go
```diff
@@ -296,20 +296,20 @@ func (lbg *lbGroup) target(ts time.Time, target int64) {
...
 				// Some kind of error during attempt to scale up
 				logrus.WithError(err).Error("Error occured during attempt to scale up")
 				return
 			}
-			lbg.target_capacity += int64(asked_for) * cp.CAPACITY_PER_RUNNER
+			lbg.target_capacity += int64(asked_for) * controlplane.CAPACITY_PER_RUNNER
 		}
 
-	} else if desiredScale <= lbg.current_capacity-cp.CAPACITY_PER_RUNNER {
+	} else if desiredScale <= lbg.current_capacity-controlplane.CAPACITY_PER_RUNNER {
```
faiq: can-we_do_someTHING-About-THIS_too?

faiq:
Looks good! Love the plugin stuff. My suggestion would be to have a factory be exposed, and then call the returning function to get an instance of a control plane manager. Right now I'm under the impression that you're not returning an instance of the object with all the maps and stuff allocated.

I think an easy test would be adding another method to the interface that is something like `Name()` that can print the name. This can make it clearer which driver we're using and can ensure that the object is there.

poolmanager/server/controlplane/plugin/noop.go
skinowski: probably needs to be initialized in init() which plugin.Open() would call, right?

poolmanager/server/controlplane/plugin/noop.go
mattstep: It's a requirement of buildmode=plugin, the symbols are only exported from the main package

poolmanager/manager.go
mattstep: I'll do a followup pr to clean all this up

poolmanager/server/controlplane/plugin/noop.go
mattstep: fixed

poolmanager/server/main.go
mattstep: yep,fixed

mattstep:
@rdallman ptal

poolmanager/server/controlplane/plugin/noop.go
rdallman: d'oh

PR #834, opened 2018-03-09
Make PureRunner an Agent so that it encapsulates its grpc server

hhexo:
**PATCH 1/2 - blocks #849** 

This makes PureRunner a type of Agent and moves the grpc server initialization into the creation function.

This allows us to avoid having to create and start the grpc server in server.go, which was a bit of a smelly leakage from the PureRunner abstraction.

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -642,7 +642,17 @@ func (s *Server) startGears(ctx context.Context, cancel context.CancelFunc) {
 	for i, ctx := range s.extraCtxs {
 		cases[i] = reflect.SelectCase{Dir: reflect.SelectRecv, Chan: reflect.ValueOf(ctx.Done())}
 	}
-	reflect.Select(cases) // Just wait for one of them. TODO: maybe some logging of which one was triggered?
+	nth, recv, wasSend := reflect.Select(cases)
+	if wasSend {
+		logrus.WithFields(logrus.Fields{
+			"ctxNumber":     nth,
+			"receivedValue": recv.String(),
+		}).Debug("Stopping because of received value from done context.")
+	} else {
+		logrus.WithFields(logrus.Fields{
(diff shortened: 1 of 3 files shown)


api/server/server.go
```diff
@@ -392,7 +392,14 @@ func WithAgentFromEnv() ServerOption {
 			if err != nil {
 				return err
 			}
-			s.agent = agent.NewSyncOnly(agent.NewCachedDataAccess(ds))
+			grpcAddr := fmt.Sprintf(":%d", s.grpcListenPort)
+			delegatedAgent := agent.NewSyncOnly(agent.NewCachedDataAccess(ds))
+			_, cancel := context.WithCancel(ctx)
```
skinowski: my suggestion here might not be an elegant way of handling this, but perhaps we could save this 'ctx' in Server object, then handle/wait for it in startGears(), eg.

```
tmp, cancel := context.WithCancel(context.Background())
s.cancelCtx = tmp
```

otherwise failure of Start() in pure runner is missed...

hhexo: Good point, I'll see what I can do about the pure runner cancellation and the context.

hhexo: I'm now thinking that if we allow extensions or implementations to run servers alongside the main one then there could be more things that would generate a cancellation context, for example if we have extensions that implement their own http server (other metrics extensions maybe?).

What if we store an array of `cancelCtx`s to which extensions can append their own, and then use `reflect.Select` to dynamically wait for any of them to be `Done()`? Do you think it's a good solution or is it too convoluted?

hhexo:
@skinowski I have added a mechanism to tell the server to consider extra contexts to wait/select on in startGears; in the end it was simpler than I expected and I believe this is also extension-friendly. What do you think?

PR #839, opened 2018-03-10
Update the Python tutorial

cartershanklin:
The Python tutorial didn't work as written.

Please consider these updates which I tested against a local (brew installed) version of fn installed on March 9 2018. Worth mentioning that func.py and sample.payload.json don't have any bearing on what "fn init" generates for you nowadays and it might make sense to get rid of this stuff.

Note1: "fn test" doesn't work due to a separate issue. Hopefully that issue will be resolved such that the tutorial as written will start working.

Note2: I have not signed the CLA and won't be able to do so until next week.

### DIFF
--- a/examples/tutorial/hello/python/README.md
+++ b/examples/tutorial/hello/python/README.md
@@ -6,31 +6,38 @@ This example will show you how to test and deploy Python code to Fn. It will als
 
 ```sh
 # Initialize your function creating a func.yaml file
-fn init --name hello-python
+fn init --runtime python3.6 hello-python
+
+# Enter the directory fn created.
+cd hello-python
 
 # Test your function. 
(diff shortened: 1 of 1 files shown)


examples/tutorial/hello/python/README.md
```diff
@@ -6,31 +6,38 @@ This example will show you how to test and deploy Python code to Fn. It will als
 
 ```sh
 # Initialize your function creating a func.yaml file
-fn init --name hello-python
+fn init --runtime python3.6 hello-python
```
denismakogon: no longer necessary, works as it was before

denismakogon: i'd rather add a note that FDK-python is only v3.6-compatible, no 2.x

Issue #836, opened 2018-03-09
docker pause errors


skinowski:
Sample snippet from test logs:

```
time="2018-03-09T18:39:28Z" level=debug msg="Starting container execution" app=zoo container=01C860Z3M9A7WHJ00000000000 cpus= format=http id=01C860Z3M9A7WHJ00000000000 idle_timeout=30 image=fnproject/fn-test-utils memory=64 route=/http
        time="2018-03-09T18:39:29Z" level=debug msg="Hot function launcher stats" action="server.handleFunctionCall)-fm" app=zoo currentStats="{[0 1 0 0] [0 0 1 0 0 0]}" id=01C860Z3M9A7WHG00000000000 isNeeded=false route=/http
        time="2018-03-09T18:39:29Z" level=debug msg="docker pause" app=zoo call_id=01C860Z3M9A7WHJ00000000000 cpus= format=http id=01C860Z3M9A7WHJ00000000000 idle_timeout=30 image=fnproject/fn-test-utils memory=64 route=/http stack=Freeze
        
        time="2018-03-09T18:39:29Z" level=info msg="starting call" action="server.handleFunctionCall)-fm" app=zoo container_id=01C860Z3M9A7WHJ00000000000 id=01C860Z3M9A7WHG00000000000 route=/http
        time="2018-03-09T18:39:29Z" level=debug msg="docker unpause" app=zoo call_id=01C860Z3M9A7WHJ00000000000 cpus= format=http id=01C860Z3M9A7WHJ00000000000 idle_timeout=30 image=fnproject/fn-test-utils memory=64 route=/http stack=Unfreeze
        
        time="2018-03-09T18:39:29Z" level=debug msg="2018/03/09 18:39:29 Container starting\n" app_name=zoo call_id=01C860Z3M9A7WHG00000000000 image=fnproject/fn-test-utils path=/http user_log=true
        time="2018-03-09T18:39:29Z" level=debug msg="2018/03/09 18:39:29 BeginOfLogs\n" app_name=zoo call_id=01C860Z3M9A7WHG00000000000 image=fnproject/fn-test-utils path=/http user_log=true
        time="2018-03-09T18:39:29Z" level=debug msg="2018/03/09 18:39:29 Received format http\n" app_name=zoo call_id=01C860Z3M9A7WHG00000000000 image=fnproject/fn-test-utils path=/http user_log=true
[...]

PR #840, opened 2018-03-10
add jaeger support, link hot container & req span 

rdallman:
* adds jaeger support now with FN_JAEGER_URL, there's a simple tutorial in the
operating/metrics.md file now and it's pretty easy to get up and running.
* links a hot request span to a hot container span. when we change this to
sample at a lower ratio we'll need to finagle the hot container span to always
sample or something, otherwise we'll hide that info. at least, since we're
sampling at 100% for now if this is flipped on, can see freeze/unfreeze etc.
if they hit. this is useful for debugging. note that zipkin's exporter does
not follow the link at all, hence jaeger... and they're backed by the Cloud
Empire now (CNCF) so we'll probably use it anyway.

screen shot 2018-03-09 at 6 06 01 pm

p.s. idk why it takes 10ms to poop out a json, that's next... :)
p.p.s. the things don't seem to link up, even if we have the link, gonna go debug with opencensus people but at least we can add the link. tried to make you a gift for #836 but i have failed you @skinowski :(
p.p.p.s. either order with #837 i'll rebase the other up


skinowski:
@rdallman ping.. merge conflict...

rdallman:
updated, thanks again!

kocic11:
On a side note, the operating/metrics.md tutorial uses wrong parameter (no FN_) and should be updated.

PR #841, opened 2018-03-10
fn: non-blocking resource tracker and notification

skinowski:
For some types of errors, we might want to notify
the actual caller if the error is directly 1-1 tied
to that request. If hotLauncher is triggered with
signaller, then here we send a back communication
error notification channel. This is passed to
checkLaunch to send back synchronous responses
to the caller that initiated this hot container
launch.

This is useful if we want to run the agent in
quick fail mode, where instead of waiting for
CPU/Mem to become available, we prefer to fail
quick in order not to hold up the caller.
To support this, non-blocking resource tracker
option/functions are now available.

The solution is not perfect as signaller channels
may be lost and it might take additional signal attempts
for that call (each costs 200msec) to get this response.

### DIFF
--- a/api/agent/resource.go
+++ b/api/agent/resource.go
@@ -23,6 +23,8 @@ const (
 	Mem1GB = 1024 * 1024 * 1024
 )
 
+var CapacityFull = errors.New("max capacity reached")
+
 // A simple resource (memory, cpu, disk, etc.) tracker for scheduling.
 // TODO: add cpu, disk, network IO for future
 type ResourceTracker interface {
@@ -34,8 +36,10 @@ type ResourceTracker interface {
 	// the channel will never receive anything. If it is not possible to fulfill this resource, the channel
 	// will never receive anything (use IsResourcePossible). If a resource token is available for the provided
 	// resource parameters, it will otherwise be sent once on the returned channel. The channel is never closed.
-	// Memory is expected to be provided in MB units.
-	GetResourceToken(ctx context.Context, memory, cpuQuota uint64, isAsync bool) <-chan ResourceToken
+	// if isNB is set, resource check is done and error token is returned without blocking.
+	// if isAsync is set, resource allocation specific for async requests is considered. (eg. always allow
+	// a sync only reserve area) Memory is expected to be provided in MB units.
+	GetResourceToken(ctx context.Context, memory, cpuQuota uint64, isAsync, isNB bool) <-chan ResourceToken
 
 	// IsResourcePossible returns whether it's possible to fulfill the requested resources on this
 	// machine. It must be called before GetResourceToken or GetResourceToken may hang.
@@ -88,11 +92,17 @@ func NewResourceTracker(cfg *AgentConfig) ResourceTracker {
 type ResourceToken interface {
 	// Close must be called by any thread that receives a token.
 	io.Closer
+	Error() error
 }
 
 type resourceToken struct {
 	once      sync.Once
+	err       error
 	decrement func()
+}
+
+func (t *resourceToken) Error() error {
+	return t.err
 }
 
 func (t *resourceToken) Close() error {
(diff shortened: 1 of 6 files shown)


rdallman:
clarification of behavior shown: the only case we want to immediately bail and not block is on resource tokens, right? not if create container gets hung up, docker pull, etc., right?

skinowski:
yes, this is only for cpu & mem allocation, AKA do not wait on cpu & mem.

rdallman:
thinking out loud: if we set the slot wait time to something pretty small, like 1ms, we would get most of the benefits here without a lot of plumbing, and without 1ms (could provide some back pressure) -- in addition it would cover cases like rogue scheduling onto a new server would quickly fail and hopefully get retried on a server with the image, with the bonus that the slot launcher on the new guy might launch a container for the next go round. i guess the hot launcher is kind of unpredictable since it's polling and will most likely (199/200 (ms) chance) sample for any given call when it's not queued (since it times out after 1ms). 

part of me thinks some kind of 2 phase 'just launch a hot container' [w/o a call] & 'run this call or timeout immediately' is in line with what we want, maybe this is the first piece of that, but it seems like we really want to either get a slot or bail which is higher up (little plumbing required). not sure, really. of course we don't have a 'just launch a hot container' button atm, but it wouldn't be so hard to add, and with our ejection it can manage scaling down itself. then the router can just run around and (with some idea of resource saturation itself, which will be required) tell 3 guys to launch a container and then try to round robin through them with a call until one can run it. thinking out loud (and prob in the wrong place)

skinowski:
I think two approaches are complementary. Shrinking slot wait is a good way to pre-warm a fleet as long as the caller (LB) has smarts to coordinate it, but slot wait time includes a set of operations which any could trigger the timeout. On the other hand, cpu & mem isolated quick fail allows the caller to rule out capacity issues only.

I think when this cpu & mem quick fail is enabled, 200 msec should be set to much faster pace. And probably not much performance impact due to reduced queuing.

hhexo:
Wow, you can send golang channels through a channel... I learn something new every day. :)

This looks like it will help a lot, but I'm still a bit confused about all the moving parts and how they play with the container creation logic. As far as I understand the producer is monitoring the stats (queued, running etc) and makes a decision to start a new container based on those; that decision seems to take into account containers that are waiting for resources to be started.

```
startWaiters := cur.containerStates[ContainerStateWait]
...
// we expect resource waiters to eventually transition
// into starters.
effectiveStarters := starters + startWaiters
```
(in `isNewContainerNeeded()`)

Is there a possibility that while a container is waiting on the resource token channel it still figures in this calculation (because until it receives the capacity error it still looks in a `ContainerStateWait`), so another call could think that it can be queued / accepted because "there's a container which will eventually start"? Is such a race between calls possible?

skinowski:
@hhexo Yes, so, at the time of this decision, there's cases where momentarily have excess idle containers, waiters, etc. which might cause agent to conclude that no new container is needed. But this should be remedied by polling from clients. For this operating mode, we should raise the polling to something faster eg. 20 msecs. In other words, isNewContainerNeeded() is allowed to make a mistake, but eventually correct it.

skinowski:
related #853

skinowski:
Ping...

I still think this is a valuable option. To be able to disable waiting on CPU/Memory. Isolating these resources provides a good way to run a function, and waiting knowing that you are waiting because
the agent will most likely be able to run this without exceeding cpu & memory capacity of that server.

Allows a simple LB layer to probe/retry agents without any significant blocking on CPU & Memory.

api/agent/resource.go
```diff
@@ -140,10 +150,85 @@ func (a *resourceTracker) GetResourceTokenWaiterCount() uint64 {
...
+	a.cpuSyncUsed += syncCPU
+
+	return &resourceToken{decrement: func() {
+		a.cond.L.Lock()
+		a.ramAsyncUsed -= asyncMem
+		a.ramSyncUsed -= syncMem
+		a.cpuAsyncUsed -= asyncCPU
+		a.cpuSyncUsed -= syncCPU
+		a.cond.L.Unlock()
+	}}
```
rdallman: we need to broadcast from this too, even if we get the resource in NB fashion, somebody could be waiting in blocking fashion by specifying the flag and we should wake them up since this is a delta (i know it's not set up like this externally, it's on or off, but it's a bug waiting to happen and not hard to close).

`allocResourcesLocked` imo should just return the resourceToken and handle the ramAsync/ramSync/cpuAsync/cpuSync bits too, they're in multiple places now but it's the same logic, and we need to have the broadcast here as well imo so it would be identical.

skinowski:
Added some comments (and a small bug fix in there...)

rdallman:
gonna merge this since your github has been compromised :)

if i regret this, i only have myself to blame!

PR #843, opened 2018-03-12
fn: timeouts and container exits should stop slot queueing

skinowski:
1) in theory it may be possible for an exited container to
requeue a slot, close this gap by always setting fatal error
for a slot if a container has exited.
2) when a client request times out or cancelled (client
disconnect, etc.) the slot should not be allowed to be
requeued and container should terminate to avoid accidental
mixing of previous response into next.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -508,6 +508,12 @@ func (s *hotSlot) Error() error {
 	return s.fatalErr
 }
 
+func (s *hotSlot) trySetError(err error) {
+	if s.fatalErr == nil {
+		s.fatalErr = err
+	}
+}
+
 func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	ctx, span := trace.StartSpan(ctx, "agent_hot_exec")
 	defer span.End()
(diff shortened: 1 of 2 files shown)


skinowski:
In this case, timeout/runner tests did not actually catch the more serious; second issue, since there was no cross check to see if response-request actually did match.

This is hard to detect for client containers as well as fn agent and the safest option to keep processing sane and isolated between requests is to terminate the container.

First issue is client http request cancellations, (connection close/rst), will be missed by the client container.

For regular timeouts, client code could manage this by tracking deadline, but this is hard as the container is frozen as soon as the client has timed out. And also timing/scheduler could impact the order of events, since fn agent can't always reliably guarantee that fn agent will always expire deadline *after* container notices that it's expired and takes action against it. Even if container tracks deadline successfully, it cannot emit a proper response since fn agent will also hit deadline soon after and might not have the chance to read the response. In addition, immediately after deadline expiry fn agent will consider the container ready for the next request, which adds additional burden on client containers since the timeout/deadline expiry may require logging to external service (or other additional processing.) 

The issue also complicates billing & tracking of processing, and 1-1 relationship between client connection context to container execution is easier than other feasible options (such as letting the current request finish/run in the container, which would exceed the allowed timeout.)

api/agent/agent.go
```diff
@@ -540,19 +546,21 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 
 	select {
 	case err := <-s.errC: // error from container
+		s.trySetError(err)
```
rdallman: this one seems redundant I guess since we're shutting the container slot goroutine down once the container exits ? tiny race b/w doing that I suppose ? this is ok, just curious

rdallman:
aside: it seems worth the time investment to rejig the slot acquisition and associated timings to be more easy to reason about their correctness. at present we're patching races. it may take a lot of scribbles and blood, but imo worth it.

api/agent/agent.go
skinowski: This is just for the sake of completeness. But in theory, the main runhot go routine may pause after queuing the errC before calling shutdown cancel function. During this time, inner slot processing go routine may have the chance to requeue the slot. But requeue can happen after slot close and there's a lot of code in submit for this to happen, so it's super unlikely.

PR #849, opened 2018-03-13
Make dataplane system test behave deterministically

hhexo:
**PATCH 2/2 - requires #834** 

This change ensures that the dataplane system test behaves the same way on all machines, by injecting fixed capacity constraints. This also makes it always use all three Pure Runners that it creates, proving that a multi-runner environment works.


api/agent/pure_runner.go
```diff
@@ -202,24 +213,24 @@ type pureRunnerCapacityManager struct {
...
 		committedCapacityUnits: 0,
 	}
 }
 
-func (prcm *pureRunnerCapacityManager) checkAndReserveCapacity(units uint64) error {
+func (prcm *pureRunnerCapacityManager) CheckAndReserveCapacity(units uint64) error {
 	prcm.mtx.Lock()
 	defer prcm.mtx.Unlock()
-	if prcm.committedCapacityUnits+units < prcm.totalCapacityUnits {
+	if prcm.committedCapacityUnits+units <= prcm.totalCapacityUnits {
```
skinowski: very minor... not likely with sane capacity numbers, but I'd revert the math here to avoid any overflow. eg;

```
if total - committed >= units {
```

PR #853, opened 2018-03-13
fn: reorg agent config

skinowski:
*) Moving constants in agent to agent config, which helps
with testing, tuning.
*) Added max total cpu & memory for testing & clamping max
mem & cpu usage if needed.

### DIFF
--- a/api/agent/config.go
+++ b/api/agent/config.go
@@ -61,9 +61,9 @@ func NewAgentConfig() (*AgentConfig, error) {
 	if cfg.EjectIdle == time.Duration(0) {
 		return cfg, fmt.Errorf("error %s cannot be zero", EnvEjectIdle)
 	}
-	if cfg.MaxLogSize > math.MaxInt32 {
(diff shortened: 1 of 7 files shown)


api/agent/config.go
```diff
@@ -10,75 +9,99 @@ import (
...
 
-	if tmp := os.Getenv("FN_MAX_LOG_SIZE"); tmp != "" {
-		cfg.MaxLogSize, err = strconv.ParseUint(tmp, 10, 64)
-		if err != nil {
-			return cfg, errors.New("error initializing max log size")
-		}
+	if cfg.EjectIdle == time.Duration(0) {
+		return cfg, fmt.Errorf("error %s cannot be zero", EnvEjectIdle)
+	}
+	if cfg.MaxLogSize > math.MaxInt32 {
```
rdallman: not important but sticks out: this isn't out of the question for logs, all things considered. 2GB signed, right? async can run 60 minutes, we should just make it int64?

PR #856, opened 2018-03-14
fn: update fn dependencies for go 1.10

skinowski:


skinowski:
Not sure if this is a bit premature, but here we go... Go 1.10...

PR #857, opened 2018-03-14
Updated Tutorials link

sachin-pikle:
Changed from fnproject.com/tutorials to fnproject.io/tutorials

The current one doesn't work so correcting the link. The current version of tutorials landing page at fnproject.io/tutorials is better than an error. :-)


carimura:
damn that was me wasn't it? I broke every rule in the book... I promise to not do it again!!

carimura:
thx for fixing

PR #860, opened 2018-03-14
fn: add storage opt size support

skinowski:
Added env FN_MAX_FS_SIZE_MB, which if defined and non-zero
is passed to docker as storage opt size. We do not validate
if this option is supported by docker currently. This is
because it's difficult to actually validate this since it
not only depends on storage driver and its backing filesystem,
but also the mount options used to mount that fs.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -597,6 +597,7 @@ func (a *agent) prepCold(ctx context.Context, call *call, tok ResourceToken, ch 
 		env:     map[string]string(call.Config),
 		memory:  call.Memory,
 		cpus:    uint64(call.CPUs),
+		fsSize:  a.cfg.MaxFsSize,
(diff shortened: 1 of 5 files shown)


skinowski:
related with #747

skinowski:
Related PR #747 had a check for this, which we could add here as well, but the check success
does not really mean size limit will really work. Given this, it might be better to leave this
to docker error handling at this point, not ideal, but I'm open to ideas:

https://github.com/fnproject/fn/pull/747/files#diff-70418a005cf583c9db18576bc95c5e68R214

rdallman:
> Related PR #747 had a check for this, which we could add here as well, but the check success
does not really mean size limit will really work. Given this, it might be better to leave this
to docker error handling at this point, not ideal, but I'm open to ideas:

that's fine I think. can we add a ticket to at least test this behavior if not add a double check? it may fail silently and act like it's imposing a limit when it's not, which could be problematic in certain deployments.

PR #864, opened 2018-03-15
fn: initial "network" option to routes

skinowski:
Basic "network" option introduction to routes. Initially,
this can be set to "default" or "disabled" to disable container
networking.

In preparation for more advanced future network options.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -598,6 +598,7 @@ func (a *agent) prepCold(ctx context.Context, call *call, tok ResourceToken, ch 
 		memory:  call.Memory,
 		cpus:    uint64(call.CPUs),
 		fsSize:  a.cfg.MaxFsSize,
+		network: call.Network,
 		timeout: time.Duration(call.Timeout) * time.Second, // this is unnecessary, but in case removal fails...
 		stdin:   call.req.Body,
 		stdout:  common.NewClampWriter(call.w, a.cfg.MaxResponseSize, models.ErrFunctionResponseTooBig),
@@ -794,6 +795,7 @@ type container struct {
 	memory  uint64
 	cpus    uint64
 	fsSize  uint64
+	network string
 	timeout time.Duration // cold only (superfluous, but in case)
 
 	stdin  io.Reader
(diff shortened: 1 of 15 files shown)


denismakogon:
So, the use case for disabled is clear to me. But what “default” network means? I guess it is a bit confusing. Does the “network” implies to the actual docker network where the Fn started? Does the network equal to vNIC?

If the network attribute only stand for an ability of the function to have an access, let’s say, to the internet should we name that as “networking_disabled”, “offile_network”?

skinowski:
In reality, default translates to docker bridge network. I thought about putting "private", "bridge", etc. in there, but I'm not sure, I'm open to ideas here.. Paging @rdallman :-)

denismakogon:
Okay, so the docker bridged network might be the Fn server config option. Here’s real-life use case - start compose deployment, make function talk to the Fn API within current compose deployment. It would basically fail to do so, because Fn containers without proprer network routing configuration.

If you want to disable the networking - that’s fine, but it’s completely different thing from what this PR brings.

So, we need to things: 
- Fn server configuration opt that stands for the network where Fn should start the containers
- routes API attribute to enabl/disable the networking.

skinowski:
@denismakogon So today, with this PR, we'll have two options;

1) as-before (default) - which is bridge-NAT
2) no network - disable network in the containers

per route.

If network is disabled for the route, those containers indeed cannot egress/ingress via network. Only useful for workloads who want to process data/report without any network I/O.

Future options in "network" could be:

1) 'use host network' - put all containers on host-net namespace
2) choose specific network name (eg. created by docker network create)

etc. etc.

But I'm trying to keep the scope narrow in this PR to only disable vs default.

skinowski:
In other words, the only information the operator needs to know is 'do I want to disable network in my functions?' today...

rdallman:
> Also, not sure if the "" (None), "default", "disabled" is proper there...

docker's default is "default" as well:

```
      --network string                 Connect a container to a network (default "default")
```

+1 following suit. should we change "disabled" to "none" to match, too?

> 'use host network' - put all containers on host-net namespace

I'm hopeful about this one making the local dev experience a lot nicer.

I kind of agree with Denis I don't think the none networking option seems very useful (considering we have hot functions that saves most of the overhead, most functions will likely call out to the network at some point, and most people already isolate their network from potential intruders), but being able to set the networking stuff in general may prove extremely useful.

skinowski:
@rdallman removed none, etc. now, it's empty (aka default) or disabled now.

skinowski:
@rdallman @denismakogon My understanding is, your are OK with the option feature itself, but 'disable' is unnecessary, and instead 'host' (use host-net-ns) would be more helpful? (I was planning to add 'host-net' as a separate PR.)

rdallman:
> My understanding is, your are OK with the option feature itself, but 'disable' is unnecessary, and instead 'host' (use host-net-ns) would be more helpful? (I was planning to add 'host-net' as a separate PR.)

I guess my interpretation of allowing users to specify the network would be that we allow any string to be passed through, as users may configure docker networks separately, e.g. `docker run --network polliwog-overlay` (we're just doing this through http api instead of cli command here) and want to use those vs. us whitelisting some set of `{"default", "none", "host"}` and documenting our own set of options, where the former option would allow them to specify any of the latter, and I would think that when we create the container it will error if the networking option is not valid to give the user some feedback but we may have to explicitly whitelist that error as a `models.APIError`.

zootalures:
I worry that this change adds a feature that  is docker-specific (fn runs docker images but does not _have_ to run on docker) creates unachievable user expectations in a distributed system (works on my machine, doesn't work in the cloud)  - including network config in Fn opens a bunch of other issues about if, or how that network might be configured 

Given that I doubt we can or should be generalising a model of networking in Fn accross all possible environments,  I prefer the approach that very infrastructure specific things (like network and storage) are left as platform extensions.

skinowski:
Closing this. Ongoing related work in prefork pool.

PR #865, opened 2018-03-16
Make PKI data and RunnerFactory public objects

gviedma:


api/runnerpool/runner_pool.go
```diff
@@ -21,6 +21,15 @@ type RunnerPool interface {
...
 }
 
+type PKIData struct {
+	Ca   string
+	Key  string
+	Cert string
+}
+
+// SecureRunnerFactory represents a factory method for constructing runners with PKI
+type SecureRunnerFactory func(addr string, pki *PKIData) (Runner, error)
```
hhexo: While we're at it, could we rename it MTLSRunnerFactory, since certs are only used for mutual tls? A "secure" runner could be implemented in other ways.

gviedma: I've made the change.

PR #866, opened 2018-03-16
Add annotations to routes and apps

### LINKED ISSUE #789: Introduce object annotations to app and route
Goal: Extend Fn resources to support attached key/value metadata set by API users or set and controlled by extensions. 

(@zootalures's edits) 
(@zootalures - renamed "metadata" to "annotations") 
# App and Route Annotations

App and Route annotations allow users building on the Fn platform to encode consumer-specific data inline with Apps and Routes.

Annotations can be used to either communicate and carry information externally (only by API users) or to communicate between external applications and Fn extensions.

### Externally defined/consumed annotations

Software using Fn as a service,  attaches non-identifying metadata annotatinos to Fn resources for subsequent reads (e.g. my reference for this function/app )

Writer : API user, Reader: API user

E.g.  platform "platX" creates/modifies functions, has an internal reference it wants to associate with an object for later retrieval (it can't query/search Fn by this attribute)

```
POST /v1/apps

 {
 ...
  "annotations" : {
    "platx.com/ref" : "adsfasdfads"
  }
 ...
 }
```

### Extensions: Allow passing data from user or upstream input to extensions (configuration of extensions)

Writer : API user, Reader: Fn platform extension  (use), API user (informational)

```
POST /v1/apps
...
   {
     ...
       "annotations" :  {
              "my_cloud_provider.com/network_id" : "network.id"
       }
     ...
}
```

###  Extensions: Allow indicating internally derived/set values to user (API extension sets/generates annotations, prevents user from changing it)

Writer : Internal platform extension, Reader: API user.

```
GET /v1/apps/myapp
...
   {
     ...
       "annotations" :  {
              "my_cloud_provider.com/create_user" : "foo@foo.com"
       }
     ...
}
```

```
PATCH /v1/apps/myapp
...
   {
     ...
       "annotations" :  {
              "my_cloud_provider.com/create_user" : "foo@foo.com"
       }
     ...
}

HTTP/1.1  400 Invalid operation

{
   "error": "annotation key cannot be changed",
}
```

## Content Examples
example : user attaches local annotations

```json
PUT /v1/apps/foo

{
  app: {
   ...
    "annotations": {
         "mylabel": "super-cool-fn",
         "myMetaData": {
           "k1": "foo",
           "number": 5000
           "array" : [1,2,3]
         }
     }
  ...
  }
}
```

User sets extension-specific annotations:

```
PUT /v1/apps/foo
{
   ...
    "annotations": {
        "example.extension.com/v1/myval" : "val"
    }
  ...
}
```

## Key Syntax and Namespaces

A key consists of any printable (non-extended) ascii characters excluding whitespace characters.

The maximum (byte) size of a key is 128 bytes (excluding quotes).

Keys are stored as free text with the object. Normatively extensions and systems using annotations *must* use a namespace prefix based on an identified domain and followed by at least one '/' character.

Systems *should* use independent annotation keys for any value that can be changed independently.

Extensions *should not* interact with annotations keys that are not prefixed with a domain they own.

## Value syntax

Values may contain any valid JSON value (object/array/string/number) except the empty string  `""` and `null`

The serialised JSON representation (rendered without excess whitespace as a string) of a single value must not exceed a 512 bytes.

## Modifying and deleting annotation keys

A key can be modified by a PATCH operation containing a partial `annotations` object indicating the keys to update (or delete)

A key can be deleted by a PATCH operation by setting its value to an empty string.

For each element that of data that can be changed independently, you *should* use a new top-level annotation key.

## Maximum number of keys

A user may not add keys in a PATCH or PUT operation if the total number of keys after changes exceeds 100 keys.

Fn may return a larger number of keys.

## Extension interaction with resource modification

An extension  may prevent a PUT,PATCH or POST operation on a domain object based on the value of an annotation passed in by a user, in this case this should result in an HTTP  400 error with an informational message indicating that an error was present in the annotations and containing the exact key  or keys which caused the error.

## Datastore impact

Type: text
Nullable by default

The maximum stored size for a given object (when stored serialized as a single JSON object) is : 
2 + (max_internal_keys  * (6 + max_key_length  + max_value_length)) 
or about 74k.

zootalures:
closes #789


api/models/metadata.go
```diff
@@ -0,0 +1,230 @@
...
+
+// metadataValue encapsulates a metadata value in the metadata map,
+// This is stored in its un-parsed JSON format for later (re-) parsing into specific structs or values
+// metadataValue objects are  immutable after JSON load
+type metadataValue struct {
+	val []byte
+}
+
+const (
+	maxMetadataValueBytes = 512
```
ericfode: Is this a reasonable limit if the metadata structure is nested? or it it always supposed to be flat?

api/models/metadata_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+var mdCases = []struct {
+	val       *testObj
+	valString string
+}{
+	{val: &testObj{Md: EmptyMetadata()}, valString: "{}"},
+	{val: &testObj{Md: EmptyMetadata().withRawKey("stringval", "\"bar\"")}, valString: "{\"metadata\":{\"stringval\":\"bar\"}}"},
+	{val: &testObj{Md: EmptyMetadata().withRawKey("intval", "1001")}, valString: "{\"metadata\":{\"intval\":1001}}"},
+	{val: &testObj{Md: EmptyMetadata().withRawKey("floatval", "3.141")}, valString: "{\"metadata\":{\"floatval\":3.141}}"},
+	{val: &testObj{Md: EmptyMetadata().withRawKey("objval", "{\"foo\":\"fooval\",\"bar\":\"barval\"}")}, valString: "{\"metadata\":{\"objval\":{\"foo\":\"fooval\",\"bar\":\"barval\"}}}"},
```
ericfode: should 
```
{val: &testObj{Md: EmptyMetadata().withRawKey("objval", "{\"foo\":\"fooval\",\"bar\":{\"barbar\":\"barbarval\"}")}, valString: "{\"metadata\":{\"objval\":{\"foo\":\"fooval\",\"bar\":{\"barbar\":\"barbarval\"}}}}"}
```
work?

api/models/metadata.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	if err != nil {
+		return err
+	}
+	v.val = buf.Bytes()
+	return nil
+}
+
+func validateField(key string, value *metadataValue) APIError {
+
```
ericfode: Should non-ascii characters/non printable chars/whitespace be allowed?

api/models/metadata.go
zootalures: it's a reasonable limit in that it keeps max metadata below 100k which was my target.

api/models/metadata_test.go
zootalures: I think so   - what looks wrong?

api/models/metadata.go
zootalures: yeah , good point - it's probabably safer to restrict than to allow here. 

I originally supposed that keys were "any utf8 string up to 128 bytes" (#789 )  -

api/models/metadata_test.go
zootalures: ah got it...

zootalures: You are missing a bracket, but yes it does work

zootalures: metadata values are "any json value except the empty string or null"  - including nested options

api/models/call.go
```diff
@@ -144,6 +144,12 @@ type Call struct {
 	// Error is the reason why the call failed, it is only non-empty if
 	// status is equal to "error".
 	Error string `json:"error,omitempty" db:"error"`
+
+	// AppMetadata contains a copy of the app metadata
+	AppMetadata Metadata `json:"app_metadata,omitempty" db:"app_metadata"`
+
+	// RouteMetadata contains a copy of the app metadata
+	RouteMetadata Metadata `json:"route_metadata,omitempty" db:"route_metadata"`
```
rdallman: not sure about persisting these in the call object, this struct is something of a mess that needs sorted out. the `db:` fields are to return from the API to the user for viewing, at present this is a very small number of fields that tend to change from call to call (timestamps, stats, status) vs. configuration which is present on a route, which are easily looked up on the route (and in the somewhat near future, these will be versioned and easier to audit older versions of). the `json:` fields double to 1) show any fields in the API to the user that also have a non-`db:"-"` and 2) to store all call configuration in an mq message for async.  in any event, adding these to each call in the db bloats the bytes we store per call, each call specifies which route it's from, and metadata will be exposed on the app and route there.

api/models/metadata.go
```diff
@@ -0,0 +1,230 @@
...
+//  // old MD remains empty
+// Metadata is lenable
+type Metadata map[string]*metadataValue
+
+// metadataValue encapsulates a metadata value in the metadata map,
+// This is stored in its un-parsed JSON format for later (re-) parsing into specific structs or values
+// metadataValue objects are  immutable after JSON load
+type metadataValue struct {
+	val []byte
+}
```
rdallman: a type alias here seems suitable, too (`type metadataValue []byte`) to add methods on top of. not sure if there's plans to add other fields; not very important.

api/models/metadata.go
```diff
@@ -0,0 +1,230 @@
...
+	"errors"
+)
+
+// Metadata encapsulates key-value metadata. The structure is immutable via its public API and nil-safe for its contract
+// permissive nilability is here to simplify updates and reduce the need for nil handling in extensions - metadata should be updated by over-writing the original object:
+//  oldMd := EmptyMetadata()
+//  newMd := oldMd.With("fooKey",1)
+//  // old MD remains empty
+// Metadata is lenable
+type Metadata map[string]*metadataValue
```
rdallman: slightly curious whether doing `map[string]interface{}` makes our lives any easier, it seems essentially to be what we want but with some logic to validate key / value sizes on top of that. if so, I assume it'd mean since we always ingest these from json through the http api (for now), we simply need to implement the key / value validations inside of the `UnmarshalJSON` logic, and this would give us slightly more control over that validation (at present, only values have a custom UnmarshalJSON, which reads the whole thing in before validation). haven't played with this concretely, I think the keys are easy (seek until `"` or n > maxKeyBytes), but the values we'd need to wrap a limit bytes reader around and unmarshal an `interface{}`, which is perhaps a little easier than reading into a bytes buffer.

as it stands, the code here to validate is vulnerable to e.g. a user sending in a 5GB key or value will allocate that space (short of any http level protections, which exist, but still likely exceed the maximums imposed here) before we validate that it is too large. I am not sure if this is an explicit goal to block this behavior.

up to you, I think it'd be possible to play with later since this is essentially `map[string]interface{}` we can change the implementation but the data would stay the same.

rdallman:
seems sane to me, some minor comments

api/models/metadata.go
zootalures: Protecting the heap wasn't wasn't the goal here, the the size validation here is more to protect the DB.  Given we aren't taking unbounded input data anywhere (I think)  I'm pretty sure a reasonable cap on ingress size in middleware and not making request size validation a concern of the business logic is the right way to go here. 

For read I want extensions  consuming metadata  to be able to bind an object/array to whatever they want (either naive map[string]interface{} or to their own structs/slices) - so forcing the unmarshalled format to the former seemed inconvenience (people end up marshalling and then unmarshalling again) . 

For write I wanted to apply validation at the point of setting  including value length checks (i.e. its an error in the extension if it sets an invalid MD key) , which entails marshalling so it seemed to make sense to marshal early here.

api/models/metadata.go
zootalures: i didn't try it but I assumed that I wouldn't be able to unmarshal a value in the same way here (can't assign to the receiver in UnmarshalJSON)  (could just be go n00bz for me though will give that ago).

zootalures:
Have dealt with most of the comments  I think - 

I've ended up doing a bit of refactoring in the API tests to try and simplify the test harness a bit, and remove the "global delete everything" step in favour of requiring a `defer s.Cleanup()` in each test.  There is more to do there. 

I've ended up puting the bulk of the e2e testing in the API tests  - getting veracity on validation is hard anywhere else (which is fine, I think). 

The SDK dep for testing is a) inconvenient as you have to round trip  fn_go  to test new API features, and also a bit fluffy w.r.t. API semantics - would prefer a pure JSON model client in the API tests for correctness, might look at this in future. 

I'm knowingly not validating pure deltas on patch which (validate irrespective of current values) - this  allows poisoned /bad existing values in the DB to block all updates.   Thought about this but I think this entails some more contract "ValidateChange(patch)" sort of thing on the data objects which i stopped at.

docs/swagger.yml
```diff
@@ -519,6 +519,11 @@ definitions:
         default: 30
         format: int32
         description: Hot functions idle timeout before termination. Value in Seconds
+      metadata:
+        type: object
```
denismakogon: i suppose this should be a `string` but not the `object`. With `string` type you can limit the size of the object that would be passed to the generated client. https://swagger.io/docs/specification/data-models/data-types/#string

docs/swagger.yml
```diff
@@ -539,9 +544,14 @@ definitions:
         readOnly: true
       config:
         type: object
-        description: Application configuration, applied to all routes.
+        description: Application function configuration, applied to all routes.
         additionalProperties:
           type: string
+      metadata:
+        type: object
```
denismakogon: same here

test/fn-api-tests/calls_test.go
```diff
@@ -42,15 +44,15 @@ func TestCallsDummy(t *testing.T) {
 		t.Error("Must fail because `dummy` call does not exist.")
 	}
 
-	DeleteApp(t, s.Context, s.Client, s.AppName)
```
denismakogon: The reason to have app provisioning and deleting on every test was to make tests independent, to make them be capable to be executed completely independently from any other test. However, this change brings the certain inconsistency between tests, so they are dependent.

zootalures: A agree with the above -  that's why I added `defer s.Cleanup()` -  now deletes any apps that were created within the test (even on a panic) - so tests are independent. .

docs/swagger.yml
zootalures: I want to support JSON natively on the API  and object works pretty nice with most rendered swagger clients (`map[string]interface{}`) in go, `Map<String,Object>` in java etc. 

The length constraints apply to individual keys and values  so having this as a string and forcing clients to re-encode/Decode to JSON seemed unnecessarily painful.

test/fn-api-tests/calls_test.go
zootalures: I've rejigged the purpose of `SetupSuite` a bit - (hence rename to `TestHarness`) - it's now a per-test case harness

docs/swagger.yml
zootalures: (I don't see a strong reason to change this)

test/fn-api-tests/calls_test.go
zootalures: Noticed an issue where I wasn't properly cleaning up apps in this case, have fixed.

api/models/metadata.go
rdallman: > For read I want extensions consuming metadata to be able to bind an object/array to whatever they want (either naive map[string]interface{} or to their own structs/slices) - so forcing the unmarshalled format to the former seemed inconvenience (people end up marshalling and then unmarshalling again) .

that is a good point. in that case `interface{}` (no map) seems easiest, even, but it's weird to work with for storage / json. bytes seems a fine decision.

api/models/metadata.go
```diff
@@ -0,0 +1,236 @@
...
+
+func (mv *metadataValue) isEmptyValue() bool {
+	sval := string(*mv)
+	return sval == "\"\"" || sval == "null"
+}
+
+// UnmarshalJSON compacts metadata values but does not alter key-ordering for keys
+func (mv *metadataValue) UnmarshalJSON(val []byte) error {
+	buf := bytes.Buffer{}
+	err := json.Compact(&buf, val)
```
rdallman: i'm not sure we should do this, specifically this way at least. nothing against the minification movement, it seems nice in theory, but it may change input for configs where spacing matters (yaml) and that kind of data seems a likely possibility in this field :( - the easy thing to do seems to just be validate `val` is utf8 and `*mv = val` ?

docs/swagger.yml
rdallman: object seems fine to me, too

rdallman:
one niggle, otherwise LGTM

api/models/metadata.go
zootalures: Hmm maybe I’ve got something wrong here, 

I assumed that for unmarshal here the byte slice is the json sub-doc corresponding to the value (in the metadata map) (i.e quoted string, array object, number), and that this byte slice is a sub-slice of the original input buffer (from the user)

My reading of the json.Compact docs was that it preserves the content model   (property ordering, string content),and that, given the metadata map is defined as JSON in the request, that whitespace was ignorable

So, the compact here is just to remove any superfluous whitespav which might be present in the metadata sub-doc from API requests , in order to get a consistent value for the (JSON) value size, I don’t think there is any way that it changes the meaning of metadata values themselves

rdallman: it at least appears to skip over `\r\n` https://golang.org/src/encoding/json/indent.go?s=284:333#L1 -- I could be wrong, didn't thoroughly read their json parser code. I tinkered with the playground for a minute and couldn't get an example to not error, either I'm completely missing what a metadata value is or json compact isn't gonna work so great for any values that have a newline: https://play.golang.org/p/U7vqPeoob0Q

api/models/metadata.go
```diff
@@ -0,0 +1,236 @@
...
+		return err
+	}
+	if err != nil {
+		return err
+	}
+	*mv = buf.Bytes()
+	return nil
+}
+
+var validKeyRegex = regexp.MustCompile("^[!-~]+$")
```
skinowski: This is not a utf-8 regex. (https://github.com/fnproject/fn/issues/789 says keys are utf-8.)

zootalures: Yep - on feedback from @ericfode  - I dropped this to printable contiguous ascii (no whitesapce)  - it's negociable but being restrictive (initially at least) seemed the safer thing to to do here.

api/models/metadata.go
zootalures: json.Compact never touches values, only interstitial whitespace between JSON tokens - the specific issue here would be :

User makes API request using indented JSON (not that they likely ever would) - the whole API request is indented: 

```
POST /v1/apps 

{
     "app" { 
           ...
          "metadata" : {
                       "md.key.1": {  
                                   "a": v1 \n with \n newlines    and arbitrary spacing ", 

                                   "b" : 10
                        }, 
                       "md.key.2": "v2"
           }
          ....
}
```

In this case the extracted Metadata doc for key "md.key.1" (passed to metadataValue.UnmarshalJSON) 
```
 {  
                                   "a": v1 \n with \n newlines    and arbitrary spacing ", 

                                   "b" : 10
                        }
```

If I removed the `json.Compact` that would be what we stored in the DB  verbatim, and that would then be was returned in subsequent API responses containing that metadata. 

I'm assuminging that (from a user's perspetive) any interstitial whitespace within the metadata values  is non-semantic (as it is on the API request that sets the metadata itself)

Stripping the whitespace within the value sub-doc avoids any confusion about storage sizes  (an indented request and a semantically identical compact request get the same size limits), and also consolidates equality (changing the spacing of your API requests doesn't change equality w.r.t updates). 

Equality is of course a little bit tricky here due to key order preservation, I opted for preserving the original key ordering within values (hence keeping the values as unparsed byte slices) as I thought that was least-surprise for consumers.


If I'd stored metadata as  `map[string]interface{}` the  compaction would happen via the internal unmarshal/marshal in the same way as with json.Compact, however Equality would then be order-independent for value maps and more importantly extensions would need to render that value interface{} back int to a byte array in order to then re-parse it into their own structs. 

I added a quick equality check to validate the correct behaviour. 

`{mustParseMd(t,

			`{
 "md.1" : {
           
   "subkey1": "value\n with\n newlines",
   "subkey2": true 

}
}`), mustParseMd(t,`{"md.1":{"subkey1":"value\n with\n newlines","subkey2":true}}`), true},`

zootalures:
Just spoke to @tteggel  and @shaunsmith  - they prefer "annotations" here - vs "metadata" - I think I agree - will make that change now

zootalures:
this should be merged along-side https://github.com/fnproject/fn_go/pull/6 (I've naively pushed that version into vendor for builds sake in this PR)

PR #868, opened 2018-03-18
Fix to broken CLI link.

shaunsmith:
Minor fix to a bug reported by an end user--corrected the relative path to the main CLI README.md file

### DIFF
--- a/docs/developers/packaging.md
+++ b/docs/developers/packaging.md
@@ -2,7 +2,7 @@
 
 ## Option 1 (recommended): Use the `fn` cli tool
 
-We recommend using the [fn cli tool](../fn/README.md) which will handle all of this for you. But if you'd like to dig in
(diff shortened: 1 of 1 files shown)


docs/developers/packaging.md
```diff
@@ -2,7 +2,7 @@
 
 ## Option 1 (recommended): Use the `fn` cli tool
 
-We recommend using the [fn cli tool](../fn/README.md) which will handle all of this for you. But if you'd like to dig in
+We recommend using the [fn cli tool](../../../cli/README.md) which will handle all of this for you. But if you'd like to dig in
```
denismakogon: Shouldn't we use relative paths? Why not just the github link?

sachin-pikle: The CLI installation is documented in two places - (1) Fn/Readme.md and (2) cli/Readme.md . If we use (1) relative path is fine. If we use (2) we should specify the absolute URL. I have updated the file to point to (1).

PR #874, opened 2018-03-21
fn: pre-fork pool for namespace/network speedup

skinowski:
For docker containers, it's possible to provide a new container with an existing namespace using "container:name" format. For example, for pass-in an existing network namespace to a new container:

```
docker run --name foo busybox sleep 600
docker run --name bar --network "container:foo" busybox echo hello
```
Which runs "bar" but using "foo"s network and user namespace.

This PR introduces a pool of asleep (dummy) containers that merely serve as a cache for network and user namespaces, similar to the 'foo' container in the example above.

This results in significant speed up in cold containers as well as burst (ramp up) handling of hot containers.

### DIFF
--- a/api/agent/drivers/docker/docker_pool.go
+++ b/api/agent/drivers/docker/docker_pool.go
@@ -1,0 +1,251 @@
+package docker
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"runtime"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/fnproject/fn/api/agent/drivers"
+	"github.com/fnproject/fn/api/common"
+	"github.com/fnproject/fn/api/id"
+
+	"github.com/fsouza/go-dockerclient"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/time/rate"
+)
+
+// Prefork Pool is used in namespace optimizations to avoid creating and
+// tearing down namespaces with every function container run. Instead, we reuse
+// already available namespaces from these running container instances. These
+// containers are not designed to run anything but placeholders for namespaces
+// such as a minimal busybox container with 'tail -f /dev/null' which blocks
+// forever. In other words, every function container is paired with a pool buddy
+// where pool buddy provides already creates namespaces. These are currently
+// network and user namespaces, but perhaps can be extended to also use pid and ipc.
+// (see docker.go Prepare() on how this is currently being used.)
+// Currently the pool is a set size and it does not grow on demand.
+
+var (
+	ErrorPoolEmpty = errors.New("docker pre fork pool empty")
+)
+
+const (
+	LimitPerSec = 10
+	LimitBurst  = 20
+
+	DefaultImage = "busybox"
+	DefaultCmd   = "tail -f /dev/null"
+)
+
+type poolTask struct {
+	id    string
+	image string
+	cmd   string
+}
+
+func (c *poolTask) Id() string                                       { return c.id }
+func (c *poolTask) Command() string                                  { return c.cmd }
+func (c *poolTask) Input() io.Reader                                 { return nil }
+func (c *poolTask) Logger() (io.Writer, io.Writer)                   { return nil, nil }
+func (c *poolTask) Volumes() [][2]string                             { return nil }
+func (c *poolTask) WorkDir() string                                  { return "" }
+func (c *poolTask) Close()                                           {}
+func (c *poolTask) Image() string                                    { return c.image }
+func (c *poolTask) Timeout() time.Duration                           { return 0 }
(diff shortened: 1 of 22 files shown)


rdallman:
${strange generic question around security}? since we're using a pool of generic containers and sharing them between different kinds of functions, there's a chance the function B gets some sugar from function A's network queue? anything else? we've mitigated this issue by other means generally in certain deployments, maybe just curiosity here.

zootalures:
Curious as to the net speedup here - To @rdallman 's point I wouldn't have considered doing this unless it was significant.

skinowski:
We can iterate on this going forward, but I prefer to merge a simple version first. Then we can open the can of worms and discuss isolation levels. For example, we could use the same net-ns across the same hot container function or not, purge them & backfill for each old/hot, etc.

zootalures:
If there are any security concerns (even if it’s flaggable) id propose not merging until they are dealt with

skinowski:
It's not enabled by default, so we could document the risks in sharing network namespaces and leave it to the operators to decide to opt in or not.

rdallman:
i think in the _ahem_ primary deployment model it doesn't add any security risks, in any event. I am sure we'd be in close contact with anyone wishing to deploy it as a multi tenant service otherwise, but it's just good to write stuff like this down somewhere so it's not just tribal knowledge. leaving it as a flag / experimental feature seems fine, was mostly curious about any threats you may have thought about. I regret saying the 's' word, sorry.

zootalures:
mkay :)  - seems fair (I'm mr no cake today)  - if you are both confident it doesn't add insecurity that's fine with me. 

FWIW I'd be  more confident if the NS containers were discarded once done  (use Cgroups as a backstop, fewer risks for  our bugs)

rdallman:
> Could you evict the network containers once you are done and backfill?

this would work, to get 'one time use' containers and in theory our stasis has relatively low container churn so the overhead wouldn't be that bad. in a worst-case scenario where every container is cold, it adds 2x overhead of create/destroy which in my xp is a lot of the cost and drastically spikes i/o, cpu, lengthens task run time, etc.

skinowski:
We can work on all these: using more bridges or one-bridge per container, discard & backfill, etc. etc. 

I'd say we keep the initial version very simple for now. We can evaluate and measure how evictions would perform for hot/cold and in various scenarios... And again not everyone needs that level of isolation and ROI on this pre-pool with backfill might not be worth it at the end especially for hot.

zootalures:
On minor points: 

K8s uses `pause` (https://github.com/kubernetes/kubernetes/blob/2f1408ebf7dfadd89d5b11579efb4149a9298936/build/pause/pause.c) for namespace sharing, which 720k smaller than busybox, has less stuff in it and doesn't have a shell, feels like a safer default/suggested choice

Even if it's experimental I think it should be doc'd somwhere (under a "here be dragons" banner)  -will get people looking at it/interested in it.

skinowski:
right busybox is also todo.. I don't personally like the tail -f & busybox.

api/agent/drivers/docker/docker_pool.go
```diff
@@ -0,0 +1,294 @@
...
+	AllocPoolId() (string, error)
+
+	// Release the id back to the pool
+	FreePoolId(id string)
+
+	// stop and terminate the pool
+	Close() error
+
+	// returns inuse versus free
+	Usage() DockerPoolStats
```
rdallman: somewhat pedantic and not a blocker: would be nice to instrument register/unregister/alloc/free with the opencensus stuff and hang things from there instead of draping one-off stats off of various things (agent, slot queue, docker pool, etc). the views stuff lets us get snapshots at any point from wherever we might like and gets auto-plumbed out to whatever exporters are draped up front instead of having to plumb this stuff out explicitly. def a learning curve and i don't think we've quite figured out a 'nice' way to bundle our views/measurements at present, either. this is easy to sweep through and do to stuff later, anyway.

skinowski:
yes, the policy right now is very relaxed, since it's experimental. If pool cannot hand out a buddy, things still work.

zootalures:
Since this is experimental - could you put something that makes this immediately  and abundantly obvious to the user in this change?  

one  or more of: 
* Make the enabling options include "EXPERIMENTAL" or similar (similar to --experimental elsewhere) - FN_EXPERIMENTAL_POOL...
* A doc indicating that this (and any other related features are experimental) 
* A big fat warning in logs saying that you are using experimental/potentially risky features.

zootalures:
LGTM, thanks!

skinowski:
I don't like how k8s folks are using nonreentrant calls in their sig handlers, but I've got this minimal thing if we need to:

https://github.com/skinowski/fn/tree/pause_container

Or we can pull k8s pause from their repo as well...

PR #875, opened 2018-03-21
tutorials link

carimura:
cc @michael-w-williams


michael-w-williams:
We should probably use https for the URL.

PR #876, opened 2018-03-21
fix id counter

rdallman:
i bring great shame to my family

### DIFF
--- a/api/id/id.go
+++ b/api/id/id.go
@@ -63,13 +63,13 @@ func New() Id {
 
 	id[8] = byte(machineID<<4) | byte((count<<4)>>60)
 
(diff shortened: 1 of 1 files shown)


skinowski:
I was wondering about the zero bytes in the ids in logs lately. :-)

api/id/id.go
```diff
@@ -63,13 +63,13 @@ func New() Id {
 
 	id[8] = byte(machineID<<4) | byte((count<<4)>>60)
```
rdallman: think i had an epic brain fart when i changed the machine id to 48 bits. this looks to use 20 bits and also be messed up. not sure how i get paid for this

PR #878, opened 2018-03-22
Add app and route models to a request's context

gviedma:
Sometimes extensions need access to the full app and route models belonging to a request. We don't want to have to resolve these multiple times via the agent.DataAccess interface, since that could result in inconsistencies due to cache evictions etc. Instead, this PR adds the app and route models to the http.Request context at the place where they are resolved.


api/agent/call.go
```diff
@@ -65,6 +65,10 @@ func FromRequest(appName, path string, req *http.Request) CallOpt {
 			route.Format = models.FormatDefault
 		}
 
+		// make app and route models available in the request context
+		ctx := context.WithValue(req.Context(), "FN_APP_MODEL", app)
+		req = req.WithContext(context.WithValue(ctx, "FN_ROUTE_MODEL", route))
+
```
rdallman: coming in through the http path this is a dupe of https://github.com/fnproject/fn/blob/master/api/server/gin_middlewares.go#L83-L91 with the same values. coming in through grpc hopefully has a middleware that does similar in the front end, which catches every handler (i.e. not just the call path). the gin middleware also uses typed keys, and can yank out of context pretty easily with helper methods (less error prone). are the front end wrappers enough or do we need to add them elsewhere now?

rdallman: vague, sorry, coffee levels rising. concrete suggestion: 

add to https://github.com/fnproject/fn/blob/master/api/common/ctx.go

```go
var (
  appContextKey = contextKey("app_context_key")
)

func WithApp(ctx context.Context, app *models.App) context.Context {
  return context.WithValue(ctx, appContextKey, app)
}

// App/AppFromContext
func App(ctx context.Context) (*models.App, bool) {
  app, ok := ctx.Value(appContextKey).(*models.App)
  return app, ok
}
```

hhexo:
We should be able to extend the RunnerCall interface (and the underlying agent.call) without significant cost, rather than using a different mechanism to pass the information.

See #879 for an alternative implementation. What do you think?

gviedma:
Closed in favor of #880

PR #879, opened 2018-03-22
Expose route and app models to RunnerCall for extensions to use it

hhexo:
This is an alternative to #878 

The internal `agent.call` is not persisted to the db, so we should be able to add route and app models to it without any significant cost. Therefore this just extends the current abstraction rather than using a different way to pass data.

### DIFF
--- a/api/agent/call.go
+++ b/api/agent/call.go
@@ -106,6 +106,8 @@ func FromRequest(appName, path string, req *http.Request) CallOpt {
 			URL:         reqURL(req),
 			Method:      req.Method,
 		}
+		c.app = app
+		c.route = route
 
 		c.req = req
 		return nil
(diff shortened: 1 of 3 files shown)


api/agent/call.go
```diff
@@ -260,6 +262,9 @@ func (a *agent) GetCall(opts ...CallOpt) (Call, error) {
 type call struct {
 	*models.Call
 
+	app   *models.App
+	route *models.Route
```
rdallman: pretty similar to the config stuff: https://github.com/fnproject/fn/blob/master/api/agent/call.go#L115-L122 - we could take one out of that book and hang it off `*models.Call`, then merge the app and route annotations into one `Annotations` bucket. notably, for async, the config goes in the json blob message to the mq (a json of `models.Call`) so that we can later use that config after we pull the message off without having to rebuild it. annotations seems to be analogous to this, and for some kind of cohesion it makes sense for now (we could decide to chop up `*models.Call` some day). thoughts?

I do think I prefer this to #878 for this use case

gviedma: Also for context there are objections to extending the models.Call object since that will be persisted with every call.

gviedma:
Closed in favor of #880

PR #880, opened 2018-03-22
Expose route and app models to RunnerCall for extensions (alternative 2)

hhexo:
Another alternative to #878 (and #879)

This allows us to merge the Annotations from app and route (which are the useful information which is missing in Call) into the models.Call. The field is ephemeral like Config so it should not be persisted and should not cause issues.

### DIFF
--- a/api/agent/call.go
+++ b/api/agent/call.go
@@ -101,6 +101,7 @@ func FromRequest(appName, path string, req *http.Request) CallOpt {
 			Memory:      route.Memory,
 			CPUs:        route.CPUs,
 			Config:      buildConfig(app, route),
+			Annotations: buildAnnotations(app, route),
 			Headers:     req.Header,
 			CreatedAt:   strfmt.DateTime(time.Now()),
 			URL:         reqURL(req),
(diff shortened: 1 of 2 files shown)


api/agent/call.go
```diff
@@ -135,6 +136,17 @@ func buildConfig(app *models.App, route *models.Route) models.Config {
 	return conf
 }
 
+func buildAnnotations(app *models.App, route *models.Route) models.Annotations {
+	ann := make(models.Annotations)
```
rdallman: tip: can pre-allocate the size perfectly here `make(models.Annotations, len(app.Annotations)+len(route.Annotations)` to avoid some number of map resizes (~$)

api/agent/call.go
```diff
@@ -135,6 +136,17 @@ func buildConfig(app *models.App, route *models.Route) models.Config {
...
+	for k, v := range app.Annotations {
+		ann[k] = v
+	}
+	for k, v := range route.Annotations {
+		ann[k] = v
+	}
```
reclaro: Annotations in route and app are independent so we don't have a guarantee that keys are unique between route and app that could be a problem with merging them

rdallman: config has the same issue. it's not clear if we should use precedence (route > app), which could be nice in certain cases. e.g. 'app' defaults to something, but 'route' can override and the code doesn't have to change to account for this. or we have separately addressable buckets and the calling code has to figure out how to resolve the conflict themselves. if we do the latter, then it's not much of a stretch to have them put in 2 different keys in 2 places since semantically, it's equivalent to the conflict resolution they'd have to do at the end anyway (this just pushes it up front). I think as long as we're clear about this behavior it's ok (and config).

gviedma:
I'm happy with this approach since it works for extensions (e.g. Call listeners) and is not persisted to the database. I also think that route annotations overriding app annotations is an acceptable solution and consistent with our approach to configuration. I'm closing #878 and #879 in favor of this one.

Issue #819, opened 2018-03-07
Namespacing in URLs


treeder:
The out of the box experience has no namespacing, but after you add authentication, there's probably a lot of people that would want that reflected in the URL (GitHub style). eg: `/ns/{namespace}/r/{func}`. This also applies to the API, would add another layer, eg: `/v1/ns/{namespace}/apps`. 

So do we make this part of base experience?  If so, the default when working locally could just fill it in with a dummy value, like: `/ns/local/r/{func}`

carimura:
Think this makes sense. many sites don't even require the /ns/ as it's always just the first value. idk if that could/should be the case here.

zootalures:
Yah we can discuss on slack - @rdallman and @vshiva  and I discussed a couple of options that can work for the use case we have in mind  without needing this 

If that sticks then whether this feature gets added is based on its own merits

jan-g:
I particularly like exposing this on the API side. There are a number of ways in which middleware might rewrite requests (much like the current DNS extension) and this gives an abstraction that can be surfaced at the API layer to express those. Then the API can sit at a different URL to the invocation point.

This kind of approach offers one avenue to model tenancy as a first-class thing, too (although there may be other ways to reflect that in the API).

rdallman:
we nixed this as the sentiment was 'why we don't need this look we're doing it this other way without first class support and the world hasn't fallen into a black hole' -- we could probably still afford to do this and it's probably going to come up again, but closing for now, we could at least doc the 'other way' maybe for potential white label / multi tenant consumers

PR #883, opened 2018-03-23
Provide GetApp and GetRoute in agent.DataAccess with original http.Request for multitenancy

gviedma:
This PR modifies the `agent.DataAccess` interface to provide full access to original `http.Request` when resolving apps and routes:
```
GetApp(ctx context.Context, appName string) (*models.App, error)
GetRoute(ctx context.Context, appName string, routePath string) (*models.Route, error)
```

This change is required in multi-tenant scenarios where it is not sufficient to know the app and route names to resolve the respective models. By providing access to the underlying `http.Request`, apps/routes can be disambiguated based on the contents of the HTTP request, including HTTP headers, Host or URL path.

### DIFF
--- a/api/agent/call.go
+++ b/api/agent/call.go
@@ -51,12 +51,12 @@ type Params []Param
 
 func FromRequest(appName, path string, req *http.Request) CallOpt {
 	return func(a *agent, c *call) error {
-		app, err := a.da.GetApp(req.Context(), appName)
+		app, err := a.da.GetApp(req, appName)
 		if err != nil {
 			return err
 		}
 
-		route, err := a.da.GetRoute(req.Context(), appName, path)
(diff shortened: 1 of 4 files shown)


zootalures:
Could you not use a middleware extension to move the relevant tenant data into the context?

gviedma:
@zootalures suggestion would work and prevents us from having to pass on the _http.Request_ inside `GetApp` and `GetRoute`. The downside is relying on special values into the context to be consumed by different layers, making it potentially more difficult to reason about the correctness of the code. Since the middleware extensions mechanism meets the multi-tenancy requirements without requiring changes to the abstraction unnecessary for the single tenancy case, I am closing this PR.

rdallman:
*sigh* /arg #819

PR #884, opened 2018-03-23
I/O Format protocol version

denismakogon:
Protocol version helps to identify whether the FDK can handle request or not.

### DIFF
--- a/api/agent/protocol/factory.go
+++ b/api/agent/protocol/factory.go
@@ -19,6 +19,7 @@ type errorProto struct {
 	error
 }
 
+func (e errorProto) Version() string                                              { return "1.0.0" }
 func (e errorProto) IsStreamable() bool                                           { return false }
 func (e errorProto) Dispatch(ctx context.Context, ci CallInfo, w io.Writer) error { return e }
 
@@ -27,7 +28,7 @@ func (e errorProto) Dispatch(ctx context.Context, ci CallInfo, w io.Writer) erro
 // It returns any protocol error, if present.
 type ContainerIO interface {
 	IsStreamable() bool
-
+	Version() string
 	// Dispatch will handle sending stdin and stdout to a container. Implementers
 	// of Dispatch may format the input and output differently. Dispatch must respect
 	// the req.Context() timeout / cancellation.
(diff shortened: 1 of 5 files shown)


api/agent/protocol/json.go
```diff
@@ -40,6 +40,7 @@ type jsonIn struct {
 	Body        string          `json:"body"`
 	ContentType string          `json:"content_type"`
 	Protocol    CallRequestHTTP `json:"protocol"`
+	Version     string          `json:"version"`
```
rdallman: `version` is too vague imo. when we have function versioning this is too easy to conflate for users. `json_protocol_version` or something is less challenging to misinterpret

denismakogon: yeah, in terms of protocol `version` is the "ok" name, but for request body let that be `protocol_version`

api/agent/protocol/factory.go
```diff
@@ -27,7 +28,7 @@ func (e errorProto) Dispatch(ctx context.Context, ci CallInfo, w io.Writer) erro
 // It returns any protocol error, if present.
 type ContainerIO interface {
 	IsStreamable() bool
-
+	Version() string
```
rdallman: I don't understand why we need to expose version from the ContainerIO interface itself. it seems very much like the version will be contained within each protocol and they will be expected to write in the version when they write in a request, but nowhere in fn seems likely to do this separately here where I would guess we would: call protocol.Disptach() and if we get back some specific fdk error that the version is too high or low we cycle through all protocols by some version? the factory method to get a protocol doesn't allow specifying version? are formats themselves versioned (e.g. `format: "json1"`), in which case version here is completely gratuitous ?

since default and http are likely never to have a different version, it doesn't seem like this is anything but anticipatory for json and the construct doesn't necessarily make sense to impose on every format.

denismakogon: okay, makes sense, only JSON protocol should have a version.

> if we get back some specific FDK error that the version is too high or low we cycle through all protocols by some version?

That shouldn't be the case. If function can't handle protocol version A, then Fn should do what it does - report back to a user with some verbose error.

> are formats themselves versioned (e.g. format: "json1"), in which case version here is completely gratuitous?

they are not, however, they must be, because any further changes to the protocol may break the FDK or make their request parsing inconsistent.

So, the idea of this PR to start versioning the protocols, because it is more than necessary for the FDKs to be consistent with the Fn server protocol implementation.

zootalures:
Closing as wontfix for now - Cloud events format will be inherently versioned.

PR #890, opened 2018-03-26
Experimental Pre-fork Pool: Recycle net ns

skinowski:
*) Rework the control loop for the containers, where now we have two states: initializing and ready. This allows us to avoid expensive & unnecessary operations for backfill/recycle phases.
*) A recycle option which discards each pool item after use.
*) Network list option to provide a list of docker bridges to spread the pool into. Each docker bridge has a linux limitation of 1023 containers, so using this list, we can exceed that limit easily.

TODO: Recycle performs poorly in cold-containers as it runs out of pool items over time. This causes churn and increases system load significantly. Need to run/tune this recycle mode for hot containers and re-evaluate.


skinowski:
Note: I'll be building/experimenting on top of this code.. Important thing here is handling of init versus ready states, which is an improvement over old code...

PR #892, opened 2018-03-26
Expose func call as the extensions API

denismakogon:
The purpose of the PR is to expose an API in the extension manner, so the developers can use this to call the functions inside the extensions, specifically, listeners (note, that existing middleware `CallFunction` can be consumed only within the middleware, but not the listeners). So, given refactoring brings capabilities of the listeners to the same level of the middleware.

There is no impact to existing API.
New method suppose to be (or become) fewer HTTP-aligned.

Please note, that, eventually, this method should become trigger-agnostic and just accept a call for the execution, so any HTTP-(or whatever)-specific configuration should happen outside of this method to give unimaginable freedom of choice.

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -898,6 +898,10 @@ func (s *Server) Datastore() models.Datastore {
 	return s.datastore
 }
 
+func (s *Server) Agent() agent.Agent {
+	return s.agent
+}
+
 // returns the unescaped ?cursor and ?perPage values
 // pageParams clamps 0 < ?perPage <= 100 and defaults to 30 if 0
 // ignores parsing errors and falls back to defaults.


api/server/runner.go
```diff
@@ -55,14 +55,16 @@ var (
 	bufPool = &sync.Pool{New: func() interface{} { return new(bytes.Buffer) }}
 )
 
-// TODO it would be nice if we could make this have nothing to do with the gin.Context but meh
-// TODO make async store an *http.Request? would be sexy until we have different api format...
-func (s *Server) serve(c *gin.Context, app *models.App, path string) error {
+// CallFunction exposed to become the API extension, to let API listeners to be capable to call a functions
+// defined by app and path
+func (s *Server) CallFunction(app *models.App, path string, req *http.Request,
+	respWriter http.ResponseWriter) (*string, io.Reader, error) {
```
rdallman: with the general idea here, thought we had done this already for some reason. this does seem an odd signature for an extension to be able to fulfill. most of our extensions are either based around http req/resp or the app/route and don't seem likely to have a handle on all of these. a lot of the stuff in here is also pretty specific to our http endpoint and not necessarily relevant to the extension caller (error handling, wait time, async, etc).

seems like:

```
var out bytes.Buffer
call, err := Server.Agent.GetCall(agent.FromRequest(req), agent.WithWriter(&out))
err := Server.Agent.Submit(call)
```

is more in line with what extensions would want, and it's flexible for the extension to do this based on whatever they have, be it a request or an app/route (we can add options to make this easier), or build a model, etc. we could tidy this up into a one liner easily enough. the proposed plumbing here just seems a little off. exposing agent is easy, not sure if it's the right plumbing, either, though.

denismakogon: Agreed, haven't thought about exposing the agent directly. So, yeah, i assume expose agent suppose to work fine for calling the functions directly.

api/server/server.go
```diff
@@ -898,6 +898,10 @@ func (s *Server) Datastore() models.Datastore {
 	return s.datastore
 }
 
+func (s *Server) Agent() agent.Agent {
+	return s.agent
+}
+
```
denismakogon: note, that `Agent()` is not a part of `fnext.ExtServer` because it brings a lot of mess in code along with circular references:
```
import cycle not allowed
package github.com/fnproject/fn
	imports github.com/fnproject/fn/api/server
	imports github.com/fnproject/fn/api/agent
	imports github.com/fnproject/fn/fnext
	imports github.com/fnproject/fn/api/agent
```
and so on. I guess simple exposable method is enough.

treeder:
What's the purpose of this?  Like what use case did you need this for?

denismakogon:
@treeder agent is required to call the functions from the API listeners.

denismakogon:
Wow, thanks, nearly forgot about this one.

PR #893, opened 2018-03-26
update thrift, opencensus, others

rdallman:
fail, fail again...

now we can use 0.6.0 opencensus, thrift was only updating to latest version, not master. have fixed both versions in for now, to prevent further thrift breakages.

release notes say they fixed jaeger, gonna try it out again and see if hot links are ready to come off the grill


rdallman:
ah, right, 1s.. sorry

PR #902, opened 2018-03-28
Try to instantiate from HTTP request in favour Event

denismakogon:
- this PR makes Agent work with the Event object that is (unfortunately) derived from the `http.Request`
 - it does make code look more HTTP-agnostic
 - step forward to multiple triggers
 - allows API Listeners to submit non-HTTP-specific events (call the function).

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -73,7 +73,7 @@ import (
 type Agent interface {
 	// GetCall will return a Call that is executable by the Agent, which
 	// can be built via various CallOpt's provided to the method.
-	GetCall(...CallOpt) (Call, error)
+	GetCall(context.Context, ...CallOpt) (Call, error)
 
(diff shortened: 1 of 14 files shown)


rdallman:
a call is already an event, it's just under a different name. we're only using request innards for headers and body, would rather yank that out than make yet another struct type which adds more indirection and complexity here. we should do something like this I think, but the cut to make 'event' inside of call doesn't seem like the right cut. it's easy to add separate builders to get a call, too, suspect we have `FromEvent(event)` eventually.

denismakogon:
Maybe, but i don't think that having an HTTP request inside the call is not what we want, eventually. At this moment, if i want to call the function i still need to create a call object from an HTTP request (even if that is a dummy). 

So, consider this PR as step forward to the `FromEvent`. But we need to make sure that HTTP request doesn't bother us, that's why i proposed new type that represents the event that leads agent can consume and call the function.

PR #911, opened 2018-04-02
move calls to logstore, implement s3

### LINKED ISSUE #482: store calls in some s3-like thing
since we do not have an infinitely(?) scalable db to work with ala spanner, if we store calls and logs in some [theoretically] infinitely scalable storage system like s3 or its api compatible friend in other clouds then our load on the [currently] sql db we are using will be really minimal, mostly just doing reads to serve caches with a very low write rate for crud stuff on apps & routes.

if we keep calls [and/or logs] in the sql db then we are going to hit scaling issues very quickly there for sharded mysql (or w/e sql db), and those sql guys generally don't like eating high write rate loads to begin with. obviously we're pushing the problem elsewhere, but blob stores tend to eat high write workloads a little better (s3 is stupid fast, on the order of milliseconds for smaller blobs, mostly network latency). 

issue here is mostly abstraction I guess, aside from thinking this is a good idea. need to get `GetCall` / `PutCall` into the `LogStore` api (with a real log store implementation). if we do this then we probably won't have to worry too too much about the sql db we're using. 

if we hook up a decent mq then it's also a relatively good option for us to send logs / calls to that to get quick response times at the end of calls and we can have a process that runs around and eats the queues and puts them into some more persistent storage (like s3), this would also give some flexibility vs just depositing into an s3 bucket but our calls / logs api would be 'eventually consistent' (most likely, unless we wanna search queues -- and we really don't).

another option is to not store call invocations / logs for sync or async requests at all, and make users add hooks to get log output and status updates (for async, specifically, sync gets output as is), which seems like an interesting and appealing option (operationally speaking, especially). 

possibly for internal discussion, but likely useful to discuss in the open world as they'll run into this problem too...

rdallman:
closes #482

the basic motivation is that logs and calls will be stored with a very high
write rate, while apps and routes will be relatively infrequently updated; it
follows that we should likely split up their storage location, to back them
with appropriate storage facilities. s3 is a good candidate for ingesting
higher write rate data than a sql database, and will make it easier to manage
that data set. can read #482 for more detailed justification.

summary:

* calls api moved from datastore to logstore
* logstore used in front-end to serve calls endpoints
* agent now throws calls into logstore instead of datastore
* s3 implementation of calls api for logstore
* s3 logs key changed (nobody using / nbd?)
* removed UpdateCall api (not in use)
* moved call tests from datastore to logstore tests
* mock logstore now tested (prev. sqlite3 only)
* logstore tests run against every datastore (mysql, pg; prev. only sqlite3)
* simplify NewMock in tests

commentary:

brunt of the work is implementing the listing of calls in GetCalls for the s3
logstore implementation. the GetCalls API requires returning items in the
newest to oldest order, and the s3 api lists items in lexicographic order.
An easy thing to do here seemed to be to reverse the
encoding of our id format to return a lexicographically descending order,
since ids are time based, reasonably encoded to be lexicographically
sortable, and de-duped (unlike created_at). This seems to work pretty well,
it's not perfect around the boundaries of to_time and from_time and a tiny
amount of results may be omitted, but to me this doesn't seem like a deal
breaker to get 6999 results instead of 7000 when trying to get calls between
3:00pm and 4:00pm Monday 3 weeks ago. Of course, without to_time and
from_time, there are no issues in listing results. We could use created at and
encode it, but it would be an additional marker for point lookup (GetCall)
since we would have to search for a created_at stamp, search for ids around
that until we find the matching one, just to do a point lookup. So, the
tradeoff here seems worth it. There is additional optimization around to_time
to seek over newer results (since we have descending order).

The other complication in GetCalls is returning a list of calls for a given
path. Since the keys to do point lookups are only app_id + call_id, and we
need listing across an app as well, this leads us to the 'marker' collection
which is sorted by app_id + path + call_id, to allow quick listing by path.
All in all, it should be pretty straightforward to follow the implementation
and I tried to be lavish with the comments, please let me know if anything
needs further clarification in the code.

The implementation itself has some glaring inefficiencies, but they're
relatively minute: json encoding is kinda lazy, but workable; s3 doesn't offer
batch retrieval, so we point look up each call one by one in get call; not
re-using buffers -- but the seeking around the keys should all be relatively
fast, not too worried about performance really and this isn't a hot path for
reads (need to make a cut point and turn this in!).

Interestingly, in testing, minio performs significantly worse than pg for
storing both logs and calls (or just logs, I tested that too). minio seems to
have really high cpu consumption, but in any event, we won't be using minio,
we'll be using a cloud object store that implements the s3 api. Anyway, mostly
a knock on using minio for high performance, not really anything to do with
this, just thought it was interesting.

I think it's safe to remove UpdateCall, admittedly this made implementing the
s3 api a lot easier. This operation may also be something we never need, it
was unused at present and was only in the cards for a previous hybrid
implementation, which we've now abandoned. If we need, we can always resurrect
from git.

Also not worried about changing the log key, we need to put a prefix on this
thing anyway, but I don't think anybody is using this anyway. in any event, it
simply means old logs won't show up through the API, but aside from nobody
using this yet, that doesn't seem a big deal breaker really -- new logs will
appear fine.

future:

TODO make logstore implementation optional for datastore, check in front-end
at runtime and offer a nil logstore that errors appropriately

TODO low hanging fruit optimizations of json encoding, re-using buffers for
download, get multiple calls at a time, id reverse encoding could be optimized
like normal encoding to not be n^2

TODO api for range removal of logs and calls


api/id/id.go
```diff
@@ -240,3 +241,42 @@ func (id *Id) UnmarshalText(v []byte) error {
...
+// of a given id, e.g. 000 -> ZZZ, which allows reversing the sort order when stored
+// contiguously since ids are lexicographically sortable. The returned string will
+// be of len(src), and assumes src is from the base32 crockford alphabet, otherwise
+// using 0xFF.
+func EncodeDescending(src string) string {
+	var buf [EncodedSize]byte
+	copy(buf[:], src)
+	for i, s := range buf[:len(src)] {
+		// XXX(reed): optimize as dec is
+		j := strings.Index(Encoding, string(s))
```
skinowski: do we need rEncoding at all? 
```
j = len(Encoding) - 1 - j
```

Or we could use a small lookup array instead of searching Encoding? We could fixup/improve later...

api/id/id.go
```diff
@@ -240,3 +241,42 @@ func (id *Id) UnmarshalText(v []byte) error {
 
 	return nil
 }
+
+// reverse encoding useful for sorting, descending
+var rEncoding = reverseString(Encoding)
```
skinowski: super minor, but IMHO this should just be a const like Encoding and we can remove the reverseString() below.

skinowski:
Almost done with this, will continue tomorrow...

api/id/id.go
skinowski: Saw the TODO notes too, we'll fixup these later...

skinowski:
I think s3 keys have max of 1024 bytes length. Just calling it out here... I think we are OK, if looking at sql.go, 256 bytes of app_id/path even with base64 overhead..

api/logs/s3/s3.go
```diff
@@ -173,6 +175,245 @@ func (s *store) GetLog(ctx context.Context, appID, callID string) (io.Reader, er
...
+		marker = callKey(filter.AppID, filter.Cursor)
+		if filter.Path != "" {
+			marker = callMarkerKey(filter.AppID, filter.Path, filter.Cursor)
+		}
+	} else if t := time.Time(filter.ToTime); !t.IsZero() {
+		// we need to use first 48 bits of id to approximate created_at, to skip over records
+		// if looking far into the past to save time.
+		ms := uint64(t.Unix()*1000) + uint64(t.Nanosecond()/int(time.Millisecond))
+		var buf id.Id
+		buf[0] = byte(ms >> 40)
```
skinowski: I think we could (should) move this code to id.go, since it's tied to id structure... Not a big deal though.

api/logs/s3/s3.go
```diff
@@ -173,6 +175,245 @@ func (s *store) GetLog(ctx context.Context, appID, callID string) (io.Reader, er
...
+		buf[1] = byte(ms >> 32)
+		buf[2] = byte(ms >> 24)
+		buf[3] = byte(ms >> 16)
+		buf[4] = byte(ms >> 8)
+		buf[5] = byte(ms)
+		mid := buf.String()[:10]
```
skinowski: should we call buf.MarshalTextTo() to translate this into base32 first?

api/logs/s3/s3.go
```diff
@@ -173,6 +175,245 @@ func (s *store) GetLog(ctx context.Context, appID, callID string) (io.Reader, er
...
+
+	result, err := s.client.ListObjects(input)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list logs: %v", err)
+	}
+
+	calls := make([]*models.Call, 0, len(result.Contents))
+
+	for _, obj := range result.Contents {
+		if len(calls) == filter.PerPage {
```
skinowski: not important, but trying to understand while if/break is necessary here (and seems out of place since there are continue statements within this loop), MaxKeys already handles this?

api/logs/s3/s3.go
```diff
@@ -173,6 +175,245 @@ func (s *store) GetLog(ctx context.Context, appID, callID string) (io.Reader, er
...
+		if filter.Path != "" {
+			fields := strings.Split(*obj.Key, "/")
+			if len(fields) != 4 {
+				return calls, fmt.Errorf("invalid key in call markers: %v", *obj.Key)
+			}
+			app = fields[1]
+			id = fields[3]
+		} else {
+			fields := strings.Split(*obj.Key, "/")
+			app = fields[1]
```
skinowski: app/id setting needs to happen after if check?

rdallman:
> I think s3 keys have max of 1024 bytes length. Just calling it out here... I think we are OK, if looking at sql.go, 256 bytes of app_id/path even with base64 overhead..

good catch, i'll document this next to it. i think we're ok too, but wasn't aware of that limitation.

api/logs/s3/s3.go
rdallman: yea, i'm with that program. `NewWithTime(t time.Time)` ought to do the trick.

api/logs/s3/s3.go
rdallman: not sure i'm tracking, the string method will call that. we could use an array and save an alloc though, can do that.

skinowski: Cool, I missed that.. :-)

api/logs/s3/s3.go
rdallman: d'oh. thanks

api/logs/s3/s3.go
rdallman: yea, max keys covers this. originally i was worried that minio API would not implement it, since the minio bindings don't offer this option. 

putting it at top of loop instead of after the continues just saves a look up of the call body itself if we don't need it. mostly just here for safety, if the s3 client lib returns 90000 results we'll be in a world of hurt anyway, but at least they're only 90000 keys.

rdallman:
updated, thanks @skinowski !

rdallman:
> One question, since we moved to from SQL to S3 for logs, what about migrations for further changes to s3 models (i assume they would happen soon enough)?

no good story here, json is quite a bit more flexible than sql schemas at least in this regard.

rdallman:
thanks!

PR #912, opened 2018-04-03
Bugfix/grpc consume eof

jan-g:
The GRPC stream will terminate with an io.EOF; ensure that we consume this in the client code.

Whilst we're at it, the RunnerCall interface strictly only requires a ReadCloser rather than the full httpRequest - the remainder of that request has been extracted into a model.Call at the point this is used.

These changes (a) make the grpc client code correct, and (b) refactor the Runner interface slightly in order that it can be reused from a command-line client.

### DIFF
--- a/api/agent/runner_client.go
+++ b/api/agent/runner_client.go
@@ -199,7 +199,11 @@ func receiveFromRunner(protocolClient pb.RunnerProtocol_EngageClient, c pool.Run
 			if body.Finished.Success {
 				logrus.Infof("Call finished successfully: %v", body.Finished.Details)
 			} else {
-				logrus.Infof("Call finish unsuccessfully:: %v", body.Finished.Details)
+				logrus.Infof("Call finished unsuccessfully: %v", body.Finished.Details)
+			}
+			// There should be an EOF following the last packet
+			if _, err := protocolClient.Recv(); err != io.EOF {
+				logrus.Errorf("Did not receive expected EOF from runner stream: %v", err)
 			}
 			close(done)
 			return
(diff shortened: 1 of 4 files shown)


api/agent/runner_client.go
```diff
@@ -199,7 +199,11 @@ func receiveFromRunner(protocolClient pb.RunnerProtocol_EngageClient, c pool.Run
 			if body.Finished.Success {
 				logrus.Infof("Call finished successfully: %v", body.Finished.Details)
 			} else {
-				logrus.Infof("Call finish unsuccessfully:: %v", body.Finished.Details)
+				logrus.Infof("Call finished unsuccessfully: %v", body.Finished.Details)
+			}
+			// There should be an EOF following the last packet
+			if _, err := protocolClient.Recv(); err != io.EOF {
+				logrus.Errorf("Did not receive expected EOF from runner stream: %v", err)
```
gviedma: Capture error with `logrus.WithError(err).Error("Did not receive expected EOF from runner stream")`

gviedma: We might also want to explicitly post the error via the channel, since receiving other messages potentially indicates a bug or issue with the state of the runner protocol messages:

`done <- error`

jan-g: Good idea; doing now.

gviedma: In fact, I think that `pure_runner.go` should be modified to do a `SendAndClose` on sending a RunnerMsg_Finished message.

api/runnerpool/runner_pool.go
```diff
@@ -42,7 +42,7 @@ type Runner interface {
 // processed by a RunnerPool
 type RunnerCall interface {
 	SlotDeadline() time.Time
-	Request() *http.Request
+	RequestBody() io.ReadCloser
```
reclaro: Probably in the future we need to use some deprecation process before changing a public interface

jan-g: Agreed on that. I'd like to go through the interfaces here once more to ensure they are as small as possible; there's no point requiring someone to construct an httpRequest if all we need is the body stream, for instance.

PR #913, opened 2018-04-03
Bugfix: unsafeBytes slices were getting GCed

jan-g:
There are alternative formulations of this, for instance see
	https://www.reddit.com/r/golang/comments/5zctpf/unsafe_conversion_between_strings_and_byte_slices/

The problem manifested in the returned values from unsafeBytes occasionally
being broken. It's possible that by keeping a reference to the `a` parameter
alive, the original code would still work - however, this definitely seems like
a fix.

(A cast to `[]byte(a)` looks increasingly attractive, for all that it'll
perform small allocations and copies.)

### DIFF
--- a/api/agent/slots.go
+++ b/api/agent/slots.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha1"
 	"encoding/binary"
 	"hash"
+	"reflect"
 	"sort"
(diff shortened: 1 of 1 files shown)


rdallman:
thanks!

Issue #915, opened 2018-04-04
multiple instances of node func.js get deployed to docker


c094728:
I'm running the sample function project setup
fn init --runtime node myproject
I retrieve the url at http://localhost:8080/r/myproject/myproject
If I pause for a while and reload the page, another instance shows up in docker ps.
These keep growing every time I pause for a while and reload
Note:
I had to stringify the results for the hello world example or it just displays [Object]
The init should have created it using stringify automatically.
return JSON.stringify(response);

denismakogon:
> If I pause for a while and reload the page, another instance shows up in docker ps

This is how hot format works in Fn, function dies as soon as it will reach out the idle timeout (one of the key concepts of the serverless - never pay for idle).

> I had to stringify the results for the hello world example or it just displays

You should do that by yourself because FDK-node doesn't know how to serialize your object.

c094728:
Thank you.  So where is the idle timeout defined?   I have 10 instances of the same container showing up in docker, all in Paused state and they have been there for over an hour.  And it still creates more if I reload the browser.

c094728:
I'm still not understanding.   I understand that a cold function will start a new container for each request.  But why are they not going away after being called,  The seem to go to paused status and remain there.  If manually I run docker unpause then the container goes away.   They won't show up forever in paused state will they?

denismakogon:
So, here's what Fn does: you call the function, Fn starts the container, your function processes the request, sends the response, Fn pauses the container for the time of the idle timeout (which is defaulted to 30 seconds), once idle timeout appeared - container dies.

Why are containers not going away? Because the function you've created from the `fn init` uses hot format by default, not the cold one.

> They won't show up forever in the paused state will they?

No, once the idle timeout happens, it's gone.

c094728:
I thought cold function was the default.  How do I change it to be a cold function.  If I defined a hot function then why is it not reusing the same container for each request and instead loading a new instance.  (if there is a delay between requests).  And they seem to never go away.

denismakogon:
> How do I change it to be a cold function?
set the format to `default` (use `fn routes update` cmd).

Fn will reuse the same container unless the request got delayed, if that delayed it will start a new container.

c094728:
[...]

PR #916, opened 2018-04-04
fn: update minimum docker version required.

skinowski:
Oracle Linux 7.4 backported versions still having issues
with freezing/terminating containers. 17.10.0-ce seems like
a resonable lowest common denominator.


skinowski:
Related: #915

PR #917, opened 2018-04-05
make annotations nullable to fix migration

zootalures:


api/datastore/sql/migrations/10_add_app_id.go
```diff
@@ -22,7 +22,7 @@ var sqlStatements = [...]string{`CREATE TABLE IF NOT EXISTS routes (
 	type varchar(16) NOT NULL,
 	headers text NOT NULL,
 	config text NOT NULL,
-	annotations text NOT NULL,
+	annotations text,
```
rdallman: we need an additional migration (`ALTER TABLE`...) on top of this anyway to handle changing this field to NOT NULL for any database that has run this, don't we? seems better to encapsulate this change there instead of modifying old migration in addition to new one. not sure, just checking.

rdallman: bump @zootalures

zootalures: Bugger, thanks for spot - added that and rebased

api/datastore/sql/migrations/12_fix_annotations_migrations_route.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"context"
+
+	"github.com/fnproject/fn/api/datastore/sql/migratex"
+	"github.com/jmoiron/sqlx"
+)
+
+func up11(ctx context.Context, tx *sqlx.Tx) error {
+	_, err := tx.ExecContext(ctx, "ALTER TABLE routes MODIFY annotations TEXT NULLABLE;")
```
rdallman: pg gives fantastically bad error messages around this, but i believe this (and the same for app) are causing CI to fail with this:

```
    	sql_test.go:94: with migrations now!
    	sql_test.go:97: database is locked
	test.go:29: 
		time="2018-04-20T19:22:19Z" level=info msg="datastore dialed" datastore=postgres max_idle_connections=256
		time="2018-04-20T19:22:19Z" level=info msg="datastore dialed" datastore=postgres max_idle_connections=256
		time="2018-04-20T19:22:19Z" level=error msg="error running migrations" error="error running migration. version: 11 direction: up err: pq: syntax error at or near \"MODIFY\"\npq: current transaction is aborted, commands ignored until end of transaction block"
		time="2018-04-20T19:22:19Z" level=info msg="datastore dialed" datastore=postgres max_idle_connections=256
		time="2018-04-20T19:22:19Z" level=error msg="error running migrations" error="database is locked"
		
	sql_test.go:94: with migrations now!
	sql_test.go:97: database is locked
```

it seems like we need to do, per https://www.postgresql.org/docs/current/static/sql-altertable.html :

```
ALTER TABLE routes ALTER COLUMN annotations DROP NOT NULL;
```

api/datastore/sql/migrations/12_fix_annotations_migrations_route.go
```diff
@@ -0,0 +1,27 @@
+package migrations
```
rdallman: i don't see `11_`

zootalures:
Thanks will look tomorrow

api/datastore/sql/migrations/12_fix_annotations_migrations_route.go
rdallman: so, of course, that doesn't work with mysql... looking into that

rdallman: and sqlite doesn't support this at all, oh boy. at least, these are all getting tossed soon enough. we could just skip it for sqlite for now i guess and leave this NOT NULL, even though it's the one most people use and most of them don't use annotations either. not sure what's best to do exactly, gonna infuse some beans in hot water and think about it

zootalures:
@rdallman  - suffice to say I feel bad about this but I think this brings the world back into alignment. 

I know this probably goes away later but it was worth it for me to do the kata of fixing my shit. 

There is an edge case due to non-transactional schema udpates on mysql where a a writer with a very old version migrating forward has a race on the null annotations fields, but i think i'm OK with that. 

(various cases in mysql migrations will cause non-recoverable/roll-backable changes I think ) 

Thanks for your patience!

zootalures:
Should have said - I've tried manual forward migrations with data from 0.3.383 -> this branch on mysql and sqlite - wasn't able to get the manual pg test working but I'm comfortable that it does work practically.

PR #919, opened 2018-04-05
fn: handleCallEnd and submit improvements

skinowski:
*) This simplifies submit() function but moves the burden
of retriable-versus-committed request handling and slot.Close()
responsibility to handleCallEnd().
*) handleCallEnd() related test fixes.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -253,6 +253,16 @@ func (a *agent) submit(ctx context.Context, call *call) error {
 	return a.handleCallEnd(ctx, call, slot, err, true)
 }
 
+func (a *agent) scheduleCallEnd(fn func()) {
+	a.wg.Add(1)
+	atomic.AddInt64(&a.callEndCount, 1)
+	go func() {
+		fn()
+		atomic.AddInt64(&a.callEndCount, -1)
+		a.wg.Done()
+	}()
+}
+
 func (a *agent) handleCallEnd(ctx context.Context, call *call, slot Slot, err error, isCommitted bool) error {
 
 	// For hot-containers, slot close is a simple channel close... No need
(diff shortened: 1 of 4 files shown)


api/agent/agent.go
```diff
@@ -238,37 +238,67 @@ func (a *agent) submit(ctx context.Context, call *call) error {
...
+			ctx, cancel := context.WithTimeout(ctx, a.cfg.CallEndTimeout)
+			call.End(ctx, err)
+			cancel()
+			atomic.AddInt64(&a.callEndCount, -1)
+			a.wg.Done()
+		}()
+
+		handleStatsEnd(ctx, err)
+		return transformTimeout(err, false)
+	}
```
rdallman: can we encapsulate this into a slot? maybe pedantic, but it was really easy to prove correct when we had `defer slot.Close(ctx)` should be easy to override the `slot.Close` with a type for async slot removal that just wraps the standard hot/cold slot

skinowski: Hmm, also noticing that lb_agent.go has similar code...

skinowski:
Don't know if my last commit improved is this much, but it's also nice to see precisely in the timeline where/how/when slot is closed in handleCallEnd().

api/agent/agent_test.go
```diff
@@ -448,14 +448,6 @@ func TestSubmitError(t *testing.T) {
...
 		t.Fatal("expected error but got none")
 	}
-
-	if cm.Status != "error" {
-		t.Fatal("expected status to be set to 'error' but was", cm.Status)
-	}
-
-	if cm.Error == "" {
-		t.Fatal("expected error string to be set on call")
-	}
```
rdallman: can we still test this? we can make the async lightspeed mode call.End optional?

skinowski: added a listener to check/test this now...

PR #920, opened 2018-04-06
fn: pure_runner deadlock

skinowski:
Blocking channel operations while holding mutex causes
deadlocks.


skinowski:
Closing this as per #928

PR #921, opened 2018-04-06
Update README.md

carimura:


carimura:
sure. plz approve/merge this then add that whenever that gets added. ;)

carimura:
btw jsut ran into this: https://github.com/cncf/presentations

probably somethin similar. can create ticket now.

PR #923, opened 2018-04-07
fn: lb_agent state trackers only apply to runners

skinowski:
State trackers do not apply to LB agent.


skinowski:
There's still the stats in stats.go, so removing this, I'll confuse us going forward.. And reduces coupling between lb and agent.

PR #925, opened 2018-04-07
Lb agent close should close shutdown channel

skinowski:


api/agent/lb_agent.go
```diff
@@ -136,6 +136,7 @@ func (a *lbAgent) Close() error {
 	ctx, cancel := context.WithTimeout(context.Background(), runnerPoolShutdownTimeout)
 	defer cancel()
 
+	close(a.shutdown)
```
rdallman: may want a sync.Once around this in case, but maybe only if someone hits it :)

PR #926, opened 2018-04-07
fn: api and systems tests port cleanup

skinowski:
*) removed unused cancel in api-test harness/server
*) removed hard coded port in getServerWithCancel along
with faulty health check code.
*) in SetupHarness() fixed code that skipped server start.

### DIFF
--- a/test/fn-api-tests/utils.go
+++ b/test/fn-api-tests/utils.go
@@ -23,7 +23,7 @@ import (
 
 const lBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
 
-func Host() string {
+func GetAPIURL() (string, *url.URL) {
 	apiURL := os.Getenv("FN_API_URL")
 	if apiURL == "" {
 		apiURL = "http://localhost:8080"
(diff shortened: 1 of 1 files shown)


test/fn-api-tests/utils.go
```diff
@@ -71,31 +97,43 @@ func getServerWithCancel() (*server.Server, context.CancelFunc) {
...
-		cancel2 = context.CancelFunc(func() {
-			cancel()
+		go func() {
+			s.Start(ctx)
 			os.Remove(tmpMq)
 			os.Remove(tmpDb)
-		})
+		}()
+
+		startCtx, startCancel := context.WithDeadline(ctx, time.Now().Add(time.Duration(10)*time.Second))
```
rdallman: i think we want this at the head of `startServer` and to be `ctx, cancel := context.WithDeadline`, not sure we need both contexts in a separate (pardon me) context since checking the server is directly related to starting the server.

test/fn-api-tests/utils.go
```diff
@@ -71,31 +97,43 @@ func getServerWithCancel() (*server.Server, context.CancelFunc) {
...
-		})
-		_, err := http.Get(apiURL + "/version")
-		for err != nil {
-			_, err = http.Get(apiURL + "/version")
-		}
-		started = true
-		cancel2 = context.CancelFunc(func() {
-			cancel()
+		go func() {
+			s.Start(ctx)
```
rdallman: should we cancel this context as well if we timeout in the following checkServer loop (I don't see it, but that seems the intention here)?

test/fn-api-tests/utils.go
rdallman: i see, we want to leave `s.Start` up but only timeout the check. ignore me.

test/fn-api-tests/utils.go
rdallman: nvm, this seems to be from the enclosing context (not super clear only from patch, my bad)

PR #928, opened 2018-04-08
fn: pure runner concurrency fixes

skinowski:
*) Moved blocking I/O operations into their own go-routines.
*) gRPC send/receive is now performed via a dedicated go-routine for each direction. Blocking
semantics are now using channels, which can we unblocked via shutdown or gRPC context timeout.

### DIFF
--- a/api/agent/pure_runner.go
+++ b/api/agent/pure_runner.go
@@ -298,6 +298,10 @@ func (ch *callHandle) spawnReceiver() {
 		for {
 			msg, err := ch.engagement.Recv()
 			if err != nil {
+				// engagement is close/cancelled from client.
+				if err == io.EOF {
+					return
+				}
 				ch.shutdown(err)
 				return
 			}
@@ -422,10 +426,10 @@ func (ch *callHandle) getTryMsg() *runner.TryCall {
 	case item := <-ch.inQueue:
 		if item != nil {
 			msg = item.GetTry()
-			if msg == nil {
-				ch.shutdown(ErrorExpectedTry)
-			}
 		}
+	}
+	if msg == nil {
+		ch.shutdown(ErrorExpectedTry)
 	}
 	return msg
 }
@@ -441,10 +445,10 @@ func (ch *callHandle) getDataMsg() *runner.DataFrame {
 	case item := <-ch.inQueue:
 		if item != nil {
 			msg = item.GetData()
-			if msg == nil {
-				ch.shutdown(ErrorExpectedData)
-			}
 		}
+	}
+	if msg == nil {
+		ch.shutdown(ErrorExpectedData)
 	}
 	return msg
 }
@@ -612,6 +616,7 @@ func (pr *pureRunner) Engage(engagement runner.RunnerProtocol_EngageServer) erro
 	}
 
 	var dataFeed chan *runner.DataFrame
+DataLoop:
 	for {
 		dataMsg := state.getDataMsg()
 		if dataMsg == nil {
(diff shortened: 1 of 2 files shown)


api/agent/pure_runner.go
```diff
@@ -27,171 +28,356 @@ import (
...
+		status:     200,
+		outQueue:   make(chan *runner.RunnerMsg),
+		doneQueue:  make(chan struct{}),
+		errQueue:   make(chan error, 1), // always allow one error (buffered)
+		inQueue:    make(chan *runner.ClientMsg),
+	}
+
+	// spawn receiver and sender go-routines. We can work
+	// concurrently on engagement Send()/Recv() separately, but multiple
+	// go-routines cannot issue Send() at the same time.
```
kmjohansen: Based upon our offline discussion, it might be worth beefing this comment up to explain that the limitation comes from GRPC.  Multiple go routines may neither send on the same stream, nor receive on the same stream.  However, it is permissible to divide send and receive as long as only one go routine handles each.  It would be nice if we had some kind of invariant available to enforce this, too.

api/agent/pure_runner.go
```diff
@@ -27,171 +28,356 @@ import (
 	"google.golang.org/grpc/peer"
 )
```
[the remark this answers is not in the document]
skinowski: Added a brief flow description.

api/agent/pure_runner.go
skinowski: added some comments and link to the thread-safety documentation in gRPC.

api/agent/pure_runner.go
```diff
@@ -27,171 +28,429 @@ import (
...
+				return
+			case <-ch.ctx.Done():
+				return
+			case data := <-input:
+				if data == nil {
+					return
+				}
+
+				if len(data.Data) > 0 {
+					_, err := io.CopyN(ch.pipeToFnW, bytes.NewReader(data.Data), int64(len(data.Data)))
```
rdallman: it does seem like we could use the data channel now to send the input on to synchronize with the other goroutine that is filling in the call body (ghost reader style, basically, we just need a `**io.Reader` that implements `io.Reader` pretty much. no need to do anything here, future simplification idea tossing

api/agent/pure_runner.go
```diff
@@ -27,171 +28,429 @@ import (
...
+
+	return input
+}
+
+// spawnReceiver starts a gRPC receiver, which
+// feeds received LB messages into inQueue
+func (ch *callHandle) spawnReceiver() {
+
+	go func() {
+		defer close(ch.inQueue)
```
rdallman: since we're using doneQueue & ctx.Done() it may be easier to not close this and rely on not receiving on this channel rather than rely on receiving a nil value on this channel (more states to track). future again.

api/agent/pure_runner.go
```diff
@@ -27,171 +28,429 @@ import (
...
-	err := ch.commitHeaders()
+// enqueueMsg attempts to queue a message to the gRPC sender
+func (ch *callHandle) enqueueMsg(msg *runner.RunnerMsg) error {
+	select {
+	case ch.outQueue <- msg:
+		return nil
+	case <-ch.ctx.Done():
+	case <-ch.doneQueue:
+	}
+	return io.EOF
```
rdallman: not `server too busy` ? this is a timeout, effectively, right?

api/agent/pure_runner.go
```diff
@@ -27,171 +28,429 @@ import (
...
+}
+
+func NewCallHandle(engagement runner.RunnerProtocol_EngageServer) *callHandle {
+
+	// set up a pipe to push data to agent Function container
+	pipeR, pipeW := io.Pipe()
+
+	state := &callHandle{
+		engagement: engagement,
+		ctx:        engagement.Context(),
```
rdallman: nitty thing, makes it hard to attenuate deadlines / metrics / $other putting context on an object vs a function parameter. we can clean up later but it's a better soon-than-later thing

PR #930, opened 2018-04-10
Move ch ring placement back from old FnLB.

skinowski:


api/runnerpool/ch_placer.go
```diff
@@ -0,0 +1,87 @@
...
+		default:
+			runners, err := rp.Runners(call)
+			if err != nil {
+				logrus.WithError(err).Error("Failed to find runners for call")
+			} else {
+				i := int(jumpConsistentHash(sum64, int32(len(runners))))
+				for j := 0; j < len(runners); j++ {
+					r := runners[i]
+
+					placed, err := r.TryExec(ctx, call)
```
rdallman: a note that this is going to flood the first runner for a given key unless TryExec is gating this (didn't look). the naive placer flooded the first node in the list to begin with but anyway, we probably need to prevent this in the starry future to shed load across multiple nodes for a given key with some probability to land on any given node from a subset of runners, hopefully based on something not just completely random.

api/runnerpool/ch_placer.go
```diff
@@ -0,0 +1,87 @@
...
+					if err != nil {
+						logrus.WithError(err).Error("Failed during call placement")
+					}
+					if placed {
+						return err
+					}
+
+					i = (i + 1) % len(runners)
+				}
```
rdallman: in the case that we timeout, this will call TryExec, timeout, and iterate through all runners and timeout [in TryExec] before exiting the loop and finding out the call timed out. seems like we should early exit this case, too?

api/runnerpool/ch_placer.go
```diff
@@ -0,0 +1,87 @@
...
+
+		select {
+		case <-ctx.Done():
+			return models.ErrCallTimeoutServerBusy
+		case <-timeout:
+			return models.ErrCallTimeoutServerBusy
+		default:
+			runners, err := rp.Runners(call)
+			if err != nil {
+				logrus.WithError(err).Error("Failed to find runners for call")
```
rdallman: return err ? otherwise we're just going to wait until the timeout and throw a timeout error ?

rdallman: ah, retry even this i guess.

api/runnerpool/ch_placer.go
jan-g: Yeah. At the moment, we're not pulling a Status back to the LB from the runner. By seguing this into the return stream from a call placement, we can keep the Runner model on the LB up-to-date with a picture of load; that's the other piece that prevents this swamping.

(The difference between this and the old fnlb is that we can push Status messages every second or so as a part of that response, meaning that an LB has a more up-to-date picture of the runner state than it did if it had to wait on XXX-Fnlb-wait.)

PR #935, opened 2018-04-11
fn: locked mutex while blocked on I/O considered harmful

skinowski:
*) Removed hold mutex while wait I/O cases these included possible
disk I/O and network I/O.

*) Error/Context Close/Shutdown semantics changed since
the context timeout and comments were misleading. Close
and Shutdown now blocks until pending sessions are done.

Context usage in Close/Shutdown was 'wait up to x secs to
report an error' which only logged the error. This is misleading
since it gives the impression that these calls can be aborted/cancelled
via context. (But in reality they spawn go-routines until the task is done.)
And the error reporting was lossy since this suppressed runner client
errors after the context expired.

Instead, the runner pool can simply log the error. And context still can be
provided in future (eg. for opencensus instrumentation) as part of the
runner pool instance itself.

### DIFF
--- a/api/agent/static_pool.go
+++ b/api/agent/static_pool.go
@@ -1,6 +1,7 @@
 package agent
 
 import (
+	"context"
 	"errors"
 	"sync"
 
@@ -121,7 +122,7 @@ func (rp *staticRunnerPool) AddRunner(address string) error {
 
 	err = rp.addRunner(r)
 	if err != nil {
-		err2 := r.Close()
+		err2 := r.Close(context.Background())
 		if err2 != nil {
 			logrus.WithError(err2).WithField("runner_addr", address).Warn("Error closing runner on AddRunner failure")
 		}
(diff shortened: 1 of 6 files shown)


api/agent/static_pool.go
```diff
@@ -57,77 +52,64 @@ func (rp *staticRunnerPool) Runners(call pool.RunnerCall) ([]pool.Runner, error)
...
 func (rp *staticRunnerPool) Shutdown(ctx context.Context) (e error) {
+
+	var toRemove []pool.Runner
+
 	rp.rMtx.Lock()
-	defer rp.rMtx.Unlock()
 
-	ctx, cancel := context.WithTimeout(context.Background(), staticPoolShutdownTimeout)
-	defer cancel()
+	toRemove = rp.runners[:]
```
rdallman: nit: could `toRemove :=` here and avoid the 3 lines of var decl, no behavior change.

hhexo:
I wish golang had scope-based `defer` so you could do RAII locking/unlocking...

Maybe I'm just paranoid, but I would recommend to move the chunks of code between a `Lock()` and `Unlock()` into sub-functions so that you can use `defer` again and be safer with respect to future changes within those chunks of code.

skinowski:
Sounds good, I'll move them to new functions.

skinowski:
Preserving context arguments in
```
RunnerPool.Shutdown(ctx)
Runner.Close(ctx)
```
to keep API the same. But looking around, it seems like consumers of this code is just passing 'context.Background()' to these calls. So, static_pool and runner_client ignores these contexts as well now.

api/agent/static_pool_test.go
```diff
@@ -11,6 +12,10 @@ func setupStaticPool(runners []string) pool.RunnerPool {
 	return NewStaticRunnerPool(runners, nil, "", mockRunnerFactory)
 }
 
+var (
+	ErrorGarbanzoBeans = errors.New("yes, that's right. Garbanzo beans...")
```
rdallman: ick, the worst

PR #936, opened 2018-04-11
fn: introducing WaitGroup suitable for shutdown/session mgmt

skinowski:


api/common/wait_utils.go
```diff
@@ -0,0 +1,110 @@
...
+   successful AddSession.
+
+   Example usage:
+
+   group := NewWaitGroup()
+
+   for item := range(items) {
+       go func(item string) {
+           if !group.AddSession() {
+               // group may be closing or full
```
rdallman: bool is cool imo

skinowski: sounds good.

PR #937, opened 2018-04-12
fn: sync.WaitGroup replacement common.WaitGroup

skinowski:
agent/lb_agent/pure_runner has been incorrectly using
sync.WaitGroup semantics. Switching these components to
use the new common.WaitGroup() that provides a few handy
functionality for common graceful shutdown cases.

From https://golang.org/pkg/sync/#WaitGroup,
    "Note that calls with a positive delta that occur when the counter
     is zero must happen before a Wait. Calls with a negative delta,
     or calls with a positive delta that start when the counter is
     greater than zero, may happen at any time. Typically this means
     the calls to Add should execute before the statement creating
     the goroutine or other event to be waited for. If a WaitGroup
     is reused to wait for several independent sets of events,
     new Add calls must happen after all previous Wait calls have
     returned."

HandleCallEnd introduces some complexity to the shutdowns, but this
is currently handled by AddSession(2) initially and letting the
HandleCallEnd() to decrement by -1 in addition to decrement by -1 in
Submit().

lb_agent shutdown sequence and particularly timeouts with runner pool
needs another look/revision, but this is outside of the scope of this
commit.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -195,11 +195,9 @@ func (a *agent) Close() error {
 }
 
 func (a *agent) Submit(callI Call) error {
-	// we allocate 2 groups one for Submit and one for CallEnd (See HandleCallEnd)
-	if !a.shutWg.AddSession(2) {
+	if !a.shutWg.AddSession(1) {
 		return models.ErrCallTimeoutServerBusy
 	}
-	defer a.shutWg.AddSession(-1)
 
 	call := callI.(*call)
 
--- a/api/agent/lb_agent.go
+++ b/api/agent/lb_agent.go
@@ -103,11 +103,9 @@ func GetGroupID(call *models.Call) string {
 }
 
 func (a *lbAgent) Submit(callI Call) error {
-	// we allocate 2 groups one for Submit and one for CallEnd (See HandleCallEnd)
-	if !a.shutWg.AddSession(2) {
+	if !a.shutWg.AddSession(1) {
 		return models.ErrCallTimeoutServerBusy
 	}
-	defer a.shutWg.AddSession(-1)
 
 	call := callI.(*call)
 


api/agent/async.go
```diff
@@ -35,15 +34,20 @@ func (a *agent) asyncDequeue() {
...
 		case model, ok := <-a.asyncChew(ctx):
 			if ok {
-				a.wg.Add(1) // need to add 1 in this thread to ensure safe shutdown
 				go func(model *models.Call) {
 					a.asyncRun(ctx, model)
-					a.wg.Done() // can shed it after this is done, Submit will add 1 too but it's fine
+					a.shutWg.AddSession(-1)
 				}(model)
+
+				// WARNING: tricky. We reserve another session for next iteration of the loop
```
rdallman: why? there was a waiter covering the entire asyncDequeue, did we remove it?

skinowski: so for-loop maintains the original +1 (from agent.go). If asyncRun go routine is spawned, then we need one more. With this, instead of defer -1, we've got for-loop return -1 in select statements. A bit hard to see.

skinowski: It basically avoids a failure case. After asyncChew, we would need to +1, but if it fails, then we drop the dequeued item, but we avoid that here by using the for-loop +1, then +1 another one after asyncRun() launch.

api/common/wait_utils.go
```diff
@@ -42,57 +46,95 @@ import (
...
+	return r.closer
+}
+
+// AddSession manipulates the session counter by
+// adding or subtracting the delta value. Incrementing
+// the session counter is not possible and will set
+// return value to false if a close was initiated.
+// It's callers responsibility to make sure addition and
+// subtraction math is correct.
+func (r *WaitGroup) AddSession(delta int64) bool {
```
rdallman: i do think it would be nice to have `(r *WaitGroup) Done()` that does a -1 dec and doesn't check isClosed, the contract for AddSession is not straight forward and is effectively 2 contracts based on the positivity of the delta. in some places, we check the return value, in some places we don't. also, did not see any `AddSession(-2)`, only -1. then `Done()` has a very clear contract that it won't fail if the waiter has been closed and it won't return anything. i guess there is then simply a panic case for calling Done() one too many times.

skinowski: yeah, especially now +2 is gone.. I'll do another PR soon for this...

PR #938, opened 2018-04-12
fn: reduce lbagent and agent dependency

skinowski:
lbagent and agent code is too dependent. This causes
any changed in agent to break lbagent. In reality, for
LB there should be no delegated agent. Splitting these
two will cause some code duplication, but it reduces
dependency and complexity (eg. agent without docker)

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -121,7 +121,7 @@ type agent struct {
 
 // New creates an Agent that executes functions locally as Docker containers.
 func New(da DataAccess) Agent {
-	a := createAgent(da, nil).(*agent)
+	a := createAgent(da).(*agent)
 	if !a.shutWg.AddSession(1) {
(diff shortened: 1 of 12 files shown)


api/agent/agent.go
```diff
@@ -119,35 +121,30 @@ type agent struct {
...
+	a := createAgent(da, nil).(*agent)
 	if !a.shutWg.AddSession(1) {
 		logrus.Fatalf("cannot start agent, unable to add session")
 	}
 	go a.asyncDequeue() // safe shutdown can nanny this fine
 	return a
 }
 
-func createAgent(da DataAccess, withDocker bool, withShutWg *common.WaitGroup) Agent {
+func createAgent(da DataAccess, withShutWg *common.WaitGroup) Agent {
```
rdallman: there are no callers that call this with a non-nil `withShutWg` now, can we nuke that one too while we're here?

PR #939, opened 2018-04-12
migratex api uses tx now instead of db

rdallman:
we want to be able to do external queries outside of the migration itself
inside of the same transaction for version checking. if we don't do this, we
risk the case where we set the version to the latest but we don't run the
table creates at all, so we have a db that thinks it's up to date but doesn't
even have any tables, and on subsequent boots if a migration slides in then
the migrations will run when there are no tables. it was unlikely, but now
it's dead.


api/datastore/sql/sql.go
```diff
@@ -151,23 +151,43 @@ func newDS(ctx context.Context, url *url.URL) (*sqlStore, error) {
...
+	// can detect whether the db did not exist and insert the latest version of
+	// the migrations BEFORE the tables are created (it uses table info to
+	// determine that).
+	//
+	// we either create all the tables with the latest version of the schema,
+	// insert the latest version to the migration table and bail without running
+	// any migrations.
+	// OR
+	// run all migrations necessary to get up to the latest, inserting that version,
+	// [and the tables exist so CREATE IF NOT EXIST guards us when we run the create queries].
```
rdallman: mostly did this for adding this note about why things are the way they are, but thought up a dreamy bug while I was in here. 11/10 level nits appreciated, my brain is a strange place that's hard to make sense of.

skinowski: Is pg failing even with 'if not exists' guards in tests?

rdallman: oh. i made a dumb dumb. yay tests

rdallman: pg is not my friend... down the rabbit hole

rdallman: found the issue https://github.com/fnproject/fn/pull/939/commits/3fba04989d9a6785fcfa5fb53d694820195ff805

skinowski: lgtm

PR #942, opened 2018-04-13
fn: LB ch and naive fixes

skinowski:
*) Naive is now a naive RR algorithm.
*) Both now checks for ctx/timeout in each attempt.

### DIFF
--- a/api/runnerpool/ch_placer.go
+++ b/api/runnerpool/ch_placer.go
@@ -30,46 +30,47 @@ func (p *chPlacer) PlaceCall(rp RunnerPool, ctx context.Context, call RunnerCall
 	sum64 := siphash.Hash(0, 0x4c617279426f6174, []byte(key))
 	timeout := time.After(call.LbDeadline().Sub(time.Now()))
 	for {
+		runners, err := rp.Runners(call)
+		if err != nil {
+			logrus.WithError(err).Error("Failed to find runners for call")
+		} else {
+			i := int(jumpConsistentHash(sum64, int32(len(runners))))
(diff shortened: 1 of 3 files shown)


api/runnerpool/ch_placer.go
```diff
@@ -30,46 +30,47 @@ func (p *chPlacer) PlaceCall(rp RunnerPool, ctx context.Context, call RunnerCall
...
+			return models.ErrCallTimeoutServerBusy
+		}
+
+		// backoff
+		select {
+		case <-ctx.Done():
+			return models.ErrCallTimeoutServerBusy
+		case <-timeout:
+			return models.ErrCallTimeoutServerBusy
+		case <-time.After(common.MinDuration(retryWaitInterval, remaining)):
```
rdallman: it may be slightly better to use a context here that attenuates existing `ctx.Done()` with `context.WithDeadline(mine(retryWaitInterval, remaining))` to tame timer leaks (where the ticker is still 'live' after this select exists, hard on resources), but we're in a loop and then there's 2 contexts to manage so i tried to bite my tongue :)

kind of side bar (no need here) the backoff strategy and code in general is mostly the same in naive and ch and it seems like we could abstract it to only abstract the selection from a list of runners instead of listing & selecting & backing off; anyway, punt-able, thinking out loud..

rdallman:
@gviedma or @hhexo merge ok? (Tolga is locked out of GH and don't wanna stick him with rebase)

rdallman:
3 weeks without an objection, I wish everything I wanted in life went like that

PR #948, opened 2018-04-19
CloudEvents I/O format support.

treeder:
This adds CloudEvent format as I/O format. 

This now has support for ingestion of CloudEvents as input via HTTP as well as the function I/O format.

### DIFF
--- a/api/agent/call.go
+++ b/api/agent/call.go
@@ -56,22 +56,26 @@ const (
 
 func FromRequest(a Agent, app *models.App, path string, req *http.Request) CallOpt {
 	return func(c *call) error {
-		route, err := a.GetRoute(req.Context(), app.ID, path)
+		ctx := req.Context()
+		route, err := a.GetRoute(ctx, app.ID, path)
 		if err != nil {
 			return err
 		}
 
+		log := common.Logger(ctx)
 		// Check whether this is a CloudEvent, if coming in via HTTP router (only way currently), then we'll look for a special header
 		// Content-Type header: https://github.com/cloudevents/spec/blob/master/http-transport-binding.md#32-structured-content-mode
 		// Expected Content-Type for a CloudEvent: application/cloudevents+json; charset=UTF-8
-		contentType := req.Header.Get("Content-type")
+		contentType := req.Header.Get("Content-Type")
 		t, _, err := mime.ParseMediaType(contentType)
 		if err != nil {
-			return fmt.Errorf("Could not parse Content-Type header: %v", err)
-		}
-		if t == ceMimeType {
-			c.IsCloudEvent = true
-			route.Format = models.FormatCloudEvent
+			// won't fail here, but log
+			log.Debugf("Could not parse Content-Type header: %v", err)
+		} else {
+			if t == ceMimeType {
+				c.IsCloudEvent = true
+				route.Format = models.FormatCloudEvent
+			}
 		}
 
 		if route.Format == "" {
--- a/api/agent/call.go
+++ b/api/agent/call.go
@@ -49,11 +49,21 @@ type Param struct {
 }
 type Params []Param
 
+const (
+	CloudEventHeader = "FN_CLOUD_EVENT"
+)
+
 func FromRequest(a Agent, app *models.App, path string, req *http.Request) CallOpt {
 	return func(c *call) error {
 		route, err := a.GetRoute(req.Context(), app.ID, path)
 		if err != nil {
 			return err
+		}
+
+		// Check whether this is a CloudEvent, if coming in via HTTP router (only way currently), then we'll look for a special header
+		if req.Header.Get(CloudEventHeader) != "" {
+			c.IsCloudEvent = true
+			route.Format = models.FormatCloudEvent
 		}
 
 		if route.Format == "" {
@@ -257,6 +267,9 @@ func (a *agent) GetCall(opts ...CallOpt) (Call, error) {
 
 type call struct {
 	*models.Call
+
+	// IsCloudEvent flag whether this was ingested as a cloud event. This may become the default or only way.
+	IsCloudEvent bool `json:"is_cloud_event"`
 
 	da             DataAccess
 	w              io.Writer
--- a/api/agent/call.go
+++ b/api/agent/call.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"mime"
 	"net/http"
 	"strings"
 	"time"
@@ -50,7 +51,7 @@ type Param struct {
 type Params []Param
 
 const (
-	CloudEventHeader = "FN_CLOUD_EVENT"
+	ceMimeType = "application/cloudevents+json"
 )
 
 func FromRequest(a Agent, app *models.App, path string, req *http.Request) CallOpt {
@@ -61,7 +62,14 @@ func FromRequest(a Agent, app *models.App, path string, req *http.Request) CallO
 		}
 
 		// Check whether this is a CloudEvent, if coming in via HTTP router (only way currently), then we'll look for a special header
-		if req.Header.Get(CloudEventHeader) != "" {
+		// Content-Type header: https://github.com/cloudevents/spec/blob/master/http-transport-binding.md#32-structured-content-mode
+		// Expected Content-Type for a CloudEvent: application/cloudevents+json; charset=UTF-8
+		contentType := req.Header.Get("Content-type")
+		t, _, err := mime.ParseMediaType(contentType)
+		if err != nil {
+			return fmt.Errorf("Could not parse Content-Type header: %v", err)
+		}
+		if t == ceMimeType {
 			c.IsCloudEvent = true
 			route.Format = models.FormatCloudEvent
 		}
--- a/api/agent/protocol/cloudevent.go
+++ b/api/agent/protocol/cloudevent.go
@@ -63,35 +63,49 @@ func (h *CloudEventProtocol) writeJSONToContainer(ci CallInfo) error {
 		return err
 	}
 
-	in := cloudEventIn{
-		CloudEvent: CloudEvent{
-			ContentType: ci.ContentType(),
-			EventID:     ci.CallID(),
-			Extensions: map[string]interface{}{
-				// note: protocol stuff should be set on first ingestion of the event in fn2.0, not here
-				"protocol": CallRequestHTTP{
-					Type:       ci.ProtocolType(),
-					Method:     ci.Method(),
-					RequestURL: ci.RequestURL(),
-					Headers:    ci.Headers(),
-				},
-				"deadline": ci.Deadline().String(),
-			},
-		},
-	}
-	if buf.Len() == 0 {
-		// nada
-		// todo: should we leave as null, pass in empty string, omitempty or some default for the content type, eg: {} for json?
-	} else if ci.ContentType() == "application/json" {
-		d := map[string]interface{}{}
-		err = json.NewDecoder(buf).Decode(&d)
+	var in cloudEventIn
+	// fmt.Printf("is cloudevent? %v\n", ci.IsCloudEvent())
+	if ci.IsCloudEvent() {
+		// then it's already in the right format, let's parse it, then modify
+		err = json.Unmarshal(buf.Bytes(), &in)
 		if err != nil {
-			return fmt.Errorf("Invalid json body with contentType 'application/json'. %v", err)
+			return fmt.Errorf("Invalid CloudEvent input. %v", err)
 		}
-		in.Data = d
 	} else {
-		in.Data = buf.String()
+		in = cloudEventIn{
+			CloudEvent: CloudEvent{
+				ContentType: ci.ContentType(),
+				EventID:     ci.CallID(),
+			},
+		}
+		if buf.Len() == 0 {
+			// nada
+			// todo: should we leave as null, pass in empty string, omitempty or some default for the content type, eg: {} for json?
+		} else if ci.ContentType() == "application/json" {
+			d := map[string]interface{}{}
+			err = json.NewDecoder(buf).Decode(&d)
+			if err != nil {
+				return fmt.Errorf("Invalid json body with contentType 'application/json'. %v", err)
+			}
+			in.Data = d
+		} else {
+			in.Data = buf.String()
+		}
 	}
+	// todo: deal with the dual ID's, one from outside, one from inside
+	if in.Extensions == nil {
+		in.Extensions = map[string]interface{}{}
+	}
+	// note: protocol stuff should be set on first ingestion of the event in fn2.0, the http router for example, not here
+	in.Extensions["protocol"] = CallRequestHTTP{
+		Type:       ci.ProtocolType(),
+		Method:     ci.Method(),
+		RequestURL: ci.RequestURL(),
+		Headers:    ci.Headers(),
+	}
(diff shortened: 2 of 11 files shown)


api/agent/protocol/cloudevent.go
```diff
@@ -0,0 +1,183 @@
...
+
+// CloudEvent is the official JSON representation of a CloudEvent: https://github.com/cloudevents/spec/blob/master/serialization.md
+type CloudEvent struct {
+	CloudEventsVersion string                 `json:"cloudEventsVersion"`
+	EventID            string                 `json:"eventID"`
+	Source             string                 `json:"source"`
+	EventType          string                 `json:"eventType"`
+	EventTypeVersion   string                 `json:"eventTypeVersion"`
+	EventTime          time.Time              `json:"eventTime"` // TODO: ensure rfc3339 format
+	SchemaURL          string                 `json:"schemaURL"`
```
denismakogon: has to be a URL to data schema according to the spec https://github.com/cloudevents/spec/blob/master/spec.md#schemaurl

api/agent/protocol/cloudevent.go
```diff
@@ -0,0 +1,183 @@
...
+	buf.Reset()
+	defer bufPool.Put(buf)
+
+	_, err := io.Copy(buf, ci.Input())
+	if err != nil {
+		return err
+	}
+
+	in := cloudEventIn{
+		CloudEvent: CloudEvent{
```
denismakogon: event type is required according to the spec https://github.com/cloudevents/spec/blob/master/spec.md#eventtype

denismakogon: cloud event version is required according to the spec https://github.com/cloudevents/spec/blob/master/spec.md#cloudeventsversion

denismakogon: source of the event is required as well according to the spec, this could be derived from the event type, so, if function was triggered with an HTTP request then we need to put a Host reader here.

api/agent/protocol/cloudevent.go
```diff
@@ -0,0 +1,183 @@
...
+				},
+				"deadline": ci.Deadline().String(),
+			},
+		},
+	}
+	if buf.Len() == 0 {
+		// nada
+		// todo: should we leave as null, pass in empty string, omitempty or some default for the content type, eg: {} for json?
+	} else if ci.ContentType() == "application/json" {
+		d := map[string]interface{}{}
```
denismakogon: i'd better leave this up to a function instead of doing this on the Fn side.
If you'd take a look at Azure EventGrid event type (which is the closest one to CloudEvents), so that data should be a type of `interface{}`, there would be no need to decode the data until that is necessary by function's needs.

treeder: The spec says it should be decoded json.

api/agent/protocol/cloudevent.go
```diff
@@ -0,0 +1,183 @@
...
+		err = json.NewDecoder(buf).Decode(&d)
+		if err != nil {
+			return fmt.Errorf("Invalid json body with contentType 'application/json'. %v", err)
+		}
+		in.Data = d
+	} else {
+		in.Data = buf.String()
```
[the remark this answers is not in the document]
treeder: I think we can probably deal with common uses cases pretty easily, and we can probably reject binary for now.

api/agent/protocol/cloudevent.go
treeder: Fixed.

treeder:
This now has support for ingestion of CloudEvents as input via HTTP as well as the function I/O format.

api/agent/protocol/cloudevent.go
denismakogon: so, if client sends a message with content type: application/json, the data has to be a JSON, but should we care? why should we recheck if that is a real json object?

api/agent/protocol/cloudevent.go
```diff
@@ -0,0 +1,207 @@
...
+		Headers:    ci.Headers(),
+	}
+	in.Extensions["deadline"] = ci.Deadline().String()
+
+	fmt.Printf("SENDING IN: %+v\n", in)
+	return json.NewEncoder(h.in).Encode(in)
+}
+
+func (h *CloudEventProtocol) Dispatch(ctx context.Context, ci CallInfo, w io.Writer) error {
+	ctx, span := trace.StartSpan(ctx, "dispatch_json")
```
rdallman: s/dispatch_json/dispatch_cloudevent/g (there's a few in this file)

api/agent/protocol/cloudevent.go
rdallman: that's just a string though, at least https://golang.org/pkg/net/url/#URL doesn't have UnmarshalJSON so we need a custom type (from non-stdlib package or our own)

rdallman: we should validate a few of these, for sure*

api/agent/protocol/cloudevent.go
```diff
@@ -0,0 +1,207 @@
...
+	// note: protocol stuff should be set on first ingestion of the event in fn2.0, the http router for example, not here
+	in.Extensions["protocol"] = CallRequestHTTP{
+		Type:       ci.ProtocolType(),
+		Method:     ci.Method(),
+		RequestURL: ci.RequestURL(),
+		Headers:    ci.Headers(),
+	}
+	in.Extensions["deadline"] = ci.Deadline().String()
+
+	fmt.Printf("SENDING IN: %+v\n", in)
```
rdallman: rm

api/agent/protocol/cloudevent.go
```diff
@@ -0,0 +1,207 @@
...
+		if buf.Len() == 0 {
+			// nada
+			// todo: should we leave as null, pass in empty string, omitempty or some default for the content type, eg: {} for json?
+		} else if ci.ContentType() == "application/json" {
+			d := map[string]interface{}{}
+			err = json.NewDecoder(buf).Decode(&d)
+			if err != nil {
+				return fmt.Errorf("Invalid json body with contentType 'application/json'. %v", err)
+			}
+			in.Data = d
```
rdallman: so we're copying the input to a buffer, decoding it and re-encoding it just to get rid of quoted strings in a json?

https://play.golang.org/p/HSk9Enz2Zp5

a note would be nice. not sure why we need all the special json treatment.

rdallman: i see that it's making the data line a string instead of an object. the intent of the cloud event seems to be to let the final thing processing the cloud event parse the data section as specified by the content type, parsing the data field as a string json blob (i.e. `json.NewDecoder(strings.NewReader(cloudEvent.Data)).Decode(&event)`) seems to be the intent, not for us to do that? a slippery slope of us doing special treatment for other content types no?

rdallman: i get that it saves an extra decode, i still am not sure it's worth it, the intent or the right thing to do (but you've been staring at the spec more than me, surely)

api/agent/protocol/cloudevent.go
```diff
@@ -0,0 +1,207 @@
...
+	// absolute (if set). it is expected that if users want to set multiple
+	// values they put it in the string, e.g. `"content-type:"application/json; charset=utf-8"`
+	// TODO this value should not exist since it's redundant in proto headers?
+	if jout.ContentType != "" {
+		rw.Header().Set("Content-Type", jout.ContentType)
+	}
+
+	// we must set all headers before writing the status, see http.ResponseWriter contract
+	if p != nil && p["status_code"] != nil {
+		sc := p["status_code"].(float64)
```
rdallman: `sc, ok` probably here too. if we could munge some types into here at some point this seems like a good place for https://golang.org/pkg/encoding/json/#Number

api/agent/protocol/cloudevent.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	var in cloudEventIn
+	// fmt.Printf("is cloudevent? %v\n", ci.IsCloudEvent())
+	if ci.IsCloudEvent() {
+		// then it's already in the right format, let's parse it, then modify
+		err = json.Unmarshal(buf.Bytes(), &in)
+		if err != nil {
+			return fmt.Errorf("Invalid CloudEvent input. %v", err)
+		}
```
rdallman: if they send in a cloud event we aren't setting it to the call id, they could leave it blank or use their own (other relevant fields like content-type, event-type, cloud-events-version, source). i think this is the intent most likely of the spec (to let the caller set these) and since this has mixed responsibilities at the moment of generating cloud events or taking them, maybe this is ok, just pointing this out if we want to fix this.

api/agent/protocol/cloudevent.go
rdallman: anyway, i guess this is going to get taken back out to the barn and shot soon enough anyway along with everything else in the 'if !IsCloudEvent` so just like my parents you can ignore me

api/agent/protocol/cloudevent.go
```diff
@@ -0,0 +1,207 @@
...
+		rw.WriteHeader(int(sc))
+	}
+
+	if ci.IsCloudEvent() {
+		// then it's already in the right format so just return it as is
+		d, err := json.Marshal(jout)
+		if err != nil {
+			return fmt.Errorf("Error marshalling CloudEvent response to json. %v\n", err)
+		}
+		_, err = rw.Write(d)
```
rdallman: `json.NewEncoder(rw).Encode(jout)` saves some bytes, afaict from http code there's no difference as both will write into the response writers `bufio.Writer` and there's no framing risk (afaict)

rdallman:
seems like build is broked `FAIL	github.com/fnproject/fn/api/agent/protocol` in CI, didn't see the actual error but must be something. 

LGTM w/ some tidying. mostly glossing over the fn-generates-the-cloud-event stuff since that's gonna get taken out to pasture. would be nice to clean up the printlns that are commented/not and the trace names need a quick sed, the rest are probably mostly me trying to drink enough coffee before noon.

api/agent/protocol/cloudevent.go
rdallman: I have found our answer I believe https://github.com/cloudevents/spec/blob/master/http-transport-binding.md#22-data-attribute

api/agent/call.go
```diff
@@ -49,13 +49,23 @@ type Param struct {
...
+
 func FromRequest(a Agent, app *models.App, path string, req *http.Request) CallOpt {
 	return func(c *call) error {
 		route, err := a.GetRoute(req.Context(), app.ID, path)
 		if err != nil {
 			return err
 		}
 
+		// Check whether this is a CloudEvent, if coming in via HTTP router (only way currently), then we'll look for a special header
+		if req.Header.Get(CloudEventHeader) != "" {
```
rdallman: the preferred method here https://github.com/cloudevents/spec/blob/master/http-transport-binding.md#22-data-attribute seems to be to use the `Content-Type` header `application/cloudevent`

rdallman:
i think we can punt on implementing the HTTP binary spec, it's WIP anyway, we can wait for it to finalize. but i do think we should use the recommended header and not do the content type funny business, per the same WIP spec. hate to start building tooling around this for all of that to have to change, there's a chance that part of the spec won't change but we're guaranteed that our current implementation doesn't implement the spec.

api/agent/protocol/cloudevent.go
treeder: We're not rechecking, this just makes sure it's part of the full object that's going into the function and not the escaped string.

api/agent/protocol/cloudevent.go
```diff
@@ -0,0 +1,183 @@
...
+	_, err := io.Copy(buf, ci.Input())
+	if err != nil {
+		return err
+	}
+
+	in := cloudEventIn{
+		CloudEvent: CloudEvent{
+			ContentType: ci.ContentType(),
+			EventID:     ci.CallID(),
+			Extensions: map[string]interface{}{
```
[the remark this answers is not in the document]
treeder: Would make sense. Maybe it's something left up to the users.

api/agent/protocol/cloudevent.go
treeder: Fixed.

api/agent/protocol/cloudevent.go
treeder: Fixed.

api/agent/protocol/cloudevent.go
treeder: The serialization page just changed, but the new page is says the same as the old. The "data" field is not an escaped string, it's the full JSON as part of the whole object: https://github.com/cloudevents/spec/blob/master/json-format.md#32-examples

treeder: So it's not left up to the final thing.

api/agent/protocol/cloudevent.go
treeder: done and returns error if invalid type.

api/agent/protocol/cloudevent.go
treeder: Ya, the caller should have at least set the required fields. We'll probably have to store a separate call_id and event_id field in our db to correlate, but we can get to that in fn2.0.

api/agent/protocol/cloudevent.go
treeder: Done.

api/agent/protocol/cloudevent.go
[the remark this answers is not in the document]
treeder: BTW, they do have a binary JSON example in here: https://github.com/cloudevents/spec/blob/master/json-format.md#32-examples . 

    "contentType" : "application/vnd.apache.thrift.binary",
    "data" : "... base64 encoded string ..."

Easy (ie: we don't have to do anything special).

api/agent/protocol/cloudevent.go
rdallman: the spec is pretty clear https://github.com/cloudevents/spec/blob/master/json-format.md#31-special-handling-of-the-data-attribute

api/agent/protocol/cloudevent.go
rdallman: got it, thanks for link

rdallman:
> Ehh, not sure why this build is failing:

rebase on master? #949 locked mysql to mysql 5.x (:latest is now 8)

treeder:
All that stuff looks so much like all the crap we've been dealing with using different formats. The part I find strange though is that if you have to set all the headers anyways, why not just pass the event in the JSON event format?  Anyways, I'll change the content-type to: `Content-Type: application/cloudevents+json; charset=UTF-8`, good find.

rdallman:
> All that stuff looks so much like all the crap we've been dealing with using different formats. The part I find strange though is that if you have to set all the headers anyways, why not just pass the event in the JSON event format? 

I'd guess so that you don't have to eat base64 decode time + byte bloat. I agree, it smells exactly like the last year of us fighting all the headers and such, not sure how best to give them the last year of dealing with that as contextual feedback before 1.0 lands. For us, it might mean we would need to handle both structured and binary in front end + container i/o + all fdks presumably -- yikes (i guess this is mitigated to front end if we translate everything to structured there, but it would mean base64 or similar). I'm assuming they'll add a protobuf transport spec, eventually, as well, so we will soon enough likely get into this, but I guess we can do all the translation to structured json format from the ingestion point and only worry about that format for functions (granted, base64 thing is still a nit). anyway... yea, would be nice to contribute to the discussion since this has ailed us.

PR #950, opened 2018-04-20
Fix missing values in context created with common.BackgroundContext

### LINKED ISSUE #949: Values added to the context by middlewares are lost through common.BackgroundContext()
During the execution path of a function call, we hang on to the context so that it can be cancelled (e.g. by timeout); however, when we execute a `call.End()` to finalize the state of the call in the DataAccess, that gets a new context with no timeout so it can complete regardless.

This new context is created by `common.BackgroundContext()`, which at the moment manually copies _some_ values from the original context. The problem with this approach is that if extension middlewares add stuff to the context that is then needed by extensions of the DataAccess, those values will _not_ be copied and the pair of "custom middleware + custom DataAccess" won't work.

There needs to be a way for `common.BackgroundContext()` to copy the values inserted by middlewares. Maybe we could store all middleware-inserted items into a specific bucket that can be identified by the code?

hhexo:
Fix missing values in context created with common.BackgroundContext by using delegation to previous context (only for Value()).

Fixes #949 .

### DIFF
--- a/api/common/ctx.go
+++ b/api/common/ctx.go
@@ -2,10 +2,9 @@ package common
 
 import (
 	"context"
+	"time"
 
 	"github.com/sirupsen/logrus"
-	"go.opencensus.io/tag"
-	"go.opencensus.io/trace"
 )
 
 type contextKey string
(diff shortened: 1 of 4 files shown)


api/common/ctx.go
```diff
@@ -34,25 +33,34 @@ func LoggerWithFields(ctx context.Context, fields logrus.Fields) (context.Contex
 	return ctx, l
 }
 
+// contextWithNoDeadline is an implementation of context.Context which delegates
+// Value() to its parent, but it has no deadline and it is never cancelled, just
+// like a context.Background().
```
rdallman: 👍 cool idea

Issue #947, opened 2018-04-18
swagger def have 401? 


rdallman:
implementers of the FN API may add extensions to do auth and want to return 401, it's sort of an open ended question whether we want to do this in the OSS swagger file and dictate how APIs should do things like auth, but also things like rate limiting, limits, etc. the benefits being that we don't have any fragmentation of clients or APIs for these specific things (begs the 'but it's inevitable for some things' question). the downside being that we dictate this and it could be inflexible, and also that for auth specifically, default OSS won't return 401 probably ever, so it's just bloat but it is pretty easily ignored.

this is potentially philosophical and may merit wider discussion short of any changes to the swagger file.

Issue #885, opened 2018-03-23
Swagger API docs are missing fields


ostrain:
README.md links to http://petstore.swagger.io/?url=https://raw.githubusercontent.com/fnproject/fn/master/docs/swagger.yml, but it seems like those docs are missing a bunch of fields. For example, http://petstore.swagger.io/?url=https://raw.githubusercontent.com/fnproject/fn/master/docs/swagger.yml#/Apps/post_apps is missing the "name" field and http://petstore.swagger.io/?url=https://raw.githubusercontent.com/fnproject/fn/master/docs/swagger.yml#/Routes/post_apps__app__routes is missing path.

Additionally, the example values aren't real examples in many cases. Things like `"image": "string",` doesn't really tell me much about what sort of string I would use for a value.

rdallman:
some of the fields such as app.name and route.path it seems we need to add a 'required' line to their specification, there may be other issues in the spec too. thanks!

rdallman:
since we're re-using the models at the bottom like `App` / `Route` it's hard to add required to these, since for things like patch they are not required. maybe we need 2 separate models, one for creation, and one for update.

Issue #870, opened 2018-03-19
Proposal: Error handling in FDK's


treeder:
While working on a demo app, I realized we don't have a nice way of dealing with errors from functions. 

Our [current function output format](https://github.com/fnproject/fn/blob/master/docs/developers/function-format.md#output) doesn't have any way to respond with an error, other than using the `protocol` section, but obviously, that's very protocol specific and not the best way to do it. 

```json
{
  "body": "{\"some\":\"output\"}",
  "content_type": "application/json",
  "protocol": {
    "status_code": 200,
    "headers": {
      "Other-Header": ["something"]
    }
  }
}
```

I propose we add some error response fields to our core output format to handle these nicely (very similar to [Lambda's format](https://docs.aws.amazon.com/apigateway/latest/developerguide/handle-errors-in-lambda-integration.html)):

```
{
  "error": {
    "message": "an error occurred, blah, blah",
    "type": "Exception",
    "trace": "stack trace"
  }
}
```

Then the server can deal with these errors appropriately based on the protocol being used. eg: For the default HTTP route, this could return a `400` or `500` (?) with the error message in the body, optionally overridden via the protocol status code. How can a developer override things for certain protocols?  Perhaps with a special Error objects that allow them to do that. 

## FDK's

FDK's should all have a nice an easy way for a user to respond with an error for an invocation. 

* Node can do something like Lambda
* Go should allow an `error` return
* Java should allow a `throw Exception`
* etc.

denismakogon:
I'm up for this one. However, i am bit disagreed with the FDK part. I would say that for Golang i would pick `panic` as the only valid case that will make FDK fill the error section of the response body, ordinary error (`error.Error`) doesn't mean much to us except that something went wrong, but doesn't mean something similar to exceptions in Python or Java , it should just lead to HTTP 500 and error body.
For FDK-go, i got used to write a wrapper:
```go
func withError(_ context.Context, in io.Reader, out io.Writer) {
	err := myHandler(in, out)
	if err != nil {
		fmt.Fprintln(os.Stderr, err.Error())
		fdk.WriteStatus(out, 500)
		out.Write([]byte(err.Error()))
		return
	}
}
```
it does great job and let the handler to return the error.

For any other languages like Java, Python i would agree that any throwable object (exception) should make the FDK write the error section of the JSON response body.

treeder:
[...]

Issue #858, opened 2018-03-14
routes with path depth > 1 unable to do point operations


rdallman:
```
$ curl localhost:8080/v1/apps/test/routes/abcd/efg
{"error":{"message":"Path not found: /v1/apps/test/routes/abcd/efg"}}
```

list works:

```
$ curl localhost:8080/v1/apps/test/routes
{"message":"Successfully listed routes","next_cursor":"","routes":[{"app_name":"test","path":"/abcd/efg","image":"busybox","memory":128,"cpus":"","type":"sync","format":"default","timeout":30,"idle_timeout":30,"created_at":"2018-03-14T00:45:39.706Z","updated_at":"2018-03-14T00:45:39.706Z"}]}
```

rdallman:
appears like we just need to modify GET `/routes/:route` to be `/routes/*route` like PUT/PATCH friends

rdallman:
closing cuz routes is ded

PR #957, opened 2018-04-25
[Proposal] Fn 0.2

rdallman:
Here be the doc. This tries to make a lot of the API changes we need to make in one fell swoop. Perhaps this is a bad idea, we can likely stage things though to where it's edible, it's nice to have a cohesive plan (if this is such... there was an attempt). There are a number of open questions at the top of the swagger2.yaml file, if you have opinions that is a great place to start after reviewing the doc. additionally, large questions remain around the log/call API, and this is a good chance to make any changes there. This doc also doesn't handle explicitly creating a separate function registry API, we can cross that bridge when we get to it, and just have a uniform function API for now. The doc itself has all the content needed for context, won't add much more here.

Feedback is more than welcome from anyone (from nits to redesigning entire components), if you would like to work on any specific pieces then let's chat about it. we should maybe break this up into smaller issues to work on, once we're in general agreement, and have more involved discussions over details on those?

I forgot to validate my swagger file, but I don't think it really holds up the discussion, i was gonna link it to comb through a rendered version here but sorry... i'll try to fix this tomorrow, i'm at the point where i have to add printlns to the validator now, boo. it probably still needs some further massaging, can add along w/ any feedback here. 

relevant issues (possibly closing): #947 #858 #885 #870 #824 #819 #817 #785 #763 #714 #574 #595 #500 #446 #281 #155 #137

### DIFF
--- a/docs/fn-2.0/README.md
+++ b/docs/fn-2.0/README.md
@@ -329,37 +329,63 @@ Picks runner to run it on and passes along event
 
 ## CLI changes
 
-### fn deploy
+### NEW IN 2.0
 
-func.yaml barely changes:
+The main changes here are related to [splitting functions and routes](https://github.com/fnproject/fn/issues/817).
+A route is now called a trigger and the route/trigger definition is no longer part of the function definition as it was before.
+
+### Function Definition File
+
+`func.yaml` barely changes, just removes route/trigger specific things. 
+
+A function does not know about triggers so it only has the information required to run the function.
 
 ```yaml
 function:
   name: yodawg
-  image: hub.docker.com/fnproject/hello
+
+  runtime: ruby
+  entrypoint: ruby func.rb
   version: 0.0.7
   memory: 42
+```
 
+`fn run` can run a function with this file only.
+
+### Trigger Definition File
+
+A `trigger.yaml` file defines the mapping from event sources to functions. A trigger
+binds one event source to one function, for instance an HTTP route to a function.
+
+```yaml
 trigger:
-  - name: myApp/sayhello
-    type: http
+  name: sayhello
+  type: http
+  # http specific param:
+  path: /sayhello
+  func: https://fnreg.com/namespace/myfunc
 ```
 
-trigger in yaml is optional. you may specify multiple triggers for each
-function. if trigger is not specified, it can be specified at runtime:
+On `fn deploy`:
 
-`fn deploy --trigger-http sayhello`
+* `trigger.yaml` must exist
+* if `func.yaml` exists:
+  * build and push function image
+  * update function registry with function definition
+  * use pushed function as `func` param for trigger
+* if `func.yaml` does not exist, `trigger.yaml` must have the `func` param set
+* Update trigger with Trigger definition
 
-since namespace is not specified, this will create a function at:
+This will create a function at:
 
 ```
-/ns/_/funcs/yodawg
+/ns/_/funcs/myfunc
 ```
 
 and a trigger at:
 
 ```
-/ns/_/apps/myApp/triggers/sayhello
+/ns/_/apps/myapp/triggers/sayhello
 {
   name: sayhello
   func: yodawg:0.0.7
@@ -367,20 +393,6 @@ and a trigger at:
 }
 ```
 
-since the above `func: ` does not specify a host name, the func will be
-expected to exist on the fn service where this trigger exists. a full url is
-also possible, see below example.
-
-it is also possible to specify a `func.yaml` with a remote function, where the
-trigger will be deployed but the function in the working directory will not be
-built and pushed to a docker registry, say if we had a func:
-
-```
-trigger:
-  - name: /sayhello
-    type: http
-    func: hub.fnproject.io/funcytown/hello
-```
 
 in all of the above examples, a user will end up with a route to call:
 
@@ -395,6 +407,27 @@ triggers, and add namespaces.
 
 ## FDKs
 
+FDKs will work in the same manner as they do now and from a user perspective probably don't need to change.
+This will be almost exactly the same as how the current JSON format works, just different fields as defined here: 
+https://github.com/cloudevents/spec/blob/master/json-format.md
+
+Handlers will take the pseudo-form:
+
+```
+handle(context, input) (output, err)
+```
+
+This form will change depending on what's right for the language, but they should all have the same 
+meaning and generally the same feel:
+
+* `context` will include all of the CloudEvent fields other than the `data` field. 
+* `input` will be the `data` field
+* `output` is the response `data` field
+* `err` is if an error occurred and will be returned with the `CloudEvent.error` field set (see above)
+
+
+### NOTES
+
 There are some decisions to make around added sugar we want to sprinkle on top
 of cloud events. We also may want to allow users to receive/output raw cloud-events
 themselves if they would like to, or at least have a way to set each of the
@@ -403,37 +436,6 @@ add additional fields) -- of course, it's possible to not use FDK and do this,
 so my thinking is that we make FDKs more to simplify things than anything.
 We can add features we'd like to FDKs as we go down the road. My thinking for
 base level FDK functionality is:
-
-FDKs will simply handle the json version of cloud-events defined here: 
-https://github.com/cloudevents/spec/blob/master/json-format.md
-
-FDKs will in the same manner as now, decode one of these at a time into a
-$programming_language object from STDIN, after receiving one, will call a user
-specified handler function, and receive a $programming_language object as
-output, which it will then encode to json on STDOUT.
-
-Handlers will take the pseudo-form:
-
-```
-Handle(CloudEvent) CloudEvent
-```
-
-where the user is forced to construct a `CloudEvent` object. Obviously, we
-should have some helpful constructors like `NewCloudEvent(contentType, body string) CloudEvent` 
-that do things like handle the json magic and that fills in most of the
-fields. This seems like a viable option and is more flexible than just trying
-to scrape up the body and have some other kind of opaque object and set
-certain fields on it (as now).
-
-there are likely other ways, but trying to keep it simple out of the gate.
-note that this is a divergence from the current cloud-event implementation,
-where the entire user output is shoved into the data section at the end.
-there's also a possibility that I'm completely misguided on what FDKs should
-look like and if you feel that way please propose a comprehensive solution and
-I'd be delighted to see it. It does seem like FDKs will basically be a for
-loop, a cloud event object definition and json decoder / encoder, and a bunch
-of getter and setter methods (potentially, lang dependent). Maybe this doesn't
-provide as much utility as it once did, but that's for us to decide.
 
 ## Proposed dicing of the pieces
 
--- a/docs/fn-2.0/README.md
+++ b/docs/fn-2.0/README.md
@@ -1,0 +1,407 @@
+# Fn 2.0
+
+## NOTE: This is a proposal for updating Fn with the following goals:
+
+* Split HTTP routes from functions
+* Support CloudEvent messages as the native format throughout
+* Simplify everything
+* Modularize components
+
+## Removals
+
+* Default, HTTP and JSON formats will be removed
+* Cold functions will be removed
+
+## Terminology
+
+Fn follows the terminology defined here: https://github.com/cloudevents/spec/blob/master/spec.md#notations-and-terminology
+
+As well as the following:
+
+### Event Sources
+
+Event sources are things that emit events and pass the event messages to Fn. Three common event sources are:
+
+* HTTP
+* Scheduler
+* Queue
+
+### Functions
+
+Functions are little programs that accept a defined input and output a defined output. They do not need to be aware of any other parts
+of the system and can run standalone; from the command line, etc. These can be written in any language and are packaged up as a container image. 
+
+* Functions read from STDIN, write to STDOUT and log to STDERR
+* Functions are packaged in container images
+* Input is a stream of CloudEvent messages in JSON format
+* Output is a stream of CloudEvent messages in JSON format
+  * With some additions to deal with things like error handling and protocol specific responses (eg: function wants to return a specific http status code if it was fired from an HTTP event source.)
+
+#### Function Error Handling
+
+Errors will be returned as a CloudEvent with an “error” attribute, no “data” attribute, describing the error. 
+
+{
+  "cloud-events-version": "0.1",
+  "event-id": "6480da1a-5028-4301-acc3-fbae628207b3",
+  "source": "http://example.com/repomanager",
+  "event-type": "http-request",
+  "event-type-version": "v1.5",
+  "event-time": "2018-04-01T23:12:34Z",
+  "content-type": "application/json",
+  “error”: {
+    "message": "an error occurred, blah, blah",
+    "type": "Exception",
+    "trace": "stack trace"
+  },
+  “extensions”: {
+    “protocol”: {
+       “headers”: …
+       Etc, same as we do now for JSON format. 
+    }
+  }
+}
+
+Error map:
+* message - required
+* type - type of error, eg IOError - optional
+* trace - stack trace - optional
+
+### Triggers
+
+Triggers bind events to functions along with configuration.
+
+Eg: HTTP event
+
+```json
+{
+    "source": "http://myapp.com/x"
+}
+```
+
+Trigger will look up the source in the triggers db then update the event with the Function information plus the app and trigger configuration.
+
+```json
+{
+    "extensions": {
+        "trigger": {
+            "function_url": "func url",
+            "config": {
+                MAP OF CONFIGS FROM APP AND TRIGGER
+            }
+        }
+    }
+}
+```
+
+It will then pass this message along to the Fn Router/LB for execution.
+
+### Apps/Applications
+
+Applications are a grouping of triggers to make up a complete application. For example, an application might have the following triggers:
+
+* http: /x -> trigger maps /x -> function f1 with config c1
+* http: /y …
+* schedule: run function f2 every day at 8am
+* queue: listen to topic t1 filtering on ff1, for every event run function f3
+
+All of these things make up the application. 
+
+## Components
+
+The following components make up the Fn platform. 
+
+* All components can run separately
+* They all have their own main
+* One main main to tie them all into one bin for a nice developer experience.
+
+### Runner
+
+The Runner does nothing more than take an event message, execute the function and return results.
+
+* Main endpoint `/run` with CloudEvent as input.
+* When called, runs a function and returns the results, as a CloudEvent.
+* If busy, returns 503.
+* Event input contains everything it needs to run, including config/secrets.
+  * TODO: Need to be sure we don’t log any secrets
+* Logs can be retrieved separately via `/log` endpoint so main response can be returned faster.
+
+### Function Registry
+
+A service that gives functions an addressable URL, eg:
+
+```
+https://functions.somewhere.com/NAMESPACE/FUNCNAME:VERSION-TAG
+```
+
+* Has metadata required to run a function provided by the creator of the function
+* All functions registered here must accept and return CloudEvent messages
+* This leads to reusable/shareable functions with no dependencies on other components
+* Could potentially run as a separate service like Docker Hub
+
+Q: Seems like a lot to just be a small layer on top of Docker Hub?
+A: It may seem this way now for just a few fields, but down the road, there could many additional things
+  like forcing functions to pass certain tests (eg: I/O formats), defining expected input parameters (can reject 
+  before starting if input is wrong), etc.
+
+### API
+
+CRUD for all the things.
+
+* Apps
+* Triggers
+* Calls
+* Logs
+
+### Function Router/LB
+
+Deals with sending event messages to the best runner.
+
+Same main endpoint as a runner, `/run` with CloudEvent input.
+
+### Trigger Manager
+
+The trigger manager is the main entrypoint for an event to enter the Fn platform.
+
+Main endpoint is `/run` with CloudEvent input. 
+
+Flow:
+
+* Input is a CloudEvent.
+* Looks up matching trigger from API based on event.source (maybe other params too).
+* Calls out to LB (or Runner if no LB) to run function.
+
+See **Trigger** above for more information.
+
+### Event Sources
+
+#### HTTP Router
+
+Reads HTTP triggers to make routing table, mapping HTTP endpoints to triggers (or directly to functions?). 
+When matching request comes in, either through trigger manager or straight to LB/runner.
+
+#### Scheduler
+
+On schedule set by user, the trigger is fired.
+
+#### Queue
+
+Listens on a stream/topic/channel, can have a filter
+For every message that passes filter, fire the trigger
+
+## Example flow
+
+HTTP
+
+### HTTP Router receives request
+
+Request comes into HTTP router. 
+Router looks at routing table for match (set on triggers). 
+If no match, return 404.
+Convert request to CloudEvent, eg:
+
+```json
+{
+  "cloud-events-version": "0.1",
+  "event-id": "6480da1a-5028-4301-acc3-fbae628207b3",
+  "source": "http://example.com/repomanager",
+  "event-type": "http-request",
+  "event-type-version": "v1.5",
+  "event-time": "2018-04-01T23:12:34Z",
+  "content-type": "application/json",
+  "data": {
+     REQUEST_BODY
+  },
+  “extensions”: {
+    “protocol”: {
+       “headers”: …
+       Etc, same as we do now for JSON format. 
+    }
+  }
+}
+```
+
+Pass event to trigger manager.
+
+### Trigger manager (TM) receives event
+
+TM looks up the matching trigger.
+Trigger has function URI + config data. 
+TM also get app config data.
+Merges config and event to get this event:
+
+```json
+{
+  "cloud-events-version": "0.1",
+  "event-id": "6480da1a-5028-4301-acc3-fbae628207b3",
+  "source": "http://example.com/repomanager",
+  "event-type": "http-request",
+  "event-type-version": "v1.5",
+  "event-time": "2018-04-01T23:12:34Z",
+  "content-type": "application/json",
+  "data": {
+     REQUEST_BODY
+  },
+  “extensions”: {
+    “protocol”: {
+       “headers”: …
+       Etc, same as we do now for JSON format. 
+    },
+    “config”: {
+        ALL APP AND TRIGGER CONFIGS
+     }
+     “function”: “https://funcs.fnproject.io/NAMESPACE/FUNC:VERSION”
+  }
+}
+```
+
+Passes event to function router (aka LB) OR runner (whichever the trigger manager is configured for, they both have the same /run endpoint. Single machine setup would be direct to runner). 
+
+### Function Router gets event
+
+Picks runner to run it on and passes along event
+
+### Runner gets event
+
+Takes function URL and looks up metadata from function registry. 
+Executes the function
+Returns CloudEvent with response as “data” body, strips out all the extension stuff and replace with response specific info.
+
+```json
+{
+  "cloud-events-version": "0.1",
+  "event-id": "6480da1a-5028-4301-acc3-fbae628207b3",
+  "source": "http://example.com/repomanager",
+  "event-type": "http-request",
+  "event-type-version": "v1.5",
+  "event-time": "2018-04-01T23:12:34Z",
+  "content-type": "application/json",
+  "data": {
+     RESPONSE_BODY
+  },
+  “extensions”: {
+    “protocol”: {
+       “headers”: …
+       Etc, same as we do now for JSON format responses 
+    },
+  }
+}
+```
+
+## Models
+
+```go
+type Function struct {
+            ID string `json:”id” db:”id”`
+	// repo-host/Namespace/Name:Tag make up the fully qualified name
+	Namespace string `json:”namespace”`
+	Name      string `json:"name" db:"name"`
+Image    string `json:”image”`
+	Resources ResourceConfig `json:”resources” db:”resources”`
+Config Config `json:”config”` // ???
+// TODO tell me there's not annotations here too?
+	CreatedAt   strfmt.DateTime `json:"created_at,omitempty" db:"created_at"`
+	UpdatedAt   strfmt.DateTime `json:"updated_at,omitempty" db:"updated_at"`
+}
+```
+
+```go
+type ResourceConfig struct {
+	Memory     uint64         `json:"memory" db:"memory"`
+	CPUs        MilliCPUs    `json:"cpus" db:"cpus"` 
+	Disk 	      uint64 	 `json:”disk” db:”disk”
+	Timeout     int32           `json:"timeout" db:"timeout"`
+	IdleTimeout int32           `json:"idle_timeout" db:"idle_timeout"`
+}
+```
+
+```go
+type Trigger struct {
+	// The grouping of routes into an app is specific to this type of trigger
+	AppID string `json:"app_id" db:"app_id"`
+	// FunctionName a fully qualified function name referencing a function in a repository. eg: `funchub.fnproject.io/jimbo/somefunc:1.2.3`
+	FunctionName string `json:"function" db:"function"`
+	// Config is required by the function, but set in the trigger, ie: the caller.
+	Config Config `json:"config,omitempty" db:"config"`
+            // Override the function’s resource config
+Resources ResourceConfig `json:"config,omitempty" db:"config"`
+	// Annotations config for environment function will be running in
+	Annotations Annotations     `json:"annotations,omitempty" db:"annotations"`
+	CreatedAt   strfmt.DateTime `json:"created_at,omitempty" db:"created_at"`
+	UpdatedAt   strfmt.DateTime `json:"updated_at,omitempty" db:"updated_at"`
+}
+```
+
+```go
+type CloudEvent struct {
+	EventType string `json:”event-type”`
+	EventTypeVersion string `json:”event-type-version”`
+	CloudEventsVersion string `json:”cloud-events-version”`
+	Source string `json:”source”`
+EventID string `json:”event-id”`	
+	EventTime strfmt.DateTime `json:”event-time”`
+	SchemaURL string `json:”schema-url”`
+	ContentType string `json:”content-type”`
+	Extensions map[string]interface{} `json:”extensions”` // map[string]string ?
+Data string `json:”data”`
+}
+```
+
+```go
+type App struct {
+	ID string `json:”id”`
+	Name string `json:”name”`
+	Config Config `json:”config”`
+	Annotations Annotations `json:”annotations”`
+}
+```
+
+### API Definitions (all components)
+
+```
+// :ns is a URL safe user-generated string, e.g. `war-party`
+GET /ns
+GET /ns/:ns
+PUT /ns/:ns
+DELETE /ns/:ns
+
+// :app is a URL safe user-generated string, e.g. `operation-kino`
+GET /ns/:ns/apps
+GET /ns/:ns/apps/:app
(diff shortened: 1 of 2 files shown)


docs/swagger2.yml
```diff
@@ -0,0 +1,1086 @@
...
+      id:
+        type: string
+        description: App ID
+        readOnly: true
+      name:
+        type: string
+        description: namespace this function resides in.
+      annotations:
+        type: object
+        description: Namespace annotations - this is a map of annotations attached to this namespace, keys must not exceed 128 bytes and must consist of non-whitespace printable ascii characters, and the seralized representation of individual values must not exeed 512 bytes
```
hibooboo2: How many keys can they have?

docs/swagger2.yml
```diff
@@ -0,0 +1,1086 @@
...
+  - http
+# will be prefixed to all paths
+# TODO(discuss): should we bump this?
+basePath: /v42
+consumes:
+  - application/json
+produces:
+  - application/json
+paths:
+  /ns:
```
hibooboo2: really excited to add namespacing. 👍 But do we need to make all endpoint now have the extra `/ns/` in front of their namespace and not just do `/{namespace}/apps/`. A little nitpicky i know but the `/ns/` seems wasteful. And then put namespace management under a different prefix.. Not sure where but just `/ns/` 👎  for me.  Since namespaces too me seem like a more permanent fixture and will not move around much.  But maybe thats just me.

hibooboo2: Will the fn server have extensions to be able to manage what happens on namespace creation deletion etc...? Seems like it should... Since that will affect things like auth for a FAAS that is using fn.. At least I feel like it probably will.

docs/fn-2.0/README.md
```diff
@@ -0,0 +1,416 @@
...
+```
+{
+  "cloud-events-version": "0.1",
+  "event-id": "6480da1a-5028-4301-acc3-fbae628207b3",
+  "source": "http://example.com/repomanager",
+  "event-type": "http-request",
+  "event-type-version": "v1.5",
+  "event-time": "2018-04-01T23:12:34Z",
+  "content-type": "application/json",
+  “error”: {
```
hibooboo2: CloudEvents spec does not seem to have Error as part of it... Would be nice if we could A stick to the format or B Get errors added to the spec if the team in charge of it seems to agree that it would be important.

docs/fn-2.0/README.md
```diff
@@ -0,0 +1,416 @@
...
+passes them off to trigger manager / LB / runner.
+
+#### Scheduler
+
+On schedule set by user, the trigger is fired.
+Scheduler would have its own API to manage schedule -> trigger mappings.
+
+#### Queue
+
+Listens on a stream/topic/channel, can have a filter
```
hibooboo2: For queue what happens with the response if there is any?

docs/fn-2.0/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+### Functions
+
+Functions are little programs that accept a defined input and output a defined output. They do not need to be aware of any other parts
+of the system and can run standalone; from the command line, etc. These can be written in any language and are packaged up as a container image. 
+
+* Functions read from STDIN, write to STDOUT and log to STDERR
+* Functions are packaged in container images
+* Input is a stream of CloudEvent messages in JSON format
+* Output is a stream of CloudEvent messages in JSON format
```
denismakogon: We do a lot of construction in order to write back a JSON representation of the CloudEvent. Shouldn't protocol dispatcher be responsible for constructing the CloudEvent object from the function's response?

docs/fn-2.0/README.md
denismakogon: That's true. If we want to return an error from the function we definitely need to use extensions.
At this moment the function is responsible for sending back complete CloudEvent object, which looks quite complex and unnecessary because the protocol dispatcher can do that for each function's call.
IMHO there is no strong requirement for a function to write back complete CloudEvent.

docs/fn-2.0/README.md
```diff
@@ -0,0 +1,416 @@
...
+       “headers”: …
+       Etc, same as we do now for JSON format. 
+    }
+  }
+}
+```
+
+Error map:
+* message - required
+* type - type of error, eg IOError - optional
```
denismakogon: i think this one is redundant because the `message` is the most verbose representation of the error type, and if the trace is present the `type` will stay unused.

docs/fn-2.0/README.md
```diff
@@ -0,0 +1,416 @@
...
+```
+
+* Has metadata required to run a function provided by the creator of the function
+* All functions registered here must accept and return CloudEvent messages
+* This leads to reusable/shareable functions with no dependencies on other components
+* Could potentially run as a separate service like Docker Hub
+
+Q: Seems like a lot to just be a small layer on top of Docker Hub?
+A: It may seem this way now for just a few fields, but down the road, there could many additional things
+  like forcing functions to pass certain tests (eg: I/O formats), defining expected input parameters (can reject
```
denismakogon: FunctionsHub is not just a docker registry. If we'd talk about public hub - it has to be similar to Docker EE hub with trusted function.

I see a FunctionsHub as a combination of Github and DockerHub. You need to see function's source code and the way it's packaged (would that be just a func file or Dockerfile).

docs/fn-2.0/README.md
denismakogon: Comparing to other serverless offerings, queue triggering is completely async (good example here would be an AWS SNS).

docs/fn-2.0/README.md
```diff
@@ -0,0 +1,416 @@
...
+### fn deploy
+
+func.yaml barely changes:
+
+```yaml
+function:
+  name: yodawg
+  image: hub.docker.com/fnproject/hello
+  version: 0.0.7
+  memory: 42
```
denismakogon: Can't we add registry auth here? See https://github.com/fnproject/fn/issues/955

docs/fn-2.0/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+trigger:
```
denismakogon: shouldn't `triggers` be a list of actual triggers?
```yaml
triggers:
 - name: myApp/sayhello
   type: http
 - name: myApp/sayhello
   type: queue
```

docs/fn-2.0/README.md
```diff
@@ -0,0 +1,416 @@
...
+of cloud events. We also may want to allow users to receive/output raw cloud-events
+themselves if they would like to, or at least have a way to set each of the
+fields and extensions (the catch being if users can define their own they may
+add additional fields) -- of course, it's possible to not use FDK and do this,
+so my thinking is that we make FDKs more to simplify things than anything.
+We can add features we'd like to FDKs as we go down the road. My thinking for
+base level FDK functionality is:
+
+FDKs will simply handle the json version of cloud-events defined here: 
+https://github.com/cloudevents/spec/blob/master/json-format.md
```
denismakogon: there is a discussion to trigger a function with more than one event, how would we handle this case?

docs/fn-2.0/README.md
```diff
@@ -0,0 +1,416 @@
...
+
+The Runner does nothing more than take an event message, execute the function and return results.
+
+* Main endpoint `/run` with CloudEvent as input.
+* When called, runs a function and returns the results, as a CloudEvent.
+* If busy, returns 503.
+* Event input contains everything it needs to run, including config/secrets.
+  * TODO: Need to be sure we don’t log any secrets
+* Logs can be retrieved separately via `/log` endpoint so main response can be returned faster.
+
```
reclaro: As regards Logs, do you think that revisiting the logging is in scope for the new version or not? Maybe I am wrong but the logging experience at the moment is not the best one from a customer point of view and from an operator. I think we can't build a system which has a fully completed logging solution but there were some interesting proposals (https://github.com/fnproject/fn/issues/337) for adding hooks so as it will be possible to offload logging to an external backend and some other open questions about the format of the log (https://github.com/fnproject/fn/issues/536).

docs/fn-2.0/README.md
hibooboo2: @denismakogon I disagree the container should send entire cloud event back on stdout since it consumes an entire one. It should just be able to easily create them using the fdk so that alot of the heavy lifting is done by the fdk.

docs/fn-2.0/README.md
hibooboo2: So is the idea with queues then that any calls have a side affect that is either in logs or in a db somewhere etc?

docs/fn-2.0/README.md
hibooboo2: Could even make it so that the metadata is in a file in the git repo and its only accepted to the Functions hub if it is in the git repo.. Then we can use any git repo link + docker image link. Then it can be deployed on prem and use private repos / images or on public cloud. Just my 2 cents.

docs/fn-2.0/README.md
hibooboo2: Could always make it so that the cli picks up what is in the config for the function and uses that as the info for deploy.. Then we do not change the func file. Darn secrets.

docs/fn-2.0/README.md
hibooboo2: Why would one triggered invocation of a function take two events? Can you explain a use case here?

denismakogon: AFAIR the idea is to have a threshold for a function invocation, i.e. the action (function call) must happen only when N events appear. In real-world deployment function, most of the times get triggered by an Event Service. It capable of defining the threshold for a function invocation.

Another use case is butch event distribution (azure supports that, azure event grid is the closest one to cloudevent spec).

P.S.: can't find cloudevent doc where that discussion happened.

hibooboo2: Hmm well I mean we could handle batch events and fire them one at a time into the container other wise make the functions spec be that you consume a list of cloud events and output a list of them as well.

hibooboo2: Seems wierd / cumbersome to handle 1 or a list though.

docs/fn-2.0/README.md
```diff
@@ -0,0 +1,416 @@
...
+* Split HTTP routes from functions
+* Support CloudEvent messages as the native format throughout
+* Simplify everything
+* Modularize components
+* Namespacing
+
+## Removals
+
+* Default, HTTP and JSON formats will be removed
+* Cold functions will be removed
```
rikgibson: How are functions/containers not built with an FDK supported, i.e. ones that accept STDIN only? I've read or been part of discussions around creating `hot wrappers` that get injected or wrapped around a "cold" container to turn it into a "hot" one - think @shaunsmith had something almost working - but no final solution or proposal yet.

docs/fn-2.0/README.md
treeder: Part of the spec is that you can add whatever you want to it. Recommended to be in extensions, but I think errors are important enough to be at the top level.

docs/fn-2.0/README.md
treeder: Lambda has these three items, they all seem useful to me.

docs/fn-2.0/README.md
treeder: IMO, I don't think we can have a service that doesn't have logs. Nobody would use it.

docs/fn-2.0/README.md
treeder: This could be private or public and source code doesn't need to be available.

docs/fn-2.0/README.md
treeder: Yes, there is no response.

docs/swagger2.yml
rdallman: yes, there will be extensions on CRUD for every API action, can clarify this, thanks. 

all of our resources specify which resource id follows it, e.g. `/functions/{function}`, the suggestion there would make namespace kind of an odd one out. This decision doesn't seem very consequential, we just need to be consistent I think, most people interface with the API through bindings, CLI and UI anyway so vanity matters a little less at this level; e.g. the UI can still have pretty URLs without `/thing/:thing` if we want that.

rdallman: additionally, we need a way for users to list namespaces (most implementations will attenuate this with some authentication layer) and putting this at `/ns` is the RESTful, natural way to do this.

docs/fn-2.0/README.md
rdallman: re: errors, i plan to open an issue shortly here https://github.com/cloudevents/spec/issues to discuss this with them, this seems like something that is useful to add. but also, yeah, the spec mentions you can add fields outside of extensions, but hopefully we can get first class support soon enough and use that. stay posted.

docs/fn-2.0/README.md
rdallman: one thought was that we could root logs on `/apps/:app/logs`, which could be more useful to users to get a 'stream' of logs across all triggers on their app vs. having to pull all of these out of individual calls (there is a TODO in the swagger for this) -- this is not trivial to implement in s3 afaik (no append). largely, this is unanswered. I'm open to ideas on improvements. From an OSS angle, to a certain degree all we can do is offer an API that is reasonable to implement and has a decent out of the box options (s3/minio), most end users of this will need something more integrated and robust than what we can offer (maybe? we could have a good OSS story around ELK instead of minio too, and maybe that's OK). I'm not sure our current offering is great and open to discussion there; this is a good opportunity to fix it even if it's largely unrelated to the core changes here. part of me thinks that a lot of the answers here are 'better tooling', most logs seem likely to go unread or streamed elsewhere to be analyzed (syslog hooks is in progress, but we need a good way to debug without that).

docs/fn-2.0/README.md
rdallman: i guess it's an option. registry auth is really a hairball since valid image names don't allow you to add creds to them, docker separated auth from the images. and there's lots of product and usability questions to answer there (e.g. if my namespace has my registry name and password already, why do i have to log in for each?) -- can we please punt on this one? it's a hair ball imo and there's enough stuff to do here (and also a non-feature atm).

docs/fn-2.0/README.md
rdallman: yea, good idea.

docs/fn-2.0/README.md
hibooboo2: Ok :)

docs/fn-2.0/README.md
hibooboo2: Interesting curious to see how we implement that..

hibooboo2: Ex: if there is no response how do we know when the container has dealt with the event.

docs/fn-2.0/README.md
hibooboo2: Yeah I agree better tooling. @treeder I was not saying to not have logs. I was saying to by default not store them.. And make it simple to change that.. Because logs can really eat up disk space and storing them locally by default in db seems meh to me. Or have an option to not store one of those would be nice.

Either way current way is not that useful IMO

docs/fn-2.0/README.md
hibooboo2: Right so I meant that git repo could be optional but to not locking on requiring docker hub specifically. Just the meta data and the docker image. But having a git repo that is linked to it as an option would be very useful :)

docs/fn-2.0/README.md
rdallman: At least initially the plan here was: 

giphy

we had some discussion around letting functions shoot themselves via extensions to get 'cold' behavior, but that would not itself solve the 'take any function that does stdin / stdout and run it like normal for 1 invocation', it would only solve the 'cold' thing.

seemingly, every other FaaS product has its own format a user must use for input/output and we would just be joining the club. Seems like we're mostly re-opening https://github.com/fnproject/fn/issues/785 discussion here, which was contentious. it would be really nice to only have 1 format, there's been quite a large amount of pain supporting multiple formats and heat levels, over time between the backend as well as for FDKs, CLI, and simply to users.

hibooboo2: Yeah there is a lot of conflicting users stories around Fn going around right now.

docs/fn-2.0/README.md
rdallman: i guess this would be up to the event source to manage that threshold and fire when ready (not fn), so to that we can now say 'sure', but it would have to fire multiple discrete events still there's no batching. since we have a 1:1 mapping of an event to a container at the moment, it seems weird to start trying to add guarantees around running multiple events through the `/run` endpoint at this point.

docs/fn-2.0/README.md
rdallman: imo it was kind of a non-goal to design a functions hub here. we just need to split functions out from the way we have it now, and have triggers set up to be able to address functions by name. we can design function hub down the road, but is it nice contextually to have that in mind here and it shows how powerful this abstraction could be.

docs/fn-2.0/README.md
rdallman: I think that internally the runner module would generate a response (it treats all invocations the same) and it's up to the queue event source to either toss out the response or it can do whatever it wants, like, say, put a message onto a different topic with the response or just the call id or post it to imgur. out of the box, we plan to have a 'queue trigger' that would probably act like our async does now, but conceptually it's not quite the same since it would be queueing events to the queue and not through fn (we can add API sugar to the queue source code if we want)

docs/fn-2.0/README.md
rikgibson: The conversation I had with @shaunsmith, we were talking about whether we can have a small binary that gets injected into a container (or the original container gets wrapped) and becomes the new entry point, takes the json/cloudevent data off the wire and then marshalls it to stdin, launching the original function entry point as a process in-container, effectively achieving cold-like behaviour from the function's perspective.  The stdin/stdout any-container any-language pitch always plays strongly at the meetups.

hibooboo2: That maybe true but the container still has to be written to use our signatures no matter what wrapping is done so it still does not provide and real value IMO. Any container may be a function but not not all containers are functions. Where as all functions are containers. Squares and Rectangles.

rdallman: @rikgibson that seems like a good plan, openfaas does similar to that. that pushes that logic to have about as small of a surface area as possible, the functions handling it individually.

rdallman:
re: @zootalures thanks, very useful feedback.

> Would this change benefit from some black-box /narrative use cases that walked through the specific new user features that it enabled? 

yes, good idea, i think we can write up something that briefly summarizes this. mostly, we should focus on end user documentation i hope instead of focusing on the context of the delta as the former will be more valuable long term.

> Also on the function registry stuff - would be good to have some lucidity by example on the good stuff that this enables (i.e. sharing functions between platforms? function hub?)

pretty much, function registry kind of a non-goal here, i can't honestly say i thought very deeply about how the exact interface should look nor if we're perfectly set up for this even with what we have, but it's something that should be near term possible if we can land some of the stuff proposed here, possibly with some minor tweaks.

> For queue triggers, can you sketch out how this could/might work with a Kafka/Kinesis queue - how do I authenticate, how to I connect etc?

added to doc, it's kinda hand wavy especially wrt auth because i'm not sure it's something we should dictate, but e.g. you would start up your queue processor with a token or permission that allows it to invoke triggers for a given app/namespace/function, the rest we're not dictating anything other than a trigger existing for that trigger type + name and a cloud event containing that info. 

> For namespaces - is the intention to enable real multi-tenancy here How will access control work in that case/ how do I integrate/mix this with an existing access control system.

we could try to sketch out an example, yeah the goal is opaquely another level of compartmentalization which vaguely maps onto something that could be used to do multi tenancy. we're more giving people a hammer here than telling them how to hang their pictures. obviously you get some sugar in that all our queries and such internally use namespaces so users wouldn't have to go rewrite all the queries to add their ideas around tenancy via extensions (hopefully?). do we just need to make a note that this could be used for multi-tenancy (i.e. the intent is not clear)?

> You're specifying components here - trigger manager seems a bit fluffy - could you clarify (e.g. for a kafka queue above) how does this scale up ?

added some bits. components are not necessarily separate processes is maybe what's not clear. we would effectively have this in some capacity, we're just giving it a name badge and abstracting it a bit (see commit for details).

> looking forward it seems odd to decouple the function and trigger resources in the API but then still leave them coupled in func.yaml - if a function is truely an independent thing (just a registered image and some config) could these be separated all the way up the stack?

we could do something completely different in the CLI, am open to ideas. decided to roll with trying to re-use func.yaml and make it flexible just to have less commands and files for end users to have to manage. it also makes sense to me to force something like an `app.yaml` to do all your trigger-function mappings in, and leave the `func.yaml` just to have function configuration. or $other_better_ideas

> Are functions immutable?

eh i mean we're fighting with docker hub on that, i don't think we can say that easily and i'm not sure it's something worth gating on fixing here to e.g. have image checksums in func/trigger definitions. the intent is to enforce function (and image) versioning i think, but it's not necessarily immutability. is this something worth pursuing at this point? this has long been an issue of 'hey i updated my image under 0.0.42 but the server is still running the old 0.0.42', is there an easy win here other than telling users to do versioning? (e.g. we could auto bump every PUT to a function and not allow this kind of behavior at all for a function, but they could pull the rug on the image still)

thanks again, hopefully that clarifies some stuff.

docs/fn-2.0/README.md
treeder: Auth should not be in files people check into source control, so they don't belong here.

rdallman:
am gonna start tinkering with https://github.com/fnproject/fn/pull/957/files#diff-a09b48bd44dbe9bb9ef1c216b3f146fbR445 -- need to tinker to see how this is gonna mash together, but initially would like to have exactly what our FDKs 'of the future' look like under the `cloudevents` format (binary and all), this exists now but the cloud event doesn't contain _all_ information inside of it (e.g. app id, timeout). after we settle on this and update FDKs, then I'd like to rip out the other (http, default, json) formats and switch to all cloud event, as it will drastically simplify the agent. I think this is relatively easily manageable, and can be done before changing triggers/functions APIs. This would break all existing functions inasmuch as they'd need to update the FDK, use an FDK or update to handle the new format -- maintaining the various formats has long been a thorn in our foot, this is more of a 'good riddance' hopefully.

rdallman:
meeting about prioritization of these wrt other things happening:

* toss namespaces out for now
* do API change of trigger/functions first
* do cloud event changes later, FDK have 2 handlers (we can stage)

i'll make changes to this doc wrt removing namespaces today hopefully. and then shift to working on scoping the API changes (vs. scoping the cloud event changes)

rdallman:
die cla bot i changed my github email

rdallman:
immediate issue created by trying to throw out namespaces is that functions do not have a bucket to exist in, they are just all in one big bucket. for someone that wants to do authenticated listing of functions they will have to implement a custom datastore that takes their namespace variable into account when making db queries, we have this issue with apps now anyway (and will continue to). this seems to be the plan anyway. not expecting to hear anything here, just yelling into the void.

hibooboo2:
Why are we removing namespacing?

rdallman:
@hibooboo2 i wonder if my email change breaks the cla bot? yay edge cases :(

hibooboo2:
It is looking to see that you signed your commit with an email that is your primary email I believe. I can make it check all emails that it can see on you. Not sure why it is saying you did not sign it. Do you want me to investigate it?

rdallman:
kinda confused why it's not working today, worked yesterday (with email change in place). will see if i can finagle it.

rdallman:
@hibooboo2 I think the CLA bot is using the earliest commit's author to check their email for the CLA bot and not the last one, so for stuff like this even if I add commits on it won't help (and for this one, since it's a collab, it's kinda mean to rebase). not a big deal, I guess if it's easy to change it to the last one it may be worth it, I'm likely a rare case though.

rdallman:
2 points that came from recent slack discussions:

* it may make sense to hang functions off of apps inside of a trigger to help users manage functions. the downsides being this hides functions inside of an app and if we end up with a function-hub-like thing then this is more challenging, we would presumably need another way to address functions. the upshot being, a user makes a function app and then creates functions and triggers in there. additionally, permissions can be attenuated better on that (and conversely, it's harder to share functions, have to re-create in each app). kind of a composability vs easy UX thing.

* hanging calls / logs off of app or function or trigger. this was an open ended TODO in the swagger file which didn't generate too much discussion. calls and logs are kind of a wart and we really want a 'pure runner' like module that only execs and returns the output + logs [+ call metadata]. this was not easy to rip out last time I tried, with our agent variations - aside from that, we still haven't really come to a conclusion on what we want this api to look like for OSS while being an interface that a larger scale deployment could presumably implement without much issue.

rdallman:
3 points:

* we could address resources by ID instead of name everywhere. we seem to have been using name for some time, this is a good place to change that if we want. github/docker allow using name, and then the underlying resource have ID, I think this is what we were going for but we hadn't fully gotten there yet. it is a bit tedious to implement the backend with the name stuff, it's really hairy (see app id), but it does look nice to have name in the UI and allows us to use PUT instead of POST for resources, as was proposed here.

mantree:
Thoughts on the three points:

1) I think Function inside App makes a lot of sense, giving a top level abstraction to hang auth and general metadata off. I'm wondering if Triggers should have an App ID as well, so it's easy to traverse the domain from app directly to Trigger. It might make more sense to go that way for certain use cases. 

Reluctant to say much on FnHub, not that familiar with the idea, but if a user takes makes use of another user's function it seems reasonable that an entry (function row) would need to exist in their domain, for calls and triggers to reference if nothing else. The FnHub, from this perspective, more providing somewhere to clone settings from, rather than a reusable execution environment.

2) Uncertain what 'pure' means here, but I think the Log (stderr from container) should exist within a Call object, referenced on a separate route via the api.

3) I think there's a mixing of API and Domain Model thinking here. The Domain Model, table design, should be completely linked via ID's imo. The API though can expose the entities via names, if that's deemed valuable. I'm not sure why that is particularly valuable here though. This is not an API users will be interacting with, all the usage comes via the CLI. The API can, therefore, be machine orientated, i.e. ID's used throughout the routes, supporting query params for name.

If there is a strong case for human-friendly urls then perhaps we could support both with different routes, or differentiate between ID and name usage, ensuring no one could create a name that looks like an ID and regexp to discover which one is in use. Maybe there's a use case I'm not yet familiar with here.

mantree:
I think 'Function Version' needs a solution. A 'call' has to reference both the exact code that executed it and the configuration it was executed with in order to be of any use. See the Docker Images table in my forks ERD for an idea in this area. Not entirely sure that is the way to go mind, might want something more generic.

mantree:
Surfacing swagger TODO's:

# TODO(?): pagination should only list resource names or the entire resource?

Paged resources normally return the entire resource. 

# TODO(?): triggers and functions OR trigs and funcs ?

I think apps, fns, triggers. *Bikeshedding alert*

# TODO(?): version change? /v2 ?

No. My understanding is that /v1 has not been released, really we should go for 0.2 or something, saving one until we're confident.

# TODO(?): remove stats from call?

They seem useful, but need to be optional. I don't think calls should link to app_id, more to fn_id. They must reference an exact code version (docker coordinates, image sha, config) somehow.

# TODO(?): amend log api? if we have /app/logs we could tail logs on an app (or trig?) and have point lookup for call still

I think log should live under call. Something like /apps/id/fns/id/calls/id/log

# TODO(?): annotations & config & event.extensions redundant to a degree? naming confusing? discuss.

Annotations and Config are different entities, imo. Annotations don't influence execution, they are labels. Config does influence the execution, being set as env vars. Not sure what event.extensions are, but they can't really be config (event data shouldn't cause a new function instance to be instantiated to execute it, for example).

# TODO(?): add /run API to swagger doc? (only covers API module atm)

I think this can be removed in favour of triggering triggers.

# TODO(?): should trigger have an 'enabled' field to easily toggle on and off via API?

Yes

# TODO: let's punt on adding a row for FuncVersions with snapshots of each func version? just a string field for now.

I think this is such a crucial aspect, see above, and we should answer this at this stage.

# TODO: put body and response objects have different required fields, etc; make 2 objects?

Yes. The extra fields being system determined I guess. Is there an extension mechanism in swagger? https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/

# TODO: change all params to $ref's

Don't really know what the norm is in Swagger, but it looks sensible.

# TODO(orcl): 204 DELETE success

Seems viable, although I personally don't like deletes at all and would prefer to use Status = Inactive. Actually deleting data makes auditing hard.

# TODO(orcl): 401 - should we define this to delineate w/ 404 even tho we don't have auth by default?

This is tricky, as a 401 indicates that a resource exists, and thereby can leak ids. Often I've simplified to not 401 and just use 404's for unauth'd as well. Potential *bikeshed*.

zootalures:
> Thank you for your submission!

*On hierarchy/setting up for a Fn-Hub sort of model* 

it's only a soft opinon but I guess my brain's preference for disentanglement is to keep a strictly hierarchical domain model for now 

Understand the argument and how this would support a FnHub  sort of thing but adding explicit indirection to the domain model  makes  a bunch of things hard that weren't before - notably working out when /how changes are propagaged, how deletions work and ar reported etc.

It's probably a bit pansy-assed but I'm all in favour of worring about sharing ownership of functions and triggers  some other time . 

My mental model here largely stands at: 
* An app is a unit of deployment , control (permission wise) and configuration
* A function is a unit of functinoality within an app, associated to it's causes by zero or more triggers. 

----

* On calls/Logs:*

My intutition is that they "belong" only to functions  and not triggers (as In when I list calls/logs for a function I see everything irrespective of the trigger) - but perhaps the trigger identifier is encoded in or attached as metadata to both (as a "cause" or something - maybe something in the CloudEvents model ) 


```
fn calls list myapp myfn 

|  time              |       trigger     |      status    | duration    |
-------------------------------------------------------
| 2012-01-01.: |    /httpRoute  |   OK           | 12ms         | 
| 2012-01-01.: |    <DEBUG>   |   OK           | 48ms         | 
| 2012-01-01.: |    queue2       |   FAILED     | 23ms         | 
```

And for logs, something morally equivalent to a logger field for the same. 

Because: 
* I  want to to be able to call a function independently of a defined trigger in the Fn model - 
       *  e.g.  via (e.g.) a debug call (over an authenticated API) as a feature later - this doesn't have an associated trigger (unless we create a synthetic one)  (e.g. I do an Fn Call on a function with no triggers) 
       * or via an external integration that is providing triggers for us but integrating directly into Fn  (i.e. we just do the callin - 
* Calls+ Logs (arguablly) could  persist beyond the lifecycle of triggers 
     * delete trigger, still want to see/account for stuff

I don't care if I  can fetch  calls/logs per app  (at least not  in  CLI for instance) - If I'm exporting my logs to a logging service then I want my app Identifier encoded as a field in my logs. 

FWIW: Off the top of my head, my spidey sense says that the modal case will be people dealing with single apps, single functions, single triggers  and that  we have to make sure that the payoff in the UX is when that is the case, 
---- 

*On names/IDs* 

Pretty sure we need  human readable/typable identifiers for functions and probably triggers  here  (unique in their context). If it's necessary for mechanical IDs then we should have them as well, not instead of.

rdallman:
sorry, didn't make it to this today, thanks both for your input - i have no meetings tomorrow and short of divine meeting intervention (the opposite of what put all these meetings there?), i will get to this first thing

mantree:
@zootalures re names/ids, it's not so much that we don't need them, it's whether they belong in the api at the route level. Name as a query param on /fns would be how people find fns by name.

My suggestion being, people aren't going to be manually typing these urls or really sharing them. It'll all go via the cli, which can resolve sort such things out.

rdallman:
> My suggestion being, people aren't going to be manually typing these urls or really sharing them. It'll all go via the cli, which can resolve sort such things out.

the main thing is that in a sense we're exposing 'webhooks' and naming these can be preferable for management, where most usage is not through the cli. this entails every resource in that chain being addressable by name instead of id (app and trigger, previously app and route). it's possible to have a mix, but it's nice to be uniform and not have to look up app names and such (as noted, a regex mix of id and name on resources is hairy). as a side note it's been nice ux wise in the cli and such to address things by name instead of having to copy and paste various ids, which is something we could fix up, but name in this case is going to have to be unique anyway. in any case, we can use ids in the backend while having the api use names, which is the plan here at least for now, definitely with an eye on function hub like thing for functions and for a decent out of the box experience for triggers without having to set up API gateway routes and such to get nice urls.

rdallman:
> TODO(?): pagination should only list resource names or the entire resource?
> TODO(?): version change? /v2 ?
> TODO(?): annotations & config & event.extensions redundant to a degree? naming confusing? discuss.

addressed: https://github.com/fnproject/fn/pull/957/commits/299aa58400f49d6d21e5686781d2b3dd6f54d9af

> I think apps, fns, triggers. Bikeshedding alert

fns and triggers seems pretty cool to me. like fn using fn terminology, downside being when someone says `fn` we have to clarify whether it's a func or a server/service they're talking about - i still like it. this is an easy rename, I have most of this plumbed with `func` but would be easy to change (hopefully posting soon...)

> TODO(?): remove stats from call?

we're sending to prometheus, we can stop storing them pretty easily and still have a handle on them. this is easy too, has come up.

> TODO(?): amend log api? if we have /app/logs we could tail logs on an app (or trig?) and have point lookup for call still

at present you can list calls on an app and filter by path, basically giving this functionality. assuming we have listing calls on an app with ability to filter by function or trigger, this allows listing across triggers/funcs which would be hindered by nesting them deeper in the url or we'd need to add an additional endpoint for this if we do end up wanting it.

> I think 'Function Version' needs a solution. 

let's spec this out, I agree we really should have something robust soon. we can figure out if we need to add it here or if we can add it in a month, but we can spec it out now to figure that out. Would you like to spec out the api bits for this on top of a function? Would be really great to be able to list all versions of a function, with the full configuration, and we could also use this to enforce docker image shas at runtime, too (we can have immutable versions, basically, if we want - this is discussed further up thread, as well). 

> TODO(?): add /run API to swagger doc? (only covers API module atm)

it's not very clearly defined, but this is basically the `/r/app/route` endpoint we have now, which is not in the swagger. this one we can kinda punt on anyone, bigger fish...

> it's only a soft opinon but I guess my brain's preference for disentanglement is to keep a strictly hierarchical domain model for now

yea, we can pretty easily add a rooted `/funcs` and move them up if we want or add a registry api separately using function ids to query into app functions later, that's true. having functions rooted seems a little odd, yea, this made more sense when it was `/ns/:ns/funcs/:func` instead of rooted `/funcs`. think we should roll with moving functions into apps, seems to be agreement on that, some functions seem useful across apps (i.e. outside of an app context), e.g. an image resize func is generally useful outside the context of an app, again this kinda made more sense with namespaces I think. anyway, on with app functions.

> My intutition is that they "belong" only to functions and not triggers

kinda discussed today IRL, textifying. the original intent of the proposal here (which is mutable) was that a trigger is the way to invoke a function and thus triggers have executions with additional configurations that may not make sense solely in the context of a function. our thinking was something like a docker image is to a docker container as a function is to a trigger; i.e. a trigger is an execution of a function, and the analogy kinda makes sense in that docker images don't have logs, containers do, in the same sense that a function doesn't have logs, a trigger does.

the oddity here is that it doesn't seem like trigger provides very much value other than being able to add additional configuration, and a case could likely be made to get rid of the trigger and just invoke functions directly - kinda get that. if we don't ever plan to let a trigger invoke multiple functions, it makes sense, but if we add the trigger construct here we can leverage it to do that, in addition to being able to configure 3rd party functions without having to copy them (maybe other stuff?).

directly to the point about calls/logs - littered various places, if we leave them on app we can add filters for function or trigger, etc and they exist on an app (i.e. if you nuke a function there's still a log on the app for it)

> Pretty sure we need human readable/typable identifiers for functions and probably triggers here (unique in their context). If it's necessary for mechanical IDs then we should have them as well, not instead of.

+1 - the app id/name thing has been a pain, was thinking about carving it out, hybrid was the big bugaboo there and any other middleware (auth). if we say, don't allow renaming or deleting then we take care of some 'issues' there. i'm not sure what to do about this situation, there's not enough zoloft to fix this one.


-------

think we're converging on a few things (like app functions), thanks for the feedback here.

rdallman:
closing to continue on #1038 free of these load times and context. will review this for anything we may have missed and swoop it up

PR #961, opened 2018-04-27
Updated cli section on triggers and FDK's.

treeder:
Some updates on triggers and FDK's.

### DIFF
--- a/docs/fn-2.0/README.md
+++ b/docs/fn-2.0/README.md
@@ -329,37 +329,63 @@ Picks runner to run it on and passes along event
 
 ## CLI changes
 
-### fn deploy
+### NEW IN 2.0
 
-func.yaml barely changes:
+The main changes here are related to [splitting functions and routes](https://github.com/fnproject/fn/issues/817).
+A route is now called a trigger and the route/trigger definition is no longer part of the function definition as it was before.
+
+### Function Definition File
+
+`func.yaml` barely changes, just removes route/trigger specific things. 
+
+A function does not know about triggers so it only has the information required to run the function.
 
 ```yaml
 function:
   name: yodawg
-  image: hub.docker.com/fnproject/hello
+
+  runtime: ruby
+  entrypoint: ruby func.rb
   version: 0.0.7
   memory: 42
+```
 
+`fn run` can run a function with this file only.
+
+### Trigger Definition File
+
+A `trigger.yaml` file defines the mapping from event sources to functions. A trigger
+binds one event source to one function, for instance an HTTP route to a function.
+
+```yaml
 trigger:
-  - name: myApp/sayhello
-    type: http
+  name: sayhello
+  type: http
+  # http specific param:
+  path: /sayhello
+  func: https://fnreg.com/namespace/myfunc
 ```
 
-trigger in yaml is optional. you may specify multiple triggers for each
-function. if trigger is not specified, it can be specified at runtime:
+On `fn deploy`:
 
-`fn deploy --trigger-http sayhello`
+* `trigger.yaml` must exist
+* if `func.yaml` exists:
+  * build and push function image
+  * update function registry with function definition
+  * use pushed function as `func` param for trigger
+* if `func.yaml` does not exist, `trigger.yaml` must have the `func` param set
+* Update trigger with Trigger definition
 
-since namespace is not specified, this will create a function at:
+This will create a function at:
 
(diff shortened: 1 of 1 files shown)


docs/fn-2.0/README.md
```diff
@@ -329,58 +329,70 @@ Picks runner to run it on and passes along event
...
+### Function Definition File
+
+`func.yaml` barely changes, just removes route/trigger specific things. 
+
+A function does not know about triggers so it only has the information required to run the function.
 
 ```yaml
 function:
   name: yodawg
-  image: hub.docker.com/fnproject/hello
```
rdallman: i think we should separate `name` and `image` still here, as it is in the API. aside from the addressability issue of e.g. `/ns/1/funcs/quay.io/rdallman/yodawg` -- i am still salty about the registry business https://github.com/fnproject/cli/pull/144 and for a cloud implementer it seems likely that many users will use a non-dockerhub registry, and we should try to be more friendly in that direction.

docs/fn-2.0/README.md
```diff
@@ -329,58 +329,70 @@ Picks runner to run it on and passes along event
...
+A `trigger.yaml` file defines the mapping from event sources to functions. A trigger
+binds one event source to one function, for instance an HTTP route to a function.
+
+```yaml
 trigger:
-  - name: myApp/sayhello
-    type: http
+  name: sayhello
+  type: http
+  # http specific param:
```
rdallman: let's just use name? the event source can strip the path to name format (i.e. strip `/`), redundant if anything and then we get into extensions for every event source type?

rdallman: i guess name has the same issue as the image if we do that. fuck me.

how about:

```
name: sayhello
source: /sayhello
```

`source` is generic and can be used across all event sources. e.g. a topic name, a schedule name, etc. where 'path' is domain specific to http.

docs/fn-2.0/README.md
```diff
@@ -329,58 +329,70 @@ Picks runner to run it on and passes along event
...
+On `fn deploy`:
 
-`fn deploy --trigger-http sayhello`
+* `trigger.yaml` must exist
+* if `func.yaml` exists:
+  * build and push function image
+  * update function registry with function definition
+  * use pushed function as `func` param for trigger
+* if `func.yaml` does not exist, `trigger.yaml` must have the `func` param set
+* Update trigger with Trigger definition
```
rdallman: 👍

docs/fn-2.0/README.md
```diff
@@ -395,6 +407,27 @@ triggers, and add namespaces.
...
+```
+handle(context, input) (output, err)
+```
+
+This form will change depending on what's right for the language, but they should all have the same 
+meaning and generally the same feel:
+
+* `context` will include all of the CloudEvent fields other than the `data` field. 
+* `input` will be the `data` field
+* `output` is the response `data` field
```
rdallman: 👎 -- if they're using the FDK I really don't think it's a far cry to ask for:

```
func handle(ctx, cloudEvent) (CloudEvent, error) {
  body := cloudEvent.Data
  if cloudEvent.ContentType != "application/xml" {
    return errors.New("NOU")
  }
  var foo struct{}
  xml.Unmarshal([]byte(body), &foo)
  
  rce := fdk.NewCloudEvent(`{"sup":"yo"}`, "application/json")
  return rce, nil
}

// additionally, a user could make a full one if they'd like: `&CloudEvent{"eventType":"kafkason", "eventTime":"01:02 1/1/1970", "data":"yo"}`
// we just have helpful constructors.
```

i'm ok to add context and error, but i/o with a CloudEvent instead of strings gives us a ton more flexibility, we're going to have to add the ability to modify fields of the returned cloud event from the fdk anyway, users may want to set content type, extensions, another event type, etc. if they're gonna be first class they should just be first class imo.

rdallman:
some yays and some nays. thanks for the edits, there was some discussion on https://github.com/fnproject/fn/pull/957 over the fdk formats and this seems like there's some different ideas here, let's try to get some more opinions on this one, the right direction is unclear.

docs/fn-2.0/README.md
rdallman: feedback I got around this: why not both. cool. let's do both.

rdallman:
alright i'll address my own comments, thanks for updates

PR #965, opened 2018-04-30
Last middleware should use the request passed by preceding middleware.

srpurani:
When we chain middlewares, the last middleware does not use the request passed by the preceding middleware. If the preceding middleware read httpRequest.Body and set a new io.ReadCloser to httpRequest.Body, the last middleware will not see this.

This change is useful when some middleware reads httpRequest.Body to
perform some logic, and assigns a new ReadCloser to httpRequest.Body
(as body can be read only once). One example is authentication middleware that needs to read the body to validate signature.

### DIFF
--- a/api/server/middleware.go
+++ b/api/server/middleware.go
@@ -93,7 +93,7 @@ func (s *Server) runMiddleware(c *gin.Context, ms []fnext.Middleware) {
 			c.Abort()
 			return
 		}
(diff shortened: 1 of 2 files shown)


rdallman:
@hibooboo2 CLA bot didn't run here? seems weird (@srpurani works here, so nbd, but still weird)

rdallman:
nvm... late to the party, that dude. with no beer!

Issue #712, opened 2018-01-22
Developer needs to be able to access startup logs and logs between function calls


treeder:
During startup, before the first request for instance, and between calls too. There may be some bug where the container looks fine, but there's some issues in the user code. 

I'm not sure the best way to do this, could be merging the in between logs with the next calls log or something.

PR #970, opened 2018-05-03
add user syslog writers to app

### LINKED ISSUE #337: add syslog logger hook for calls
users should be able to log their functions to wherever they want directly. we already have the code for this in common, we just need to wire it in to the func logging code.

rdallman:
users may specify a syslog url[s] on apps now and all functions under that app
will spew their logs out to it. the docs have more information around details
there, please review those (swagger and operating/logging.md), tried to
implement to spec in some parts and improve others, open to feedback on
format though, lots of liberty there.

design decision wise, I am looking to the future and ignoring cold containers.
the overhead of the connections there will not be worth it, so this feature
only works for hot functions, since we're killing cold anyway (even if a user
can just straight up exit a hot container).

syslog connections will be opened against a container when it starts up, and
then the call id that is logged gets swapped out for each call that goes
through the container, this cuts down on the cost of opening/closing
connections significantly. there are buffers to accumulate logs until we get a
`\n` to actually write a syslog line, and a buffer to save some bytes when
we're writing the syslog formatting as well. underneath writers re-use the
line writer in certain scenarios (swapper). we could likely improve the ease
of setting this up, but opening the syslog conns against a container seems
worth it, and is a different path than the other func loggers that we create
when we make a call object. the Close() stuff is a little tricky, not sure how
to make it easier and have the ^ benefits, open to idears.

this does add another vector of 'limits' to consider for more strict service
operators. one being how many syslog urls can a user add to an app (infinite,
atm) and the other being on the order of number of containers per host we
could run out of connections in certain scenarios. there may be some utility
in having multiple syslog sinks to send to, it could help with debugging at
times to send to another destination or if a user is a client w/ someone and
both want the function logs, e.g. (have used this for that in the past,
specifically).

this also doesn't work behind a proxy, which is something i'm open to fixing,
but afaict will require a 3rd party dependency (we can pretty much steal what
docker does). this is mostly of utility for those of us that work behind a
proxy all the time, not really for end users.

there are some unit tests. integration tests for this don't sound very fun to
maintain. I did test against papertrail with each protocol and it works (and
even times out if you're behind a proxy!).

closes #337

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -1,8 +1,10 @@
 package agent
 
 import (
+	"bytes"
 	"context"
 	"io"
+	"log/syslog"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -669,8 +671,21 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	// buffering overflows (json to a string, http to a buffer, etc)
 	stdoutWrite := common.NewClampWriter(stdoutWritePipe, s.maxRespSize, models.ErrFunctionResponseTooBig)
 
+	// get our own syslogger with THIS call id (cheap), using the container's already open syslog conns (expensive)
+	// TODO? we can basically just do this whether there are conns or not, this is relatively cheap (despite appearances)
+	buf1 := bufPool.Get().(*bytes.Buffer)
+	buf2 := bufPool.Get().(*bytes.Buffer)
+	defer bufPool.Put(buf1)
+	defer bufPool.Put(buf2)
+
+	sw := newSyslogWriter(call.ID, call.Path, call.AppID, syslog.LOG_ERR, s.container.syslogConns, buf1)
+	var syslog io.WriteCloser = &nopCloser{sw}
+	syslog = newLineWriterWithBuffer(buf2, syslog)
+	defer syslog.Close()                            // close syslogger from here, but NOT the call log stderr OR conns
+	stderr := multiWriteCloser{call.stderr, syslog} // use multiWriteCloser for its error ignoring properties
+
 	proto := protocol.New(protocol.Protocol(call.Format), stdinWrite, stdoutRead)
-	swapBack := s.container.swap(stdinRead, stdoutWrite, call.stderr, &call.Stats)
+	swapBack := s.container.swap(stdinRead, stdoutWrite, stderr, &call.Stats)
 	defer swapBack() // NOTE: it's important this runs before the pipes are closed.
 
 	errApp := make(chan error, 1)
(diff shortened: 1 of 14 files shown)


rdallman:
/me i dream of a day where i won't have to click ci rebuild over IOPipes (is this test providing value? i might just get rid of it, it doesn't seem worth it now, there's `TestPipesAreClear` to make sure we aren't leaving garbage in there, though IOPipes tests 2 other things).

this may also close #712 - we also have mr freeze now so it's kinda moot, the container won't do anything between calls really, it could still spew some logs on startup though. i'm not really sure we want to have a container logs api, the call logs api is already getting nailed up for target practice, just getting the logs out to syslog seems pretty good to me ?

rdallman:
looks pretty good in papertrail (the format is kinda optimized for that -- it's what I know)

screen shot 2018-05-02 at 5 53 52 pm

rdallman:
53s4vjs

bump @zootalures @jan-g @hhexo (i think DP is gonna use this? idk i'm out of the loop)

rdallman:
thanks for reviewing.

> Should we execute this dial in parallel for all syslog sinks 

yea, this could be a good idea. trying to keep it simple for a first pass, also suspect most users will use 1 or maybe 2-3 tops, so N is really small, but could be 1-2s for a tls conn half way around the world too I guess. 

> and also in background to avoid adding up to container start latency?

slight concern about doing it in parallel to starting the container, as this could mean that the logger could miss some logs. In particular, for containers whose launches subsequently fail relatively quickly. I think it's best that we ensure we don't miss any logs (hard to explain/debug) and in practice w/ hot functions eating the latency should be mitigated mostly (though, that depends on how optimistically we launch).

skinowski:
I'm concerned about adding any latency to Prepare/Run. We can make a call on if we should make these logs lossy or not (config opt?), but I think the connect process can start very early in the processing. (We can hoist that to a per hotLauncher() or some conn pool orthogonal to these later if we want, but runHot() seems OK for now.) And we already ignore partial writes/errors in syslog Write. If freezer enabled (it normally should be), then the connection is not used until exec() since we block IO until exec(). So, for common case, we could initiate a background early connect, and in exec() either block the IO (if not lossy) or do non-blocking IO (lose everything until eventually connect succeeds) without delaying processing...  If not lossy, we can perhaps flag syslog Write() failures as error... Just thinking out loud here. :-)

rdallman:
so, it's possible to do all of that. i can't honestly say i agree that it's worth it, but we don't really have any data to point at now to say one way or another. we're going to be presumably adding a strategy to pre-emptively launch containers to get around start time in any case, for the high throughput installs. for this one, more worried about imposing some connect timeout policy than anything, as in the average case I expect this is an entire order of magnitude smaller than the docker start time, for udp even less (dns time, if any?). but, sticking with the theme of not imposing limits in OSS (it seems like a decent use case for annotations, to me, but there's no other plumbing). I think adding timeout addresses the real concern, when without a timeout even if we go optimize to the hilt, plumb a syslog conn pool manager that starts in one place and gets waited on in another, we still end up waiting longer than someone might like to get our stinkin logs out just b/c the dial takes forever (if blocking).

the idea of a bit to set lossy logs and unlossy logs seems like a really slippery slope i'd prefer to avoid, aside from being yet-another-knob-nobody-will-use. we should just pick a sane default I think. I see the support tickets already: "hey my container says 'yo' on startup but on April 13th, 2021 at 11:01 GMT my container started but didn't print 'yo', I set non-lossy logs, is life still worth living?". UDP will inherently be lossy, making the slope even more slippery. it seems hard to grok for users. i wish we had a larger community of prod users, we could poll for stuff like this.

getting logs out from the container on startup is key I think, if we have to pick one. we could hold up handing out a slot until the dial goes through, but we're still in the dial takes too long case potentially and it's relatively complex to plumb.

skinowski:
I'm OK with moving on with the current code, and if Prepare/Run is impacted we can revisit.

rdallman:
added a trace so we can track this, we'll probably need a latency distribution, am hoping to get back to that stuff next week and have some ideas for getting that out of traces

Issue #955, opened 2018-04-25
An explicit way to setup a Fn against the auth-protected private registry


denismakogon:
## problem
According to our [operation docs](https://github.com/fnproject/fn/blob/master/docs/operating/private_registries.md) and [docker registry auth dock](https://docs.docker.com/engine/reference/commandline/login/#privileged-user-requirement) the only way to make docker client be authorized is to create a config file within the special path (known as `DOCKER_CONFIG`), no matter would that be a pointer to the keychain or explicit user/password definition.

So, there is no explanation/doc how to set up the vanilla/upstream Fn how to enable the authorization for the private registries without hacking into the operating system of the Fn docker image (i.e. mounting the volume with docker config or rebuilding the Fn from source with config included).

The question appeared on [public slack](https://fnproject.slack.com/archives/C6FL02Q02/p1524639509000229): how to set up the Fn to talk to the registry on GCP.

## possible way(s) to solve

The simplest short-term solution is to write a doc that explains how to "make-it-work" on k8ts cluster with an auth-protected registry (i.e. how to create a volume with the config file, how to build Fn from source with docker config).

## alternative

we work towards context management and splitting function from triggers, can't we allow each function to define the registry auth as well just to move away from a need to work with volumes or rebuilding the Fn image.

shaunsmith:
+1 on being able to configure the registry at the app and/or function level.

rdallman:
i think #971 and #1015 close this one, thanks

PR #971, opened 2018-05-03
improve private auth docs a little

rdallman:


rdallman:
related to #955 - this resolves the main concern there around having to set up a new fn image just to use private images (i.e. you don't).

i do still think we can do better, but let's discuss that on that.

zootalures:
Awesome - @derekschultz  - I'll take a look at integrating this with fn-helm

rdallman:
thanks

PR #972, opened 2018-05-03
remove flaky tests

rdallman:
if we want them back, we can dig them out of git instead of some poor soul
uncommenting them 10 years from now and spending 3 months on failing CI builds
trying to figure out how a test that breaks doesn't mean the code's broke.

these tests are notoriously flaky and hard to understand/fix, they also test
very specific agent behaviors all the way through the front end when it may be
easier to test them in unit tests instead (should we so choose). at least,
since the behaviors tested aren't being changed very often, these are only
serving to provide negative value in time wasted re-running the test suite
[since them failing doesn't really indicate the code being wrong].

the `IOPipes` test is partially covered by `TestPipesAreClear` which hasn't
cropped up as being as flaky, but it tests less behaviors. it is not easy tt o
understand, either. while i think we learned a lot from these tests, they
haven't been a great citizen of our test suite at large, i figure if we need
to change runner behavior in the future we can maybe make another go at it.


rdallman:
the last 2-3 weeks this has been ~unbearable, gotta kick the bucket 2-3 times even for docs changes

rdallman:
cc @skinowski ^ we can discuss when you are back :D

skinowski:
No worries, I may re-send some of this as a new PR and re-run/clean them up until they are reliable. (And preferrably in agent_test.go if I can.)

PR #982, opened 2018-05-09
fn: allow specified docker networks in functions

skinowski:
If FN_DOCKER_NETWORK is specified with a list of
networks, then agent driver picks the least used
network to place functions on.

### DIFF
--- a/api/agent/drivers/docker/docker.go
+++ b/api/agent/drivers/docker/docker.go
@@ -6,10 +6,12 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"os"
 	"path"
 	"strings"
+	"sync"
 	"time"
 
 	"go.opencensus.io/stats"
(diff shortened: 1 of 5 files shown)


api/agent/drivers/docker/docker.go
```diff
@@ -48,11 +50,13 @@ func (r *runResult) Error() error   { return r.err }
...
-	docker   dockerClient // retries on *docker.Client, restricts ad hoc *docker.Client usage / retries
-	hostname string
-	auths    map[string]docker.AuthConfiguration
-	pool     DockerPool
+	conf         drivers.Config
+	docker       dockerClient // retries on *docker.Client, restricts ad hoc *docker.Client usage / retries
+	hostname     string
+	auths        map[string]docker.AuthConfiguration
+	pool         DockerPool
+	networksLock sync.Mutex
```
kmjohansen: I assume that this protects the networks map, but it would be great if you would add a comment that states exactly what should be covered by this mutex.

skinowski: done.

PR #989, opened 2018-05-10
fn: lb and pure-runner with non-blocking agent

skinowski:
*) In LB and runner gRPC communication, removed ACK sending
from pure-runners. Now, upon TryCall, pure-runners can quickly call
Agent Submit(). This is good since at this stage pure-runner
already has all relevant data to initiate the call.
*) Removed pure-runner capacity tracking code. This did
not play well with internal agent resource tracker. 'Too-busy' responses
from Agent Submit() can now be retried.
*) Agent is set to non-blocking mode in pure-runner, this allows
for a faster 'Too-busy' response from Agent Submit() if the request
gets blocked on CPU & Memory resources.
*) LB immediately streams http body to runners. This is optimistic
common case. But streaming is interrupted if NACK is received.
*) For retriable requests, request body IO is handled by GetBody function,
which is set by LB. This allows for reading http Body safety multiple times
in case of retries.
*) Idempotenty/retry is similar to previous code. After initial success
in Engament, after attempting a TryCall, unless we receive NACK,
we cannot retry that call.
*) ch and naive places now wraps each TryExec with a cancellable
context to clean up gRPC contexts quicker.


hhexo:
I like this implementation, but I am still concerned about buffering the whole body of the http request on the LB side.

The LB is going to handle multiple requests - possibly for multiple tenants. If I send several malicious requests with gigabytes of body, and I have enough bandwidth to do so within the function timeout, I could OOM the LB, affecting not only my functions but also other people's. Granted, it has to be a lot of bandwidth (e.g. 64GB / 2 minutes = 500MB/sec), but still...

This buffering is used only for retries, and hopefully we get a NACK pretty quickly if we have to retry (after all, we just wait for the capacity check). Is it possible to use a ring buffer and hope that we receive the NACK before it wraps around (and error out with a "server too busy" if we have to retry but it has already wrapped around)?

This is the scenario I'm thinking about:
- LB receives request
- LB tries first runner, starts streaming and filling the ring buffer
- LB receives NACK, stops streaming, ring buffer is half full
- LB tries second runner, starts streaming the data in the buffer first
- LB receives NACK before any more data is needed, stops streaming, ring buffer is still half full
- LB tries third runner, starts streaming the data in the buffer first
- Then it seems to work, so LB resumes reading the request body and streaming / buffering it
- The buffer wraps around, we lose the first bit of the body
- At this point, if the runner has worked then everything is fine, but if the LB receives a very late NACK then it must error out because it cannot retry anymore.

I'm hoping that the vast majority of requests will satisfy these two conditions:
- The body is small anyway and probably just fits into the maximum ring buffer size
- If it doesn't, the NACK responses are fast enough that we don't need to buffer too much
Since there is an upper bound to the ring buffer size, a malicious actor will OOM the LB only if they send millions of requests - at which point some higher level rate limiting will kick in anyway.

api/agent/lb_agent.go
```diff
@@ -179,6 +190,45 @@ func (a *lbAgent) Submit(callI Call) error {
...
+		return nil
+	}
+
+	// WARNING: we need to handle IO in a separate go-routine below
+	// to be able to detect a ctx timeout. When we timeout, we
+	// let gin/http-server to unblock the go-routine below.
+	errApp := make(chan error, 1)
+	go func() {
+
+		buffer, err := ioutil.ReadAll(r.Body)
```
kmjohansen: Mentioned this in a prior review, but is an unbounded ReadAll safe here?  Is there something else enforcing the maximum size of the request body, or do we need to do it ourselves?

api/agent/lb_agent.go
```diff
@@ -179,6 +190,45 @@ func (a *lbAgent) Submit(callI Call) error {
...
+		if err != nil {
+			errApp <- err
+			return
+		}
+
+		r.Body = ioutil.NopCloser(bytes.NewBuffer(buffer))
+
+		// GetBody does not mutate the state of the request body
+		r.GetBody = func() (io.ReadCloser, error) {
+			return ioutil.NopCloser(bytes.NewBuffer(buffer)), nil
```
kmjohansen: Also mentioned this in another review, but this seems to violate the contract of NewBuffer.  According to the documentation the Buffer object created on line 214 is the owner of `buffer`.  I understand that we're doing this to avoid a copy, but it makes me nervous.

kmjohansen:
@hhexo I like the idea of pre-allocating and re-using some fixed amount of storage.  That means the garbage collector should have less work to do if we re-use instead of free and reallocate.  IMO we should either buffer everything and be willing to retry, or just buffer the request header and only re-try if we get a nack after sending the first frame.  There are enough edge cases here that we may want to start simple first.

hhexo:
I have another question... consider this scenario:

- The LB receives a request
- The LB tries a runner
- The LB streams the body, and it never receives a NACK so it thinks the runner is ok...
- ... but the runner was actually defective and it kept the connection open but never even ran the function
- The LB times out the call after a while

In this scenario, the LB has no information on whether it can or cannot retry the call (we don't know if the function had actually started), and the only option is to fail when it times out.
If the protocol required an explicit ACK instead, the code that expects the ACK could time out earlier than the function timeout (this is analogous to the slot allocation timeout in single-node fn), therefore giving the possibility of a retry, because the LB knows that the runner has not started the function.

Removing the explicit ACK curtails our options somewhat. Is the scenario of a defective runner rare enough that we accept the possibility of letting some calls fail in this way? (We'd lose all function calls issued to the runner before some other mechanism detects the faulty runner)

skinowski:
@hhexo Yes, runner failure is assumed/considered rare. If runner has failed, engagement will also fail which should prevent excessive loss of function calls.

Even with presence of Ack, failures could occur and functions can fail in the same way. Ack merely attempts to narrow this window, where new code has a somewhat large window for these losses to occur, but in practice I think the difference is not worth to delay execution for common case.

api/agent/lb_agent.go
```diff
@@ -171,6 +188,17 @@ func (a *lbAgent) Submit(callI Call) error {
...
 	statsDequeueAndStart(ctx)
 
+	// pre-read and buffer request body if already not done based
+	// on GetBody presence.
+	buf, err := a.setRequestBody(ctx, call)
+	if err != nil {
+		logrus.WithError(err).Error("Failed to process call body")
+		return a.handleCallEnd(ctx, call, err, true)
+	}
+	if buf != nil {
```
kmjohansen: Does this block need to be swapped with the `err != nil {` block?  I ask because it looks like `setRequestBody` can return a buffer even if it returns an error.  Is the `bufPool.Put()` needed to free this back up?  If so, it _looks_ like you have the potential for a leak here.

api/agent/lb_agent.go
```diff
@@ -183,6 +211,48 @@ func (a *lbAgent) Submit(callI Call) error {
...
+		if err != nil {
+			errApp <- err
+			return
+		}
+
+		r.Body = ioutil.NopCloser(bytes.NewBuffer(buf.Bytes()))
+
+		// GetBody does not mutate the state of the request body
+		r.GetBody = func() (io.ReadCloser, error) {
+			return ioutil.NopCloser(bytes.NewBuffer(buf.Bytes())), nil
```
kmjohansen: I'm still unclear about how this is safe.  Are we getting away with this because the NopCloser allows only reading from the buffer?

api/agent/pure_runner.go
```diff
@@ -678,8 +638,29 @@ func DefaultPureRunner(cancel context.CancelFunc, addr string, da DataAccess, ce
 	return NewPureRunner(cancel, addr, da, cert, key, ca, nil)
 }
 
-func NewPureRunner(cancel context.CancelFunc, addr string, da DataAccess, cert string, key string, ca string, gate CapacityGate) (Agent, error) {
-	a := createAgent(da)
+func ValidatePureRunnerConfig() AgentOption {
```
kmjohansen: Thanks for adding this

api/agent/runner_client.go
```diff
@@ -90,120 +100,191 @@ func (r *gRPCRunner) TryExec(ctx context.Context, call pool.RunnerCall) (bool, e
...
 	err = runnerConnection.Send(&pb.ClientMsg{Body: &pb.ClientMsg_Try{Try: &pb.TryCall{ModelsCallJson: string(modelJSON)}}})
 	if err != nil {
 		logrus.WithError(err).Error("Failed to send message to runner node")
-		return false, err
-	}
-	msg, err := runnerConnection.Recv()
-	if err != nil {
-		logrus.WithError(err).Error("Failed to receive first message from runner node")
-		return false, err
+		return true, err
```
kmjohansen: Do we need to do anything to close the runnerConnection on error?

api/agent/lb_agent.go
skinowski: Switching this to bytes.Reader() which should *really* be read-only.

api/agent/runner_client.go
skinowski: I don't think so, corresponding generated code intentionally leaks the stream (runner.pb.go Engage()) and returns nil. And we'll be cancelling this context also. We should be OK.

api/agent/lb_agent.go
skinowski: Arrgh. :-)

PR #991, opened 2018-05-11
Fix to make custom headers work in test CallFN function

ostrain:
I noticed the EnvAsHeader function was incorrect, and that the only test that really exercises it was also broken in a way that made it pass when it shouldn't.

The fixed test doesn't pass because the attempt to invoke the function fails with: `Failed to pull image 'denismakogon/os.environ': pull access denied for denismakogon/os.environ, repository does not exist or may require 'docker login'`

@denismakogon, looks like that container isn't publicly readable -- could you make that image publicly accessible somewhere?


denismakogon:
@ostrain the images you referred to has gone, so i recommend to make a new one to cover this case.

skinowski:
fn-api-tests are decomissioned, closing.

PR #998, opened 2018-05-15
fn: agent MaxRequestSize limit

skinowski:
Currently, LimitRequestBody() exists to install a
http request body size in http/gin server. For production
enviroments, this is expected to be used. However, in agents
we may need to verify/enforce these size limits and to be
able to assert in case of missing limits is valuable.
With this change, operators can define an agent env variable
to limit this in addition to installing Gin/Http handler.

http.MaxBytesReader is superior in some cases as it sets
http headers (Connection: close) to guard against subsequent
requests.

However, NewClampReadCloser() is superior in other cases,
where it can cleanly return an API error for this case alone
(http.MaxBytesReader() does not return a clean error type
for overflow case, which makes it difficult to use it without
peeking into its implementation.)

For lb agent, upcoming changes rely on such limits enabled
and using gin/http handler (http.MaxBytesReader) makes such
checks/safety validations difficult.

### DIFF
--- a/api/common/io_utils.go
+++ b/api/common/io_utils.go
@@ -11,9 +11,9 @@ type clampWriter struct {
 	overflowErr error
 }
 
-func NewClampWriter(buf io.Writer, maxResponseSize uint64, overflowErr error) io.Writer {
-	if maxResponseSize != 0 {
-		return &clampWriter{w: buf, remaining: int64(maxResponseSize), overflowErr: overflowErr}
+func NewClampWriter(buf io.Writer, max uint64, overflowErr error) io.Writer {
+	if max != 0 {
+		return &clampWriter{w: buf, remaining: int64(max), overflowErr: overflowErr}
 	}
 	return buf
 }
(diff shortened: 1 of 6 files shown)


api/common/io_utils.go
```diff
@@ -34,6 +34,39 @@ func (g *clampWriter) Write(p []byte) (int, error) {
...
+	if g.remaining <= 0 {
+		return 0, g.overflowErr
+	}
+	if int64(len(p)) > g.remaining {
+		p = p[0:g.remaining]
+	}
+
+	n, err := g.r.Read(p)
+	g.remaining -= int64(n)
+	if g.remaining <= 0 {
```
kmjohansen: Not sure if your callers expect this, but if `len(p)` is 1024, and `max` was 1024, then a Read(p) will return `1024, g.overflowErr` on the first invocation.  If the callers are checking for err being non-nil, this may confuse them.  OTOH, if they're checking size before checking for errors then this is probably fine.

reclaro: Personally, I find confusing returning the error if the case that @kmjohansen described. Looking at the implementation of MaxBytesReader in the case (max=len(p)) it doesn't return an error: https://golang.org/src/net/http/request.go?s=32861:32938#L1038 (unless the Read itself returns an error). So I think people don't expect to get an error back in that case, what do you think? Once we take a decision it would be great to have a simple unit test to assert what we have decided, implementing the different scenarios.

hhexo: I agree, reading up to and including the full max size should be fine.

PR #1001, opened 2018-05-16
add DELETE to allowed cors methods

kppk:
Default cors allowed headers is missing DELETE method, this change sets the AllowMethods explicitly, doesn't rely on cors middleware default.

/cc @adoublebarrel @rdallman


adoublebarrel:
I'll wait for second endorsement before I hit that merge button though.

PR #1002, opened 2018-05-16
Wrap custom datastore with metrics and validator

kppk:
When the server is configured using `WithDatastore` function fn doesn't use the validator/metric emitter from api/datastore/internal/datastoreutil.

This change makes validator and metrics emitter mandatory for custom datastore implementation by introducing new function `Wrap` to datastore, which is then called from `WithDatastore` function.

/cc @rdallman @zootalures @adoublebarrel @jan-g

### DIFF
--- a/api/datastore/datastore.go
+++ b/api/datastore/datastore.go
@@ -18,7 +18,11 @@ func New(ctx context.Context, dbURL string) (models.Datastore, error) {
 		return nil, err
 	}
 
-	return datastoreutil.MetricDS(datastoreutil.NewValidator(ds)), nil
+	return Wrap(ds), nil
+}
+
+func Wrap(ds models.Datastore) (models.Datastore) {
+	return datastoreutil.MetricDS(datastoreutil.NewValidator(ds))
 }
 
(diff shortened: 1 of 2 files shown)


skinowski:
Wrapping metrics should be optional/explicit there IMHO. Maybe we could expose that wrapper instead? Because if a user synthesizes a datastore using New(), then pass that datastore to WithDatastore(), it'll be wrapped again.

kppk:
@skinowski makes sense -- on the other hand, this way we can say that FN guarantees the same set of metrics every time.

I get your point, especially with having fnext/datastore which might need more detailed metrics anyway.

Do we agree that every datastore should be wrapped with validator?

denismakogon:
I do have doubts making metrics and validation optional. Most of the times we need validation for an app/route, so the validation is required just to prevent from weird SQL queries, but have no clue when I woundn’t need metrics.

skinowski:
We can (should) enforce metrics/validation. I'm only (slightly) concerned about user errors where "a user synthesizes a datastore using api/datastore/datastore.New(), then passes that datastore to WithDatastore()".  Not sure if this is a big concern..

PR #1004, opened 2018-05-17
Replace panic on enqueue for LB agent with error.

msgodf:


hhexo:
Shall we add a unit test for it too?

PR #1007, opened 2018-05-18
fn: lb & pure-runner slot hash id communication

skinowski:
With this change, LB can pre-calculate the slot hash
key and pass it to runners. If LB knows/calculates
the slot hash ids, then it can also make better
estimates on which runner can successfully execute
it especially when status messages from runner
include a small summary of idle slots for a given
slot hash id. (TODO)


skinowski:
@kmjohansen 

1) I have not measured the performance impact. Increased network rpc size by 20 byte string. But in
retry cases LB calculates this once. Expecting to see some CPU increase on LBs, and some CPU decrease in runners with slight network trx bytes increase.
2) hostagent doesn't keep much state regarding this. If a slot key does not refer to something that doesn't exist, it's treated as a 'cold-start' and a corresponding slot queue is created.
3) In theory, we could save these keys to the DB. But we haven't done this so far because maintaining these in DB (or a cache) is painful and often it's easier to recalculate it. (if a config changes for that route or app, slot hash needs to be updated, annotations & config are merged, etc.)

PR #1008, opened 2018-05-18
Implement graceful shutdown of agent.DataAccess

gviedma:
As part of an agent's graceful shutdown, we should also flush and close any resources associated with its underlying _DataAccess_ implementation. This includes adding the ability to close the underlying Datastore, Logstore and MessageQueue, which are extended to implement _io.Closer_.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -84,6 +84,7 @@ type Agent interface {
 	Submit(Call) error
 
 	// Close will wait for any outstanding calls to complete and then exit.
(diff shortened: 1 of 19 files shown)


rdallman:
for a full node we just have to make sure we're waiting for all requests to finish before shutting down the db / mq / logstore / data access / etc etc -- if we're not doing that, it's not the fault of this anyway, it just came to mind thinking about shutdown sequence. thanks for this

PR #1012, opened 2018-05-22
fn: size restricted tmpfs /tmp and read-only / support

skinowski:
*) read-only Root Fs Support
*) removed CPUShares from docker API. This was unused.
*) docker.Prepare() refactoring
*) added docker.configureTmpFs() for size limited tmpfs on /tmp
*) tmpfs size support in routes and resource tracker
*) fix fn-test-utils to handle sparse files better in create file

### DIFF
--- a/api/agent/call.go
+++ b/api/agent/call.go
@@ -119,6 +119,7 @@ func FromRequest(a Agent, app *models.App, path string, req *http.Request) CallO
 			Priority:    new(int32), // TODO this is crucial, apparently
 			Timeout:     route.Timeout,
 			IdleTimeout: route.IdleTimeout,
+			TmpFsSize:   route.TmpFsSize,
 			Memory:      route.Memory,
 			CPUs:        route.CPUs,
 			Config:      buildConfig(app, route),
@@ -252,7 +253,9 @@ func (a *agent) GetCall(opts ...CallOpt) (Call, error) {
 		return nil, errors.New("no model or request provided for call")
 	}
 
-	if !a.resources.IsResourcePossible(c.Memory, uint64(c.CPUs), c.Type == models.TypeAsync) {
+	mem := c.Memory + uint64(c.TmpFsSize)
+
+	if !a.resources.IsResourcePossible(mem, uint64(c.CPUs), c.Type == models.TypeAsync) {
 		// if we're not going to be able to run this call on this machine, bail here.
 		return nil, models.ErrCallTimeoutServerBusy
 	}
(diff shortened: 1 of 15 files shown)


zootalures:
Is it feasible to to add an e2e test to witness the FS limit here?

zootalures:
I have some minor concerns about the impact of making this per-route configurable  - but understand that while this is necessarily tmpfs that this make sense.

rdallman:
seems like a good use case for annotations to configure ?

api/agent/call.go
```diff
@@ -252,7 +253,9 @@ func (a *agent) GetCall(opts ...CallOpt) (Call, error) {
 		return nil, errors.New("no model or request provided for call")
 	}
 
-	if !a.resources.IsResourcePossible(c.Memory, uint64(c.CPUs), c.Type == models.TypeAsync) {
+	mem := c.Memory + uint64(c.TmpFsSize)
```
kmjohansen: I was originally on the fence about whether we should account for this, but after looking at `runc` and the kernel's `memcg` implementation, I think this is mandatory for correct behavior.  The cgroup's total page allocation count is charged against the memory limit in the memory cgroup, so if we under provision here then the user won't be able to use all of tmpfs.  The werid thing is that this lets them consume more of any pages, so if they have a piggy app they could use all of their allocation plus our extra for tmp, and then not get everything available for transient storage.  That said, file backed and regular anon pages are also charged against this total memory cap, so the behavior at the limit is going to seem strange no matter what.

api/agent/config.go
```diff
@@ -102,6 +107,10 @@ func NewAgentConfig() (*AgentConfig, error) {
 		cfg.EnableNBResourceTracker = true
 	}
 
+	if _, ok := os.LookupEnv(EnvEnableReadOnlyRootFs); ok {
```
kmjohansen: Does this imply that we need a change in the pure runner to enable this feature?

skinowski:
Annotations sounds good. Then operators could choose to put this as app or route based setting.

api/agent/config.go
skinowski: Yes, runner env would need an update.

api/agent/call.go
skinowski: Yeah, I felt this is a bit aggressive here as well, but safer than over committing IMHO.

skinowski:
@rdallman FYI, @zootalures has some reservations on using annotations for this. Leaving this PR as is, I think per route config is OK and gives enough flexibility to operators.

kmjohansen:
Are there any corresponding changes that need to be made to the fn-api or cli to support this?  Relatedly: should we enable by default and instead use the environment to allow users to disable?

skinowski:
I'd say we leave it disabled by default. This is the least disruptive for open-source users, cli, etc.

PR #1013, opened 2018-05-24
datastore no longer implements logstore

rdallman:
the underlying implementation of our sql store implements both the datastore
and the logstore interface, however going forward we are likely to encounter
datastore implementers that would mock out the logstore interface and not use
its methods - signalling a poor interface. this remedies that, now they are 2
completely separate things, which our sqlstore happens to implement both of.

related to some recent changes around wrapping, this keeps the imposed metrics
and validation wrapping of a servers logstore and datastore, just moving it
into New instead of in the opts - this is so that a user can have the
underlying datastore in order to set the logstore to it, since wrapping it in
a validator/metrics would render it no longer a logstore implementer (i.e.
validate datastore doesn't implement the logstore interface), we need to do
this after setting the logstore to the datastore if one wasn't provided
explicitly.

* splits logstore and datastore metrics & validation logic
* `make test` should be `make full-test` always. got rid of that so that
nobody else has to wait for CI to blow up on them after the tests pass locally
ever again.
* additional logstore validation logic

need to move some of the other logstore metrics around, but i know there is another person poking at metrics at the moment, so will wait on that. this is mostly just shuffling with no changes to logic aside from additional validation in the logstore validator (that was the goal, at least). noticed this was an issue when adding some datastore stuff, so did this first.


api/datastore/datastore.go
```diff
@@ -13,12 +13,7 @@ import (
...
-	ds, err := newds(ctx, dbURL) // teehee
-	if err != nil {
-		return nil, err
-	}
-
-	return Wrap(ds), nil
+	return newds(ctx, dbURL) // teehee
 }
 
 func Wrap(ds models.Datastore) models.Datastore {
```
mantree: Two thoughts on this:
 * Does the system ever want to create a non-metric creating datastore? If not then this should be moved in 'newds' so it can't be forgotten.
* Wrap is very generic. DecorateMetrics, or even AddMetrics would help readability, for me at least.

Makefile
```diff
@@ -43,7 +43,7 @@ test-extensions: test-basic
 test-basic: checkfmt pull-images fn-test-utils
 	./test.sh
 
-test: checkfmt pull-images test-basic test-middleware test-extensions
+test: checkfmt pull-images test-basic test-middleware test-extensions test-api test-system
```
skinowski: :+1:

api/datastore/datastore.go
skinowski: CP may rely on this Wrap() being inside New() and in WithDatastore()

rdallman: hmm, confused by the wording there. if a caller uses `WithDatastore` with their own datastore, then the behavior here should remain unchanged - that is, the provided datastore will get wrapped [once], just in New() instead of WithDatastore().

PR #1018, opened 2018-05-30
Swagger modifications to new shape

mantree:


docs/swagger2.yml
```diff
@@ -27,119 +27,133 @@ consumes:
...
-          description: List of funcs.
+          description: List of fns.
           schema:
-            $ref: '#/definitions/FuncsWrapper'
+            $ref: '#/definitions/FnsWrapper'
+        default:
+          description: Unexpected error
+          schema:
+            $ref: '#/definitions/Error'
+    post:
```
rdallman: was hoping to avoid post semantics, and this mostly hinges on using ids as the primary identifier I understand. 2 threads open for this, I'll post in the other one on id vs name.

docs/swagger2.yml
```diff
@@ -27,119 +27,133 @@ consumes:
...
         default:
           description: Unexpected error
           schema:
             $ref: '#/definitions/Error'
-    put:
-      operationId: "UpsertFunc"
-      summary: "Create or update a func."
-      description: "Creates a func if it does not exist, or updates the func if it does."
+
+  /apps/{app}/fns/{fn}/triggers:
```
rdallman: most list endpoints now have a filter to e.g. `/apps/app/triggers?fn=myfunctionname` to list triggers for a certain function. we can change to this if we want, we should just be consistent. less handlers / smaller api ftw to me, this has 2 spouts to list triggers per function if we add this endpoint.

docs/swagger2.yml
```diff
@@ -189,28 +203,28 @@ paths:
...
             $ref: '#/definitions/Error'
         default:
           description: Unexpected error
           schema:
             $ref: '#/definitions/Error'
-    patch:
+    post:
       operationId: "UpdateApp"
-      summary: "Updates an app."
+      summary: "Updates an app"
```
rdallman: we do want to remove patch, was going forward with only having GET/PUT/DELETE for each resource, would be nice to have idempotent create or update (post doesn't necessarily suggest such semantics)

docs/swagger2.yml
```diff
@@ -448,15 +396,16 @@ paths:
           schema:
             $ref: '#/definitions/Error'
 
-  /apps/{app}/calls:
+  /apps/{app}/fn/{fn}/calls:
```
rdallman: not sure we should nest this deep for calls. if we leave them on app you can list calls on an app, which may come from various triggers (and thus functions), for causality this could be useful e.g. function A calls function B. we can further sort down by trigger or function or image, but leaving these hanging off of an app allows the additional flexibility there.

mantree: The idea behind pushing these down is to account for the scale of calls that will occur at an app level. Why deployed this data will most likely be shipped to a dedicated metrics store, which will be interacted with directly.

Putting it down will give cli users something to work with and something that can be supported at scale.

mantree:
* Remove /fns/{fn}/triggers, add fnName query param to /apps/{app}/triggers.
* Names are now used everywhere. The server side will have to resolve the mapping id->name on querying, which shouldn't be too taxing.
* Swagger validates.
* Remove 'wrappers' they didn't add anything to my mind.
* deletes 204.

I think there are two outstanding thoughts, captured as todos at the top, but here to:

# TODO(?): add /run API to swagger doc? (only covers API module atm)
# TODO(orcl): 401 - should we define this to delineate w/ 404 even tho we don't have auth by default?

Least resistance might be to just add 401 as a possible response code to everything.

mantree:
One thing actually, fn versions. My current thinking is the 'image' field on a Fn, is a full docker image description like:

repo/image:version

Then the Call has the executing fn image sha on it. This means that 'version' is not it's own field.

mantree:
* Remove /fns/{fn}/triggers, add fnName query param to /apps/{app}/triggers.
* Names are now used everywhere. The server side will have to resolve the mapping id->name on querying (names are mutable, therefore db schemas should use ids), which shouldn't be too taxing.
* Swagger validates.
* Remove 'wrappers' they didn't add anything to my mind.
* deletes 204.
* All resources have a potential 401 support, although no relevant in a local situation, any deployment of the system is going to want to do this.

I think there are two outstanding thoughts, captured as todos at the top, but here to:

TODO(?): add /run API to swagger doc? (only covers API module atm)

Re: use of PUTS: They do make things simple, but open the door to some data loss on concurrent updates and require complete information for map setting, i.e. when wanting to add an element to a configuration the entire configuration has to be sent, and it will clobber any value set by another user that might have occurred since you fetched the full set. As long as everyone's aware of that and is taking this decision with that risk in mind, in the name of progress I'm ok with it.

rdallman:
> Swagger validates.

you the man.

> Remove 'wrappers' they didn't add anything to my mind.

the nice thing is that you can have an object like:

```go
type Resp struct {
  App models.App `json:"app"`
  Error string `json:"error"` 
}
```

where for things like a call there is a real error field on the call but also a possible error getting it -- it makes them more easily extensible without breaking, but yes to a certain degree it's api design bike shedding. might we keep this the way it was? things like apps/calls won't break unless we do this 

> but open the door to some data loss on concurrent updates and require complete information for map setting

with versioning this should be easy to put a thumb on. and if this comes up that users want patch after a while instead of only get->put we could add it.

will pore over the file itself in a bit i hope, thanks for updates

docs/swagger2.yml
```diff
@@ -27,673 +19,414 @@ consumes:
...
       config:
         type: object
-        description: Trigger configuration - overrides application configuration
-        additionalProperties:
-          type: string
-      annotations:
-        type: object
-        description: Trigger annotations - this is a map of annotations attached to this trigger, keys must not exceed 128 bytes and must consist of non-whitespace printable ascii characters, and the seralized representation of individual values must not exeed 512 bytes
-        additionalProperties:
-          type: object
```
rdallman: config and annotations should remain here I think, and the resource config stuff below should be top level on the func (rather than in the `config:` section

docs/swagger2.yml
```diff
@@ -27,673 +19,414 @@ consumes:
...
-      config:
-        type: object
-        description: Application function configuration, applied to all triggers.
-        additionalProperties:
-          type: string
+        description: unique name for this trigger, used to identify this trigger
+      type:
+        type: string
+        description: class of trigger, e.g. schedule, http
+        enum: [http, stream]
```
rdallman: string? we aren't necessarily restricting the domain here and users could 'roll their own' yea? e.g. i want a mongodb trigger (and the ability to list triggers by type may be something we end up with eventually)

docs/swagger2.yml
```diff
@@ -313,9 +327,9 @@ definitions:
       image_id:
         type: string
         description: "docker image id, sha hash"
-      revision:
+      revision_id:
```
rdallman:
```
revision:
  type: int
```

seems inline w/ what we've discussed, keeping it a string here in case we change our minds about monotonic int?

rdallman:
etag stuff looks good

mantree:
This conversation has been rolled into the 0.2 work.

PR #1019, opened 2018-05-30
fn: read-only root fs becomes default

skinowski:
This change sets root filesystem as read-only by default.
This behavior can be disabled via setting the
    FN_DISABLE_READONLY_ROOTFS
environment variable.

### DIFF
--- a/api/agent/config.go
+++ b/api/agent/config.go
@@ -31,7 +31,7 @@ type AgentConfig struct {
 	PreForkNetworks         string        `json:"pre_fork_networks"`
 	EnableNBResourceTracker bool          `json:"enable_nb_resource_tracker"`
 	MaxTmpFsInodes          uint64        `json:"max_tmpfs_inodes"`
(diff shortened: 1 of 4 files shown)


skinowski:
I'm going to update/add this to docs. The original option was undocumented.

api/agent/config.go
```diff
@@ -56,14 +56,13 @@ const (
...
 	EnvMaxTmpFsInodes          = "FN_MAX_TMPFS_INODES"
-	EnvEnableReadOnlyRootFs    = "FN_ENABLE_READONLY_ROOTFS"
+	EnvDisableReadOnlyRootFs   = "FN_DISABLE_READONLY_ROOTFS"
 
 	MaxDisabledMsecs = time.Duration(math.MaxInt64)
 
 	// defaults
 
-	DefaultHotPoll     = 200 * time.Millisecond
-	DefaultNBIOHotPoll = 20 * time.Millisecond
```
skinowski: was unused.

PR #1021, opened 2018-05-31
fn: user friendly timeout handling changes

skinowski:
Timeout setting in routes now means "maximum amount
of time a function can run in a container".

Total wait time for a given http request is now expected
to be handled by the client. As long as the client waits,
the LB, runner or agents will search for resources to
schedule it.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -274,7 +274,9 @@ func (a *agent) submit(ctx context.Context, call *call) error {
 
 	statsDequeueAndStart(ctx)
 
-	// pass this error (nil or otherwise) to end directly, to store status, etc
+	// Pass this error (nil or otherwise) to end directly, to store status, etc.
+	// Reusing http.Request.Context for slot.exec() since call.Start() updates
+	// this with execution deadline.
 	err = slot.exec(call.req.Context(), call)
 	return a.handleCallEnd(ctx, call, slot, err, true)
 }
@@ -380,7 +382,9 @@ func handleStatsEnd(ctx context.Context, err error) {
 // request type, this may launch a new container or wait for other containers to become idle
 // or it may wait for resources to become available to launch a new container.
 func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
-	// start the deadline context for waiting for slots
+	// start the deadline context for waiting for slots, if slot deadline
+	// is not set, then this means, we will wait as long as the client is
+	// connected (via http.Request.Context())
 	if !call.slotDeadline.IsZero() {
 		tmp, cancel := context.WithDeadline(ctx, call.slotDeadline)
 		ctx = tmp
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -228,10 +228,7 @@ func (a *agent) Submit(callI Call) error {
 
 	call := callI.(*call)
 
-	ctx, cancel := context.WithDeadline(call.req.Context(), call.execDeadline)
-	call.req = call.req.WithContext(ctx)
-	defer cancel()
-
+	ctx := call.req.Context()
 	ctx, span := trace.StartSpan(ctx, "agent_submit")
 	defer span.End()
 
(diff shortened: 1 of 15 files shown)


api/agent/agent.go
```diff
@@ -277,8 +274,10 @@ func (a *agent) submit(ctx context.Context, call *call) error {
 
 	statsDequeueAndStart(ctx)
 
-	// pass this error (nil or otherwise) to end directly, to store status, etc
-	err = slot.exec(ctx, call)
+	// Pass this error (nil or otherwise) to end directly, to store status, etc.
+	// Reusing http.Request.Context for slot.exec() since call.Start() updates
+	// this with execution deadline.
```
rdallman: this seems pretty hairy, is there any way to make this more explicit? move https://github.com/fnproject/fn/pull/1021/files#diff-02ea3c2166e8abe7a3c4684f68d78db5L231 to here?

api/agent/call.go
```diff
@@ -254,40 +254,45 @@ func (a *agent) GetCall(opts ...CallOpt) (Call, error) {
...
-
-	c.slotDeadline = slotDeadline
-	c.execDeadline = execDeadline
+	if c.Type == models.TypeAsync {
+		// *) for async, slotDeadline is also c.Call.timeout. This is because we would like to
+		// allocate enough time for docker-pull, slot-wait, docker-start, etc.
+		// and also make sure we have c.Call.Timeout inside the container as well..
+		// *) for sync, there's no slotDeadline, the timeout is controlled by http-client
+		// context (or runner gRPC context)
+		c.slotDeadline = time.Now().Add(time.Duration(c.Call.Timeout) * time.Second)
```
rdallman: just move this to creating a context here for async https://github.com/fnproject/fn/pull/1021/files#diff-02ea3c2166e8abe7a3c4684f68d78db5L231 and get rid of slot and exec deadline fields? we don't need the state later down the line anywhere else ?

api/agent/call.go
```diff
@@ -387,6 +387,15 @@ func (c *call) Start(ctx context.Context) error {
...
 	}
 
+	// We are about to execute the function, set container Exec Deadline (call.Timeout)
+	// For LB, this does not apply, since execution time is on runners.
+	if c.cancelCtx == nil && !c.isLB {
+		execDeadline := time.Now().Add(time.Duration(c.Timeout) * time.Second)
+		ctx, cancel := context.WithDeadline(ctx, execDeadline)
+		c.req = c.req.WithContext(ctx)
+		c.cancelCtx = cancel
+	}
```
rdallman: makes sense to do this in submit i guess?

rdallman:
not a huge fan of the mutable `call.req` to swap context stuff in / out, there's the context on methods too and this seems brittle (the goal was to surface `call.Req.Context()` early and use it as a standard `ctx` param) - if some better way comes to mind, would prefer it, but if not just ignore me and ship it.

skinowski:
@rdallman a bit cleaner with latest commit I think...

api/agent/agent.go
```diff
@@ -752,8 +761,10 @@ func (a *agent) prepCold(ctx context.Context, call *call, tok ResourceToken, ch
 
 	call.containerState.UpdateState(ctx, ContainerStateStart, call.slots)
 
+	deadline := time.Now().Add(time.Duration(call.Timeout) * time.Second)
+
 	// add Fn-specific information to the config to shove everything into env vars for cold
-	call.Config["FN_DEADLINE"] = strfmt.DateTime(call.execDeadline).String()
+	call.Config["FN_DEADLINE"] = strfmt.DateTime(deadline).String()
```
rdallman: this value won't match up very well, e.g. if it gets hung up waiting on a resource token. think for hot this gets built at call create time, can reuse that value. also, i'm not sure anybody is really using this, though the fdk-go is so anyone using that

skinowski: So, the new timeout semantics (FN_DEADLINE) start at slot.exec() time for hot, although for cold it's not super accurate, it's close...

rdallman: right, the other way would be the issue. pull would throw this off a good bit?

skinowski: We could split this and move pull/prep into prepCold and rest to slot.exec(), but cold is going away...

skinowski:
Ignoring cold-deadline issue for now as cold is slated to be decommissioned.

PR #1024, opened 2018-06-01
fn: fixup temp dir read/write permissions if tmp fs size is not set.

skinowski:
When TmpFsSize is not set in a route, docker fails to create a /tmp
mount that is writable. Forcing docker to explicitly to this if
read-only root directory is enabled (default).

### DIFF
--- a/api/agent/agent_test.go
+++ b/api/agent/agent_test.go
@@ -632,10 +632,106 @@ func TestGetCallReturnsResourceImpossibility(t *testing.T) {
 	}
 }
 
-func TestTmpFsSize(t *testing.T) {
-	// TODO it may be a good idea to mock out the http server and use a real
-	// response writer with sync, and also test that this works with async + log
+//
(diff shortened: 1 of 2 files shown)


denismakogon:
worked fine, thanks!

denismakogon:
But still here would be bad UX in cases when the max size of the FS would not enough. A good example here would a function I've done (it downloads a video file from S3 store and function has no idea of how big video file can be unless there's particular check whether a function can fit a file), OpenCV can't read a video file from in-memory stream, it can only accept a file path (it's weird, indeed, but still one of the possible use cases).

So, i do understand that we can't let a funciton decide how big FS it wants (expect route configuration), but there has to be a way to let function know of how much memory to use it has, similar to RAM.

Thoughts?

skinowski:
@denismakogon Routes have TmpFsSize config param to adjust /tmp size limit. If unset, /tmp is unlimited, but in reality tied to available system memory.

PR #1029, opened 2018-06-02
fn: build tweaks

skinowski:
*) removes git-tag/git-push from circle CI. Here a separate manual tag/PR is required to trigger CircleCI
to release dind and fn images.
*) optimizations provided by changed-since-last.sh script is reverted for time-being.


rdallman:
thanks for digging into this.

to fix the master builds failing all we'd need to do is comment out the `deploy:` section of the CI config, right? this would get us a green check mark, I believe? I'm hesitant to gut / change a lot of the auto deploy process expecting to do it manually, would be nice to keep it the same but try to automate it in another way. At least, it's not clear this is the intent of this patch, but maybe so?

master is the only protected branch we have, as far as I know. I believe we have a number of alternatives to maintain the automated release process albeit none seem to be as convenient as what we have. the easiest solution I can think of in my jet lagged state, is for our deploy CI section to create a 'release' branch automatically and bump the version and push it there (at which point, we need to figure out how to automate getting it into the tip of master, but at least the release process of docker images, etc. is happening and has the same code as master). it would be relatively easy to post PRs of release branch into master, there may be other/better ways and we may be able to securely automate that to keep the same bump version for every change that lands behavior - open to ideas around that or automating this in general.

skinowski:
Closing this (it was an experiment), as we'll likely not use this approach...

PR #1033, opened 2018-06-04
fn: LB agent: reduce 'Too Busy' error logs

skinowski:
With this PR, runner client translates too busy errors
from gRPC session and runner itself into Fn error type.
Placers now ignore this error message to reduce unnecessary
logging.


skinowski:
@kojustin I think they need to become metrics. A counter for number of 'too busy' runner responses, etc.

kojustin:
yeah a metric would be best!

Issue #1030, opened 2018-06-02
expose FN_TMPFSSIZE and FN_TMPFS


skinowski:
Introduce two new env variables to expose tmpfs size and location to applications. For example:

FN_TMPFSSIZE=128
FN_TMPFS=/tmp

in MB.

rdallman:
this feels like something leaking from the service into OSS and if I recall from internal briefings long term tmpfs isn't something we're planning to rely on. `disk` seems like a knob we want to expose for users to turn on each func, which perhaps is tmpfs at some point and a real disk at other times, but for compat reasons and for simplicity reasons we should probably try to hash out what we really want to support here as breaking things like env vars is challenging.

skinowski:
Also, similar to FN_MEMORY, FN_TMPSIZE will not be actual /tmp size available (but what's configured for that route/function) since between calls in hot containers, /tmp will persist until hot container is dead. Their provided value in env (as config) is a bit dubious, so maybe we can punt on them for time being.

PR #1034, opened 2018-06-04
fn: swagger update for tmpfs size

skinowski:


rdallman:
same question as in #1030 -- long term viability / desire of this as a feature was left murky

skinowski:
Let's hold off on this one... Until config/operational config decision is finalized.

skinowski:
@shaunsmith was asking about documentation.

zootalures:
We're going to remove user configuration this  and CPU configuration from the APIs in the trigger change -  as discussed previously it's unlikely that honoring these by guarantee (without adding substantial complexity to scheduling) will work in a scaled up system .  For both we'll set these by configuration in the server as a function of memory initially .

rdallman:
wrote that up (at last) https://github.com/fnproject/fn/issues/1106 i'm not sure it was well circulated that it was planned in triggers changes, feel free to add [0.2] to title or 'waiting' tag depending on scope

skinowski:
Sounds good. This PR was a consistency change to keep route configuration: cpu/mem/tmpfssize together. Since tmpfssize was merged in already, it should be in swagger. When we remove it, we can remove it along with cpu, etc.

PR #1035, opened 2018-06-05
fn: lb-agent: Initial TryCall result can be retriable.

skinowski:
Before this change, we assumed data may end up in a container
once we placed a TryCall() and if gRPC send failed, we did not
retry. However, a send failure cannot result in data in a
container, since only upon successful receipt of a TryCall can
pure-runner schedule a call into a container. Here we trust
gRPC and if gRPC layer says it could not send a msg, then
the receiver could not have received the message fully.


skinowski:
In other words, Sending tryCall can only fail (on sender side), if stream is broken, and if stream is broken, receiver cannot receive a full tryCall message and cannot place a call via agent.Submit().

PR #1036, opened 2018-06-05
add functions CRUD api & model

rdallman:
this simply plumbs the function API model and CRUD endpoints, without doing
any removal of existing constructs in fn. that is, it is possible to
create/read/update/delete functions without the ability to invoke them at all,
whatsoever. the hope is that this reiterates that we have too many test
suites, reiterates that my karma is far more negative than originally thought,
and this allows the CLI and other datastore implementers to get unblocked on
this, and we can then begin the re-plumbing of agent to invoke functions
rather than routes on top of this.

note that this does not reflect extremely recent discussions around certain
fields on a func, it should be rather easy to modify this code to change/add
any fields that we want to. we need to put them on an app as we've decided,
which is pretty straightforward from this, too. The other thing that sticks
out is that PUT behaves very much like PATCH and there is maybe a discussion
to have around that (esp. keeping in mind that we're adding revisions).

we can quickly rename this to `fns` too, nothing final here.

### DIFF
--- a/api/datastore/sql/sql.go
+++ b/api/datastore/sql/sql.go
@@ -86,6 +86,22 @@ var tables = [...]string{`CREATE TABLE IF NOT EXISTS routes (
 	app_id varchar(256) NOT NULL,
 	log text NOT NULL
 );`,
+
+	`CREATE TABLE IF NOT EXISTS funcs (
+	id varchar(256) NOT NULL,
+	name varchar(256) NOT NULL,
+	image varchar(256) NOT NULL,
+	format varchar(16) NOT NULL,
+	cpus int NOT NULL,
+	memory int NOT NULL,
+	timeout int NOT NULL,
+	idle_timeout int NOT NULL,
+	config text NOT NULL,
+	annotations text NOT NULL,
+	created_at varchar(256) NOT NULL,
+	updated_at varchar(256) NOT NULL,
+	PRIMARY KEY (name)
+);`,
 }
 
 const (
@@ -93,6 +109,7 @@ const (
 	callSelector      = `SELECT id, created_at, started_at, completed_at, status, app_id, path, stats, error FROM calls`
 	appIDSelector     = `SELECT id, name, config, annotations, syslog_url, created_at, updated_at FROM apps WHERE id=?`
 	ensureAppSelector = `SELECT id FROM apps WHERE name=?`
+	funcSelector      = `SELECT id, name, image, format, cpus, memory, timeout, idle_timeout, config, annotations, created_at, updated_at FROM funcs`
 
 	EnvDBPingMaxRetries = "FN_DS_DB_PING_MAX_RETRIES"
 )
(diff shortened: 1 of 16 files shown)


api/datastore/sql/sql.go
```diff
@@ -86,13 +86,30 @@ var tables = [...]string{`CREATE TABLE IF NOT EXISTS routes (
...
+	format varchar(16) NOT NULL,
+	cpus int NOT NULL,
+	memory int NOT NULL,
+	timeout int NOT NULL,
+	idle_timeout int NOT NULL,
+	config text NOT NULL,
+	annotations text NOT NULL,
+	created_at varchar(256) NOT NULL,
+	updated_at varchar(256) NOT NULL,
+	PRIMARY KEY (name)
```
mantree: id should be the primary key, name should have a UNIQUE constraint.

api/datastore/sql/sql.go
```diff
@@ -689,6 +706,154 @@ func (ds *SQLStore) GetRoutesByApp(ctx context.Context, appID string, filter *mo
...
+		} else {
+			// update it
+			dst.Update(fn)
+			err = dst.Validate()
+			if err != nil {
+				return err
+			}
+			fn = &dst // set for query & to return
+
+			query = tx.Rebind(`UPDATE funcs SET
```
mantree: Does this mean update structs must contain all values?

api/datastore/sql/sql.go
```diff
@@ -915,6 +1035,46 @@ func buildFilterCallQuery(filter *models.CallFilter) (string, []interface{}) {
...
+	if filter == nil {
+		return "", nil
+	}
+	var b bytes.Buffer
+	var args []interface{}
+
+	// where(fmt.Sprintf("image LIKE '%s%%'"), filter.Image) // TODO needs escaping, prob we want prefix query to ignore tags
+	args = where(&b, args, "id>?", filter.Cursor)
+
+	fmt.Fprintf(&b, ` ORDER BY id ASC`)
```
mantree: The swagger specifies results are sorted alphabetically. Though not entirely sure this executes those requests.

api/models/datastore.go
```diff
@@ -59,6 +59,24 @@ type Datastore interface {
...
+	// GetFuncs returns a list of funcs, applying any additional filters provided.
+	GetFuncs(ctx context.Context, filter *FuncFilter) ([]*Func, error)
+
+	// GetFunc returns a function by name. Returns ErrDatastoreEmptyFuncName if funcName is empty.
+	// Returns ErrFuncsNotFound if a func is not found.
+	// TODO(reed): figure out addressable by id or name biz. iff 1 query, name works.
+	GetFunc(ctx context.Context, funcName string) (*Func, error)
+
+	// RemoveFunc removes a function. Returns ErrDatastoreEmptyFuncName if funcName is empty.
+	// Returns ErrFuncsNotFound if a func is not found.
```
mantree: FuncNotFound isn't an error... it's expected behaviour, maybe this is a goism i'm not used to.

api/models/datastore.go
```diff
@@ -59,6 +59,24 @@ type Datastore interface {
...
+	// TODO(reed): should we allow rename if id provided?
+	PutFunc(ctx context.Context, fn *Func) (*Func, error)
+
+	// GetFuncs returns a list of funcs, applying any additional filters provided.
+	GetFuncs(ctx context.Context, filter *FuncFilter) ([]*Func, error)
+
+	// GetFunc returns a function by name. Returns ErrDatastoreEmptyFuncName if funcName is empty.
+	// Returns ErrFuncsNotFound if a func is not found.
+	// TODO(reed): figure out addressable by id or name biz. iff 1 query, name works.
+	GetFunc(ctx context.Context, funcName string) (*Func, error)
```
mantree: Going to need more flex here.

api/models/func.go
```diff
@@ -0,0 +1,224 @@
...
+	IdleTimeout int32 `json:"idle_timeout,omitempty" db:"idle_timeout"`
+}
+
+// SetDefaults sets zeroed field to defaults.
+func (f *Func) SetDefaults() {
+	if f.ID == "" {
+		f.ID = id.New().String()
+	}
+
+	if f.Memory == 0 {
```
mantree: Aren't these all natural go zero values? Only ID and time seem like they need to be here.

api/server/server.go
```diff
@@ -884,6 +884,11 @@ func (s *Server) bindHandlers(ctx context.Context) {
 			v1.GET("/apps", s.handleAppList)
 			v1.POST("/apps", s.handleAppCreate)
 
+			v1.GET("/funcs", s.handleFuncsList)
```
mantree: Should we just chuck all this under /v0.2 or something similar?

api/datastore/sql/sql.go
rdallman: they don't have to as `dst.Update(fn)` will fill in all fields from the existing, updating any modified. one thing of note is that as it stands this keeps the strange config deletion failure which seems to get us a bug report every 6-8 weeks, we can change that (it would change app/route configs as well, so left for now), need to discuss UX as e.g. `fn apps config set myKey myValue` won't work as expected anymore in CLI

api/models/datastore.go
rdallman: just rev yea?

api/datastore/sql/sql.go
rdallman: yea i waffled back and forth on this. can switch it back to that, boy do ids make great cursors though

api/models/datastore.go
rdallman: on delete (well, most endpoints) for all of our resources we return not found if it didn't exist. as far as go errors, we encode http statuses in these fancy errors https://github.com/fnproject/fn/blob/master/api/models/error.go as well as implementing `models.APIError` to hide errors we're less sure of from users.

api/models/func.go
rdallman: we call this on the func that we deserialized from the request body, so e.g. memory could be set to 42 and we don't want to override it if that's the case

api/server/server.go
rdallman: that was an option, i think we decided amongst ourselves that since v1 isn't really v1 we could just use v1

api/datastore/sql/sql.go
mantree: Config deletion? It is a problem with complete update style apis, but fixing it can be an expense people tend not to like.

api/server/server.go
mantree: thinking of terms of allowing both the new and old to sit side by side for a bit, but maybe big bang is the way to go

api/datastore/sql/sql.go
rdallman: sorry, I mean config key delete doesn't work if you simply update a config that omits a key. so if a user sends in a config `{"a":"v", "b":"v"}` when config `{"c":"v"}` exists, you get config `{"a":"v", "b":"v","c":"v"}` and it seems most of our users want the first config shown here to be the final config, not the one I've shown that simply adds new keys to existing. as it stands, to delete a key, users have to do `{"a":""}` (send an empty value), and thus far this has not been something intuited by users.

rdallman:
one thing to figure out maybe while we're at this point is do we want to use function ids in the backend (datastore) or names? currently this uses names, and app id is, well, app id, if we want to do things that way. Not clear the best path forward there.

api/datastore/sql/sql.go
mantree: Ah ok, the opposite implementation.

I think if we have CAS, we can do the first version, i.e. just replace the entire config rather than merge.

rdallman:
@mantree @riconnon do we want to merge this to `fn-v2.0`? we can use it as a WIP branch to stage work into if we'd like.

afaict the immediate work to do here:

* rename to `fns`
* put on app
* maybe use id in same vein as app id?

with maybe now or a little bit later:

* revisions (across apps, funcs, and triggers - maybe do all at once)
* status (across funcs and triggers - maybe do all at once)
* change config behavior
* docker image id - we want the client to provide this, probably sep discussion item, not urgent

rdallman:
we can close, assume #1062 supercedes

PR #1038, opened 2018-06-05
[WIP] fn v0.2

rdallman:
we've decided to go back in time from #957 with new hotness. notable deltas:

* revisions on every resource for CAS PUT action
* docker shas in func for 'immutable' functions [by rev]
* call functions fns instead of funcs, cuz :yodawg: i heard you use fn to run your fns
* invoke fns and not triggers, triggers do not have config/annotations
* fns exist under an app instead of rooted / on namespace, toss out func registry dreams for now
* murder logs/calls, plans to provide debug invoke to get logs and metrics to get same telemetry
* trigger manager from 2.0 thing is no longer a thing
* fn/trigger enabled/disabled bit 

we are working to implement this spec on this branch, feedback is very welcome.


rdallman:
i've added the previous swagger we had in as the `swagger02-final.yaml` - this is the swagger we're working towards implementing, with the `swagger02.yaml` file being the swagger file we expect to have implemented (for use in client bindings / api docs) and that we can evolve.

rdallman:
these are all encapsulated as issues in https://github.com/fnproject/fn/projects/1 now and #1086 supersedes the additional work to implement that spec here, closing

PR #1044, opened 2018-06-07
fn: introducing hot function max requests limit

skinowski:
Other than idle_timeout, hot-functions have a potential
of staying alive unbounded amount of time. In certain routes,
it may be desirable to limit total request count.

Setting max_requests to 1 also simulates a cold-function behavior
(but with superior formats: json, cloudevents, http) since
only one execution will be performed in a given container.
This change is also a preliminary step for decommissioning cold-format
since max_requests parameter can now meet cold requirements.

In busy deployments, max_requests can be valuable workaround
for resource leaks in problematic routes.

In future, max_time can also be added to limit the total
execution time.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -842,6 +842,7 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
 	ctx, shutdownContainer := context.WithCancel(ctx)
 	defer shutdownContainer() // close this if our waiter returns, to call off slots
 	go func() {
+		remaining := call.MaxRequests
 		defer shutdownContainer() // also close if we get an agent shutdown / idle timeout
 
 		for {
@@ -870,6 +871,13 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
 			if slot.fatalErr != nil {
 				logger.WithError(slot.fatalErr).Info("hot function terminating")
 				return
+			}
+
+			if remaining != 0 {
+				remaining--
+				if remaining == 0 {
+					return
+				}
 			}
 		}
 	}()
--- a/api/agent/agent_test.go
+++ b/api/agent/agent_test.go
@@ -845,6 +845,181 @@ func TestTmpFsSize(t *testing.T) {
 	}
 }
 
+//
+// Here we clamp MaxRequests to 2 and run 4 requests. We expect
+// two hot-containers each taking about 2 requests.
+//
+func TestMaxRequestsI(t *testing.T) {
+	appName := "myapp"
+	path := "/hello"
+	url := "http://127.0.0.1:8080/r/" + appName + path
+
+	app := &models.App{Name: appName}
+	app.SetDefaults()
+
+	reqCount := 4
+	maxRequests := uint64(2)
+
+	// let's create a MaxRequests 1 route. And every request should be
+	// handled by a new hot-container.
+	ds := datastore.NewMockInit(
+		[]*models.App{app},
+		[]*models.Route{
+			{
+				Path:        path,
+				AppID:       app.ID,
+				Image:       "fnproject/fn-test-utils",
+				Type:        "sync",
+				Format:      "http", // this _is_ the test
+				Timeout:     5,
+				IdleTimeout: 100,
+				MaxRequests: maxRequests,
+				Memory:      64,
+			},
+		},
+	)
+
+	ls := logs.NewMock()
+	a := New(NewDirectDataAccess(ds, ls, new(mqs.Mock)))
+	defer checkClose(t, a)
+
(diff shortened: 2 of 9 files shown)


skinowski:
Alternatively, we could:

1) add max_requests (and also a max_exectime) to limit all containers in a agent cfg.

This allows a service to safely upper bound both without adding per-route settings.

Then we lose the cold-like execution in a route. (Does anyone really need cold-like execution?) or
such settings on specific routes but not others?

rdallman:
not sure that this will fall out as expected with the 'default' format specifically, the main issue being that when a currently 'default' image starts it can exit before the `runHot` thread would have a chance to receive a slot for it and attach to its pipes (if it ignores `stdin`) - meaning the call doesn't have any output when the container did have output - currently default is the way it is due to this (poorly documented, sorry). i'm not sure that trying to figure out such a race is worth it. I know there have been a number of discussions around 'how do we get default format behavior if default format isn't baked into fn', though nothing is turning up there are a couple threads. (sorry this reply sucks, I will elaborate more tomorrow if I can).

the other discussion i saw the decision log around the knobs for a route specifically, haven't had time to digest this either sorry. i'm pretty firmly in the 'minimal number of knobs' category myself, but i think me and my flimsy backbone have already been defeated to a certain degree (all is not lost?) - a story around how do we have the knobs without the user seeing them seems like a productive discussion worth engaging in, everybody wins (that is my main concern with having them).

skinowski:
To be clear, MaxRequests=1 does not allow default format in hot-containers. It just allows execute-once support in json/cloudevents/http formats.

kmjohansen:
I'm generally in favor of instituting a max-requests policy for containers.  Other security sensitive software enforces a finite lifetime on their processes to mitigate bugs, leaks, and other issues.  At least at a high-level, I think this idea makes sense.

From a customer standpoint, however, it's going to be intrusive to have every Nth request generate a cold-container start, at least until we're able to optimize our cold start times.

I'm not a huge fan of having lots of knobs either.  My preference would be to have the software auto-tune, or choose sane defaults.  However, there are cases where we can expect to have values that need to be overridden on a per-customer basis.  In that case, I'd rather we make that easy to observe but I'm flexible about the exact mechanism.  For this particular question, either agent or route configuration seems most sensible.  The argument in favor of agent is that as a security feature we won't be able to have hostile software change the parameter for us in the message payload.  The argument in favor of a per-route option presumes that it would be more important for us to modify this for some functions, in case the customer is severely impacted by our default.

As far as cold containers go, I suspect that it's really only useful to us internally as a testing tool to keep the `max-requests=1` behavior.

denismakogon:
> In busy deployments, max_requests can be a valuable workaround
for resource leaks in problematic routes.

Indeed, however, there is at least one use cases when a function should be aware of how many times it can be called - "self-called" or "self-triggered" function, the idea here that the function can reschedule itself certain number of times before it dies (a function calls itself if requirement match after certain time passed).

Here's the use case, trigger a function only when the bucket has the particular amount of objects. So, the function queries the store, if the requirement met - end or do whatever is necessary, if not - sleep and call itself one more time.

So, the feature you'd like to introduce should not actually affect the use case I've described, because it is still valid. However, the max number of requests could be used by a function to limit the number of times it can reschedule itself.

rdallman:
i believe this was discussed somewhere, is there a reason to prefer a count of executions to cap container lifetime vs. a time based solution? for the 1 case, max executions makes sense, but for something like the 100 case, max executions seems like it could lead to the container being around for anywhere between a few seconds to hours (say, the function is called with 30 minutes between invocations and idle timeout is higher than that value). if we end up with a max number of executions settings, and a 'max active duration' to cap the lifetime of a container based on time, we would have 2 knobs to do very similar things and it gets hard to reason about if both are set, etc. may be worth discussing that before adding one or the other (or maybe not)

kmjohansen:
I'm not sure that it's an either or situation.  I think having a time based expiration policy makes sense.  I consider the request-count based expiration policy to be complimentary.  It works best in a situation where an attacker needs multiple tries in order to guess the layout of memory, leak something in the right place to overwrite later, or just generally exhaust resources.  You're absolutely correct that this has the potential to impact busy customers, and it may be something that we want to disable or tune up.  I consider it to be good security policy, but we can argue about whether it makes sense to be a mandatory setting -- I certainly take your point about it exacerbating the impact of container start time.

skinowski:
Most similar systems have both of these.. I think we need both and I can add time-based max-exectime to this PR as well... (after we decide where these would go...)

skinowski:
Closing this. I still like the idea of bounded execution with maximum time & maximum number of requests, etc. But the PR is old and we can rework this later if it comes up as high priority.

PR #1046, opened 2018-06-07
Adding a way to inject a request ID

reclaro:
It is very useful to associate a request ID to each incoming request,
this change allows to provide a function to do that via Server Option.
The change comes with a default function which will inspect the incoming
request to see if there is a header with a request ID, if it's there it
uses that value as request ID if it is not there it will generate a new
request ID. The request ID is put in the request context along with a
common logger which always logs the request-id


api/common/request_id_util.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/gin-gonic/gin"
+	"github.com/sirupsen/logrus"
+)
+
+type key string
+
+const (
+	ridLabel = "request-id"
+	// RidKey is the name of the key saved in the context for holding the request ID
+	RidKey    = key(ridLabel)
```
rdallman: nit: this should be unexported to force usage through certain methods

rdallman:
this seems like the right idea to me. needs to add the same middleware component to the lb, am not sure where that is at the moment.

api/server/server_options.go
```diff
@@ -10,6 +10,8 @@ import (
 
 type ServerOption func(context.Context, *Server) error
 
+type RIDHandler func(string) func(c *gin.Context)
```
kppk: Thinking out loud here:

Would it make sense to have stronger semantics around RID? 

We know it is header (and context value), we want to set it if it is not set, we want some string value provider if the header is not set, we want the header parameter name to be configurable

What about having something like:

`type RIDProvider struct {
	Name string
	ProvideNew func() string
}`

and then:

`WithRIDProvider(ridProvider RIDProvider) ServerOptions`

This way we would not need to leak the gin.Context and also the 'if the value not set, set new value' logic might be handled by fn server.

Gopkg.toml
```diff
@@ -81,3 +81,7 @@ ignored = ["github.com/fnproject/fn/cli"]
 [[constraint]]
   name = "github.com/dchest/siphash"
   version = "1.1.0"
+
+[[constraint]]
+  branch = "master"
+  name = "golang.org/x/net"
```
rdallman: don't see a corresponding dep update here, this is implicit? can remove i think

api/common/ctx.go
```diff
@@ -9,6 +9,16 @@ import (
...
 
+//RidKey is the context key to use to store a request ID
+type RidKey string
+
+const ridCtxKey = RidKey("fn_request_id")
+
+//RIDContextKey returns the context key to use for dealing with request IDs
+func RIDContextKey() RidKey {
+	return ridCtxKey
+}
```
rdallman: couple little things to clean up here. casing should be consistent on `RID`, `s/RidKey/RIDKEY/`. from usage, i'm not sure we want to provide alternate keys here? if we do, maybe it should be a configurable var (plumbing is never fun) -- this seems like a thing we should punt on since we don't need it right now.

the key is needed only in `common`, we shouldn't expose the type for that (the type is generally there to keep context value keys from conflicting if they end up being the same string/int value. e.g. `keyAType("a") != `otherKeyAType("a")`) - the methods hide the usage anyway, and shouldn't be needed. if we use a const for the `RIDContextKey` value it can be of either type we need it for in various places, as opposed to having to cast it to a string from the `agent` package (`agent` should get a `string` and not have to deal with this cast).

```go
// unexported type
type ridKey string

// since this is a const, it can be of type ridKey or string, when needed.
const RIDContextKey = "fn_request_id"
```

rdallman: it's generally the case that type declarations should be the fully named type and only local variables should have the shorthand for it, such as `rid` representing a `RequestID` - it would be great to expand the exported `RIDContextKey` to `RequestIDContextKey` or `RequestIDKey` such here, as well (for example, in the `agent` caller, it's hard to know what an `RID` is immediately upon looking at it - this should be the case.

api/common/request_id_util.go
```diff
@@ -0,0 +1,33 @@
+package common
+
+import (
+	"context"
+
+	"github.com/fnproject/fn/api/id"
+)
+
+const (
+	maxLength = 32
```
rdallman: would possibly remove this since it's likely only bound to make us slip up and find a conflicting request id early on (this is never fun to debug, have lived that one). the basis would be that since we're creating the request ids, we shouldn't need to verify their length when we retrieve them - the counter point being more nefarious people could inject really long request ids if they want, but the boundary between lb and server is relatively secure in this respect (using https) and the chance of mishap likely exceeds the chance of users attempting to inject bad request ids. 

a possibly less error prone magic number may be something more along the lines of `1024`, which still limits the memory usage, but it's not likely we'll ever generate an id that is 1024 byte (32 on the other hand is much more likely to be exceeded).

api/common/request_id_util.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+	if len(ridFound) > maxLength {
+		// we truncate the rid to the value specified by maxLenght const
+		return ridFound[:maxLength]
+	}
+	return ridFound
+}
+
+//IncomingRID extract the request id from ctx
+func IncomingRID(ctx context.Context) string {
```
rdallman: `RequestIDFromContext` is more conventional, the preferred way to structure seems to be more something akin to `requestid.FromContext` where we'd have a `requestid` package, but it's unclear if we need/want that.

api/common/request_id_util.go
rdallman: this is effectively unused since `id.New().String()` is a static length - we can remove this.

api/server/server.go
```diff
@@ -167,6 +169,11 @@ func NewFromEnv(ctx context.Context, opts ...ServerOption) *Server {
 	opts = append(opts, WithNodeCert(getEnv(EnvCert, "")))
 	opts = append(opts, WithNodeCertKey(getEnv(EnvCertKey, "")))
 	opts = append(opts, WithNodeCertAuthority(getEnv(EnvCertAuth, "")))
+	ridProvider := &RIDProvider{
+		HeaderName:   getEnv(EnvRidHeader, "fn-request-id"),
+		RIDGenerator: common.FnRequestID,
```
rdallman: could inline this one even: `RIDGenerator: func() string { return id.New().String() }` - pretty terse, one less exposed method from common

rdallman: nvm, not as easy the way it's set up now. i'd think we'd want the same `if ridFound { return rid } else { return gen() }` logic for any rid provider, though, and maybe `RIDProvider` should only be a generator as opposed to a branch to handle a request id existing. I don't have full context here - if that's what we need, then that is perfectly ok.

api/server/server.go
```diff
@@ -167,6 +169,11 @@ func NewFromEnv(ctx context.Context, opts ...ServerOption) *Server {
 	opts = append(opts, WithNodeCert(getEnv(EnvCert, "")))
 	opts = append(opts, WithNodeCertKey(getEnv(EnvCertKey, "")))
 	opts = append(opts, WithNodeCertAuthority(getEnv(EnvCertAuth, "")))
+	ridProvider := &RIDProvider{
+		HeaderName:   getEnv(EnvRidHeader, "fn-request-id"),
```
rdallman: in `agent/runner_client` we're using `fn_request_id` from `common.RIDContextKey()` but this is `fn-request-id` - am not sure if we want to be consistent, pointing this out.

rdallman:
overall LGTM, some minor janitorial comments. thanks for adding this here :)

skinowski:
I think the changes need a bit clarification as who adds the id and when. Because as is, server.NewFromEnv() seems to add a handler regardless of type (lb, runner, pure-runner, api).
IMHO, two points:

1) the rid generator (which seems to be intended for LB only) should not try to fetch id from
untrusted client http headers. It should always generate the id.

2) The rid generator option should not be part of server.NewFromEnv() by default because
it needs to be optional. (see: system tests, which only adds it to LB, which I think is what we want.)

api/common/ctx.go
reclaro: ok thanks. I'll use the const and get rid of the exported type. I am going to add a new method in common called WithRequestID to store a request ID into the context.

api/common/request_id_util.go
reclaro: Renamed to RequestIDFromContext, thanks

api/server/server.go
reclaro: I see it more as a generator so I'd prefer to leave it as it is, it gives us more flexibility

api/server/server.go
reclaro: good catch, thanks.

grpcutil/dial.go
```diff
@@ -81,3 +85,27 @@ func CreateCredentials(certPath, keyPath, caCertPath, certCommonName string) (cr
...
 }
+
+// RIDStreamServerInterceptor is a gRPC stream interceptor which gets the request ID out of the context and put a logger with request ID logged into the common logger in the context
+func RIDStreamServerInterceptor(srv interface{}, stream grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+	ctx := stream.Context()
+	newStream := grpc_middleware.WrapServerStream(stream)
+	md, ok := metadata.FromIncomingContext(ctx)
+	if ok {
+		ridKeyToString := string(common.RIDContextKey())
+		rid := md[ridKeyToString][0]
```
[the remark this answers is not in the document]
reclaro: right! I've extrapolated the common code in a function and added some tests for it. Thanks

grpcutil/dial.go
```diff
@@ -81,3 +85,27 @@ func CreateCredentials(certPath, keyPath, caCertPath, certCommonName string) (cr
...
+	}
+	return handler(srv, newStream)
+}
+
+// RIDUnaryServerInterceptor is an unary gRPC interceptor which gets the request ID out of the context and put a logger with request ID logged into the common logger in the context
+func RIDUnaryServerInterceptor(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+	md, ok := metadata.FromIncomingContext(ctx)
+	if ok {
+		ridKeyToString := string(common.RIDContextKey())
+		rid := md[ridKeyToString][0]
```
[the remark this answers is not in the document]
reclaro: Ditto

api/agent/runner_client.go
```diff
@@ -117,7 +118,9 @@ func (r *gRPCRunner) TryExec(ctx context.Context, call pool.RunnerCall) (bool, e
 		// If we can't encode the model, no runner will ever be able to run this. Give up.
 		return true, err
 	}
-
+	// Adding the request-Id to the gRPC metadata
+	mp := metadata.Pairs(string(common.RIDContextKey()), common.IncomingRID(ctx))
```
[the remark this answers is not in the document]
reclaro: good point, done and done.

api/common/request_id_util.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+// FnRequestID returns the passed value if that is not empty or too long otherwise it generates a new unique ID
+func FnRequestID(ridFound string) string {
+	if ridFound == "" {
+		return id.New().String()
+	}
+	if len(ridFound) > maxLength {
+		// we truncate the rid to the value specified by maxLenght const
+		return ridFound[:maxLength]
```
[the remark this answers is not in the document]
reclaro: you don't like this and @rdallman  doesn't either so I am thinking that I am wrong :) As the request ID can be passed in the incoming request, I wanted to have a way to guard against insane big values passed. Looking at the HTTP RFC there is no limit on the length of a header field but the common pattern here is that the check for the size of headers is a webserver's job and in fact, most of them have default limits...Well yes you have convinced me to get rid of the truncation.

api/common/request_id_util.go
reclaro: I'll remove it.

api/common/request_id_util.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+	return ridFound
+}
+
+//IncomingRID extract the request id from ctx
+func IncomingRID(ctx context.Context) string {
+	rid, found := ctx.Value(RIDContextKey()).(string)
+	if !found {
+		log := Logger(ctx)
```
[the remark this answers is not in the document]
reclaro: I'll return an empty string

api/server/server.go
```diff
@@ -167,6 +169,11 @@ func NewFromEnv(ctx context.Context, opts ...ServerOption) *Server {
...
+		RIDGenerator: common.FnRequestID,
+	}
+	opts = append(opts, WithRIDProvider(ridProvider))
```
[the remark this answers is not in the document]
reclaro: There is no validation at the moment but implementing this feature as a server option should give the flexibility to inject a custom request ID generator which can perform validation. 
The implementation offered here is a basic out-of-the-box implementation, probably it is better to not enable ID by default, I am happy to remove the code here so the new feature is not automatically enabled.

reclaro:
@treeder and @skinowski thank you both for the great reviews. With reference to @skinowski final comment I think we can change the signature for the RIDGenerator function which could accept a HTTP.Request as param instead of a string, doing that we delegate to the RIDProvider decide if the RID will be looked up in the headers, somewhere else or just generate a new ID from scratch. What do you think?

api/common/request_id_util.go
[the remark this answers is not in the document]
skinowski: Yes, a production server must go through configuring proper gin/http-server settings which includes header size, how long we wait for this header, I/O read/write timeouts, etc. etc. We can assume this will be covered unless we want to clamp on a specific size (eg. 2k) and return an error. Either way is fine...

PR #1048, opened 2018-06-07
ID should be primary key on apps model and name should be unique.

riconnon:
This allows for renaming apps in the future.

### DIFF
--- a/api/datastore/sql/migrations/15_alter_apps.go
+++ b/api/datastore/sql/migrations/15_alter_apps.go
@@ -1,0 +1,84 @@
+package migrations
+
+import (
+	"context"
+
+	"github.com/fnproject/fn/api/datastore/sql/migratex"
+	"github.com/jmoiron/sqlx"
+)
+
+func up15(ctx context.Context, tx *sqlx.Tx) error {
+	_, err := tx.ExecContext(ctx, "ALTER TABLE apps RENAME TO old_apps;")
+	if err != nil {
+		return err
+	}
+	newTable := `CREATE TABLE apps (
+	id varchar(256) NOT NULL PRIMARY KEY,
+	name varchar(256) NOT NULL UNIQUE,
+	config text NOT NULL,
+	annotations text NOT NULL,
+	created_at varchar(256),
+	updated_at varchar(256),
+	syslog_url text
+);`
+	_, err = tx.ExecContext(ctx, newTable)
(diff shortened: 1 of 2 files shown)


api/datastore/sql/migrations/15_alter_apps.go
```diff
@@ -0,0 +1,84 @@
...
+	_, err := tx.ExecContext(ctx, "ALTER TABLE apps RENAME TO old_apps;")
+	if err != nil {
+		return err
+	}
+	newTable := `CREATE TABLE apps (
+	id varchar(256) NOT NULL PRIMARY KEY,
+	name varchar(256) NOT NULL UNIQUE,
+	config text NOT NULL,
+	annotations text NOT NULL,
+	created_at varchar(256),
```
mantree: Nit pick, but these should be NOT NULL to right?

api/datastore/sql/migrations/15_alter_apps.go
```diff
@@ -0,0 +1,84 @@
...
+		return err
+	}
+	insertQuery := `INSERT INTO apps(id,name,config,annotations,created_at,updated_at,syslog_url)
+					SELECT  id,name,config,annotations,created_at,updated_at,syslog_url FROM old_apps;`
+
+	_, err = tx.ExecContext(ctx, insertQuery)
+	if err != nil {
+		return err
+	}
+
```
mantree: Worth a count comparison? make sure everythings made it? Probably not that big a deal really.

api/datastore/sql/migrations/15_alter_apps.go
riconnon: Perhaps, but that's unrelated if so.

api/datastore/sql/migrations/15_alter_apps.go
riconnon: Pretty sure we'd get an error if we had a problem.

api/datastore/sql/migrations/15_alter_apps.go
rdallman: ah, good idea

rdallman: there may exist some fns that actually do have apps with a null created_at, since it was an addition a while ago (last november?), but it seems unlikely

PR #1056, opened 2018-06-08
Extract register view functionality

### LINKED ISSUE #1041: Operational metrics: slice and dice different ways
There are a number of potential deployment options for fn: eg, a single-tenant on-prem version, or a multi-tenant hosted service, amongst others.

We want to be able to support operations metrics across a range of different cases.

At the moment, the exported views are okay for a single-tenant version (they use app and route as the axes for metric labels). This makes sense to me if I'm the only consumer; I'd like to be able to see the ways my various functions are behaving (or not).

However, in the multi-tenant case, there are problems here. App and route are not unique to a tenant (potentially). They are open-ended sets; this may be in conflict with the time-series backend (eg, prometheus gives some advice about not multiplying label values without necessity).

As a multi-tenant service operator, I'd like alternative options: either logging these kinds of stats per tenant, or simply aggregating across the board.

We need to pull the view creation out of the init() functions and explicitly call it from a main(); that way, people repurposing Fn can replace the metric exposure with their own flavour.

peterj:
Fixes #1041

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -20,7 +20,6 @@ import (
 	"github.com/go-openapi/strfmt"
 	"github.com/sirupsen/logrus"
 	"go.opencensus.io/stats"
-	"go.opencensus.io/stats/view"
 	"go.opencensus.io/trace"
 )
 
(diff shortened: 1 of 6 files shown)


rdallman:
branch is out of date I believe with the morning merge, apologies, will click the big green button once it's fixed up - thanks

PR #1058, opened 2018-06-12
fn: introducing lb placer basic metrics

skinowski:
This change adds basic metrics to naive and consistent
hash LB placers. The stats show how many times we scanned
the full runner list, if runner pool failed to return a
runner list or if runner pool returned an empty list.

Placed and not placed status are also tracked along with
if TryExec returned an error or not. Most common error
code, Too-Busy is specifically tracked.

If client cancels/times out, this is also tracked as
a client cancel metric.

For placer latency, we would like to know how much time
the placer spent on searching for a runner until it
successfully places a call. This includes round-trip
times for NACK responses from the runners until a successful
TryExec() call. By excluding last successful TryExec() latency,
we try to exclude function execution & runner container
startup time from this metric in an attempt to isolate
Placer only latency.

### DIFF
--- a/api/runnerpool/ch_placer.go
+++ b/api/runnerpool/ch_placer.go
@@ -7,11 +7,12 @@ import (
 	"context"
 	"time"
 
+	"github.com/fnproject/fn/api/common"
 	"github.com/fnproject/fn/api/models"
 
 	"github.com/dchest/siphash"
-	"github.com/fnproject/fn/api/common"
 	"github.com/sirupsen/logrus"
+	"go.opencensus.io/stats"
 )
 
 type chPlacer struct {
@@ -30,28 +31,34 @@ func NewCHPlacer() Placer {
 // the LB_WAIT to drive placement decisions: runners only accept work if they have the capacity for it.
 func (p *chPlacer) PlaceCall(rp RunnerPool, ctx context.Context, call RunnerCall) error {
 
+	startTime := time.Now()
 	log := common.Logger(ctx)
 
 	// The key is just the path in this case
 	key := call.Model().Path
 	sum64 := siphash.Hash(0, 0x4c617279426f6174, []byte(key))
 
+OutTries:
 	for {
 		runners, err := rp.Runners(call)
 		if err != nil {
+			stats.Record(ctx, errorPoolCountMeasure.M(0))
 			log.WithError(err).Error("Failed to find runners for call")
 		} else {
-			i := int(jumpConsistentHash(sum64, int32(len(runners))))
-			for j := 0; j < len(runners); j++ {
+			if len(runners) == 0 {
+				stats.Record(ctx, emptyPoolCountMeasure.M(0))
+			}
 
-				select {
-				case <-ctx.Done():
-					return models.ErrCallTimeoutServerBusy
-				default:
+			i := int(jumpConsistentHash(sum64, int32(len(runners))))
+
+			for j := 0; j < len(runners); j++ {
+				if ctx.Err() != nil {
+					break OutTries
 				}
 
 				r := runners[i]
 
+				attemptTime := time.Now()
 				tryCtx, tryCancel := context.WithCancel(ctx)
 				placed, err := r.TryExec(tryCtx, call)
 				tryCancel()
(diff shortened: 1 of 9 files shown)


api/runnerpool/placer_stats.go
```diff
@@ -0,0 +1,59 @@
...
+	err := view.Register(
+		createView(fullScanCountMeasure, view.Count(), tagKeys),
+		createView(errorPoolCountMeasure, view.Count(), tagKeys),
+		createView(emptyPoolCountMeasure, view.Count(), tagKeys),
+		createView(cancelCountMeasure, view.Count(), tagKeys),
+		createView(placedErrorCountMeasure, view.Count(), tagKeys),
+		createView(placedOKCountMeasure, view.Count(), tagKeys),
+		createView(retryTooBusyCountMeasure, view.Count(), tagKeys),
+		createView(retryErrorCountMeasure, view.Count(), tagKeys),
+		createView(placerLatencyMeasure, view.Distribution(1, 10, 25, 50, 200, 1000, 10000, 60000), tagKeys),
```
kppk: I did something similar in api/server/stats.go

The idea was to register views only if the server is started with `WithPrometheus` option.
All the views are public, so one can register selected views if `WithPrometheus` is *not* set.

api/runnerpool/placer_stats.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"github.com/sirupsen/logrus"
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+)
+
+var (
+	fullScanCountMeasure     = stats.Int64("lb_placer_fullscan_count", "LB Placer Full RunnerList Scan Count", "")
```
kppk: This looks like the placer count stats can use tags? the tags key could be something like 'status'?

skinowski: So, on OSS side, typically the tags are fn_appname, fn_path, but for multi-tenant environments, I think there will be no tags. Leaving this to the caller of RegisterPlacerViews().

api/runnerpool/placer_stats.go
kmjohansen: Showing the limits of my own knowledge with this question.  Looked at the docs for `view.Distribution()` but it wasn't clear if this gives us multiple values per-bucket, or a single value?  Can the upstream metrics collector re-sort this to generate better percentile values if we need?

api/runnerpool/placer_stats.go
kppk: I don't think that fn_appname should be tag.

see https://prometheus.io/docs/practices/naming/
`CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.`

api/runnerpool/ch_placer.go
```diff
@@ -30,28 +31,34 @@ func NewCHPlacer() Placer {
...
 
+OutTries:
 	for {
 		runners, err := rp.Runners(call)
 		if err != nil {
+			stats.Record(ctx, errorPoolCountMeasure.M(0))
 			log.WithError(err).Error("Failed to find runners for call")
 		} else {
+			if len(runners) == 0 {
+				stats.Record(ctx, emptyPoolCountMeasure.M(0))
```
kmjohansen: Should this code `break OutTries`?  It would save you from incrementing a bunch of the subsequent metrics, which may not be applicable in the case no runners exist.  E.g. do we want to record full-scan for a zero length runner list?

api/runnerpool/ch_placer.go
```diff
@@ -60,20 +67,47 @@ func (p *chPlacer) PlaceCall(rp RunnerPool, ctx context.Context, call RunnerCall
...
 		// backoff
 		select {
 		case <-ctx.Done():
-			return models.ErrCallTimeoutServerBusy
+			break OutTries
 		case <-time.After(p.rrInterval):
 		}
 	}
+
+	// Cancel Exit Path / Client cancelled/timedout
```
kmjohansen: Should this code be conditional on the length of the runners list being greater than zero?  Also worth noting: this changes the behavior of this function in the case that no runners are found.  You'll now return `ErrCallTimeoutServerBusy` when previously it would have returned `nil`

api/runnerpool/naive_placer.go
```diff
@@ -27,23 +28,28 @@ func NewNaivePlacer() Placer {
...
+
+OutTries:
 	for {
 		runners, err := rp.Runners(call)
 		if err != nil {
+			stats.Record(ctx, errorPoolCountMeasure.M(0))
 			log.WithError(err).Error("Failed to find runners for call")
 		} else {
+			if len(runners) == 0 {
+				stats.Record(ctx, emptyPoolCountMeasure.M(0))
```
kmjohansen: Same question here about breaking out early.

api/runnerpool/naive_placer.go
```diff
@@ -52,16 +58,43 @@ func (sp *naivePlacer) PlaceCall(rp RunnerPool, ctx context.Context, call Runner
...
-			return models.ErrCallTimeoutServerBusy
+			break OutTries
 		case <-time.After(sp.rrInterval):
 		}
 	}
+
+	// Cancel Exit Path / Client cancelled/timedout
+	stats.Record(ctx, cancelCountMeasure.M(0))
+	stats.Record(ctx, placerLatencyMeasure.M(int64(time.Now().Sub(startTime)/time.Millisecond)))
+	return models.ErrCallTimeoutServerBusy
```
kmjohansen: And, same question here about the both the recording of stats and return of `ErrCallTimeoutServerBusy` in the case where no servers are found.

api/runnerpool/placer_stats.go
skinowski: Yes, RegisterPlacerViews() leaves that optional. OSS Fn typically has fn_appname/fn_path, which could lead to high cardinality, but unless we do a sweep of all stats and change them, I think fn_appname/fn_path on OSS side is OK. I think this can be done with separate PR to sweep and change all stats...

skinowski: FYI:
https://github.com/fnproject/fn/blob/master/cmd/fnserver/main.go#L25

api/runnerpool/placer_stats.go
skinowski: Distribution seems to get multiple values in the buckets.

skinowski:
@kppk @kmjohansen Does it look better now?

One important, but unrelated with metrics change here is rp.Runners(call) error return
action. I've changed the semantics here a bit and documented it in the interface. That is,
if an error is returned, we do not need to retry-loop.

api/server/server.go
```diff
@@ -468,6 +468,12 @@ func WithAgentFromEnv() ServerOption {
 				placer = pool.NewNaivePlacer()
 			}
 
+			// If prometheus is enabled, add LB placer metrics to the views
+			if s.promExporter != nil {
```
rdallman: don't think we should be doing this as a general rule of thumb -- which exporters we have attached shouldn't affect which metrics we collect (easy to fix...) -- if we need a way to turn these off we kinda have it half plumbed with the register views methods throughout now, seems like we could do similar to as was done in #1057 to have the default set of views and let the user either import the package that turns on all our views or they can add them individually themselves. anyway, not too big a deal

skinowski: @rdallman thanks for pointing this out, saw your PR and just merged it. :-)

PR #1059, opened 2018-06-12
should turn on views regardless of exporter

rdallman:
https://github.com/fnproject/fn/pull/1058/files#r194913723

maybe i missed some context, but in general think this is the right thing to do


skinowski:
@rdallman Thanks for catching this.

PR #1060, opened 2018-06-13
Add support for Function and Trigger domain objects 

mantree:
First stage of move toward new trigger /Functions API


api/models/trigger.go
```diff
@@ -0,0 +1,175 @@
...
+//go:generate jsonenums -type=TriggerType
+type TriggerType int
+
+const (
+	Unknown TriggerType = iota
+	HTTP
+)
+
+type Trigger struct {
+	ID          string          `json:"id" db:"id"`
```
zootalures: Have we decided to ditch routers.Headers  ? 

Also should "type" be validated (e.g. dynamically based on pluggable trigger types)

zootalures: I guess my thoughts are : 

for HTTP there are a clear set of possible  configuration values: e.g.: 

* CORS configuration 
* HTTP method selectors (in future) (e.g. different Fns for PUT/GET) 
* Current "Jack in my headers" behaviour (which may be a bit broken in many ways) 

Was wondering about the relative mertits of : 

* Hard coding these on the Trigger Object as sub-classes (e.g. HttpTrigger < Trigger) via Swagger polymorphism 
* Setting up some independent Annotation specs for these 

The former is obviously tighter but scaling this up to other classes of event source might be trigger (CronTrigger looks easy, but QueueTrigger looks harder) - the latter feels too loose in general - especially for HTTP having unbound/typed properties for core values seems wrong.

rdallman: > Have we decided to ditch routes.Headers ?

yes. we can add it back if needed but it's not a required feature

> Also should "type" be validated (e.g. dynamically based on pluggable trigger types)

yea i think it should just be a flexible string, then we allow custom triggers out of the box. extensions could possibly validate a limited set (say, a service that only supports x and y), but we probably shouldn't in fn core

api/datastore/sql/sql.go
```diff
@@ -122,8 +122,9 @@ const (
 	ensureAppSelector = `SELECT id FROM apps WHERE name=?`
 	fnSelector        = `SELECT id, name, app_id,  image, format, cpus, memory, timeout, idle_timeout, config, annotations, created_at, updated_at FROM fns`
 
-	triggerIDSelector = `SELECT * FROM triggers WHERE id=?`
-	triggerSelector   = `SELECT * FROM triggers`
+	triggerIDSelector   = `SELECT * FROM triggers WHERE id=?`
+	triggerNameSelector = `SELECT * FROM triggers `
+	triggerSelector     = `SELECT * FROM triggers`
```
mantree: duplication here, should reduce to one

api/models/trigger.go
mantree: I don't like the idea of a flexible string for the trigger type. (Http, HTTP, http, etc, etc, needless pain).

I'm not sure how there could be a 'custom' trigger without it existing in a code base, somehow. In the json transfer the names are used ('HTTP' for example), it's only internally the int value is used. So there's not much danger of someone porting something over and it changing type.

I think annotation specs should come in as soon as we've nailed down what they need to be for HTTP. They should be open for extension though.

docs/swagger_v2.yml
```diff
@@ -0,0 +1,663 @@
...
+      responses:
+        204:
+          description: Trigger successfully deleted.
+        default:
+          description: Error
+          schema:
+            $ref: '#/definitions/Error'
+
+definitions:
+  Route:
```
rdallman: can nuke

docs/swagger_v2.yml
```diff
@@ -0,0 +1,663 @@
...
+        type: string
+        description: Unique trigger identifier
+        readOnly: true
+      name:
+        type: string
+        description: unique name for this trigger, used to identify this trigger
+      type:
+        type: string
+        description: class of trigger, e.g. schedule, http
+        enum: [http, stream]
```
rdallman: string here? an extension could limit the domain here but i'm not sure in core that we should do it.

docs/swagger_v2.yml
```diff
@@ -216,6 +216,10 @@ paths:
           description: Function definition
           schema:
             $ref: '#/definitions/Fn'
+        404:
```
mantree: You don't really need these, the default handles all of this. I removed them earlier.

api/models/trigger.go
```diff
@@ -71,6 +71,9 @@ var (
 	ErrTriggerMissingFnID = err{
 		code:  http.StatusBadRequest,
 		error: errors.New("Missing Fn ID on Trigger")}
+	ErrTriggerFnIDNotSameApp = err{
```
mantree: Who need foreign key constraints!

zootalures:
+1 from me

PR #1062, opened 2018-06-14
Add functions as a "first class" object in the data model

riconnon:
This is the first PR in a series of work to restructure the data model.

This simply introduces the objects and APIs for functions without making them do anything.


api/datastore/mock.go
```diff
@@ -194,6 +197,83 @@ func (m *mock) batchDeleteRoutes(ctx context.Context, appID string) error {
 	return nil
 }
 
+func (m *mock) PutFn(ctx context.Context, fn *models.Fn) (*models.Fn, error) {
+	if fn.AppID == "" {
+		return nil, models.ErrFnsMissingAppID
+	}
```
rdallman: this should be in the `validator` Datastore implementation

api/server/server.go
```diff
@@ -899,13 +899,25 @@ func (s *Server) bindHandlers(ctx context.Context) {
...
 					withAppCheck.DELETE("/routes/*route", s.handleRouteDelete)
+
 					withAppCheck.GET("/calls/:call", s.handleCallGet)
 					withAppCheck.GET("/calls/:call/log", s.handleCallLogGet)
 					withAppCheck.GET("/calls", s.handleCallList)
+
+					withAppCheck.GET("/triggers", s.handleFnsList)
+					withAppCheck.GET("/triggers/:trigger", s.handleFnsGet)
+					withAppCheck.PUT("/triggers/:trigger", s.handleFnsPut)
+					withAppCheck.DELETE("/triggers/:trigger", s.handleFnsDelete)
```
rdallman: may leave these out for #1060 ?

api/models/fn.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+// Fn contains information about a function configuration.
+type Fn struct {
+	// ID is the generated resource id.
+	ID string `json:"id" db:"id"`
+	// Name is a user provided name for this fn.
+	Name string `json:"name" db:"name"`
+	// AppID is the ID of the app this fn belongs to.
+	AppID string `json:"app_id" db:"app_id"`
```
rdallman: this field is inconsistent with the swagger which indicates a field for `app_name` and not `app_id`. it would be nice if this were the entire app object but the tradeoff is that requires a join. in any case, since we don't have 'get app by id' endpoint (we want everything name-addressable), this should probably be `app_name` so that users can look it up, possibly with an additional field like `AppID string json:"-" db:"app_id"` to hide the app id from the api, but to store it for use so we can look up the real name underneath.

api/models/datastore.go
```diff
@@ -59,6 +59,24 @@ type Datastore interface {
...
+	GetFns(ctx context.Context, filter *FnFilter) ([]*Fn, error)
+
+	// GetFn returns a function by name. Returns ErrDatastoreEmptyFnName if funcName is empty.
+	// Returns ErrFnsNotFound if a func is not found.
+	// TODO(reed): figure out addressable by id or name biz. iff 1 query, name works.
+	GetFn(ctx context.Context, funcName string) (*Fn, error)
+
+	// RemoveFn removes a function. Returns ErrDatastoreEmptyFnName if funcName is empty.
+	// Returns ErrFnsNotFound if a func is not found.
+	RemoveFn(ctx context.Context, funcName string) error
```
rdallman: these all (except PutFn which has it in the model) need an arg for `appID` I think (like Route) -- maybe we could get away with by name but we never quite sorted that out so I'm not sure, appID is the safe route.

mantree: I think at this level it should use FnID, the mapping over from the name, appId pair should happen higher up.

api/datastore/sql/sql.go
```diff
@@ -82,13 +82,31 @@ var tables = [...]string{`CREATE TABLE IF NOT EXISTS routes (
 	app_id varchar(256) NOT NULL,
 	log text NOT NULL
 );`,
+
+	`CREATE TABLE IF NOT EXISTS fns (
```
mantree: How come we've got both a migration and this? Seems like a duplication.

api/datastore/sql/sql.go
```diff
@@ -667,6 +685,168 @@ func (ds *SQLStore) GetRoutesByApp(ctx context.Context, appID string, filter *mo
 	return res, nil
 }
 
+func (ds *SQLStore) PutFn(ctx context.Context, fn *models.Fn) (*models.Fn, error) {
+	err := ds.Tx(func(tx *sqlx.Tx) error {
+		query := tx.Rebind(fmt.Sprintf(`%s WHERE name=?`, fnSelector))
```
mantree: This needs to be by either (appId, FnName) or just use FnID here. I'm in favour of just using FnID

api/datastore/sql/sql.go
```diff
@@ -893,6 +1028,46 @@ func buildFilterCallQuery(filter *models.CallFilter) (string, []interface{}) {
...
+	if filter == nil {
+		return "", nil
+	}
+	var b bytes.Buffer
+	var args []interface{}
+
+	// where(fmt.Sprintf("image LIKE '%s%%'"), filter.Image) // TODO needs escaping, prob we want prefix query to ignore tags
+	args = where(&b, args, "id>?", filter.Cursor)
+
+	fmt.Fprintf(&b, ` ORDER BY id ASC`)
```
mantree: Should this not be 'name'? I think swagger states that as the sorting

api/datastore/internal/datastoretest/test.go
```diff
@@ -490,6 +490,211 @@ func Test(t *testing.T, dsf func(t *testing.T) models.Datastore) {
...
+		if err != nil {
+			t.Fatalf("Test GetFns: error: %s", err)
+		}
+		if len(fns) != 1 {
+			t.Fatalf("Test GetFns: expected result count to be 1 but got %d", len(fns))
+		} else if fns[0].Name != testFn.Name {
+			t.Fatalf("Test GetFns: expected `func.Name` to be `%s` but it was `%s`", testFn.Name, fns[0].Name)
+		}
+
+		fns, err = ds.GetFns(ctx, &models.FnFilter{PerPage: 2, Cursor: fns[0].ID})
```
mantree: Can't see any name like queries or image match tests....

api/models/fn.go
```diff
@@ -0,0 +1,235 @@
...
+
+	if !f.Equals(original) {
+		f.UpdatedAt = strfmt.DateTime(time.Now())
+	}
+}
+
+type FnFilter struct {
+	AppID string // this is exact match
+	Image string // this is exact match
+
```
mantree: Don't we need a few more things here? Name for example?

api/models/datastore.go
```diff
@@ -59,6 +59,24 @@ type Datastore interface {
 	// ErrDatastoreEmptyRoutePath when routePath is empty. Returns ErrRoutesNotFound when no route exists.
 	RemoveRoute(ctx context.Context, appID, routePath string) error
 
+	// PutFn inserts a new function if one does not exist, applying any defaults necessary, or
+	// updates a function that exists under the same name. Returns ErrDatastoreEmptyFn if func is nil,
```
mantree: Not a big fan of this idea. Can see it causing users a lot of trouble with people over writing others changes.

api/datastore/sql/sql.go
zootalures: sad fact of the way that migration works now - you need both and they have to describe the same target state.

api/models/fn.go
zootalures: lets add later as needed - I'd go so far as to stay `Image` is a feature unbacked by need.

api/server/fns_delete.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/gin-gonic/gin"
+)
+
+func (s *Server) handleFnsDelete(c *gin.Context) {
+	ctx := c.Request.Context()
+
+	fn := c.Param(api.Fn)
+	appName := c.Param(api.CApp)
+
+	appID, err := s.datastore.GetAppID(ctx, appName)
```
rdallman: this should already be set as a param by https://github.com/fnproject/fn/blob/master/api/server/gin_middlewares.go#L172 if we want to use `MustGet` here, it saves an additional lookup. route is this way - am not sure if we are diverging from the routes way, which could be ok too - just throwing this out.

rdallman:
this LGTM. i think it has everything hooked up now: extensions, tests, front end, db, etc. at worst, if we missed something, we can fix it but other work builds on this so would be nice to land. thanks everyone.

* need to figure out if we want to work off master or feature branch. this means we have to potentially add migrations over minor bugs for something that isn't hooked up - but that's not bad.
* kind of odd that everything is by app id in the backend but func which has ids in a similar manner is all name in the backend. i suspect the middleware delete-recreate issue exists. i'm ok with putting a stake in the mud to move forward, we have gone back and forth on this for days now (i'm not sure my merged analogy works)

api/datastore/sql/sql.go
```diff
@@ -765,8 +793,8 @@ func (ds *SQLStore) PutFn(ctx context.Context, fn *models.Fn) (*models.Fn, error
 				config = :config,
 				annotations = :annotations,
 				updated_at = :updated_at
-			WHERE name=:name AND app_id=:app_id;`)
-		}
+			    WHERE id=:id AND app_id=:app_id;`)
```
mantree: don't think this is right. Especially without the ID being validated as present in the validator.

api/models/fn.go
```diff
@@ -28,8 +28,6 @@ type Fn struct {
 	ID string `json:"id" db:"id"`
 	// Name is a user provided name for this fn.
 	Name string `json:"name" db:"name"`
-	// AppName is the name of the app this fn belongs to.
-	AppName string `json:"app_name" db:"app_name"`
```
mantree: So no app_name in the json, despite it being what powers the whole api?

api/server/fn_listeners.go
```diff
@@ -0,0 +1,76 @@
+package server
+
+import (
+	"context"
+
+	"github.com/fnproject/fn/api/models"
+	"github.com/fnproject/fn/fnext"
+)
+
+type fnListeners []fnext.FnListener
```
mantree: What is all this about?

fnext/listeners.go
```diff
@@ -51,9 +51,25 @@ type RouteListener interface {
 	// AfterRouteUpdate called after route updated in datastore
 	AfterRouteUpdate(ctx context.Context, route *models.Route) error
 	// BeforeRouteDelete called before route deleted from the datastore
-	BeforeRouteDelete(ctx context.Context, appName string, routePath string) error
+	BeforeRouteDelete(ctx context.Context, appId string, routePath string) error
```
mantree: Why are Routes being changed? These should be being left as is.

mantree:
The triggers branch has been rebased on to this, but we are going to do more work on the Fns stuff there. So, we will be landing both together from the triggers branch.

Therefore, this can be closed without merging.

Issue #1067, opened 2018-06-19
PR and issue templates


rdallman:
to aid contributors and users in creating issues that will help us diagnose issues faster with less back and forth and to help users create great pull requests. it's not always easy given a blank slate, and there's typical information we will always want in certain contexts, so we might as well ask for it up front.

msgodf:
I know you mention it in the template, but it might be worth making it clear that this is for _bug_ issues.

PR #1068, opened 2018-06-19
create issue_template.md

rdallman:
part 1 of #1067 - this will help users create issues with all (most of?) the context we need without having to ask. it also includes a policy to close bugs that are left stale after 7 days, which is convenient to keep issues from growing like a weed.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,54 @@
+<!--
+If you are reporting a new issue, make sure that we do not have any duplicates
+already open. You can ensure this by searching the issue list for this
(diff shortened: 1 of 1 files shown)


rdallman:
I haven't tested this or done it before, but the instructions are straight out of https://help.github.com/articles/manually-creating-a-single-issue-template-for-your-repository/ and this mostly follows Docker's template (thanks whale foundation), whose files are located in this location... their template made me want to add this to begin with, I think that it's really useful (specifically, asking for _expected_ results in addition to what happened, and getting a version)

PR #1072, opened 2018-06-19
Mask DB password in logs

### LINKED ISSUE #1071: Sensitive information should be logged with DEBUG level?
When starting the server, the connection credential is written with "INFO" level. Should the level be "DEBUG"?

```
time="2018-06-19T04:14:04Z" level=info msg="Connecting to DB" url="root:root@tcp(mysql:3306)/fn"
```

zootalures:
fixes #1071

### DIFF
--- a/api/datastore/sql/sql.go
+++ b/api/datastore/sql/sql.go
@@ -143,7 +143,7 @@ func (sqlLogsProvider) New(ctx context.Context, u *url.URL) (models.LogStore, er
 func newDS(ctx context.Context, url *url.URL) (*SQLStore, error) {
 	driver := url.Scheme
 
-	log := common.Logger(ctx)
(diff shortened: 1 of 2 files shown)


api/datastore/sql/sql.go
```diff
@@ -143,7 +143,7 @@ func (sqlLogsProvider) New(ctx context.Context, u *url.URL) (models.LogStore, er
 func newDS(ctx context.Context, url *url.URL) (*SQLStore, error) {
 	driver := url.Scheme
 
-	log := common.Logger(ctx)
+	log := common.Logger(ctx).WithFields(logrus.Fields{"url": common.MaskPassword(url)})
```
msgodf: Is there a chance that `url` could be be `nil`?

zootalures: @msgodf  bump

PR #1081, opened 2018-06-21
Bye bye openapi

rdallman:
What I did:

### multiple reasons to remove the api tests:

* awkward dependency with fn_go meant generating bindings on a branched fn to
vendor those to test new stuff. this is at a minimum not at all intuitive,
worth it, nor a fun way to spend the finite amount of time we have to live.
* api tests only tested a subset of functionality that the server/ api tests
already test, and we risk having tests where one tests some thing and the
other doesn't. let's not. we have too many test suites as it is, and these
pretty much only test that we updated the fn_go bindings, which is actually a
hassle as noted above and the cli will pretty quickly figure out anyway.
* fn_go relies on openapi, which relies on mgo, which is deprecated and we'd
like to remove as a dependency. openapi is a _huge_ dep built in a NIH
fashion, that cannot simply remove the mgo dep as users may be using it.
we've now stolen their date time and otherwise killed usage of it in fn core,
for fn_go it still exists but that's less of a problem.

### created common.DateTime

same thing as strfmt.DateTime but sans mgo, and sans json. json uses TextMarshaler, which the strfmt.DateTime implements as well and is more efficient than their json marshaler. strfmt is a good case of NIH gone wrong, keeping this as ammo. this gets rid of our openapi dep.

### update the deps

removals:

* easyjson
* mgo
* go-openapi
* mapstructure
* fn_go
* purell
* go-validator

also, had to lock docker. we shouldn't use docker on master anyway, they
strongly advise against that. had no luck with latest version rev, so i locked
it to what we were using before. until next time.

the rest is just playing dep roulette, those end up removing a ton tho


rdallman:
futzing with system tests... half a mind to remove those too, i can't seem to get them working one way locally or another way in CI... not fun

skinowski:
LGTM, but are we sure removed api tests are already covered in the remaining tests in fn?

rdallman:
i cross checked, in varying capacities everything is covered i believe. summary: 

* much of it is app/route POST/PUT/GET testing alone and 200/404 cases - better coverage in server/
* call/log api testing (runner_test.go has a ton of coverage)
* hot json testing (but no other formats - well covered also, and every format)
* annotations have similar behavioral tests in annotations_test.go, simply not through api. app api has some coverage for setting them.

rdallman:
thanks

zootalures:
Thanks, looks good - We still have a gap in verifying that the server strongly complies with both its swagger spec and generaly behaviour but I think this wasn't the right hammer for it .

PR #1083, opened 2018-06-21
fn: applications should be aware of reserved writable space

skinowski:
Similar to FN_MEMORY, we pass FN_TMPSIZE to function config.

### DIFF
--- a/api/agent/call.go
+++ b/api/agent/call.go
@@ -154,6 +154,7 @@ func buildConfig(app *models.App, route *models.Route) models.Config {
 	// TODO: might be a good idea to pass in: "FN_BASE_PATH" = fmt.Sprintf("/r/%s", appName) || "/" if using DNS entries per app
 	conf["FN_MEMORY"] = fmt.Sprintf("%d", route.Memory)
 	conf["FN_TYPE"] = route.Type
+	conf["FN_TMPSIZE"] = fmt.Sprintf("%d", route.TmpFsSize)
 
 	CPUs := route.CPUs.String()
 	if CPUs != "" {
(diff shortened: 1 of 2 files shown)


skinowski:
@kmjohansen For time being, it's not configurable. Hardcoded to /tmp.

skinowski:
I understand that this approach here is limited, but this is consistent with FN_MEMORY, FN_CPU. These are passed in as part of call configuration structure, for example, for json:

https://github.com/fnproject/fn/blob/master/api/agent/protocol/json.go#L27

TMPFSSIZE is part of Route API and it's a single entity. When volumes are implemented, we'd need to extend/supplement this. Right now, TMPFSSIZE is either unlimited (0), or set to a specific limit.

I'm ok with holding off on this until we decide on a better strategy as well.

kmjohansen:
My major concerns here are two-fold.  The first is the vast proliferation of environment variables.  It seems like it might be easier to encode configuration information into a runtime neutral format that any FDK can parse and return to the application.  The second is the set of issues around providing static definitions for resources that are dynamically available.  If, for example, we provided an interface that allowed the caller to get both the max and available transient storage in a single call, that would satisfy me.  The concern here is that any caller that depends upon the environment variable when the availability of the resource is dynamic is broken by definition.

skinowski:
@kmjohansen These are actually properly/formally formatted in json/cloud-events protocol as function-config. (They only have to be env variables in cold-functions, which has no structured protocol. Because of this, I keep calling these env variables and I've accidentally caused some confusion.) FDKs actually parse these and offer them as config data.

I also agree with dynamic nature of the resource. But these are to be treated as "function configuration" not as a way to dynamically determine how much memory or /tmp space a function has at a given time. In other words, these tell the function which "shape" it is currently running on. We expose FN_CPU, FN_MEMORY along with configuration set in that route, which is also merged with the application configuration. In other words, we communicate almost all configuration params in a route to the function:

https://github.com/fnproject/fn/blob/master/api/agent/call.go#L142

skinowski:
Based on ongoing discussion:

https://github.com/fnproject/fn/issues/1106

we might eventually hide these: cpu/tmpsize from users. But in case we don't, I'd like to keep the triplet: cpu/mem/tmpfs consistent & merged in & exposed.

PR #1085, opened 2018-06-24
Add a finer-grained view for placer latency metrics

jan-g:
This is a small tweak to the placer latency stats. If we have a cluster of values
around the 1-2s mark, then having a single relatively broad bucket that captures
the (1s, 10s] range will obscure that. In particular, typical Prometheus quartile
estimates may be distorted by this bucket size.

We should probably do something better here - either let the end-user inject their distributions, or perhaps generate a log-linear spread across likely values with far more intervals.


skinowski:
LGTM, but minor concern regarding adding too many buckets.

jan-g:
Agreed; advice from statisticians wrt bucket counts versus from the prom guys is in conflict. All we can tell from quantiles is "the nth percentile lies in this bucket", though. We probably need a better scheme to let us tune these across the board.

PR #1086, opened 2018-06-24
HTTP Triggers hookup

zootalures:
This doesn't add invoke yet - and that will necessarily pay down some of the refactoring debt created here but I wanted to get black-box testing and functionality in place for HTTP triggers before we started the larger invoke refactor. 

This also adds "unique by appId, source, type" to triggers and tests. 

A lot of the changes are an initial refactor of the data store interfaces  with a view to removing: 
* NoPStore from the pure runner 
* Store calls from the Agent (Agent as a proxy for the DB) 
* Storage operations from agent.call 

None of these are completely payed down yet so it's a bit entangled at the moment - Many of these things can wash out of invoke.

### DIFF
--- a/api/agent/data_access.go
+++ b/api/agent/data_access.go
@@ -237,3 +237,13 @@ func (da *directDataAccess) Close() error {
 	return da.mq.Close()
 
 }
+
+type noAsyncEnqueueAccess struct{}
+
+func (noAsyncEnqueueAccess) Enqueue(ctx context.Context, mCall *models.Call) error {
+	return models.ErrAsyncUnsupported
+}
+
+func NewUnsupportedAsyncEnqueueAccess() EnqueueDataAccess {
+	return &noAsyncEnqueueAccess{}
+}
--- a/api/agent/data_access.go
+++ b/api/agent/data_access.go
@@ -11,6 +11,7 @@ import (
 	"github.com/patrickmn/go-cache"
 )
 
+//ReadDataAccess represents read operations required to operate a load balancer node
 type ReadDataAccess interface {
 	GetAppID(ctx context.Context, appName string) (string, error)
 	// GetAppByID abstracts querying the datastore for an app.
@@ -21,20 +22,23 @@ type ReadDataAccess interface {
 	GetRoute(ctx context.Context, appID string, routePath string) (*models.Route, error)
 }
 
+//DequeueDataAccess abstracts an underlying dequeue for async runners
 type DequeueDataAccess interface {
 	// Dequeue will query the queue for the next available Call that can be run
 	// by this Agent, and reserve it (ultimately forwards to mq.Reserve).
 	Dequeue(ctx context.Context) (*models.Call, error)
 }
 
+//EnqueueDataAccess abstracts an underying enqueue for async queueing
 type EnqueueDataAccess interface {
 	// Enqueue will add a Call to the queue (ultimately forwards to mq.Push).
 	Enqueue(ctx context.Context, mCall *models.Call) error
 }
 
 // CallHandler consumes the start and finish events for a call
+// This is effectively a callback that is allowed to read the logs -
+// TODO Deprecate this - this could be a CallListener except it also consumes logs
 type CallHandler interface {
-	io.Closer
 	// Start will attempt to start the provided Call within an appropriate
 	// context.
 	Start(ctx context.Context, mCall *models.Call) error
@@ -44,23 +48,15 @@ type CallHandler interface {
 	Finish(ctx context.Context, mCall *models.Call, stderr io.Reader, async bool) error
 }
 
-// DataAccess abstracts the datastore and message queue operations done by the
-// agent, so that API nodes and runner nodes can work with the same interface
-// but actually operate on the data in different ways (by direct access or by
-// mediation through an API node).
+// DataAccess is currently
 type DataAccess interface {
 	ReadDataAccess
 	DequeueDataAccess
 	CallHandler
-
-	// Close will wait for any pending operations to complete and
-	// shuts down connections to the underlying datastore/queue resources.
-	// Close is not safe to be called from multiple threads.
-	//io.Closer
 }
 
 // CachedDataAccess wraps a DataAccess and caches the results of GetApp and GetRoute.
-type CachedDataAccess struct {
+type cachedDataAccess struct {
 	ReadDataAccess
 
 	cache        *cache.Cache
@@ -68,7 +64,7 @@ type CachedDataAccess struct {
 }
 
 func NewCachedDataAccess(da ReadDataAccess) ReadDataAccess {
-	cda := &CachedDataAccess{
+	cda := &cachedDataAccess{
 		ReadDataAccess: da,
 		cache:          cache.New(5*time.Second, 1*time.Minute),
 	}
@@ -83,11 +79,11 @@ func appIDCacheKey(appID string) string {
 	return "a:" + appID
 }
 
-func (da *CachedDataAccess) GetAppID(ctx context.Context, appName string) (string, error) {
+func (da *cachedDataAccess) GetAppID(ctx context.Context, appName string) (string, error) {
 	return da.ReadDataAccess.GetAppID(ctx, appName)
 }
 
-func (da *CachedDataAccess) GetAppByID(ctx context.Context, appID string) (*models.App, error) {
+func (da *cachedDataAccess) GetAppByID(ctx context.Context, appID string) (*models.App, error) {
 	key := appIDCacheKey(appID)
 	app, ok := da.cache.Get(key)
 	if ok {
@@ -107,7 +103,7 @@ func (da *CachedDataAccess) GetAppByID(ctx context.Context, appID string) (*mode
 	return app.(*models.App), nil
 }
 
-func (da *CachedDataAccess) GetRoute(ctx context.Context, appID string, routePath string) (*models.Route, error) {
+func (da *cachedDataAccess) GetRoute(ctx context.Context, appID string, routePath string) (*models.Route, error) {
 	key := routeCacheKey(appID, routePath)
 	r, ok := da.cache.Get(key)
 	if ok {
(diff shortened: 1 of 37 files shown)


api/datastore/sql/sql.go
```diff
@@ -445,15 +511,17 @@ func (ds *SQLStore) GetAppByID(ctx context.Context, appID string) (*models.App,
 
 // GetApps retrieves an array of apps according to a specific filter.
 func (ds *SQLStore) GetApps(ctx context.Context, filter *models.AppFilter) ([]*models.App, error) {
-	res := []*models.App{}
+	var res []*models.App
```
rdallman: for json we were at least in v1 returning an empty list. a nil list will be nil in json. we should probably have a package level var to skip the alloc, but that's the rationale if we're keeping that behavior

api/datastore/sql/sql.go
```diff
@@ -658,6 +730,148 @@ func (ds *SQLStore) GetRoutesByApp(ctx context.Context, appID string, filter *mo
...
+	})
+
+	if err != nil {
+		return nil, err
+	}
+	return fn, nil
+}
+
+func (ds *SQLStore) GetFns(ctx context.Context, filter *models.FnFilter) ([]*models.Fn, error) {
+	var res []*models.Fn // for json empty list
```
rdallman: `res := []*models.Fn{}` for json empty list ;)

api/datastore/sql/sql.go
```diff
@@ -667,6 +881,48 @@ func (ds *SQLStore) GetRoutesByApp(ctx context.Context, appID string, filter *mo
 	return res, nil
 }
 
+func (ds *SQLStore) GetFnByID(ctx context.Context, fnID string) (*models.Fn, error) {
```
rdallman: open ended question: it may be wise to require an app id as well as an fn id here to avoid security bugs where a user may be able to access another user's function which they shouldn't have access to by supplying the correct id. this would become an issue from some auth middleware, presumably, but would manifest as an issue at this level (because at auth time you wouldn't know the function doesn't exist on a certain other app until you look it up). 

for example:

* Bob has an app `poodles`, with function id `4242`. 
* Julie has an app `corgis` with a few functions on it. 
* Bob is Julie's sworn nemesis (ok, maybe not relevant). 
* Julie calls the api at `/v2/apps/corgis/fns/4242`, is authenticated against the `corgis` app correctly, and grabs Bob's function `4242` out of the db. Julie then sheathes Bob's beautiful poodle Fifi in the thick of Nordic winter (and maybe adopts it so that this is more of a rescue than a tale of malevolence?).

in any event, I have seen this before! maybe, we'd like to guard against it differently, but it's an issue at some level no less and the cost of adding app id to each of these endpoints is relatively minimal, as its already given in the API and we'd just be passing it down -- maybe, there are better ways though.

api/models/datastore.go
```diff
@@ -59,8 +56,44 @@ type Datastore interface {
 	// ErrDatastoreEmptyRoutePath when routePath is empty. Returns ErrRoutesNotFound when no route exists.
 	RemoveRoute(ctx context.Context, appID, routePath string) error
 
-	// GetDatabase returns the underlying sqlx database implementation
-	GetDatabase() *sqlx.DB
```
rdallman: the extension stuff (and in particular https://github.com/fnproject/ext-auth) was using this. i'm no fan either, but it may be better to wage that battle another time (up to you)

api/server/server.go
```diff
@@ -912,27 +937,59 @@ func (s *Server) bindHandlers(ctx context.Context) {
...
 			}
 
+			cleanv2 := engine.Group("/v2")
+			v2 := cleanv2.Group("")
+			v2.Use(s.apiMiddlewareWrapper())
+
+			{
+				v2.GET("/apps", s.handleAppList)
+				v2.POST("/apps", s.handleAppCreate)
+				v2.GET("/apps/:appId", s.handleAppGet)
```
rdallman: in the process of linting... would appreciate doing `appID` / `fnID` / etc. while you're in here (lint will yell), though I can do a sweep afterwards for a low low price for you my friend *merchant voice*

api/server/runner.go
```diff
@@ -57,7 +62,133 @@ var (
...
+	if err != nil {
+		handleV1ErrorResponse(c, err)
+	}
+}
+
+// handleTriggerHttpFunctionCall2 executes the function and returns an error
+// Requires the following in the context:
+// * "app"
+// * "source"
+func (s *Server) handleTriggerHttpFunctionCall2(c *gin.Context) error {
```
rdallman: perhaps I missed it or it's still WIP (apologies), I don't see another handler that will allow invoking a function for any trigger type w/ cloud event data, only a handler for http triggers (which will mostly be serviced by an lb and can use the former, but we probably should have for convenience as well for 'getting started'). just didn't see anything and pointing this out in case, sure that something is in the works i'm just not sure of what :)

rdallman:
know it's WIP but left a few hopefully helpful comments, sorry if it's just something you hadn't gotten to yet and i'm just nagging, forgive me i was curious

api/server/runner.go
zootalures: yep, still WIP , haven't looked at cloud event endpoint - imagine that that would be a different URL

api/datastore/sql/sql.go
zootalures: ah good point !
I just auto-refactoerd there. 

We should check in DS tests here.

zootalures:
Have rebased down, some of this stuff went in on the triggers merge - so might need fixing frlz on master 
(sorry this was distantly branched of the original `triggers` branch - i rebased and sqashed back which probably explains why @rdallman is the author) - will fix .

zootalures:
Just doing another PR to fix the original feedback here,

api/datastore/sql/sql.go
zootalures: My view is that access control is orthogonal to the API structure -  makes more sense (in my per-haps 90s oriented API design mentality) to solve authz on a resource-level properly rather than relying on API to help you here . 

Of course - getting the access control right (when you need access to the resource to assert on policy) is a harder problem that I leave as an excercise to the implementor of that feature. 

All of this is consolidating at the Data store level right now which is good as it's sound w.r.t most ACL decisions  but bad as it adds more complexity there. 

I /think/ you can plausibly implement a reasonably power Authz policy based on the data store abstraction by making decision after the subordinate call is complete but before the interceptor returns - this is what I've seen in the past (i.e. you can't make an ACL on a object till you've loaded it)

api/models/datastore.go
zootalures: Hmm good point - I think I chose to keep this broken in the very short term. 

One option might be to return the root data source from the stack via this interface and then let the extensions up-cast to something that gives them the DB

api/server/server.go
zootalures: yep

api/server/runner_httptrigger.go
```diff
@@ -0,0 +1,145 @@
...
+		// we could filter that error out here too as right now it yells a little
+		if err == models.ErrCallTimeoutServerBusy || err == models.ErrCallTimeout {
+			// TODO maneuver
+			// add this, since it means that start may not have been called [and it's relevant]
+			c.Writer.Header().Add("XXX-FXLB-WAIT", time.Now().Sub(time.Time(model.CreatedAt)).String())
+		}
+		return err
+	}
+
+	// TODO TRIGGERWIP  Not sure I want this evil magic in my life here - shoud remove
```
rdallman: i think we need to just do this content type detection as a middleware for all endpoints on the incoming request and outgoing response if we're going to do it at all. the CRUD api needs it too, it just assumes json and doesn't error on content type or anything. anyway, feel free to remove, but i think some of the examples and maybe tests rely on this behavior for better or for worse so it may not be worth it (we can hide it). it's better than defaulting to json :)

api/models/call.go
```diff
@@ -149,6 +149,12 @@ type Call struct {
 
 	// App this call belongs to.
 	AppID string `json:"app_id" db:"app_id"`
+
+	// App this call belongs to.
```
mantree: Fix comment

api/models/call.go
```diff
@@ -149,6 +149,12 @@ type Call struct {
...
+	TriggerID string `json:"trigger_id" db:"trigger_id"`
+
+	// App this call belongs to.
```
mantree: again

api/models/trigger.go
```diff
@@ -71,46 +77,63 @@ func ValidTriggerType(a string) bool {
 }
 
 var (
+	//ErrTriggerIDProvided indicates that a trigger ID was specified when it shouldn't have been
```
mantree: ouch, this looks painful!

api/server/runner_httptrigger_test.go
```diff
@@ -0,0 +1,622 @@
...
+	srv := testServer(ds, &mqs.Mock{}, logDB, rnr, ServerTypeFull)
+
+	for i, test := range []struct {
+		path          string
+		body          string
+		expectedCode  int
+		expectedError error
+	}{
+		{"/t/app/route", "", http.StatusNotFound, models.ErrAppsNotFound},
+		{"/t/myapp/route", "", http.StatusNotFound, models.ErrTriggerNotFound},
```
mantree: No success test?

zootalures: Yep -about 30 of them further down the file

api/models/trigger.go
zootalures: https://github.com/golang/lint/issues/263 
It's the style guide

api/models/call.go
zootalures: Fixed (and actually started populating these, which I'd missed)

api/models/call.go
zootalures: fixed.

api/agent/data_access.go
```diff
@@ -183,22 +200,18 @@ func (da *directDataAccess) Finish(ctx context.Context, mCall *models.Call, stde
 	if async {
 		// XXX (reed): delete MQ message, eventually
 		// YYY (hhexo): yes, once we have the queued/running/finished mechanics
-		// return da.mq.Delete(ctx, mCall)
+		// return cda.mq.Delete(ctx, mCall)
```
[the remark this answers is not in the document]
zootalures: meh-the compiler will do that for free

PR #1095, opened 2018-06-26
Fixed up a couple of incorrect response codes

rikgibson:
Fixes a couple of incorrect response codes (204 instead of 200), and replaces an occurrence of "route" with "function".

### DIFF
--- a/api/server/fns_delete.go
+++ b/api/server/fns_delete.go
@@ -18,5 +18,5 @@ func (s *Server) handleFnDelete(c *gin.Context) {
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


zootalures:
The 204s are what the API should be returning for deletes - we should fix the code to match swagger there.

rikgibson:
https://stackoverflow.com/questions/2342579/http-status-code-for-update-and-delete

I can concur with 204.

rikgibson:
I'll revert and look at fixing the code instead.

PR #1099, opened 2018-06-28
Set shortcodes in trigger response entities based on config or request URL

zootalures:


zootalures:
am not super happy with how this turns out but I think 

A) trigger URLS ought to be absolute
B) the default should be for them to be dynamic in (e.g. local Fn server) rather than persisted. 

had a play with doing this in a DS interceptor but  getting the httpReq context (whilst technically available from ctx.Value(0) seemed... worse

rdallman:
needs rebase^

PR #1102, opened 2018-06-29
List Cursor management moved into datastore layer.

mantree:
Pushes datastore cursor creation and manipulation down into the datastore level, from the api layer. This gives complete freedom to datastore implementations to handle cursors in whatever format they desire.

Also changes the generations of names in the test suite so they aren't created in the same order as the id's, this was hiding the fact that triggers where not in fact being sorted by Name as per spec. This change led to a few edits in older tests.

### DIFF
--- a/api/datastore/mock.go
+++ b/api/datastore/mock.go
@@ -80,13 +80,14 @@ func (m *mock) GetApps(ctx context.Context, filter *models.AppFilter) (*models.A
 	// sort them all first for cursoring (this is for testing, n is small & mock is not concurrent..)
 	sort.Sort(sortA(m.Apps))
 
+	var cursor string
 	if filter.Cursor != "" {
 		s, err := base64.RawURLEncoding.DecodeString(filter.Cursor)
 		if err != nil {
 			return nil, err
 		}
 		logrus.Error(s)
-		filter.Cursor = string(s)
+		cursor = string(s)
 	}
 
 	var apps []*models.App
@@ -94,7 +95,7 @@ func (m *mock) GetApps(ctx context.Context, filter *models.AppFilter) (*models.A
 		if len(apps) == filter.PerPage {
 			break
 		}
-		if strings.Compare(filter.Cursor, a.Name) < 0 {
+		if strings.Compare(cursor, a.Name) < 0 {
 			if filter.Name != "" && filter.Name != a.Name {
 				continue
 			}
@@ -331,12 +332,21 @@ func (m *mock) GetFns(ctx context.Context, filter *models.FnFilter) (*models.FnL
 
 	funcs := []*models.Fn{}
 
+	var cursor string
+	if filter.Cursor != "" {
+		s, err := base64.RawURLEncoding.DecodeString(filter.Cursor)
+		if err != nil {
+			return nil, err
+		}
+		cursor = string(s)
+	}
+
 	for _, f := range m.Fns {
 		if filter.PerPage > 0 && len(funcs) == filter.PerPage {
 			break
 		}
 
-		if strings.Compare(filter.Cursor, f.Name) < 0 &&
+		if strings.Compare(cursor, f.Name) < 0 &&
 			(filter.AppID == "" || filter.AppID == f.AppID) &&
 			(filter.Name == "" || filter.Name == f.Name) {
 			funcs = append(funcs, f)
@@ -346,7 +356,8 @@ func (m *mock) GetFns(ctx context.Context, filter *models.FnFilter) (*models.FnL
 
 	var nextCursor string
 	if len(funcs) > 0 && len(funcs) == filter.PerPage {
-		nextCursor = funcs[len(funcs)-1].Name
+		last := []byte(funcs[len(funcs)-1].Name)
+		nextCursor = base64.RawURLEncoding.EncodeToString(last)
 	}
 
 	return &models.FnList{
@@ -463,12 +474,13 @@ func (s sortT) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
 func (m *mock) GetTriggers(ctx context.Context, filter *models.TriggerFilter) (*models.TriggerList, error) {
 	sort.Sort(sortT(m.Triggers))
 
+	var cursor string
 	if filter.Cursor != "" {
 		s, err := base64.RawURLEncoding.DecodeString(filter.Cursor)
 		if err != nil {
 			return nil, err
 		}
-		filter.Cursor = string(s)
+		cursor = string(s)
 	}
 
 	res := []*models.Trigger{}
@@ -478,7 +490,7 @@ func (m *mock) GetTriggers(ctx context.Context, filter *models.TriggerFilter) (*
 		}
 
 		matched := true
-		if filter.Cursor != "" && t.Name <= filter.Cursor {
+		if filter.Cursor != "" && t.Name <= cursor {
 			matched = false
 		}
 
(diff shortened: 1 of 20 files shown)


api/datastore/mock.go
```diff
@@ -75,33 +76,42 @@ func (s sortA) Len() int           { return len(s) }
...
 	// sort them all first for cursoring (this is for testing, n is small & mock is not concurrent..)
 	sort.Sort(sortA(m.Apps))
 
+	if filter.Cursor != "" {
+		s, err := base64.RawURLEncoding.DecodeString(filter.Cursor)
+		if err != nil {
+			return nil, err
+		}
+		logrus.Error(s)
+		filter.Cursor = string(s)
```
zootalures: You are modifyng a user-supplied parameter here - cursor shoud probably be a var from heron in

api/datastore/mock.go
```diff
@@ -315,7 +325,7 @@ func (s sortF) Len() int           { return len(s) }
 func (s sortF) Less(i, j int) bool { return strings.Compare(s[i].Name, s[j].Name) < 0 }
 func (s sortF) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
 
-func (m *mock) GetFns(ctx context.Context, filter *models.FnFilter) ([]*models.Fn, error) {
+func (m *mock) GetFns(ctx context.Context, filter *models.FnFilter) (*models.FnList, error) {
 	// sort them all first for cursoring (this is for testing, n is small & mock is not concurrent..)
 	sort.Sort(sortF(m.Fns))
```
zootalures: does the fn cursor not get base64'd ?

api/datastore/mock.go
```diff
@@ -438,20 +457,28 @@ func (m *mock) GetTriggerByID(ctx context.Context, triggerId string) (*models.Tr
...
-func (m *mock) GetTriggers(ctx context.Context, filter *models.TriggerFilter) ([]*models.Trigger, error) {
+func (m *mock) GetTriggers(ctx context.Context, filter *models.TriggerFilter) (*models.TriggerList, error) {
 	sort.Sort(sortT(m.Triggers))
 
+	if filter.Cursor != "" {
+		s, err := base64.RawURLEncoding.DecodeString(filter.Cursor)
+		if err != nil {
+			return nil, err
+		}
+		filter.Cursor = string(s)
```
zootalures: don't modify input params here

api/datastore/sql/sql.go
```diff
@@ -508,11 +509,16 @@ func (ds *SQLStore) GetAppByID(ctx context.Context, appID string) (*models.App,
...
 
-	if filter.NameIn != nil && len(filter.NameIn) == 0 { // this basically makes sure it doesn't return ALL apps
-		return res, nil
+	if filter.Cursor != "" {
+		s, err := base64.RawURLEncoding.DecodeString(filter.Cursor)
+		if err != nil {
+			return nil, err
+		}
+		logrus.Error(s)
+		filter.Cursor = string(s)
```
zootalures: as  above - feels like the decoding should be in buildFiterAppQuery..

api/datastore/sql/sql.go
```diff
@@ -508,11 +509,16 @@ func (ds *SQLStore) GetAppByID(ctx context.Context, appID string) (*models.App,
 }
 
 // GetApps retrieves an array of apps according to a specific filter.
-func (ds *SQLStore) GetApps(ctx context.Context, filter *models.AppFilter) ([]*models.App, error) {
-	res := []*models.App{} // for JSON empty list
+func (ds *SQLStore) GetApps(ctx context.Context, filter *models.AppFilter) (*models.AppList, error) {
+	res := &models.AppList{}
```
zootalures: See comment on LHS for (JSON encoding reasons)  the list needs to be initialised as an empty slice or it comes out as `null` in JSON

api/datastore/sql/sql.go
```diff
@@ -844,12 +855,20 @@ func (ds *SQLStore) UpdateFn(ctx context.Context, fn *models.Fn) (*models.Fn, er
 	return fn, nil
 }
 
-func (ds *SQLStore) GetFns(ctx context.Context, filter *models.FnFilter) ([]*models.Fn, error) {
-	res := []*models.Fn{} // for JSON empty list
+func (ds *SQLStore) GetFns(ctx context.Context, filter *models.FnFilter) (*models.FnList, error) {
+	res := &models.FnList{}
```
zootalures: see above

api/datastore/sql/sql.go
```diff
@@ -844,12 +855,20 @@ func (ds *SQLStore) UpdateFn(ctx context.Context, fn *models.Fn) (*models.Fn, er
...
 	if filter == nil {
 		filter = new(models.FnFilter)
 	}
 
+	if filter.Cursor != "" {
+		s, err := base64.RawURLEncoding.DecodeString(filter.Cursor)
+		if err != nil {
+			return nil, err
+		}
+		filter.Cursor = string(s)
```
zootalures: as above

api/datastore/sql/sql.go
```diff
@@ -1323,29 +1345,35 @@ func buildFilterTriggerQuery(filter *models.TriggerFilter) (string, []interface{
...
 		args = append(args, filter.PerPage)
 	}
 
 	return b.String(), args
 }
 
-func (ds *SQLStore) GetTriggers(ctx context.Context, filter *models.TriggerFilter) ([]*models.Trigger, error) {
-	res := []*models.Trigger{} // for JSON  empty list
+func (ds *SQLStore) GetTriggers(ctx context.Context, filter *models.TriggerFilter) (*models.TriggerList, error) {
+	res := &models.TriggerList{}
```
zootalures: as above

api/datastore/sql/sql.go
```diff
@@ -1323,29 +1345,35 @@ func buildFilterTriggerQuery(filter *models.TriggerFilter) (string, []interface{
...
 		filter = new(models.TriggerFilter)
 	}
 
-	filterQuery, args := buildFilterTriggerQuery(filter)
+	if filter.Cursor != "" {
+		s, err := base64.RawURLEncoding.DecodeString(filter.Cursor)
+		if err != nil {
+			return nil, err
+		}
+		filter.Cursor = string(s)
```
zootalures: as above

api/server/apps_list.go
```diff
@@ -12,26 +12,21 @@ func (s *Server) handleAppList(c *gin.Context) {
...
 	if err != nil {
 		handleErrorResponse(c, err)
 		return
 	}
 
-	var nextCursor string
-	if len(apps) > 0 && len(apps) == filter.PerPage {
-		last := []byte(apps[len(apps)-1].Name)
-		nextCursor = base64.RawURLEncoding.EncodeToString(last)
+	if len(apps.Items) > 0 && len(apps.Items) == filter.PerPage {
```
zootalures: This seems wrong - cursors shoudl live wholly in the data store.

api/datastore/mock.go
mantree: looks like it doesn't... it certainly should!

api/datastore/mock.go
mantree: mutability is terrible

api/datastore/sql/sql.go
mantree: it should, but that doesn't have an error return...

api/datastore/sql/sql.go
mantree: should be a test for this

mantree: deffo as it doesn't look like that solution actually worked...

PR #1107, opened 2018-07-02
fn: lb-agent: Add Runner Scheduler/Execution Stats

skinowski:
LB agent reports lb placer latency. It should also report
how long it took for the runner to initiate the call as
well as execution time inside the container if the runner
has accepted (committed) to the call.


skinowski:
@kmjohansen Parse failures that return empty objecjs should fail the time.Time.IsZero() checks and should be ignored.

PR #1110, opened 2018-07-04
Add errors for trigger and function

reclaro:
This change adds two errors for trigger and function not found in the model errors


mantree:
Are these actually used anywhere?

zootalures:
These are already present - check triggers.go (ErrTriggerNotFound) - and fn.go (ErrFnsNotFound)

reclaro:
Thanks @zootalures I missed them, sorry for the noise.

PR #1116, opened 2018-07-11
fn: runner status and docker load images

skinowski:
Introducing a function run for pure runner Status
calls. Previously, Status gRPC calls returned active
inflight request counts with the purpose of a simple
health checker. However this is not sufficient since
it does not show if agent or docker is healthy. With
this change, if pure runner is configured with a status
image, that image is executed through docker. The
call uses zero memory/cpu/tmpsize settings to ensure
resource tracker does not block it.

However, operators might not always have a docker
repository accessible/available for status image. Or
operators might not want the status to go over the
network. To allow such cases, and in general possibly
caching docker images, added a new environment variable
FN_DOCKER_LOAD_FILE. If this is set, fn-agent during
startup will load these images that were previously
saved with 'docker save' into docker.


api/agent/pure_runner.go
```diff
@@ -540,6 +546,15 @@ func (pr *pureRunner) handleTryCall(tc *runner.TryCall, state *callHandle) error
 		return err
 	}
 
+	// Just to be safe. Status image is special and exempt from agent scheduler
+	// resource tracker. We need to make sure normal functions calls do not use it
+	// to get around the resource scheduler.
+	if pr.statusImageName != "" && c.Image == pr.statusImageName {
```
kmjohansen: Does this mean that we're effectively reserving an image by name that the customer will not be able to use?  Is it possible for them to create another image with an identical name and cause a conflict that they won't obviously be able to resolve here?

api/agent/pure_runner.go
```diff
@@ -626,11 +641,112 @@ DataLoop:
...
+				// This is because call.End() is executed asynchronously.
+				result.CompletedAt = common.DateTime(time.Now()).String()
+			}
+		}
+	}
+
+	// Status images should not output excessive data since we echo the
+	// data back to caller.
+	body, _ := ioutil.ReadAll(resp.Body)
+
```
kmjohansen: Worth adding a comment that we're truncating this to the last 64 bytes of the response body?

api/agent/pure_runner.go
```diff
@@ -626,11 +641,112 @@ DataLoop:
...
+	// Try to fetch ctx deadline
+	deadline, ok := ctx.Deadline()
+	if ok {
+		c.Timeout = int32(math.Min(float64(deadline.Sub(start)/time.Second), float64(1)))
+	} else {
+		c.Timeout = 10
+	}
+	c.IdleTimeout = 1
+
+	// TODO: reliably shutdown this container after executing one request.
```
kmjohansen: Is this reliable shutdown in scope for this PR?  If no, do we run any risk of having these status containers linger and behave in unexpected ways?

api/agent/pure_runner.go
skinowski: Yes, this has to be a reserved image name (plus tag) that customers cannot use. But operators should be able  to configure a private image name/tag here to ensure such image name clashes do not occur. (eg. 192.0.2.0:9999/fn-status-checker with a private non-routable IP/port prefix)

api/agent/pure_runner.go
skinowski: Sounds good.

api/agent/pure_runner.go
skinowski: Absence of 'execute-once' (aka maxRequests = 1) will make these tests more efficient since if status container is not idle for more than 1 seconds, it'll continue handling queries. We probably want to test a mixture of both cases, so as a follow up, I'd expect us to use a maxRequests=2 to test both fresh hot container *and* hot container handling subsequent queries.

skinowski: Or maxRequests=1 to keep things simple for time being. Until then, IdleTimeout=1 is best we can do here.

api/agent/drivers/docker/docker_client.go
```diff
@@ -227,6 +229,24 @@ func filterNoSuchContainer(ctx context.Context, err error) error {
 	return err
 }
 
+func (d *dockerWrap) LoadImages(ctx context.Context, filePath string) error {
```
kmjohansen: Based upon our offline discussion, would you be willing to add a comment here that explains the intricacies of save/load and how we need to ensure that the saved image that we load actually bears a tag?

skinowski: OK, added a link to docker-save documentation in options.md. Docker-save documentation shows how to cherry pick tags.

api/agent/pure_runner.go
```diff
@@ -626,11 +640,113 @@ DataLoop:
...
 // implements RunnerProtocolServer
 func (pr *pureRunner) Status(ctx context.Context, _ *empty.Empty) (*runner.RunnerStatus, error) {
-	return &runner.RunnerStatus{
-		Active: atomic.LoadInt32(&pr.inflight),
-	}, nil
+
+	results := &runner.RunnerStatus{}
+
+	if pr.statusImageName != "" {
+		pr.runStatusCall(ctx, results)
```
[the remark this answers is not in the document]
skinowski: For time being, I think we could keep this simple. I've considered returning cached results here and limiting the rate, but that may obscure/complicate operational issues. I could lock this to ensure one-at-a-time query though... Hmm... I'll submit an update...

skinowski:
@zootalures @kmjohansen Added a mutex to force one-call-a-time for Status gRPC.

api/agent/pure_runner.go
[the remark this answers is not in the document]
skinowski: I've decided not to limit concurrency via mutex/channel at this point. I'd vote for leaving this as is and reiterating/reevaluating on it if this becomes an issue. Also IdleTimeout will serve queries via hot container if rate is more than 1 rps.

api/agent/pure_runner.go
```diff
@@ -668,6 +784,16 @@ func PureRunnerWithAgent(a Agent) PureRunnerOption {
 	}
 }
 
+func PureRunnerWithStatusImage(imgName string) PureRunnerOption {
```
kmjohansen: Thanks for adding the note to options.md.  That does point out where the documentation for save and load reside, but it doesn't make it clear that the correct operation of the status image is dependent upon loading an image with tags correctly embedded.  Would you please add a comment here, or somewhere else if this isn't the appropriate place, that makes this clear.  I'm trying to avoid having somebody spend a lot of time debugging a runner that won't load a status image, simply because the image loaded isn't bearing any tags.  Something like the following would suffice, assuming the description is factually accurate:
```
// PureRunnerWithStatusImage returns a PureRunnerOption that annotates a PureRunner with a
// statusImageName attribute.  This attribute names a file that is loaded by docker when a status check
// gRPC message is received by the runner.  Correct operation of the status image requires that an image be
// loaded by the DOCKER_LOAD_FILE environment variable.  In addition, this docker image must have been
// created with the tag name that is passed as the argument in imgName.  See docs/operating/options.md for
// additional details.
```

skinowski: Added comments to clarify the PureRunnerWithStatusImage() usage.

api/agent/pure_runner.go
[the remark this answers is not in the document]
kmjohansen: Personally, I like @zootalures suggestion.  Bursts of status requests seem potentially problematic.  If this code were to serialize runs of the status image and then return the status value to any callers who blocked while it ran, that would lower the load induced from running multiple status functions, correct?

Looking at it from the other side, what data would we need to collect in order to determine that concurrent status request are becoming a problem?

skinowski:
I will include more tests, please do not merge.

skinowski:
@kmjohansen @zootalures Ping.. Ping... Ready for review again.

PR #1118, opened 2018-07-11
String helper for annotations

riconnon:

### DIFF
--- a/api/models/annotations.go
+++ b/api/models/annotations.go
@@ -160,6 +160,18 @@ func (m Annotations) Get(key string) ([]byte, bool) {
 	return nil, false
 }
 
+// GetString returns a string value if the annotation value is a string, otherwise an error
+func (m Annotations) GetString(key string) (string, error) {
+	if v, ok := m[key]; ok {
+		var s string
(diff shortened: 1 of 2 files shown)


api/models/annotations.go
```diff
@@ -160,6 +160,18 @@ func (m Annotations) Get(key string) ([]byte, bool) {
...
+// GetString returns a string value if the annotation value is a string, otherwise an error
+func (m Annotations) GetString(key string) (string, error) {
+	if v, ok := m[key]; ok {
+		var s string
+		if err := json.Unmarshal([]byte(*v), &s); err != nil {
+			return "", err
+		}
+		return s, nil
+	}
+	return "", errors.New("Annotation not found")
```
zootalures: It might be useful to be able to distinghish between "annotation not found" and "error unmarshalling annotation" 

Without adding a third return parameter the norm here would be an identifiable error in the API (ErrAnnotationNotFound and return that)

riconnon: I started to do that then realised that they don't really associate with specific HTTP errors.

PR #1120, opened 2018-07-12
Adding support for building on wercker

lake-of-dreams:
- Link to issue this resolves
Currently fnproject builds on circleci. This PR demonstrates changes that will be required to build it in wercker
- What I did
Following are the changes required to build fnproject as is on wercker.

**wercker.yml** - Added a wercker.yml on lines of existing .circlecli/config.yml. Few things have changed for example we no more require to install `golang` as the base `box` being used is `golang:1.10` itself.Also changed `CIRCLE_BRANCH` to equivalent `WERCKER_GIT_BRANCH` . Not sure about the usage and need of `CIRCLE_PR_REPONAME` but if required - we can try to find equivalents in wercker for this as well.

**agent_test.go** - Two tests were modified - `TestTmpFsRW` and `TestTmpFsSize`. Both of these tests were trying to check if `/proc/mounts` contains mount points for '/tmp' and `/` both. When running in wercker - only `/tmp` entry was there in `/proc/mounts`. Also the attribute string was not exact as being compared (`rw,nosuid,nodev,noexec,relatime` in `TestTmpFsRW` and `rw,nosuid,nodev,noexec,relatime,size=1024k,nr_inodes=1024` in `TestTmpFsSize`) . Therefore changes these tests slightly to not look for `/` entry in '/proc/mounts' and also verify presence of each attribute for `/tmp` individually rather than as a whole string.

**release.sh, images/dind/release.sh, images/fn-test-utils/release.sh** - In `release.sh` and `images/dind/release.sh` - `git status` command was failing in wercker with a dirty tree error - 
showing
```
/vendor/git.apache.org/thrift.git/tutorial/erl/client.sh
/vendor/github.com/docker/docker/CONTRIBUTING.md
```
as being modified but not staged. Could not understand why these deps were checked in in vendor folder and also why git was giving a error - not sure if any of the build steps inside Makefile is touching these files. Therefore commented all the `git` related code in `release.sh` and `images/dind/release.sh`. Also some code was changed in these scripts to tag and push to repositories under DOCKER_USER in dockerhub for testing instead of pushing to fnproject itself.

**test.sh** - When running in wercker - we create a custom network for docker containers and any containers that requires access to each other ( even all the wercker `steps` are run in a container only called `box`) - has to be on that network. Therefore changed the code that started the mysql, minio and postgres containers by adding a `--network=$DOCKER_NETWORK_NAME`. Also as an advantage of this custom network - we no longer need the `host` function to deduce IP address of these containers as these can now simply be referred by the container name as will be visible in the script.

Also to test `fnproject/swagger` it was required that `pwd` is mounted at `/go/src/github.com/fnproject/fn` in the container - but as explained earlier since we run all the steps inside a docker container only - to make data available to any other container - we need to use docker `volume`. Therefore changes were done to copy first `pwd` to a docker `volume` using an ephemeral container and then attach that container to `fnproject/swagger` to work as expected

- How I did it
As you can see this has been done in fork of fnproject and may or may not be merged directly to your main repo based on whether you would want to cut over to wercker completely or want to run circleci and wercker in parallel for some time. But this should give you an fare idea of the possibilities.!

- How to verify it
I have created a pipeline in wercker staging at https://dev.wercker.com/ravi-r-singhal/fn/ and anyone with access to wercker staging should be able to view the application and builds


lake-of-dreams:
Separated wercker related changes to separate files ( apart from agent_test.go_ and now builds on both wercker staging - https://dev.wercker.com/ravi-r-singhal/fn/runs/build/5b487dac5edcca00064ec9cb and circleci - https://circleci.com/gh/fnproject/fn/4031

carimura:
Thanks for this @lake-of-dreams. We're definitely open to running the build process on Wercker as discussed, but we shouldn't have staging/dev inside OSS master. When can this be run on Wercker production for anyone to access?

lake-of-dreams:
@carimura - The feature is now released in wercker prod and I have also created a pipeline for the fork that I had created at https://app.wercker.com/ravi-r-singhal/fn/runs

I have made you admin for the application above.


This feature is not public yet and is only allowed to users of certain `organizations` based on a flag. I have enabled it for `fnproject` organization in app.wercker and any users under that org should be able to create pipelines. So after changes from my fork are merged to your branch - you may please create an application in your org from master branch of fnproject.

PR #1130, opened 2018-07-18
fn: cleanup of docker private registry code

skinowski:
Start using URL parsed ServerAddress and its subdomains
for easier image ensure/pull in docker driver. Previous
code to lookup substrings was faulty without proper
URL parse and hostname tokenization. When searching
for a registry config, if image name does not contain
a registry and if there's a private registry configured,
then search for hub.docker.com and index.docker.io. This
is similar to previous code but with correct subdomain
matching.

### DIFF
--- a/api/agent/drivers/docker/docker.go
+++ b/api/agent/drivers/docker/docker.go
@@ -45,6 +45,11 @@ type runResult struct {
 	status string
 }
 
+type driverAuthConfig struct {
+	auth       docker.AuthConfiguration
+	subdomains map[string]bool
+}
+
 func (r *runResult) Error() error   { return r.err }
 func (r *runResult) Status() string { return r.status }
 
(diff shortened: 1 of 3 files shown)


zootalures:
@skinowski   Can you clarify exactly what the expected behaviour is here (which domains should work/shouldn't work) and explain what this is fixing?

skinowski:
The PR fixes the issue in: #1129 

A TODO in the code also states the problem:

https://github.com/fnproject/fn/blob/master/api/agent/drivers/docker/docker.go#L315

Similary registryURL() uses strings.Contains() to check for subdomains:

https://github.com/fnproject/fn/blob/master/api/agent/drivers/docker/registry.go#L11

For example, with image name "docker.io/myrepo/myimage" or "**hub**hub.docker.com**.tr**/myrepo/myimage), the following calls are problematic:

```
strings.HasSuffix("https://index.docker.io**/v2/**", "docker.io")
strings.HasSuffix("https://**zoo.spicy**docker.io", "docker.io")
strings.Contains("**hub**hub.docker.com**.tr**", "hub.docker.com")
```

With this PR, the string parsing is now being changed to URL parse auth.ServerAddress during agent startup. After this, the subdomains are saved into a config to be looked up during image pull.

PR #1134, opened 2018-07-24
fn: tests for private repo auth and rename DOCKER_AUTH

skinowski:
Renamed DOCKER_AUTH with FN_ prefix to clarify the purpose. Docker
does not use this variable.

New tests to clarify the repo/auth-config behavior.

### DIFF
--- a/api/agent/drivers/docker/registry_test.go
+++ b/api/agent/drivers/docker/registry_test.go
@@ -1,7 +1,10 @@
 package docker
 
 import (
+	"strings"
 	"testing"
+
+	"github.com/fsouza/go-dockerclient"
 )
 
 func verify(expected []string, checks map[string]bool) bool {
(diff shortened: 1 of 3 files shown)


api/agent/drivers/docker/registry_test.go
```diff
@@ -1,7 +1,10 @@
 package docker
 
 import (
+	"strings"
 	"testing"
+
+	"github.com/fsouza/go-dockerclient"
```
ericfode: Is this a new dep?

api/agent/drivers/docker/registry_test.go
```diff
@@ -57,3 +60,54 @@ func TestRegistrySubDomains(t *testing.T) {
 		t.Fatalf("subdomain results failed expected[%+v] != results[%+v]", exp, res)
 	}
 }
+
+func TestRegistryEnv(t *testing.T) {
+
+	testCfg := `{
+	"auths":{
+		"https://my.registry.com":{"auth":"Y29jbzpjaGVlc2UK"},
+		"https://my.registry.com:5000":{"auth":"Y29jbzpjaGVlc2UK"},
```
ericfode: No email and password style auths needed?

skinowski: Yes, not testing those... We pass that directly to fsouza/docker currently...

api/agent/drivers/docker/registry_test.go
skinowski: No, we use it currently in driver/docker.

PR #1137, opened 2018-07-25
fn: New timeout for LB Placer

skinowski:
Previously, LB Placers worked hard as long as
client contexts allowed for. Adding a Placer
config setting to bound this by 30 seconds by
default.

The new timeout is not accounted during actual
function execution and only applies to the amount
of wait time in Placers when the call is not
being executed.


skinowski:
Increased the default timeout to 360 secs. @zootalures

Issue #1141, opened 2018-07-30
Support loading environment variables from file


jgh-:
I am attempting to use Docker Secrets to propagate secret data such as database credentials, however this data is propagated via files mounted directly in the container by the docker process and not easily put into environment variables prior to starting the fnserver process (at least with just manipulating compose files).

One good solution to this that Postgres supports, for example, is having each env variable have a corresponding `_FILE` variant (e.g. `POSTGRES_USER_FILE`) where it simply slurps the value of that file.

PR #1142, opened 2018-07-30
Support _FILE postfixes for environment variables

jgh-:
- Link to issue this resolves
https://github.com/fnproject/fn/issues/1141

- What I did
`getEnv` and `getEnvInt` now test to see if `{key}_FILE` is set after testing `{key}`.  If it is set it reads the value from the specified file.

- How I did it
I searched for the point where the environment variables were loaded and added a small feature to it.

- How to verify it
Build the docker image as usual, then
```sh
echo 8081 > /tmp/test_var
docker run -v /tmp/test_var:/run/secrets/test_var -v /var/run/docker.sock:/var/run/docker.sock -e FN_PORT_FILE=/run/secrets/test_var -i {image}
```
You should see Fn server running on port 8081

Support `_FILE` postfixes for environment variables to be loaded from file.

### DIFF
--- a/api/server/init.go
+++ b/api/server/init.go
@@ -2,12 +2,12 @@ package server
 
 import (
 	"context"
+	"io/ioutil"
 	"os"
 	"os/signal"
-	"io/ioutil"
-	"strings"
 	"runtime"
 	"strconv"
+	"strings"
 	"syscall"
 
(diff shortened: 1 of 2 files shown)


delabassee:
@jgh- 's OCA is approved!

Issue #1150, opened 2018-08-01
Custom Route Handlers have no Access to Server's route serve method


gviedma:
Extensions that implement custom trigger handling using `server.WithoutHTTPTriggerEndpoints()` are also required to also implement route handling. Implementing custom route handlers is not currently possible since the Server method `serve(c *gin.Context, app *models.App, route *models.Route)` is private and cannot be invoked outside of the server package.  A version of this method (e.g. ServeRoute) should be made public as is already the case with `ServeHTTPTrigger`.

PR #1151, opened 2018-08-01
Expose ServeRoute method on Server to allow extensions to plugin custom route handling

gviedma:
Fixes issue https://github.com/fnproject/fn/issues/1150 by exposing a public method `func (s *Server) ServeRoute(c *gin.Context, app *models.App, route *models.Route) error` on Server that can be invoked by custom route handlers.


rdallman:
we should reconsider what to do with https://github.com/fnproject/fn/blob/master/fnext/middleware.go#L14 as well (the `*server.Server` implements this), it's for the same thing but slightly different, in general though we should only use methods on interfaces most likely for stuff like this.

PR #1152, opened 2018-08-01
Add server API to disable hybrid API on API servers

zootalures:
In a customised server where runners get state via a customised data store  we want to be able to bypass the Hybrid API to reduce the API server footprint.


zootalures:
Yeah - hope so.

PR #1154, opened 2018-08-03
fn: stats view/distribution improvements

skinowski:
*) View latency distribution is now an argument
in view creation functions. This allows easier
override to set custom buckets. It is simplistic
and assumes all latency views would use the same
set, but in practice this is already the case.
*) Moved API view creation to main, this should not
be enabled for all node types. This is consistent with
the rest of the system.
*) Some stats with Distribution did not have their buckets
configured, this is now plumbed in.

### DIFF
--- a/api/agent/stats.go
+++ b/api/agent/stats.go
@@ -5,10 +5,11 @@ import (
 	"strings"
 	"time"
 
+	"github.com/fnproject/fn/api/common"
+
 	"github.com/sirupsen/logrus"
 	"go.opencensus.io/stats"
 	"go.opencensus.io/stats/view"
-	"go.opencensus.io/tag"
 )
 
 // TODO add some suga:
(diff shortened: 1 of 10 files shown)


skinowski:
This is a bit haphazard PR, but trying to get these into a somewhat consistent state.

api/agent/stats.go
```diff
@@ -82,70 +83,70 @@ const (
...
 
 // RegisterContainerViews creates and register containers views with provided tag keys
-func RegisterContainerViews(tagKeys []string) {
+func RegisterContainerViews(tagKeys []string, latencyDist []float64) {
 	// Create views for container measures
 	for i, key := range containerGaugeKeys {
 		if key == "" {
 			continue
 		}
-		v := createView(containerGaugeMeasures[i], view.Count(), tagKeys)
```
skinowski: a bugfix here... Sum vs Count...

PR #1156, opened 2018-08-03
remove agent.call.req field

rdallman:
in order to open the pearly gates of cloud events we need to rip the http
request specific stuff out of the agent's plumbing. this makes less of an
improvement than maybe I would've liked, but at least that part of it is
accomplished. agent.call.RequestBody is the 'standard' way to get a handle on
the request body now, mostly due to lb_agent. this replaces the previous logic
for caching the request body for a call to use when a placer retries with
something that doesn't require an extra goroutine, this should reduce latency
through the lb agent and is in some ways simpler (and others not).
additionally, this requires Submit and GetCall to take a context argument,
which is now plumbed too.

this also extracts a decent chunk from the delta of #1138

### DIFF
--- a/api/agent/call.go
+++ b/api/agent/call.go
@@ -126,7 +126,7 @@ func FromRequest(app *models.App, route *models.Route, req *http.Request) CallOp
 			SyslogURL:   syslogURL,
 		}
 
-		return setCallPayload(ctx, req.Body, c)
+		return c.setCallPayload(ctx, req.Body, req.GetBody)
 	}
 }
 
@@ -194,7 +194,7 @@ func FromHTTPTriggerRequest(app *models.App, fn *models.Fn, trigger *models.Trig
 			SyslogURL:   syslogURL,
 		}
 
-		return setCallPayload(ctx, req.Body, c)
+		return c.setCallPayload(ctx, req.Body, req.GetBody)
 	}
 }
 
@@ -331,8 +331,8 @@ func reqURL(req *http.Request) string {
 func FromModel(mCall *models.Call) CallOpt {
 	return func(ctx context.Context, c *call) error {
 		c.Call = mCall
-		c.originalBody = strings.NewReader(c.Payload)
-		return nil
+		f := func() (io.ReadCloser, error) { return ioutil.NopCloser(strings.NewReader(c.Payload)), nil }
+		return c.setCallPayload(ctx, strings.NewReader(c.Payload), f)
 	}
 }
 
@@ -340,7 +340,7 @@ func FromModel(mCall *models.Call) CallOpt {
 func FromModelAndInput(mCall *models.Call, in io.ReadCloser) CallOpt {
 	return func(ctx context.Context, c *call) error {
 		c.Call = mCall
-		return setCallPayload(ctx, in, c)
+		return c.setCallPayload(ctx, in, nil)
 	}
 }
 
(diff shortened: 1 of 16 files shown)


api/agent/call.go
```diff
@@ -406,13 +390,32 @@ func (c *call) Extensions() map[string]string {
...
+		c.cachedBody = buf
 	}
-	return c.req.Body
+
+	reader := io.TeeReader(c.originalBody, c.cachedBody)
+	// read anything in our cache first, then read the rest of the original body
+	// (which will get added to our cached body for the next time around)
+	reader = io.MultiReader(bytes.NewReader(c.cachedBody.Bytes()), reader)
+
+	return ioutil.NopCloser(reader)
```
rdallman: this may or may not be racy, which makes me think that it may be, it's kinda slick vs the previous tho so i'm trying to run with it. the properties are nice vs. the old read the entire thing into the buffer before sending (and extra goroutine to copy & timeout on), but I fear ) a data race on the underlying byte slice & ) a race where if 2 threads successfully read from the reader, say, 1 timed out but read to the teereader past the deadline, the second could end up with a 'gap' in the body (and this one in particular a mutex would not necessarily fix unless we serialized all readers in their entirety). thinking.

skinowski: I like this too, but I think it is indeed not safe. I'll take a look at it again when I'm rested, but maybe we should just accept the difficulty of this and just block and fill the buffer. We lose the nice streaming behavior and possibly take some hit on latency, but in practice I don't think this matters.

rdallman: i think i'm gonna make a reader that on the first read locks & dumps the entire req into a buffer (as you say), this is still a slight change in behavior from blocking & reading the entire request body into the buffer _before_ sending the body bytes (reduces memory overhead / latency in lb, esp in cases where placement fails entirely), as well as changing the abstraction to wrap things up into an io.Reader (this won't require an extra goroutine, since io is done elsewhere respecting the ctx timeout we won't have to explicitly do this). in any case, may be able to keep the teereader, but need to fiddle a bit.

rdallman:
@skinowski i've changed the cacher to a 'more sane' thing than the mr tee business if you wouldn't mind taking a peep https://github.com/fnproject/fn/pull/1156/commits/ed810d2276473d6a1d3dbc1bb0af4725048dd2be. it seems somewhat simple, i'm still reeling with not being able to read a reader when i want to but hey bats could be taking over the major cities as we speak so life could be worse.

@mantree would be nice to merge this if possible prior to additional invoke plumbing... the thinking was we rip req out, then it's easy to make a `GetCall(FromEvent(cloudevent.CloudEvent))` without having to create a phony http.Request just to make the agent happy. the FromEvent bits don't need a whole lot and we can probably use most of the existing cloud event bits that exist already.

api/agent/pure_runner.go
```diff
@@ -678,17 +668,40 @@ func (pr *pureRunner) Engage(engagement runner.RunnerProtocol_EngageServer) erro
 	return err
 }
 
+func readData(state *callHandle) {
+	dataFeed := state.spawnPipeToFn()
+
+	var i int
+	for {
+		i++
```
skinowski: i unused?

api/agent/call.go
```diff
@@ -368,28 +338,82 @@ func (a *agent) GetCall(opts ...CallOpt) (Call, error) {
...
+
+	// only allow one read of the buffer ever. everybody else must wait. this is
+	// entirely fucked up and can deadlock all succeeding threads for this body,
+	// but go ahead and try just reading the reader into a buffer and see if life
+	// is still worth living. TODO this leaves a situation where if the first reader
+	// doesn't finish then nobody else gets the body ever. NO SOUP FOR YOU
+	done := make(chan struct{}, 1)
+	done <- struct{}{} // prime it
+
+	buf := bufPool.Get().(*bytes.Buffer)
```
skinowski: I think this leaks. Should there be a corresponding Put()?

api/agent/call.go
```diff
@@ -368,28 +338,82 @@ func (a *agent) GetCall(opts ...CallOpt) (Call, error) {
...
+			return ioutil.NopCloser(strings.NewReader(stringer.String())), nil
+		}
+		return nil
+	}
+
+	// only allow one read of the buffer ever. everybody else must wait. this is
+	// entirely fucked up and can deadlock all succeeding threads for this body,
+	// but go ahead and try just reading the reader into a buffer and see if life
+	// is still worth living. TODO this leaves a situation where if the first reader
+	// doesn't finish then nobody else gets the body ever. NO SOUP FOR YOU
```
skinowski: I think this is fine, all of such I/O reads would typically be done in a go-routine, where we have to assume that first-reader would close/abort properly unlocking the followers as well.

Another weird case here is when a non-eof error happens during ReadFrom(), the followers will block.
We might want to expose ReadFrom() error code for the followers are well.

rdallman: > But when a non-eof error happens during ReadFrom(), with early return
the followers will block. We might want to close to channel and
expose ReadFrom() error code for the followers are well.

good idea

api/agent/call.go
rdallman: can add one to the returned io.Closer `Close()` instead of the nop. d'oh. it won't leak, per se, it'll get GC'd but we're stealing from the pool without giving back (and probably not allocating new ones only while throughput is decreasing).

api/agent/call.go
rdallman: done, but the comment stays... there's still a case where the first request could get cut short that the second request could in theory pick up and finish flushing to the cache, that's just where we get into race territory...

PR #1159, opened 2018-08-06
Add Entity Equality Tests

mantree:
Creates a test that aims to assert that the Equals and Clone functions
for the App struct actually work.

The bulk of the code is spent creating a gopter generator for the App
Struct. See information of generative or property based testing for
explainations on that topic, but basically it's an object that is
capable of creating a stream of unique instances of the given struct.

With the generator we then make three assertions:
 1) App's are always equal to themselves.
 2) A .Clone() of an App is Equal to the original App.
 3) A .Clone() of an App that has a field modified is not equal to the
 orignal.

The third property is the worse for implementation, as it does not
generate the field to modify, it simply loops all fields for each generated
App, and checks Equals always breaks.

Break testing shows that this would have caught earlier bugs in Equals
due to field addition. It will add to the work to add further fields,
generators have to be manually specified for each field, but that
seems a worthy cost.


rdallman:
a bit pedantic, any chance of doing this without the added dependency (aside from more to manage, it's more cognitive burden for every dev that has to futz with these tests to learn a 3rd party API and this one isn't widely known amongst gophers)? 

`reflect` isn't my fav thing in the world but it is in stdlib, this is just for testing, and have used `https://golang.org/pkg/reflect/#DeepEqual` in the past to test stuff like this and it works OK.

mantree:
I think this could largely be done with test/quick, but gopter is far more able (has shrinking, more exstensible) and I wanted to get used to it.

This is trying to assert some core properties of the code, which I think examples would miss. I don't think it can be done without a lot of reflection, and although the gopter stuff looks heavy when I tried without it I just ended up with a half baked pants version of the map that gets fed into the structs gen.

I'm quite interested in the conversation around this approach as well. Once the generators are written they can be reused and I'd look to becoming more property driven in the future.

zootalures:
I think a few people have raised using PBT for contracts in the past, there are a few places I think it could help out, (e.g data store and tests maybe where you have a couple of dimensions of variation on a couple of properties it can simplify/strengthen valid extent testing) 

I think I’m pro if we can reuse the generators there against if not. 

Can you double check that the licensing of this and any subdeps are sound

riconnon:
gopter is under MIT and there are no additional transitive dependencies.

One thing though, is we should pin the gopter version in Gopkg.toml I guess to avoid the version changing and causing unexpected failures in tests.

riconnon:
@rdallman @zootalures 
Seems like you two disagreed regarding the use of gopter.

I think that we can re-use the generators in other test areas as long as we check that the relevant parts are exported.

rdallman:
i have PTSD from bad experiences maintaining non-standard go tests in the past over a long period of time (https://github.com/smartystreets/goconvey, to be exact, so not this), it was not easy to ramp up new people and they would just go write their own new test suite they could understand -- a problem we have, already, as well. it would be great to use `testing/quick` if possible since it's stdlib, though I'll admit it's somewhat 'obscure' for most go devs in any case -- if gopter has additional features we're using I can understand (are we? motivation? alternatives?) -- that's just my preference, I won't block this and agree about the general usefulness of PBT, also thanks for adding it.

mantree:
testing/quick is pretty limited, it doesn't really have generators, so doesn't support shrinking, which is pretty key for PBT imo. Although that said gopter doesn't shrink struct generators, so that's not a huge win here.

I get the reservations, but when thinking about this I couldn't think of another way of ensuring that an added struct field was also added to Equals/Clone, without relying on a human to remember something. With this it will prompt them pretty quickly to sort it out.

mantree:
@rdallman @zootalures someone want to give this the green light then?

api/models/fn_test.go
```diff
@@ -0,0 +1,109 @@
...
+	fieldGens["ID"] = gen.AlphaString()
+	fieldGens["Name"] = gen.AlphaString()
+	fieldGens["AppID"] = gen.AlphaString()
+	fieldGens["Image"] = gen.AlphaString()
+	fieldGens["Config"] = configGenerator()
+	fieldGens["ResourceConfig"] = resourceConfigGenerator(t)
+	fieldGens["Annotations"] = annotationGenerator()
+	fieldGens["CreatedAt"] = datetimeGenerator()
+	fieldGens["UpdatedAt"] = datetimeGenerator()
+	fieldGens["Format"] = gen.AlphaString()
```
rdallman: sorry this has been on the backburner, I'll get better at my flagged emails again one day (send help). maintenance of the fields here concerns me, but the expectation is that if a field gets added or removed then these tests will fail and someone will have to go in and update these tests, right? I'd suppose that's a good thing overall, shouldn't be able to just go add a field willy nilly without having to add tests yea? Roundabout I'm asking whether it's possible to do this without the generators, almost fuzz style (but with working ones), and get the same effect but without the code maintenance overhead here.

mantree: Yeah, it's kind of the point to highlight that new fields need a bit more than just adding to the struct.

I think you could do something without the generators, but it wouldn't explore as much as this version does. The payoff is slight here, but I'd like to add more of these style tests in the future.

PR #1160, opened 2018-08-07
fn: SSL config adjustments

skinowski:
SSL related FN_NODE_CERT (and related) settings are
not very clear today. Removing this in favor of a
simple map of tls.Config objects. Three keys are
provided for this map:

TLSGRPCServer
TLSAdminServer
TLSWebServer

which correspond to server TLS settings for the
associated services. These three service keys
are detected during startGears() execution and
SSL/TLS is enabled accordingly in web/admin listeners.

Operators/implementers can further add more
keys to the map and add their own TLS config.

pool.PKIData is now removed. This is a very limited TLS
configuration and tls.Config in crypto/tls provides this
already.

Providing some utility functions to cover the old functionality
in common/tls_utils which now exposes a few simple functions
to set ca, clientCA, cert/key, etc.

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/tls"
 	"encoding/base64"
 	"errors"
 	"fmt"
@@ -104,15 +105,6 @@ const (
 	// EnvJaegerURL is the url of a jaeger node to send traces to.
 	EnvJaegerURL = "FN_JAEGER_URL"
 
-	// EnvCert is the certificate used to communicate with other fn nodes.
-	EnvCert = "FN_NODE_CERT"
-
-	// EnvCertKey is the key for the specified cert.
-	EnvCertKey = "FN_NODE_CERT_KEY"
-
-	// EnvCertAuth is the CA for the cert provided.
-	EnvCertAuth = "FN_NODE_CERT_AUTHORITY"
-
 	// EnvRIDHeader is the header name of the incoming request which holds the request ID
 	EnvRIDHeader = "FN_RID_HEADER"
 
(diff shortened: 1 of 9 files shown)


skinowski:
Basic idea here is TLS/SSL is complicated and should be left to downstream implementer.

api/server/server.go
```diff
@@ -983,14 +935,20 @@ func (s *Server) startGears(ctx context.Context, cancel context.CancelFunc) {
 	installChildReaper()
 
 	server := http.Server{
-		Addr:    listen,
-		Handler: &ochttp.Handler{Handler: s.Router},
+		Addr:      listen,
+		Handler:   &ochttp.Handler{Handler: s.Router},
+		TLSConfig: s.tlsConfigs[TLSWebServer],
```
ostrain: Is this going to crash if the key is not present in the s.tlsConfigs map?

api/server/server.go
```diff
@@ -1003,12 +961,18 @@ func (s *Server) startGears(ctx context.Context, cancel context.CancelFunc) {
 		adminListen := fmt.Sprintf(":%d", s.adminListenPort)
 		logrus.WithField("type", s.nodeType).Infof("Fn Admin serving on `%v`", adminListen)
 		adminServer := http.Server{
-			Addr:    adminListen,
-			Handler: &ochttp.Handler{Handler: s.AdminRouter},
+			Addr:      adminListen,
+			Handler:   &ochttp.Handler{Handler: s.AdminRouter},
+			TLSConfig: s.tlsConfigs[TLSAdminServer],
```
ostrain: Same here

skinowski: this should return zero value (https://blog.golang.org/go-maps-in-action) which is nil in this case. Later we check for not-nil to see if we need to enable SSL.

PR #1165, opened 2018-08-08
WIP: Fn Invoke take 3

mantree:
Just this afternoons effort to start from the front so far.

I've copied up the event from the other PR, but I do think that there is a strong case for a library here.

Other than that I'm basically copying what has been done for http triggers and just trying to get the most boring stuff done.


rdallman:
i've started a library https://github.com/fnproject/cloudevent which has basic support for transforming an http.Request into a cloudevent by the spec's 0.1 standard. I'd like to use it, as well, for not only this but cli and fdk-go too -- we can modify to needs relatively easily. I believe that the copied over code from 1138 was also copied over from said library and tweaked slightly. 

will take a further look today

rdallman:
one thing i'm not sure about is whether we are going to have to update the lb_agent/pure_runner models to accommodate arbitrary cloud events (as #1138 does) -- eventually we likely want to move to #1138 style but for the intermediary state idk how much we can get away with. i'd be OK with merging something that works for a 'standard' agent before getting to that point too, it would unblock some work potentially and keep the deltas smaller.

mantree:
I think smallest Delta possible! And really should be noting down these further steps somewhere...

api/event/event.go
```diff
@@ -0,0 +1,240 @@
...
+//const (
+//	BodyTypeNone   = BodyType(iota)
+//	BodyTypeJSON   = BodyType(iota)
+//	BodyTypeString = BodyType(iota)
+//	BodyTypeBinary = BodyType(iota)
+//)
+
+// TODO :  Ideally this would be able to pass an arbitrary byte stream in its body (i.e. not be subject to JSON limitations)   but it's not now
+// Currently this only accepts valid JSON bodies, or non-json content that is valid UTF-8
+// TODO: Would really prefer this to be an interface with stronger correctness by guarantee
```
rdallman: i'm not sure we need to support binary _in_ a json. spec release 0.2 maybe clarifies this some more (i.e. base64).

rdallman:
I made https://github.com/fnproject/cloudevent/pull/3 (PTAL) to fill in for the validate stuff going away, that could be useful, though all of those fields we don't really _need_ in order to run a function, it would be nice to have the cli send them in so users don't have to futz with them and then require them against the API in general.

mantree:
This have been over taken by events, represented in #1168

PR #1166, opened 2018-08-08
Change the syslog format to use app_name instead of app_id

### LINKED ISSUE #1146: change syslog format to use app name rather than app id
I haven't looked at the plumbing but this should be relatively straight forward to switch over.

the motivation being that since users specify app name they will be searching for app names in their logs rather than an app id, and it's an extra step to look up an app id for a given app name before going to search the logs. 

moreover, we should allow users to specify a syslog line format with some DSL to specify various things they do or don't want (such as call id, app name, function name, function id, trigger name, etc. etc.), but this is less pressing at the moment and the thinking is app should default to name in the logs, in any case.

peterj:
This fixes #1146. 

I added the AppName to models.Call in order to be able to write it out in the logs.

Verified by running unit tests and manually setting the app sys-log and verifying correct logs are being sent to papertrail.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -750,7 +750,7 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	defer bufPool.Put(buf1)
 	defer bufPool.Put(buf2)
 
-	sw := newSyslogWriter(call.ID, call.Path, call.AppID, call.AppName, syslog.LOG_ERR, s.container.syslogConns, buf1)
+	sw := newSyslogWriter(call.ID, call.Path, call.AppName, syslog.LOG_ERR, s.container.syslogConns, buf1)
 	var syslog io.WriteCloser = &nopCloser{sw}
 	syslog = newLineWriterWithBuffer(buf2, syslog)
 	defer syslog.Close()                            // close syslogger from here, but NOT the call log stderr OR conns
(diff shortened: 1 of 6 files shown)


zootalures:
Not super important but I think we should be using resource IDs as canonical identifiers in things like this going forward  rather than display names. (fn_id, app_id, trigger_id)

zootalures:
Not super important but I think we should be using resource IDs as canonical identifiers in things like this going forward  rather than display names. (fn_id, app_id, trigger_id)

rdallman:
i guess the v2 api itself is moving this direction, though the cli retains the name-friendly commands (hence the issue filed, after some discussion). there hasn't been a general consensus on the whole matter really, but maybe since v2 is moving that way we should put a stick in the mud

peterj:
I can add the app_id back and have it in addition to the name. Looking at the actual logs, this:

```
Aug 08 12:32:25 fn logger: call_id=01CMDGD889R2M0250ZJ0000001 func_name=/testapp app_id=01CM8P0Y60R2M023RZJ0000001 **app_name=petersapp** 2018/08/08 19:32:25 In main function
```

seems more human understandable than this:

```
Aug 08 12:32:25 fn logger: call_id=01CMDGD889R2M0250ZJ0000001 func_name=/testapp app_id=01CM8P0Y60R2M023RZJ0000001 2018/08/08 19:32:25 In main function
``` 

As my next step, if I see the above line is probably to figure out what the app name was...

rdallman:
> As my next step, if I see the above line is probably to figure out what the app name was...

yea, that was the thinking. in most cases the func name probably gives it away without needing to, eventually we have to make a similar decision over func name or id, at least with this patch it's somewhat consistent (calls don't have names, only ids, so they're an exception). in the general concerns bucket i'm not sure how worth it it is to just have everything (both name and id), at some level it costs a significant amount of wasted bytes, and typically logging systems are billed by bytes logged. eventually, we probably need to let users define the logging format they want, but that's not very pressing.

peterj:
Is there a place in the CLI where app_id is required, instead of an app name?

denismakogon:
@peterj in CLI we don't use app IDs, only names. Names are unique per user's space (consider a multi-tenancy).

peterj:
@zootalures are you ok with merging this as is?

zootalures:
Think so but I think we end up having to come back to this.

PR #1168, opened 2018-08-09
Adding Fn invoke endpoint that works just like triggers endpoint

mantree:
A version of Fn invoke that works exactly like the trigger endpoint point.

### DIFF
--- a/api/agent/call.go
+++ b/api/agent/call.go
@@ -202,6 +202,75 @@ func FromHTTPTriggerRequest(app *models.App, fn *models.Fn, trigger *models.Trig
 	}
 }
 
+// Sets up a call from an http trigger request
+func FromHTTPFnRequest(app *models.App, fn *models.Fn, req *http.Request) CallOpt {
+	return func(c *call) error {
+		ctx := req.Context()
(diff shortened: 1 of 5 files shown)


rdallman:
think i get the basic idea here, do the same thing as triggers api but without the trigger configuration bits. that line of reasoning is sensible i think for interfacing with the agent to get something working, in the interest of looking forward is it a lot of work for us to restrict the domain of the http API to only accept cloud events? is the cloudevent library unworkable hence resurrecting the event logic from earlier https://github.com/fnproject/fn/pull/1165 ?

(will fiddle with this today)

rdallman:
i guess part of this exercise was to scope out exactly what that might look like exactly, we're just now trying to skip doing that to have something working is what it seems like.

skinowski:
Why is the resource in models only have Memory?

mantree:
@skinowski that and timeout are going to be the only configurables, i think shapes are going to come into this area.

rdallman:
> Why is the resource in models only have Memory?

good catch, yea we need to iron that out

PR #1172, opened 2018-08-15
Fns now annotated with invoke urls, as per triggers

mantree:
Clone of the trigger work to inject invoke urls into the annotations
on a fn when it is returned from the server.

### DIFF
--- a/api/server/fns_get.go
+++ b/api/server/fns_get.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/fnproject/fn/api"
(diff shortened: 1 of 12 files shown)


zootalures:
sorry haven't got time to do this in detail -  I think this can go in as it stands 

Only question that came to mind was that If every function always has an endpoint shouldn't  this be an attribute on the resource rather than an annotation?

mantree:
I think we had this chat when it came to triggers. I think it came down to the fact it's not really a piece of entity metadata, as in the end user can not edited it, and it could conceivably change out side of their control.

Not sure if I'm remembering that right, but it was something like that

zootalures:
yeah - for triggers it's optional and dependent on the trigger type - my argument is that for functions it's for everywhere. 

That said the cost of adding a new field to fn is relatively high now so....

api/server/fns_get.go
```diff
@@ -9,11 +10,23 @@ import (
...
+
 	f, err := s.datastore.GetFnByID(ctx, c.Param(api.ParamFnID))
 	if err != nil {
 		handleErrorResponse(c, err)
 		return
 	}
 
+	app, err := s.datastore.GetAppByID(ctx, f.AppID)
+	if err != nil {
+		handleErrorResponse(c, fmt.Errorf("unexpected error - fn app not available: %s", err))
```
rdallman: need return here, since we're writing the response back from this line

PR #1173, opened 2018-08-15
removes extension example

### LINKED ISSUE #1163: extensions: no data store provider found for storage url sqlite3
**Description**
Certain problems appear while building an extended fnserver.

1. Use fn-ext-example tutorial and build
2. Build a server
3. Launch a server

```
INFO[2018-08-08T13:55:00+03:00] Setting log level to                          level=info
FATA[2018-08-08T13:55:00+03:00] Error during server opt initialization.       error="no data store provider found for storage url sqlite3:///Users/denismakogon/Documents/oracle/go/src/github.com/fnproject/fn-ext-example/data/fn.db"

```

I expected to get a working fnserver as if i'd build and start ordinary fnserver.

**Additional information you deem important (e.g. issue happens only occasionally):**

fnserver version: 0.3.58
OS env is clear, nothing interfer with Fn-related env vars.

rdallman:
this was not obvious / well documented in our extensions getting started docs
/ examples, this attempts to clarify the matter so that users don't get a main
that blows up in their face at runtime :)

closes #1163


rdallman:
I guess we have that too. any preference? would prefer to only have one if they're about the same level (just a simple getting started)

denismakogon:
Let’s make sample app the only piece of code to reference since we already have a link to that repo and drop whatever we have in Fn. Thoughts?

rdallman:
we can do that. I must say that I don't think the separate repo is very discoverable and think our `examples/` is more easily discoverable for users in general (we should be consistent). I find the current state of our docs hard to navigate as someone familiar with them even, and it's hard for things like extensions to be found when they reside under `docs/contributors/` (users have to search through every dir in `docs/` first level). 

additionally, the logspam example only has listeners around a call and not around app, route, etc. so they illustrate different things. it would be nice to have them in similar places. the tutorials repo may be an OK place too, I don't think the every example gets a repo thing will scale so well though.

rdallman:
done, removed example here, and: https://github.com/fnproject/docs/compare/update-ext-docs?expand=1

rdallman:
i guess this code did highlight the listeners better than the example repo. i'll look to pollenate to fn-ext-example

rdallman:
@denismakogon so big benefit of keeping extensions here is that we can test it to make sure it doesn't break, too... seems like a big benefit, the external repo I don't fully get the value of over just having a directory here, let me know what you think

PR #1174, opened 2018-08-16
Removing the cloudevent fn format changes

mantree:


rdallman:
replace, maybe, but remove?

rdallman:
gonna close... under discussion & no context is provided here

PR #1175, opened 2018-08-17
WIP initial cut at cleaning old docker images

ericfode:
- Link to issue this resolves

- What I did
Added a control loop to clean up images, as well as an LRU to keep track of them.

- How I did it
 ~~Added an LRU, on eviction from the lru a message is passed on a channel to a goroutine that takes care of cleaning up the image.~~
* Added a slice that keeps track of the images that the docker agent tells it about. On each image, there is a count of how often it has been used and when it was last used.  
* Every 5 minutes a clean up worker asks the image cache for any evictable images in order of "most evictable-ness. (this is defined as timeSinceLastUse/numberOfUses, and not locked).
* If the deletion is a success the image is removed from the cache. 
* When a cookie is created for an execution the image is "locked" in the image cache until the cookie is cleaned up ~~**STILL PENDING HOW TO CLEAN THIS UP**~~
 * When an image is used for a container the numberOfUses is incremented.

- How to verify it
Run 100 functions with different 1GB images, if they all worked the evictor did it's job.

Clean up after old functions when there is no more space.

robot


api/agent/drivers/docker/docker.go
```diff
@@ -102,11 +125,27 @@ func NewDocker(conf drivers.Config) *DockerDriver {
 		}
 	}
 
+	imageEvictChan := make(chan docker.APIImages)
+	onEvictor := OnImageEvict(imageEvictChan)
+	go NewImageCleaner(driver, imageEvictChan)
+
+	//20GB
+	driver.imageLRU = NewLRU(1024*1024*1024*1024*20, onEvictor)
```
skinowski: TODO: fetch this from Agent.Config -> Driver.Config

api/agent/drivers/docker/docker.go
```diff
@@ -102,11 +125,27 @@ func NewDocker(conf drivers.Config) *DockerDriver {
...
+
 	if conf.DockerLoadFile != "" {
 		err = loadDockerImages(driver, conf.DockerLoadFile)
 		if err != nil {
 			logrus.WithError(err).Fatalf("cannot load docker images in %s", conf.DockerLoadFile)
 		}
+
+		ctx, _ := common.LoggerWithFields(context.Background(), logrus.Fields{"stack": "listDockerImages"})
+		images, err := driver.docker.ListImages(ctx)
```
skinowski: Oh, loadDockerImages() should be excluded from LRU evictor.

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,103 @@
...
+// Add adds the provided key and value to the cache, evicting
+// an old item if necessary.
+func (c *Cache) Add(key string, value d.APIImages) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// Already in cache?
+	if ee, ok := c.cache[key]; ok {
+		c.ll.MoveToFront(ee)
+		ee.Value.(*entry).value = value
```
skinowski: It's a weird edge case, but if value.Size is updated, then we'd need to update totalSize and run LRU logic too.

api/agent/drivers/docker/docker.go
ericfode: Ahh, because we never want to delete any that we load?

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,103 @@
...
+		return
+	}
+
+	// Add to cache if not present
+	ele := c.ll.PushFront(&entry{key, value})
+	c.cache[key] = ele
+	c.totalSize += value.Size
+
+	if c.TotalSize() > c.maxSize {
```
skinowski: here, this should probably run in a loop until totalSize <= maxSize, but also in a go-routine. We should unlock the mutex before as well. Or perhaps move this out of here into a separate nanny go-routine that gets signalled at every LRU.Add() where we could retry failed docker-removes too.

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,103 @@
...
+		return ele.Value.(*entry).value, true
+	}
+	return
+}
+
+// RemoveOldest removes the oldest item in the cache and returns its key and value.
+// If the cache is empty, the empty string and nil are returned.
+func (c *Cache) RemoveOldest() (key string, value d.APIImages) {
+	c.mu.Lock()
+	c.onEvict(value)
```
skinowski: we should call onEvict() with mutex unlocked.

rdallman:
we do need something like this, thanks for digging into this. if I can be of use, the last time I've seen something along these lines get used there were 2 issues that I'm not sure are tackled here (and maybe they are, I'm just throwing this out there):

1) listing images means that images that are based on other images will cause images that get used more frequently to get removed because the image they depend on is not getting used often and gets LRU'd (and transitively, the one that _does_ get used, as well). it may be possible to work around this, but iirc requires us to keep our own LRU list of the images rather than relying on docker's (stored and loaded on boot, too, perhaps) -- perhaps this has changed and images have some new useful fields in this dept (it's been ~2 years)
2) force removal can cause issues when not serialized with any calls to create containers (using some kind of semaphore), causing container creation to fail. this may also have changed, and is relatively easy to test for, but was the reason we ended up opting to just rotate servers completely after a set number of hours instead of fiddling with the cache bits (a not-great answer that had other causes, too, it would certainly be nice to think we could run servers for prolonged periods, software should work like this)

ericfode:
Chatted with @kmjohansen, he pointed out that I could prune at startup (to catch the case where someone had restarted their runner) and only add images to the LRU that are actually requested by the user. 

For comments from @rdallman  
1. Good catch "list all" was the wrong operation and we can get around it with the above strategy. 
2. Removed force remove, just add the image back to the LRU if there is an issue deleting it. 

Now there needs to be something to keep this from looping too much in the case the entire LRU is in use...

zootalures:
Quick q (am in a field so can’t review properly) but does this avoid deleting the pre-loaded image(s) used for status checks and netNs pre forking

api/agent/drivers/docker/docker.go
skinowski: Yes, these are special, we should not touch them. :-)

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,105 @@
...
+		c.ll.MoveToFront(ele)
+		return ele.Value.(*entry).value, true
+	}
+	return
+}
+
+// RemoveOldest removes the oldest item in the cache and returns its key and value.
+// If the cache is empty, the empty string and nil are returned.
+func (c *Cache) RemoveOldest() (key string, value d.APIImages) {
+	c.onEvict(value)
```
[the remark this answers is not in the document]
kmjohansen: but really that leaves you vulnerable to a race where that image can be re-inserted into the cache once you've unlocked and are evicting.  We almost want something like findOldest() and then have onEvict actually do the removal once it succeeds.

kmjohansen:
@ericfode I talked with @skinowski who convinced me that I've given you bad advice about pruning at startup.  It has the potential to make start take a long time just from the prune operation itself, and it also has the side effect of potentially generating a pull storm for all containers that launch right after restart.

It might be better to just populate the LRU on boot by doing a list, and then only add from the images that are actually requested by customer code.

skinowski:
Had a chat with @kmjohansen on this, in summary, we think:

1) we can't run prune images since it purges indiscriminately. In case of restart, we risk docker-pull flood.
2) we can perhaps use list-images (like before) and keep track of them in LRU. (Except for DockerLoadFile images) (list-images with all=false probably addresses the issue pointed out by @rdallman)
3) when we queue a image remove job, we might need to order operations or block/wait requests/jobs if requests for that same image are between 'docker-pull' and 'docker-create' to avoid errors.
4) LRU should asynchronously run if a threshold is reached in total image size usage.

skinowski:
And of course, by default this should be disabled since we do not know how users are running docker ecosystem (dind, shared, swarm, etc.)

kmjohansen:
Another part of the discussion that @skinowski and I had revolved around how to handle insertion and removal of items from the LRU, and what to do if we receive a request to execute a function while we're trying to delete its image.

From work on previous systems, I've found it easier to only insert items into this type of cache when they're actually eligible for eviction.  So, in that sense, it may be worth deferring cache insertion until after we've evicted the hot functions using image.

It may also be beneficial to flag an item as being in the process of being deleted, so that if a new request arrives for this image, and we've already sent it down to the evictor, we can block (or fail?) the inbound request until the delete completes.  Similarly,  functions in the LRU that aren't yet being deleted can be removed from the list if a new request for them arrives.

Since the delete process is fairly expensive, it may make sense to only trigger this periodically, and only re-try failed deletes on a subsequent invocation.

skinowski:
One minor note: we probably want to run list-images after loading the images in DockerLoadFile, and LRU should record the list of images that are loaded by driver.loadDockerImages() and always skip removing those. In other words, images loaded by DockerLoadFile param should not be evicted. Since these are being pre-loaded, it is likely that they can't be pulled.

api/agent/drivers/docker/docker.go
```diff
@@ -373,6 +404,8 @@ func (drv *DockerDriver) run(ctx context.Context, container string, task drivers
 	attachSuccess := make(chan struct{})
 	mwOut, mwErr := task.Logger()
 
+	drv.imageLRU.Get(container)
```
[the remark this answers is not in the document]
ericfode: The goal of calling that was to get it up to the top of the LRU

kmjohansen:
>What about using this https://docs.docker.com/engine/reference/commandline/system_prune/

IIUC, system prune is going to do more than you really want, since it may also remove networks and other objects that aren't currently in use.

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,108 @@
...
+		return
+	}
+
+	// Add to cache if not present
+	ele := c.ll.PushFront(&entry{key, value})
+	c.cache[key] = ele
+	c.totalSize += value.Size
+	c.mu.Unlock()
+
+	for c.TotalSize() > c.maxSize {
```
skinowski: should probably be a loop here as we don't know single evict will satisfy the total size <= max size requirement.

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,108 @@
...
+// Add adds the provided key and value to the cache, evicting
+// an old item if necessary.
+func (c *Cache) Add(key string, value d.APIImages) {
+	c.mu.Lock()
+
+	// Already in cache?
+	if ee, ok := c.cache[key]; ok {
+		c.ll.MoveToFront(ee)
+		c.totalSize = (c.totalSize - ee.Value.(d.APIImages).Size) + value.Size
+		ee.Value.(*entry).value = value
```
skinowski: c.mu.Unlock()

api/agent/config.go
```diff
@@ -102,6 +104,7 @@ func NewConfig() (*Config, error) {
 		MaxCallEndStacking: 8192,
 		PreForkImage:       "busybox",
 		PreForkCmd:         "tail -f /dev/null",
+		MaxImageCacheSize:  20 * 1024 * 1024,
```
skinowski: I think we might want to initialize this to zero.(unlimited).

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,108 @@
...
+	key   string
+	value d.APIImages
+}
+
+// New returns a new cache with the provided maximum items.
+func NewLRU(maxSize int64, onEvict ImageEvictor) *Cache {
+	return &Cache{
+		ll:      list.New(),
+		cache:   make(map[string]*list.Element),
+		onEvict: onEvict,
```
skinowski: maxSize not initialized? I think we also (here or at a higher layer in driver code) need to handle the case where maxSize is zero. That should be unlimited where evictions get disabled.

api/agent/drivers/docker/docker.go
```diff
@@ -102,6 +130,22 @@ func NewDocker(conf drivers.Config) *DockerDriver {
...
 	}
 
+	imageEvictChan := make(chan docker.APIImages, 10)
+	onEvictor := OnImageEvict(imageEvictChan)
+	go NewImageCleaner(driver, imageEvictChan)
+
+	logrus.Infof("Max cache size: %v\n", conf.MaxImageCacheSize)
+	driver.imageLRU = NewLRU(int64(conf.MaxImageCacheSize), onEvictor)
+
+	go func() {
```
skinowski: so, if the go-routine completes after loadDockerImages() below, we might accidentally clear on-disk requested images.

api/agent/drivers/docker/docker_client.go
```diff
@@ -216,6 +219,18 @@ func (d *dockerWrap) LoadImages(ctx context.Context, filePath string) error {
 	})
 }
 
+func (d *dockerWrap) ListImages(ctx context.Context) ([]docker.APIImages, error) {
+	ctx, span := trace.StartSpan(ctx, "list_docker_images")
+	defer span.End()
+	return d.docker.ListImages(docker.ListImagesOptions{All: false})
+}
+
+func (d *dockerWrap) PruneImages(ctx context.Context) (*docker.PruneImagesResults, error) {
```
skinowski: Unused?

api/agent/config.go
```diff
@@ -108,6 +110,7 @@ func NewConfig() (*Config, error) {
 		MaxCallEndStacking: 8192,
 		PreForkImage:       "busybox",
 		PreForkCmd:         "tail -f /dev/null",
+		MaxImageCacheSize:  2000 * 1024 * 1024,
```
skinowski: default should probably be zero (unlimited)

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,195 @@
+package docker
+
```
zootalures: When you're done doc comments are a must for this type

api/agent/drivers/docker/docker.go
```diff
@@ -54,16 +54,60 @@ func (r *runResult) Error() error   { return r.err }
...
+	timer := time.NewTimer(time.Minute * 5)
+	for {
+		select {
+		case <-context.Done():
+			{
+				return nil
+			}
+		case <-timer.C:
+			{
+				if !dockerDriver.imageCache.OverFilled() {
```
skinowski: maybe get the size that exceeds the limit, then based on that remove images necessary to meet that?

api/agent/drivers/docker/docker.go
```diff
@@ -209,6 +274,7 @@ func (drv *DockerDriver) unpickNetwork(c *cookie) {
 
 func (drv *DockerDriver) CreateCookie(ctx context.Context, task drivers.ContainerTask) (drivers.Cookie, error) {
 
+	drv.imageCache.Lock(task.Image())
```
skinowski: Cookie close() needs Unlock(). And also recursive lock is needed as pointed out by @zootalures

api/agent/drivers/docker/docker.go
```diff
@@ -102,13 +146,34 @@ func NewDocker(conf drivers.Config) *DockerDriver {
...
 
+	go NewImageCleaner(driver, context.Background())
+
+	logrus.Infof("Max cache size: %v\n", conf.MaxImageCacheSize)
+	driver.imageCache = NewCache(int64(conf.MaxImageCacheSize))
+
+	imagesBeforeLoad, err := driver.docker.ListImages(context.Background())
+
 	if conf.DockerLoadFile != "" {
 		err = loadDockerImages(driver, conf.DockerLoadFile)
```
skinowski: it might be easier to modify loadDockerImages to get a list.. (And it would be nice to log the names of those images anyway.) then remove those from imagesBeforeLoad.

api/agent/drivers/docker/docker.go
```diff
@@ -102,13 +146,34 @@ func NewDocker(conf drivers.Config) *DockerDriver {
 		}
 	}
 
+	go NewImageCleaner(driver, context.Background())
```
skinowski: this probably shouldn't start before loadDockerImages() just to be safe.

api/agent/drivers/docker/docker.go
skinowski: it might be easier to modify loadDockerImages to get a list.. (And it would be nice to log the names of those images anyway.) then remove those from imagesBeforeLoad.

api/agent/drivers/docker/docker.go
```diff
@@ -102,13 +146,34 @@ func NewDocker(conf drivers.Config) *DockerDriver {
...
+
 	if conf.DockerLoadFile != "" {
 		err = loadDockerImages(driver, conf.DockerLoadFile)
 		if err != nil {
 			logrus.WithError(err).Fatalf("cannot load docker images in %s", conf.DockerLoadFile)
 		}
 	}
 
+	go func() {
+		images, err := driver.docker.ListImages(context.Background())
```
skinowski: probably don't need to double ListImages(). It's safer to get the list of images from loadDockerImages().

api/agent/drivers/docker/docker.go
```diff
@@ -54,16 +54,60 @@ func (r *runResult) Error() error   { return r.err }
...
 	networksLock sync.Mutex
 	networks     map[string]uint64
 }
 
+func NewImageCleaner(dockerDriver *DockerDriver, context context.Context) error {
+	opts := docker.RemoveImageOptions{}
+	opts.Force = true
+	opts.NoPrune = false
+	opts.Context = context
+	timer := time.NewTimer(time.Minute * 5)
```
skinowski: https://golang.org/pkg/time/#Ticker might be more convenient here, but not a big deal.

api/agent/drivers/docker/docker.go
ericfode: It looked like loadDockerImages does not return a list of what it loaded.

api/agent/drivers/docker/docker.go
ericfode: I don't think that is supported in the dockerapi. Docker.LoadImage(...) err is the signature of the underlying function.

api/agent/agent.go
```diff
@@ -218,6 +218,7 @@ func NewDockerDriver(cfg *Config) (drivers.Driver, error) {
 		PreForkUseOnce:       cfg.PreForkUseOnce,
 		PreForkNetworks:      cfg.PreForkNetworks,
 		MaxTmpFsInodes:       cfg.MaxTmpFsInodes,
+		MaxImageCacheSize:    cfg.MaxImageCacheSize,
```
kmjohansen: If you're still on board with switching to an approach that looks at the amount of free space on a volume, would it be worth it to change this to a percentage of space free/in-use instead of using an absolute size?  This would allow you to compute the amount of free space needed without needing to rely on knowing the absolute size of the docker volume as a config parameter.

api/agent/drivers/docker/docker.go
```diff
@@ -54,16 +54,59 @@ func (r *runResult) Error() error   { return r.err }
...
 	networksLock sync.Mutex
 	networks     map[string]uint64
 }
 
+func NewImageCleaner(dockerDriver *DockerDriver, context context.Context) error {
+	opts := docker.RemoveImageOptions{}
+	opts.Force = true
+	opts.NoPrune = false
+	opts.Context = context
+	ticker := time.NewTicker(time.Minute * 5)
```
kmjohansen: Nit: would you please change the '5' to a const variable.  `ImageCleanScanMinutes = 5` or whatever is stylistically consistent for this code, perhaps?

api/agent/drivers/docker/docker.go
```diff
@@ -54,16 +54,59 @@ func (r *runResult) Error() error   { return r.err }
...
+	for {
+		select {
+		case <-context.Done():
+			{
+				return nil
+			}
+		case <-ticker.C:
+			{
+				if !dockerDriver.imageCache.OverFilled() {
```
kmjohansen: Another nit: OverFilled() is only currently used here, and it has a not flipping the condition.  Would you change the function name so that it reads instead of "not overfilled" to something like "space available"?

api/agent/drivers/docker/docker.go
```diff
@@ -102,13 +145,36 @@ func NewDocker(conf drivers.Config) *DockerDriver {
...
+	if conf.MaxImageCacheSize != 0 {
+		driver.imageCache = NewCache(int64(conf.MaxImageCacheSize))
+
+		go func() {
+			images, err := driver.docker.ListImages(context.Background())
+			if err != nil {
+				logrus.WithError(err).Fatalf("cannot list docker images %s", err)
+			}
+			for _, i := range images {
+				if Contains(imagesBeforeLoad, i) {
```
kmjohansen: is it possible to have `loadDockerImages()` return the id of the image that it loaded?  If so, would it make sense to simplify this check to just lock the loaded docker image, or are we expecting to have more than one here?

api/agent/drivers/docker/docker.go
```diff
@@ -102,13 +145,36 @@ func NewDocker(conf drivers.Config) *DockerDriver {
...
 		err = loadDockerImages(driver, conf.DockerLoadFile)
 		if err != nil {
 			logrus.WithError(err).Fatalf("cannot load docker images in %s", conf.DockerLoadFile)
 		}
 	}
 
+	if conf.MaxImageCacheSize != 0 {
+		driver.imageCache = NewCache(int64(conf.MaxImageCacheSize))
+
+		go func() {
```
kmjohansen: Do we ever care about when this function completes?  Read: wondering if there's a need to add a channel to wait for it, in case we need to ensure the cache has been populated.  Just makes my race-condition-spidey-sense tingle a bit, but that could also be allergies.

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,223 @@
...
+// NewEntry constructs a an entry from a docker d.APIImages object.
+func NewEntry(value d.APIImages) Entry {
+	return Entry{
+		lastUsed: time.Now(),
+		locked:   make(map[*interface{}]*interface{}),
+		uses:     0,
+		image:    value}
+}
+
+// New returns a new cache with the provided maximum items.
```
kojustin: doccomment doesn't match function name

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,223 @@
...
+
+func (c *Cache) unlock(ID string, key interface{}) {
+	for _, i := range c.cache {
+		if i.image.ID == ID {
+			delete(i.locked, &key)
+		}
+	}
+}
+
+// Add adds the provided key and value to the cache, evicting
```
kojustin: comment is stale, there is no key ?

kojustin: does this evict?

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,223 @@
...
+	if c.contains(value) {
+		c.mark(value.ID)
+		return
+	}
+	c.cache = append(c.cache, NewEntry(value))
+}
+
+// Checks the total size of the docker data dir.
+func (c *Cache) TotalSize() int64 {
+	return 0
```
kojustin: is this a TODO?

api/agent/drivers/docker/docker.go
```diff
@@ -101,13 +148,36 @@ func NewDocker(conf drivers.Config) *DockerDriver {
...
 		err = loadDockerImages(driver, conf.DockerLoadFile)
 		if err != nil {
 			logrus.WithError(err).Fatalf("cannot load docker images in %s", conf.DockerLoadFile)
 		}
 	}
 
+	if conf.MaxImageCacheSize != 0 {
+		driver.imageCache = NewCache()
+
+		go func(context context.Context) {
```
skinowski: I think there's a timing issue here. Since go-routine, this means, new images are being pulled as this is running... So, a user function can potentially be flagged as 'base image'

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,223 @@
...
+
+// Returns true if there is to much space consumed.
+func (c *Cache) OverFilled() bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return c.TotalSize() > c.maxSize
+}
+
+// Checks for evictable images ordered by score.
+func (c *Cache) Evictable() EntryByAge {
```
kojustin: EntryByAge is a slice type, are there any restrictions on the length here? Should this be `[]Entry`?

api/agent/drivers/docker/docker.go
```diff
@@ -101,13 +148,36 @@ func NewDocker(conf drivers.Config) *DockerDriver {
 		}
 	}
 
+	imagesBeforeLoad, err := driver.docker.ListImages(context.Background())
+
 	if conf.DockerLoadFile != "" {
 		err = loadDockerImages(driver, conf.DockerLoadFile)
```
skinowski: I think there's an output stream in args:

https://godoc.org/github.com/fsouza/go-dockerclient#LoadImageOptions

if this works, it's better than hitting the list images twice.

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,223 @@
...
+	for _, i := range c.cache {
+		if i.image.ID == ID {
+			i.locked[&key] = &key
+			return nil
+		}
+	}
+	return errors.New("Image not found in cache")
+}
+
+// Check to see if an image by id as any locks on it.
```
kojustin: typo: as -> has?

api/agent/drivers/docker/imageCache.go
ericfode: Its EntryByAre so it can be sorted.

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,223 @@
...
+	mu      sync.Mutex
+	cache   EntryByAge
+	maxSize int64
+}
+
+// Entry contains when the image was last used, how many times the image has been used,
+// The image metadata, and a set of tokens that are locking the image in question.
+type Entry struct {
+	lastUsed time.Time
+	locked   map[*interface{}]*interface{}
```
kojustin: I like this idea. 

It looks like you are not using the map key anywhere, could this be a slice or set?

api/agent/drivers/docker/docker.go
ericfode: We have to parse raw logs if we use that. Is that something we are interested in?

api/agent/drivers/docker/imageCache.go
kojustin: ok, i see what is going on here. Maybe the comment could be something like "Evictable returns all evictable image, ordered by score'

api/agent/drivers/docker/imageCache.go
ericfode: I wanted a set, but I could not find a built-in one. so I figured a hash where the key == value is roughly the same.

api/agent/drivers/docker/docker.go
```diff
@@ -53,16 +53,63 @@ func (r *runResult) Error() error   { return r.err }
...
+	opts := docker.RemoveImageOptions{}
+	duopts := docker.DiskUsageOptions{}
+	opts.Force = true
+	opts.NoPrune = false
+	opts.Context = context
+	ticker := time.NewTicker(dockerDriver.conf.ImageCacheCleanInterval)
+	for {
+		select {
+		case <-context.Done():
+			{
```
kojustin: is this gofmt brace style?

api/agent/drivers/docker/docker.go
skinowski: yeah, it doesn't look very structured. :-(

api/agent/drivers/docker/imageCache.go
skinowski: So the main logic just sweeps and removes images. The fact that the image list is sorted is not used, is it?

api/agent/drivers/docker/docker.go
```diff
@@ -53,16 +53,63 @@ func (r *runResult) Error() error   { return r.err }
...
+			}
+		case <-ticker.C:
+			{
+				du, err := dockerDriver.docker.DiskUsage(duopts)
+				if err != nil {
+					logrus.WithError(err).Error("attempting to check disk usage")
+				}
+				if du.LayersSize > maxSize {
+					toEvict := dockerDriver.imageCache.Evictable()
+					for _, i := range toEvict {
```
skinowski: This is a simple for-loop. Evict list sorted or not doesn't matter, right?

api/agent/drivers/docker/imageCache.go
ericfode: It is now. I forgot to add it after getting the disk usage API to work.

api/agent/drivers/docker/docker.go
ericfode: Some will probably fail. I want it to try the most likey to be useful evict first. As well as be able to abort the loop if enough space has been freed. On the other hand if its more complicated then it's worth we can totally delete it.

api/agent/drivers/docker/docker.go
```diff
@@ -53,17 +53,71 @@ func (r *runResult) Error() error   { return r.err }
...
 }
 
+// NewImageCleaner builds an evicter that loops every ImageCacheCleanInterval, checkes to see
+// if there is more space consumed then the maxSize allowed by configuration. If there is
+// NewImageCleaner checks the image cache for the list of evicitable images and then tries
+// in order of most evicitable to remove the image. If the disk consumption constraint is
+// satisfied the for loop breaks.
+func NewImageCleaner(context context.Context, dockerDriver *DockerDriver, maxSize int64) error {
+	opts := docker.RemoveImageOptions{}
+	duopts := docker.DiskUsageOptions{}
```
skinowski: we should set the ctx inside DiskUsageOptions.

ericfode:
> What is api/datastore/sql/sqlite_test_dir ? I think it's committed by accident.

You were correct. It is now deleted.

api/agent/config.go
```diff
@@ -122,10 +127,12 @@ const (
...
-		MinDockerVersion: "17.10.0-ce",
-		MaxLogSize:       1 * 1024 * 1024,
-		PreForkImage:     "busybox",
-		PreForkCmd:       "tail -f /dev/null",
+		MinDockerVersion:        "17.10.0-ce",
+		MaxLogSize:              1 * 1024 * 1024,
+		PreForkImage:            "busybox",
+		PreForkCmd:              "tail -f /dev/null",
+		MaxImageCacheSize:       0,
+		ImageCacheCleanInterval: 0,
```
skinowski: minor nit here, this is initialized to zero, which we don't need to do in go-lang. But below interval is defaulted to 5 mins. Could you pls clean it up?

docker-compose.yml
```diff
@@ -59,6 +59,7 @@ services:
       FN_MQ_URL: "redis://mq:6379/"
       FN_LOGSTORE_URL: "s3://admin:password@logstore:9000/us-east-1/fnlogs"
       FN_DOCKER_NETWORKS: "fn-network"
+      FN_MAX_IMAGE_CACHE_SIZE: "2000000"
```
skinowski: I'd leave docker compose as uninitialized (not enabled) by default. I think this is not needed.

api/agent/drivers/docker/docker.go
```diff
@@ -102,13 +164,41 @@ func NewDocker(conf drivers.Config) *DockerDriver {
...
+				if Contains(requiredImages, i) {
+					driver.imageCache.Add(i)
+				} else {
+					driver.imageCache.Add(i)
+					driver.imageCache.Lock(i.ID, "baseimage")
+				}
+			}
+		}(context.Background())
+
+		go NewImageCleaner(context.Background(), driver, int64(conf.MaxImageCacheSize))
```
skinowski: I think we should start this go-routine after the list processing above is completed, maybe at line 196?

api/agent/drivers/docker/docker.go
```diff
@@ -102,13 +164,41 @@ func NewDocker(conf drivers.Config) *DockerDriver {
...
+		go func(context context.Context) {
+			liopts := docker.ListImagesOptions{All: false}
+			liopts.Context = context
+			images, err := driver.docker.ListImages(liopts)
+
+			if err != nil {
+				logrus.WithError(err).Fatalf("cannot list docker images %s", err)
+			}
+			for _, i := range images {
+				if Contains(requiredImages, i) {
```
skinowski: I think this should be "not Contains"?

api/agent/drivers/docker/docker.go
```diff
@@ -102,13 +164,41 @@ func NewDocker(conf drivers.Config) *DockerDriver {
...
+	if conf.MaxImageCacheSize != 0 {
+		driver.imageCache = NewCache()
+
+		go func(context context.Context) {
+			liopts := docker.ListImagesOptions{All: false}
+			liopts.Context = context
+			images, err := driver.docker.ListImages(liopts)
+
+			if err != nil {
+				logrus.WithError(err).Fatalf("cannot list docker images %s", err)
```
skinowski: This is tricky... Maybe log and retry this? This step could be part of NewImageCleaner() too, it could retry for a while? If we crash here, we might crash in the middle of processing functions.

api/agent/drivers/docker/docker.go
```diff
@@ -51,17 +51,79 @@ func (r *runResult) Error() error   { return r.err }
...
 
+// NewImageCleaner builds an evicter that loops every ImageCacheCleanInterval, checkes to see
+// if there is more space consumed then the maxSize allowed by configuration. If there is
+// NewImageCleaner checks the image cache for the list of evicitable images and then tries
+// in order of most evicitable to remove the image. If the disk consumption constraint is
+// satisfied the for loop breaks.
+func NewImageCleaner(context context.Context, dockerDriver *DockerDriver, maxSize int64) error {
+	opts := docker.RemoveImageOptions{}
+	duopts := docker.DiskUsageOptions{}
+	opts.Force = true
```
skinowski: Given that we track in-use, do we still need Force=true?

api/agent/drivers/docker/docker.go
```diff
@@ -51,17 +51,79 @@ func (r *runResult) Error() error   { return r.err }
...
+		case <-context.Done():
+			return nil
+		case <-ticker.C:
+			du, err := dockerDriver.docker.DiskUsage(duopts)
+			if err != nil {
+				logrus.WithError(err).Error("attempting to check disk usage")
+			}
+			if du.LayersSize > maxSize {
+				toEvict := dockerDriver.imageCache.Evictable()
+				for _, i := range toEvict {
```
skinowski: is there a race here? We get list of evictable, but these could then be added back as not-evictable while we are attempting to remove them?

rdallman: yea, need to hold lock over removal, also need lock over all dependent parent images due to docker rm behavior, which may be tricky. eg where each letter is an image:

     a
   /   \
x      y

there's some ordering of events that goes like this: x is the only child of base image a, x is scheduled for eviction, a is now a node with no edges and docker can remove it, concurrently to that removal of x or a, lock is taken on y, which also is a child of a, a is removed while y is pulled, this results in container creation failing, which may result in user felt errors / added latency as well as possible weird states in the image eviction cache we need to handle. not clear how best to handle this, but it's the main tricky problem we need to solve here - as it is, locks are only held over function end images and not any layers it may depend on - this could be okay ultimately if we flesh out the behavior really well, but it's potentially tricky.

api/agent/drivers/docker/docker.go
rdallman: force may cause issues, particularly if for whatever reason an image is a dependency of another function's image - should remove this.

api/agent/drivers/docker/docker.go
rdallman: another thing that comes to mind is that it's possible that an agent initiates an image removal, agent panics/dies and restarts, then restarts and loads image into image cache and it ends up getting removed even while it has the lock held in its own process. this one is pretty whack, makes me think we should try to use all docker commands to manage without having our own locks too if we can, but it needs some reasoning about... if we retry container create on certain errors we may in the worst case have a container that fails a container create and retries, it's added latency but it's possibly correct without us having to manage our own lock process over our own logical docker image cache replica we're managing... need to think on it tho

api/agent/drivers/docker/docker.go
```diff
@@ -51,17 +51,79 @@ func (r *runResult) Error() error   { return r.err }
...
+					err := dockerDriver.docker.RemoveImage(i.image.ID, opts)
+					if err != nil {
+						logrus.WithError(err).Errorf("Could not remove image: %v because: %v", i, err)
+					} else {
+						dockerDriver.imageCache.Remove(i.image)
```
rdallman: a bit out there, but it's possible that this ends up in inconsistencies between images that exist on disk and the image caches list of images (ie we have a fugitive image) - eg docker can succeed in removing an image but have a network error telling us about it. then we have an image in the cache that isn't there. i don't think it's the end of the world, maybe there's a pathological case in there, but it's not easy to reason about either.

api/agent/drivers/docker/docker.go
```diff
@@ -102,13 +164,41 @@ func NewDocker(conf drivers.Config) *DockerDriver {
...
+			}
+			for _, i := range images {
+				if Contains(requiredImages, i) {
+					driver.imageCache.Add(i)
+				} else {
+					driver.imageCache.Add(i)
+					driver.imageCache.Lock(i.ID, "baseimage")
+				}
+			}
+		}(context.Background())
```
rdallman: any chance we can manage these 2 goroutines with the driver.Close() method? for testing this could get out of hand if we launch these but never shut them down. nitty but would be nice to extract this one to a separate function to call too out of NewDocker()

api/agent/drivers/docker/docker.go
rdallman: test would be great on this too

api/agent/drivers/docker/docker.go
rdallman: would be nice to fatalf in the same goroutine as NewDocker if we're going to do it (and I agree that we should in this case probably, it probably means docker is dead or misconfigured)

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,210 @@
...
+	uses     int64
+	image    d.APIImages
+}
+
+// The score is the time since last use divided by the number of total uses.
```
rdallman: comments on exported methods should start with the name of the method (and matching casing), eg `Score is the time since...` to make lint happy (I appreciate the comments here a lot)

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,210 @@
...
+
+package docker
+
+import (
+	"errors"
+	"sort"
+	"sync"
+	"time"
+
+	d "github.com/fsouza/go-dockerclient"
```
rdallman: s/d/docker/

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// Cache is a list of Entries. Stored as type EntryByAge so that sort works.
+// The mutex is present to control access access the list of images and their locks
+// The Cache may be used in concurrent contexts
+type Cache struct {
```
rdallman: probably want to either make all of these objects and methods private or make a package off of the drivers package with this stuff like `drivers/lrucache` where we could theoretically lru cache any driver's images (i don't want to think about pluggable driver image remove methods really at this hour though, up to you)

api/agent/drivers/docker/imageCache.go
```diff
@@ -0,0 +1,210 @@
...
+	sort.Sort(ea)
+	return ea
+}
+
+// Len returns the number of items in the cache.
+func (c *Cache) Len() int {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	return len(c.cache)
+}
```
rdallman: unit tests please for this cache :) lots :)

api/agent/drivers/docker/docker.go
ericfode: There is already a in the ListImages function, should we retry the retries?

skinowski:
closing this, we'll start fresh with a new PR.

PR #1176, opened 2018-08-18
proposal: attempt a cohesive cloud events plan

rdallman:
there are very few TODOs, but some. mostly, there are concrete suggestions for
them. to date, this plan exists in old docs in closed down issues that are now
stale, so this is a new plan with a smaller delta that we can follow, with an
attempt to approximate the actual work involved at the bottom.

this is trying to have the simplest possible cloud event format, without us
modifying the things into oblivion (only http trigger does this as proposed)

it seems kinda reasonable if we are willing to commit, so maybe we should
mostly talk about the commitment part, technically speaking there's not a lot
to these things. i have little success in extracating these kinds of things
into an english document, so we'll see how this goes...

### DIFF
--- a/docs/developers/proposal-ce-format.md
+++ b/docs/developers/proposal-ce-format.md
@@ -1,0 +1,168 @@
+# simple cloud-event container model
+
+This aims to be as simple as possible, the basic idea being that we support a
+very minimally modified default cloud event 0.1 format. Honestly, it's not a
+lot different than what we have already, this just writes it down somewhere so
+that we can bicker over the finer points on github to act as proof of work for
+the latest and greatest cryptocurrency in addition to the powers that be.
+
+## json format:
+
+TODO idk where to put this? here for illustrative purposes...
+
+```
+{
+  "eventType": "fn/http",
+  "cloudEventsVersion": "0.1",
+  "eventID": "6480da1a-5028-4301-acc3-fbae628207b3",
+  "source": "http://example.com/repomanager",
+  "eventTypeVersion": "1.0",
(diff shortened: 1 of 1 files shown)


mantree:
My main thought in this area is we don't want to support "structured" cloud events, nor any format that involves marshalling the request body.

I'm thinking the call delivery model should be much closer to a proxy, using metadata (either from http headers, or preconfigured options with regard to event consumption) to route the request. The less the platform has to do with the request the better.

### support binary by using a cloud event in a request over a socket

Seems like the way to go, if we're going to cause a wide-ranging change let's do it to move to http-server model, which has lots of nice benefits.

rdallman:
adding streaming presents issues at a few levels, one being limits and the other being gluing the lb/runner/etc, the current lb_agent won't take well to streaming since we need to buffer for retries, it opens a big can of worms. we've tried to support streaming in the past, out of the box even, but ended up having to buffer for various reasons -- that was before grpc lb_agent even, which makes things slightly more complex. i'm not sure that going to cloud events precludes us from having something streamable, their http transport in unstructured mode is streamable, and if we switch to a port on the container we could in theory stream it all the way down and back out too. that being said, for e.g. invoke w/ debug we'd need a structured return, and we could support structured or unstructured for the http trigger endpoint maybe, it's almost like it'd be a 'cloudevent http trigger' endpoint for unstructured, where on invoke we could do similar just not an open ended http request like it is now -- invoke would immediately be a structured or unstructured cloud event (but idk about with `debug` mode added)

PR #1177, opened 2018-08-20
Add annotation to trigger on create if endpoints are enabled

### LINKED ISSUE #1178: Triggers and FNs do not get annotated on Create(Only get and list.)
**Description**
Trigger and FN annotators are not used on Create.
1. Create a trigger of fn... the response does not have an annotation with the endpoint when it should.

No annotation

Have an annotation on create.

hibooboo2:
- What I did
Add trigger annotation so that you do not have to get the trigger to have its http endpoint

- How I did it
Called the annotator from trigger create

- How to verify it
Create a trigger and check annotations for the http endpoint

Annotate httpendpoint to trigger on create

https://giphy.com/embed/CgfteGy1ZPYPu

Fixes #1178

### DIFF
--- a/api/server/fns_create.go
+++ b/api/server/fns_create.go
@@ -1,15 +1,16 @@
 package server
 
 import (
-	"fmt"
 	"net/http"
 
+	"github.com/fnproject/fn/api/common"
 	"github.com/fnproject/fn/api/models"
 	"github.com/gin-gonic/gin"
 )
 
 func (s *Server) handleFnCreate(c *gin.Context) {
 	ctx := c.Request.Context()
+	log := common.Logger(ctx)
 
 	fn := &models.Fn{}
 	err := c.BindJSON(fn)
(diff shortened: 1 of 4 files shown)


ericfode:
Is there a test we can add for this?

hibooboo2:
I am looking into it right now.. I think I can.

api/server/fns_create.go
```diff
@@ -26,5 +27,17 @@ func (s *Server) handleFnCreate(c *gin.Context) {
 		handleErrorResponse(c, err)
 	}
 
+	app, err := s.datastore.GetAppByID(ctx, fnCreated.AppID)
```
srpurani: Why are we fetching it again from the database?

hibooboo2: We do not have the app.. Just the appID and annotators take the full app.

hibooboo2: As to if we can make the annotators not require full app that is a different question that I have no clue as to the answer for.

rdallman: it's okay for this endpoint to look up the app after create, though it would be nice if we made sure that if the resource gets created we return 202/200 instead of error'ing for e.g. app lookup/trigger (for GET this is less of an issue since it's a read-only endpoint)

srpurani: I was wondering if we can avoid another lookup of app in the same request flow, perfectly fine if not.

rdallman: not really without changing the annotator contract, which maybe should be more flexible (punt), but right now it's not. this endpoint is pretty low throughput and not necessarily latency sensitive at that level though, so it's ok I think

hibooboo2: Ill fix the resp codes. @rdallman

gviedma:
Looks good, this will be very useful.

PR #1179, opened 2018-08-22
disable user function logs at debug level configuration

### LINKED ISSUE #1147: ability to toggle function logs if FN_LOG_LEVEL=debug
at the moment, we are using `FN_LOG_LEVEL=debug` to spill all function logs to stderr in addition to any fn logs that are debug level. while this can be useful in certain cases, there will come a day where a service provider may want to enable debug logs for a particular fn node but not want the user function logs to be spilled to stderr. thus, we should have some kind of toggle:

`FN_LOG_FUNCTIONS=true` - an env var that allows function logs to be spat to stderr [iff FN_LOG_LEVEL=debug]

we can set this to default on (as now) or off with the ability to turn it on. for the sake of making a decision, let's say we should turn this flag off by default. this should be relatively easy as well, it's mostly plumbing a config option through the agent.

rdallman:
- Link to issue this resolves

closes #1147 

- What I did

added an env var FN_DISABLE_DEBUG_USER_LOGS to keep user logs from being logged at debug level. whether they're being logged to stderr is dependent on log level, even if it's info this will be a speedup, and if the level is in debug, it's likely a production installation will not want to be able to access these (yet still want debug logs for the system). 

- How to verify it

start it up

add ability to disable user function logs to debug logs

### DIFF
--- a/api/agent/agent_test.go
+++ b/api/agent/agent_test.go
@@ -342,7 +342,7 @@ func TestLoggerIsStringerAndWorks(t *testing.T) {
 	// TODO test limit writer, logrus writer, etc etc
 
 	var call models.Call
-	logger := setupLogger(context.Background(), 1*1024*1024, &call)
+	logger := setupLogger(context.Background(), 1*1024*1024, true, &call)
 
 	if _, ok := logger.(fmt.Stringer); !ok {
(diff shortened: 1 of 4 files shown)


rdallman:
`make test` no longer works for me and it's a sadness I'm having trouble confronting :(

PR #1181, opened 2018-08-22
#1180 Provides pluggable option to modify HTTP headers of a function call

### LINKED ISSUE #1180: Provide pluggable option to modify HTTP response from calling a function
Currently, when a function executes, it's output is buffered and eventually committed to the HTTP response returned to the caller. There is no possibly of intercepting the response to alter it via middleware or otherwise, since the response is committed by the time the response is handled by the agent.

This functionality is required in enterprise environments. For example, when running a multi-tenant service with DNS subdomains partitioned by app, we need to disallow setting of cookies by functions at the top level DNS domain. Performing such sanitization of the HTTP response is currently not possible via an extension mechanism.

gviedma:
- Link to issue this resolves
#1180 

- What I did
I added a server option that provides access to headers so they can be added/removed prior to committing a response. See `WithFunctionResponseModifier`.

- How to verify it
I added tests that use the above option for triggers, fn invoke and route calls.

Provides pluggable option to modify HTTP headers of a function call

### DIFF
--- a/api/models/error.go
+++ b/api/models/error.go
@@ -206,10 +206,13 @@ var (
 		code:  http.StatusBadRequest,
 		error: fmt.Errorf("Invalid annotation change, new key(s) exceed maximum permitted number of annotations keys (%d)", maxAnnotationsKeys),
 	}
-
 	ErrAsyncUnsupported = err{
 		code:  http.StatusBadRequest,
 		error: errors.New("Async functions are not supported on this server"),
+	}
+	ErrCopyResponseFromBufferFailed = err{
+		code:  http.StatusInternalServerError,
+		error: errors.New("Failed to process buffered response"),
 	}
 )
 
--- a/api/server/runner.go
+++ b/api/server/runner.go
@@ -123,28 +123,7 @@ func (s *Server) ServeRoute(c *gin.Context, app *models.App, route *models.Route
 		return err
 	}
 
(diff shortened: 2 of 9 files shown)


zootalures:
I don't think this is necessary unless you are planning to modify the response body  you can modify outbound headers by subverting the responsewriter within a middleware with one that messes with headers when they are committed: 

```
// CookieMesser does something to headers when the response gets committed
type CookieMesser struct {
	sub http.ResponseWriter
	committed bool
}

func (cm *CookieMesser) Header() http.Header{
	return cm.sub.Header()
}

func (cm *CookieMesser) Write(b []byte) (int,error){
	if !cm.committed {
		cm.WriteHeader(http.StatusOK)
	}
	return cm.sub.Write(b)
}

func (cm *CookieMesser) WriteHeader(sc int){
	cm.committed = true
	cm.sub.Header().Del("Cookie")
	cm.sub.WriteHeader(sc)
}
....
	funcServer.AddRootMiddlewareFunc(func(next http.Handler) http.Handler {
		return http.HandlerFunc(func(rw http.ResponseWriter,req *http.Request){
			next.ServeHTTP(&CookieMesser{sub:rw},req)
		})
	})
....
```

gviedma:
Closing since this can be done with root middleware as described in the comments.

PR #1189, opened 2018-08-28
fn: introducing docker-syslog driver as default logger

skinowski:
With this change, fn-agent prefers docker-syslog driver
for logging stdout/stderr from containers. The advantage
of this is to offload it to docker itself instead of
streaming stderr along with stdout, which gets multiplexed
through single connection via docker-API.

The change will need support from FDKs in order to log
correct call-id and supress '\n' that splits syslog lines.


skinowski:
Before:

```
Aug 27 21:27:01 fnserver 2 2018-08-27T21:27:01Z fn - - - - call_id=01CNYMGSZ1NGA000RZJ0000003 func_name=/fn-http-func app_name=fn-http-func 
2018/08/27 21:27:01 Received format http
Aug 27 21:27:01 fnserver 2 2018-08-27T21:27:01Z fn - - - - call_id=01CNYMGSZ1NGA000RZJ0000003 func_name=/fn-http-func app_name=fn-http-func 
2018/08/27 21:27:01 Received request main.AppRequest{SleepTime:0, ResponseCode:200, ResponseContentType:"", JasonContentType:"", EchoContent
:"hello", IsDebug:true, IsCrash:false, ReadFile:"", ReadFileSize:0, CreateFile:"", CreateFileSize:0, AllocateMemory:0, LeakMemory:0, Respons
eSize:0, InvalidResponse:false}
Aug 27 21:27:01 fnserver 2 2018-08-27T21:27:01Z fn - - - - call_id=01CNYMGSZ1NGA000RZJ0000003 func_name=/fn-http-func app_name=fn-http-func 
2018/08/27 21:27:01 Received headers map[Fn_call_id:[01CNYMGSZ1NGA000RZJ0000003] Fn_method:[POST] Fn_request_url:[http://localhost:8080/r/fn
-http-func/fn-http-func] Content-Type:[application/json] User-Agent:[curl/7.61.0] Content-Length:[79] Accept:[*/*] Fn_deadline:[2018-08-27T2
1:27:11.857Z]]
Aug 27 21:27:01 fnserver 2 2018-08-27T21:27:01Z fn - - - - call_id=01CNYMGSZ1NGA000RZJ0000003 func_name=/fn-http-func app_name=fn-http-func 
2018/08/27 21:27:01 Received config map[PATH:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin HOSTNAME:05b75bd86a2f FN_MEMORY:12
8 FN_TMPSIZE:0 FN_FORMAT:http HOME:/root FN_TYPE:sync FN_APP_NAME:fn-http-func FN_PATH:/fn-http-func]
```

After:
```
Aug 28 00:46:19 fnserver func_name=/fn-http-func,app_name=fn-http-func,e97dd9aff479[17]: 2018/08/28 00:46:19 Received format http
Aug 28 00:46:19 fnserver func_name=/fn-http-func,app_name=fn-http-func,e97dd9aff479[17]: 2018/08/28 00:46:19 Received request main.AppRequest{SleepTime:0, ResponseCode:200, ResponseContentType:"", JasonContentType:"", EchoContent:"hello", IsDebug:true, IsCrash:false, ReadFile:"", ReadFileSize:0, CreateFile:"", CreateFileSize:0, AllocateMemory:0, LeakMemory:0, ResponseSize:0, InvalidResponse:false}
Aug 28 00:46:19 fnserver func_name=/fn-http-func,app_name=fn-http-func,e97dd9aff479[17]: 2018/08/28 00:46:19 Received headers map[Fn_request_url:[http://localhost:8080/r/fn-http-func/fn-http-func] User-Agent:[curl/7.61.0] Content-Length:[79] Fn_call_id:[01CNYZXQMYNGA000RZJ0000003] Fn_deadline:[2018-08-28T00:46:29.806Z] Fn_method:[POST] Accept:[*/*] Content-Type:[application/json]]
Aug 28 00:46:19 fnserver func_name=/fn-http-func,app_name=fn-http-func,e97dd9aff479[17]: 2018/08/28 00:46:19 Received config map[PATH:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin FN_TMPSIZE:0 FN_FORMAT:http HOME:/root HOSTNAME:f4931fa91618 FN_APP_NAME:fn-http-func FN_PATH:/fn-http-func FN_MEMORY:128 FN_TYPE:sync]
Aug 28 00:46:19 fnserver func_name=/fn-http-func,app_name=fn-http-func,e97dd9aff479[17]: HTTP/1.1 200 OK
Aug 28 00:46:19 fnserver func_name=/fn-http-func,app_name=fn-http-func,e97dd9aff479[17]: Content-Length: 624
Aug 28 00:46:19 fnserver func_name=/fn-http-func,app_name=fn-http-func,e97dd9aff479[17]: 
Aug 28 00:46:19 fnserver func_name=/fn-http-func,app_name=fn-http-func,e97dd9aff479[17]: {"request":{"responseCode":200,"echoContent":"hello","isDebug":true},"header":{"Accept":["*/*"],"Content-Length":["79"],"Content-Type":["application/json"],"Fn_call_id":["01CNYZXQMYNGA000RZJ0000003"],"Fn_deadline":["2018-08-28T00:46:29.806Z"],"Fn_method":["POST"],"Fn_request_url":["http://localhost:8080/r/fn-http-func/fn-http-func"],"User-Agent":["curl/7.61.0"]},"config":{"FN_APP_NAME":"fn-http-func","FN_FORMAT":"http","FN_MEMORY":"128","FN_PATH":"/fn-http-func","FN_TMPSIZE":"0","FN_TYPE":"sync","HOME":"/root","HOSTNAME":"f4931fa91618","PATH":"/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"},"data":{}}
```

skinowski:
@rdallman @zootalures @kmjohansen  Before I spend more time on this PR, I'd like to ask an early review of this approach. This PR is a POC for now.

In summary, docker-syslog works. However it seems to log both stdout/stderr. As we discussed, FDK will still need to handle call_id addition to the logging and perhaps it should handle supressing '\n' from output.

stdout/stderr merging may be a showstopper for this. The above before/after is from an http function, and you can see that http output is also logged.

Do we still want to proceed with this?

kmjohansen:
@skinowski I'm generally okay with this approach modulo a few nits.  At a higher level, if we want to enable people to drop daemons into containers, it's not clear whether we can expose an _actual_ syslog connection on their behalf.  In that scenario, I would have expected the application to have closed stdin, stdout, stderr along the lines of a `daemon(3)` syscall: https://www.freebsd.org/cgi/man.cgi?query=daemon&sektion=3&apropos=0&manpath=FreeBSD+11.2-RELEASE+and+Ports

In terms of stdout/stderr merging: it's probably not ideal, but a lot of other system utilities seem to default to this behavior on linux distros that use systemd.  It's standard/default setting is to merge both stdout and stderr and then spew that to journald: https://www.freedesktop.org/software/systemd/man/systemd.exec.html#StandardError=

rdallman:
> perhaps it should handle supressing '\n' from output.

this means just to aggregate lines until the fdk sees `\n` before sending, right? this could be okay to ignore, even if it looks weird in papertrail, et al. we gotta keep fdk contract minimal (despite all efforts to the contrary), and users can learn to log lines. as long as it's not doing anything weird. 

the stdout thing seems ok too, it's not that big of a deal and makes sense w/ socket io - as long as we still get stdout, and tests seem to pass. 

also think we should remove the other syslog code and not feature flag this, it's in git we can find it.

for future context: the real reason we're doing this is because we're attenuating network permissions to fn's container, while the container running dockerd that fn talks to has the network permissions necessary to send to a user's container.

api/agent/drivers/docker/cookie.go
```diff
@@ -25,6 +26,56 @@ type cookie struct {
...
+
+	// Let's stream logs back to agent if docker syslog is disabled.
+	if !c.drv.conf.EnableDockerSyslog {
+		c.opts.Config.AttachStderr = true
+		return
+	}
+
+	conf := c.task.LoggerConfig()
+
+	// If docker-syslog is enabled and Task has no URL, we discard logs.
```
rdallman: this seems wrong, too... we still want to get stderr out right now for storing logs (locally) and eventually for debug mode we'll likely want the same? doesn't seem like the right place to be doing this, the creator of the cookie should pick this stuff

api/agent/agent.go
```diff
@@ -1051,11 +1065,17 @@ func newHotContainer(ctx context.Context, call *call, cfg *Config) (*container,
 	stderr := common.NewGhostWriter()
 	stdout := common.NewGhostWriter()
 
-	// these are only the conns, this doesn't write the syslog format (since it will change between calls)
-	syslogConns, err := syslogConns(ctx, call.SyslogURL)
-	if err != nil {
-		// TODO we could write this to between stderr but between stderr doesn't go to user either. kill me.
-		common.Logger(ctx).WithError(err).WithFields(logrus.Fields{"app_id": call.AppID, "path": call.Path, "image": call.Image, "container_id": id}).Error("error dialing syslog urls")
+	var syslogCon io.WriteCloser
+	if !cfg.EnableDockerSyslog {
```
kmjohansen: Same question about the conditional on `EnableDockerSyslog` here too.  Should this skip the not?  Am I misreading?

api/agent/config.go
```diff
@@ -143,6 +147,9 @@ func NewConfig() (*Config, error) {
 	if _, ok := os.LookupEnv(EnvDisableDebugUserLogs); ok {
 		cfg.DisableDebugUserLogs = true
 	}
+	if _, ok := os.LookupEnv(EnvEnableDockerSyslog); ok {
+		cfg.EnableDockerSyslog = true
```
kmjohansen: This will enable syslog whenever the environment variable is present, regardless of its value.  Should this code check that the value in the env variable evaluates to something truthy?

skinowski:
@rdallman @kmjohansen This is a bit more cleaned up now.

api/agent/drivers/driver.go
```diff
@@ -86,6 +86,18 @@ type RunResult interface {
...
+// Logger Configuration for container
+type LoggerConfig struct {
+	// Log Sink URL
+	URL string
+
+	// Application Name for Log Tags
+	AppName string
+
+	// Function Name for Log Tags
+	FuncName string
```
rdallman: the driver package itself shouldn't have any knowledge of functions or apps, it only speaks in containers. could we change `AppName` and `FuncName` to `Tag string` as it's represented in the syslog config and pass in the string tag we'd like from the caller rather than configuring it from these variables? this is more flexible to future changes, too.

rdallman: i guess it's a little awkward with the `{{.Id}}` that the caller doesn't know about either. maybe the contract should just be to prepend whatever is provided ?

rdallman: upon further thought, should we be exposing the container id to the syslog logs? it makes sense maybe for us to be able to relate calls to a certain container for debugging stuff on the operational side, but do we want to expose to users this kind of information? most platforms seem to hide and google cloud functions even successfully masks it from users completely. for most users, they likely won't care at all though in the general case and the container id is just wasted bytes. it's also likely to spark questions (maybe with or without) around the nature of 'why did this function run here and this one there' (weak arg here)

skinowski: Maybe not {{.ID}}, but I think we need some kind of hot-container id, which is helpful to search in logs. We are expecting call-id to be added to these logs by FDKs, but call-id does not cover all (between calls, etc.). I could expose agent container id.

rdallman: in the hot container startup iirc we're logging to fn's stderr the container id + call id so that we can associate operationally, agree that this is useful. for user syslogs is this useful / should they care / should we hide it is kinda the open question

skinowski: I'm thinking like UDP syslog sink and interleaved logs. Maybe expose something like ctx_id and use agent container id for its value? Especially stdout/stderr with new lines. Or we can assume that FDK will tag everything (stdout+stderr) with a call_id?

skinowski:
@rdallman @kmjohansen minor changes. (removed container id and switched to RFC2454 and updated docs.)

PR #1191, opened 2018-08-28
fn: add http.Server options for web/admin/grpc services in server

skinowski:
This allows to time limit slow/malicious clients when
reading HTTP headers. In GetBody() buffering, same timeout
can be used to time limit to give consistent I/O wait
limits for the service in addition to per handler
imposed limits we already have.

### DIFF
--- a/api/server/gin_middlewares.go
+++ b/api/server/gin_middlewares.go
@@ -155,7 +155,7 @@ func apiMetricsWrap(s *Server) {
 
 	r := s.Router
 	r.Use(measure(r))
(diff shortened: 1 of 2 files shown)


rdallman:
thanks... I wonder if we should just have `WithServer(*http.Server) Option` ? i've grown to hate this functional config thing, i imagine there are quite a few timeouts and things we need to set and this might get tedious. read/write timeout still need setting, too, for example

reclaro:
I am wonder if we should document that this setting is available somewhere.

PR #1193, opened 2018-09-03
WIP: Copy 'logs' and 'calls' across from swagger v1 to swagger v2.

JadeRedworth:
Copied across the 'logs' and 'calls' from swagger v1 to swagger v2 so they are still supported when we remove swagger v1.

### DIFF
--- a/docs/swagger_v2.yml
+++ b/docs/swagger_v2.yml
@@ -145,6 +145,75 @@ paths:
           schema:
             $ref: '#/definitions/Error'
 
+  /log:
+    get:
+      operationId: "GetCallLogs"
+      summary: "Get Call Logs For An Application."
+      description: "Returns the call logs for a specific Call."
+      tags:
+        - Call
+        - Log
+      parameters:
+        - $ref: '#/parameters/CallIDQuery'
+      responses:
+        200:
+          description: Log found
+          schema:
+            $ref:  '#/definitions/Log'
+        404:
+          description: Log not found.
+          schema:
+            $ref: '#/definitions/Error'
+
+  /calls:
+    get:
+      summary: Get app-bound calls.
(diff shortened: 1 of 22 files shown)


docs/swagger_v2.yml
```diff
@@ -591,6 +695,141 @@ definitions:
         type: string
         readOnly: true
 
+  CallsWrapper:
```
riconnon: Should we remove the wrappers to align with other v2 API semantics?

docs/swagger_v2.yml
```diff
@@ -145,6 +145,110 @@ paths:
...
+        200:
+          description: Call found
+          schema:
+            $ref:  '#/definitions/CallWrapper'
+        404:
+          description: Call not found.
+          schema:
+            $ref: '#/definitions/Error'
+
+  /apps/{app}/calls:
```
riconnon: In v2 we had a flat API hierarchy.
To align with this we'd have `/calls` with a mandatory query parameter of `app` rather than `/apps/{app}/calls`

riconnon:
Additionally... this MR copies the swagger definitions but doesn't actually make the functionality available under the `/v2` route.

zootalures:
Good start - but I think we ought to do the legwork to bring Calls and Logs into the functions world: 

so: 
the data model would be : 

* app
   * function (many)
       * trigger (many)
       *  call  (many)
           *  log  (one) 

Suggested changes: 
* List calls only by fn_id
* List logs only by call_id
* Add fn_id to call 
* Remove/deprecate `path` from call


This brings the API close to the event data changes that we started a few months ago -

JadeRedworth:
@zootalures sure, sounds good!

mantree:
Has been superseeded by #1207

PR #1194, opened 2018-09-04
Gracefully handles client request cancelations, instead of treating treating them as server errors

gviedma:
Currently, if a client closes an HTTP connection to the server in the middle of a request, this results in the server logging and attempting to respond with an HTTP 500 error. Instead, we should log that the request was canceled and stop further processing.

### DIFF
--- a/api/agent/runner_client.go
+++ b/api/agent/runner_client.go
@@ -358,7 +358,7 @@ DataLoop:
 			break DataLoop
 
 		default:
-			log.Error("Ignoring unknown message type %T from runner, possible client/server mismatch", body)
+			log.Errorf("Ignoring unknown message type %T from runner, possible client/server mismatch", body)
 		}
 	}
 
--- a/api/agent/runner_client.go
+++ b/api/agent/runner_client.go
@@ -192,8 +192,8 @@ func (r *gRPCRunner) TryExec(ctx context.Context, call pool.RunnerCall) (bool, e
 
 	recvDone := make(chan error, 1)
 
-	go receiveFromRunner(ctx, runnerConnection, call, recvDone)
-	go sendToRunner(ctx, runnerConnection, call)
+	go receiveFromRunner(ctx, runnerConnection, r.address, call, recvDone)
+	go sendToRunner(ctx, runnerConnection, r.address, call)
 
 	select {
 	case <-ctx.Done():
@@ -208,11 +208,11 @@ func (r *gRPCRunner) TryExec(ctx context.Context, call pool.RunnerCall) (bool, e
 	}
 }
 
-func sendToRunner(ctx context.Context, protocolClient pb.RunnerProtocol_EngageClient, call pool.RunnerCall) {
+func sendToRunner(ctx context.Context, protocolClient pb.RunnerProtocol_EngageClient, runnerAddress string, call pool.RunnerCall) {
 	bodyReader := call.RequestBody()
 	writeBuffer := make([]byte, MaxDataChunk)
 
-	log := common.Logger(ctx)
+	log := common.Logger(ctx).WithField("runner_addr", runnerAddress)
 	// IMPORTANT: IO Read below can fail in multiple go-routine cases (in retry
 	// case especially if receiveFromRunner go-routine receives a NACK while sendToRunner is
 	// already blocked on a read) or in the case of reading the http body multiple times (retries.)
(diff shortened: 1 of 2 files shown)


api/server/error_response.go
```diff
@@ -28,6 +28,11 @@ func simpleError(err error) *models.Error {
 // TODO delete me !
 func handleV1ErrorResponse(ctx *gin.Context, err error) {
 	log := common.Logger(ctx)
+	if ctx.Err() == context.Canceled {
```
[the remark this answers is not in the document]
gviedma: I don't think so. The contract for an http request context on the server side is that it will be closed when the client closes the connection.

```
// For incoming server requests, the context is canceled when the
  // client's connection closes, the request is canceled (with HTTP/2),
  // or when the ServeHTTP method returns
```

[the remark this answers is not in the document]
gviedma: I've updated the check to use ctx.Request.Context() instead of the gin.Context, so we should be good now.

api/server/error_response.go
```diff
@@ -63,9 +68,14 @@ func writeV1Error(ctx context.Context, w http.ResponseWriter, statuscode int, er
...
 }
 
-// handleV1ErrorResponse used to handle response errors in the same way.
+// HandleErrorResponse used to handle response errors in the same way.
 func HandleErrorResponse(ctx context.Context, w http.ResponseWriter, err error) {
 	log := common.Logger(ctx)
+	if ctx.Err() == context.Canceled {
+		log.Info("request canceled")
+		return
+	}
```
rdallman: bit nitty: for either of the error response handlers, might we move this down to just preceding `WriteError`, which should also log the error? this keeps control flow simpler for logging when we could get a canceled context + some other error that may explain slightly better how the cancellation happened, and no early return to worry about. i.e. last statement for this function:

```go
if ctx.Err() != context.Canceled {
  WriteError(ctx, w, statusCode, err)
}
```

I guess this means we were getting an EOF from writing the response back? it seems like this would also induce an EOF on the client end itself if we did get a cancellation for a reason other than the client hanging up? I'm a bit curious about the root cause of the EOF anyway, I'm not sure us skipping writing back a 500 to the client is going to resolve the EOF?

api/agent/runner_client.go
```diff
@@ -208,11 +208,11 @@ func (r *gRPCRunner) TryExec(ctx context.Context, call pool.RunnerCall) (bool, e
 	}
 }
 
-func sendToRunner(ctx context.Context, protocolClient pb.RunnerProtocol_EngageClient, call pool.RunnerCall) {
+func sendToRunner(ctx context.Context, protocolClient pb.RunnerProtocol_EngageClient, runnerAddress string, call pool.RunnerCall) {
 	bodyReader := call.RequestBody()
 	writeBuffer := make([]byte, MaxDataChunk)
 
-	log := common.Logger(ctx)
+	log := common.Logger(ctx).WithField("runner_addr", runnerAddress)
```
rdallman: i guess for a not-well-configured install this is necessary? most logging systems will give this to us tho just by way of sending the logs in. I suspect it would be nice to hang this off of the root logger somewhere in the server package with something like hostname to get this for all logs? can slide it in for debugging I guess, it just seems like it will be a duplicate at some point (maybe not?)

api/server/error_response.go
rdallman: (with a comment about this, preferably)

api/agent/runner_client.go
gviedma: I'm not sure that will work. These lines are being logged on the lb nodes (grpc client), not the runner nodes (grpc server) so the hostname/ip address would always be the same and correspond to the lb. I'd like us to log what runner we are connected to rather than what host is acting as client here.

rdallman: ah, that makes sense, thanks for explanation

api/server/error_response.go
gviedma: I don't see another reason that we would have a request context being canceled other than the client aborted the request. This fix is not about fixing EOFs on the client side (those are most likely caused by proxies or LBs timing out in front of the server and beyond the current scope). The intention with the fix is to stop us from logging errors/pushing http code stats for client disconnects as if we had an internal server error, which in this case it wasn't.

gviedma: I agree on the nit. Will fix it tomorrow.

gviedma: +1 on adding stats for disconnects as well. This is going to be the first step, just fixing our logs. More PRs will follow to extend our stats :)

gviedma: Actually, going over the nit, I realized that it won't work. I want to bypass the line 
```
		log.WithError(err).WithFields(logrus.Fields{"stack": string(debug.Stack())}).Error("internal server error")
```
to avoid logging client disconnects as internal server errors. So I think early return is necessary I'm afraid. Unless you have other suggestions?

rdallman: ah, gotcha. i was advocating for that i guess. not sure i understand, i guess the error will show 'canceled' but the log message is still internal server error which is misleading since the client hung up and it's not the server's fault? this just slightly prunes the logs to make it not look like the server is on fire so that when 'internal server error' is getting rattled off we know it's actually something? i buy that. carry on

gviedma: Yep, that's it :)

PR #1195, opened 2018-09-04
use tini to run every container

### LINKED ISSUE #1101: Use a process supervisor on all functions by default
**Description**

Currently we invoke functions directly via the equivalent of docker run,  this invokes the entrypoint as PID 1 within the container. 

This can cause problems for languages/code that do not set up the appropriate signal handling for a pid1 process which in turn can cause zombie processes to leak, eventually exhausting the resources of the sever. 

https://hackernoon.com/my-process-became-pid-1-and-now-signals-behave-strangely-b05c52cc551c?gi=9eddcdf48d16

Not tested directly but certain that this could catch people out when using many FDKs (python, node) or using arbitrary binaries. 

Spawning child processes via exec will not be reaped, resulting in zombies

Child processes are reaped. 

**What to do** ? 

I think a simple solution here is to always run functions with the default `docker-init` wrapper  in all environments, this provides consistency   of experience and provides the most protection against accidental process leaks. 

As of recent versions of docker this binary is installed as part of the docker distribution

within the docker agent we'd need to add InitBinary calls to the container construction call, sadly this also means reving our docker API dependency.

rdallman:
fixes #1101

additional context:

* this was introduced in docker 1.13 (1/2017), we require docker 17.10
(10/2017), this should not have any issues dependency-wise, as `docker-init`
is in the docker install from that point in time. unless explicitly removed,
it should be in the dind container we use as well...
* the PR that introduced this to docker is
https://github.com/moby/moby/pull/26061 for additional context
* it may be wise to put this through some paces, if anybody has any...
interesting... function containers. the tests seem to work fine, however, and
this shouldn't be something users have to think about (?) at all, just
something that we are doing. this isn't the default in docker for
compatibility reasons, which is maybe a yellow flag but I am not sure tbh
* i've defaulted this to 'on' with the config bit to turn it off, in case it runs into
issues it should be easy to flip it back off without having to deal with a build, etc.
__we need to test on osx and win__ and we can default it to 'off' and turn it on
for any 'serious' deployments if there are issues to keep the out of the box xp working,
it does seem to work on osx (per linked issue above), but i haven't tested that or win myself.
it seems to add about 50ms to the start time, which seems worth it.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -219,6 +219,7 @@ func NewDockerDriver(cfg *Config) (drivers.Driver, error) {
 		PreForkNetworks:      cfg.PreForkNetworks,
 		MaxTmpFsInodes:       cfg.MaxTmpFsInodes,
 		EnableReadOnlyRootFs: !cfg.DisableReadOnlyRootFs,
+		EnableTini:           !cfg.DisableTini,
 	})
 }
 
--- a/api/agent/config.go
+++ b/api/agent/config.go
@@ -33,6 +33,7 @@ type Config struct {
 	EnableNBResourceTracker bool          `json:"enable_nb_resource_tracker"`
 	MaxTmpFsInodes          uint64        `json:"max_tmpfs_inodes"`
 	DisableReadOnlyRootFs   bool          `json:"disable_readonly_rootfs"`
+	DisableTini             bool          `json:"disable_tini"`
 	DisableDebugUserLogs    bool          `json:"disable_debug_user_logs"`
 }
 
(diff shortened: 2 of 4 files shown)


zootalures:
Works fine on mac , LGTM

rdallman:
thanks for review... see if any windows people come out of the woodwork we can flip it relatively easily

zootalures:
Docker4Windows uses basically the same LK  VM images as d4m so I "expect it will be fine"

rdallman:
> "expect it will be fine"

famous last words!

PR #1196, opened 2018-09-05
fn: handle client connection close errors

skinowski:
Handle context Cancel (client closed connection) with special error code HTTP 444. We merely set the status code in this case to make sure stats/metrics record the error code.


gviedma:
Looks good, thanks for making this change @skinowski !

PR #1197, opened 2018-09-08
fn: docker driver stats/metrics for prometheus

skinowski:
With this change, we start tracking latencies in docker wrapper. This information is already in opencensus spans, however for environments where there's no open tracing server available, it's beneficial to have them in prometheus as well.

Docker exit codes are also now being tracked, which replaces OOM only counter now.

Exit codes, for example are as follows:
```
# HELP fn_docker_exits docker exit counts
# TYPE fn_docker_exits counter
fn_docker_exits{api_name="docker_wait_container",exit_status="ctx_canceled",fn_appname="fn-http-func",fn_path="/fn-http-func"} 1
fn_docker_exits{api_name="docker_wait_container",exit_status="exit_137",fn_appname="fn-http-func",fn_path="/fn-http-func"} 2
```

and:

```
# HELP fn_docker_api_error docker api errors
# TYPE fn_docker_api_error counter
fn_docker_api_error{api_name="docker_inspect_image",fn_appname="fn-http-func",fn_path="/fn-http-func"} 1
fn_docker_api_error{api_name="docker_wait_container",fn_appname="fn-http-func",fn_path="/fn-http-func"} 1
# HELP fn_docker_api_latency Docker wrapper latency
# TYPE fn_docker_api_latency histogram
fn_docker_api_latency_bucket{api_name="docker_attach_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="1"} 2
fn_docker_api_latency_bucket{api_name="docker_attach_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="10"} 3
fn_docker_api_latency_bucket{api_name="docker_attach_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="50"} 3
fn_docker_api_latency_bucket{api_name="docker_attach_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="100"} 3
fn_docker_api_latency_bucket{api_name="docker_attach_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="250"} 3
fn_docker_api_latency_bucket{api_name="docker_attach_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="500"} 3
fn_docker_api_latency_bucket{api_name="docker_attach_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="1000"} 3
fn_docker_api_latency_bucket{api_name="docker_attach_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="10000"} 3
fn_docker_api_latency_bucket{api_name="docker_attach_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="60000"} 3
fn_docker_api_latency_bucket{api_name="docker_attach_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="120000"} 3
fn_docker_api_latency_bucket{api_name="docker_attach_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="+Inf"} 3
fn_docker_api_latency_sum{api_name="docker_attach_container",fn_appname="fn-http-func",fn_path="/fn-http-func"} 1
fn_docker_api_latency_count{api_name="docker_attach_container",fn_appname="fn-http-func",fn_path="/fn-http-func"} 3
fn_docker_api_latency_bucket{api_name="docker_create_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="1"} 0
fn_docker_api_latency_bucket{api_name="docker_create_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="10"} 0
fn_docker_api_latency_bucket{api_name="docker_create_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="50"} 0
fn_docker_api_latency_bucket{api_name="docker_create_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="100"} 0
fn_docker_api_latency_bucket{api_name="docker_create_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="250"} 3
fn_docker_api_latency_bucket{api_name="docker_create_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="500"} 3
fn_docker_api_latency_bucket{api_name="docker_create_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="1000"} 3
fn_docker_api_latency_bucket{api_name="docker_create_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="10000"} 3
fn_docker_api_latency_bucket{api_name="docker_create_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="60000"} 3
fn_docker_api_latency_bucket{api_name="docker_create_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="120000"} 3
fn_docker_api_latency_bucket{api_name="docker_create_container",fn_appname="fn-http-func",fn_path="/fn-http-func",le="+Inf"} 3
fn_docker_api_latency_sum{api_name="docker_create_container",fn_appname="fn-http-func",fn_path="/fn-http-func"} 438
fn_docker_api_latency_count{api_name="docker_create_container",fn_appname="fn-http-func",fn_path="/fn-http-func"} 3
```

### DIFF
--- a/api/agent/drivers/docker/docker.go
+++ b/api/agent/drivers/docker/docker.go
@@ -14,7 +14,6 @@ import (
 	"sync"
 	"time"
 
-	"go.opencensus.io/stats"
 	"go.opencensus.io/trace"
 
 	"github.com/coreos/go-semver/semver"
@@ -368,33 +367,20 @@ func dockerMsg(derr *docker.Error) string {
 // The docker driver will attempt to cast the task to a Auther. If that succeeds, private image support is available. See the Auther interface for how to implement this.
 func (drv *DockerDriver) run(ctx context.Context, container string, task drivers.ContainerTask) (drivers.WaitResult, error) {
 
-	attachSuccess := make(chan struct{})
 	mwOut, mwErr := task.Logger()
-
-	waiter, err := drv.docker.AttachToContainerNonBlocking(ctx, docker.AttachToContainerOptions{
-		Success:      attachSuccess,
+	waiter, err := drv.docker.AttachToContainer(ctx, docker.AttachToContainerOptions{
 		Container:    container,
+		InputStream:  task.Input(),
 		OutputStream: mwOut,
 		ErrorStream:  mwErr,
 		Stream:       true,
 		Stdout:       true,
 		Stderr:       true,
 		Stdin:        true,
-		InputStream:  task.Input()})
-
+	})
 	if err != nil && ctx.Err() == nil {
 		// ignore if ctx has errored, rewrite status lay below
 		return nil, err
-	}
-
-	// Sync up with NB Attacher above before starting the task
-	if err == nil {
-		// WARNING: the I/O below requires docker hijack function to honor
-		// the contract below, specifically if an error is not returned
-		// from AttachToContainerNonBlocking, then max blocking time
-		// here should be what drv.docker dialer/client config was set to.
-		<-attachSuccess
-		attachSuccess <- struct{}{}
 	}
 
 	// we want to stop trying to collect stats when the container exits
@@ -588,6 +574,7 @@ func (w *waitResult) wait(ctx context.Context) (status string, err error) {
 	// just say it was a timeout if we have [fatal] errors talking to docker, etc.
 	// a more prevalent case is calling wait & container already finished, so again ignore err.
 	exitCode, _ := w.drv.docker.WaitContainerWithContext(w.container, ctx)
+	defer w.drv.docker.RecordWaitContainerResult(ctx, exitCode)
 
 	w.waiter.Close()
 	err = w.waiter.Wait()
@@ -618,7 +605,6 @@ func (w *waitResult) wait(ctx context.Context) (status string, err error) {
 	case 0:
 		return drivers.StatusSuccess, nil
 	case 137: // OOM
-		stats.Record(ctx, dockerOOMMeasure.M(1))
 		common.Logger(ctx).Error("docker oom")
 		err := errors.New("container out of memory, you may want to raise route.memory for this route (default: 128MB)")
 		return drivers.StatusKilled, models.NewAPIError(http.StatusBadGateway, err)
--- a/api/agent/drivers/docker/docker.go
+++ b/api/agent/drivers/docker/docker.go
@@ -574,7 +574,7 @@ func (w *waitResult) wait(ctx context.Context) (status string, err error) {
 	// just say it was a timeout if we have [fatal] errors talking to docker, etc.
 	// a more prevalent case is calling wait & container already finished, so again ignore err.
 	exitCode, _ := w.drv.docker.WaitContainerWithContext(w.container, ctx)
-	defer w.drv.docker.RecordWaitContainerResult(ctx, exitCode)
+	defer RecordWaitContainerResult(ctx, exitCode)
 
 	w.waiter.Close()
 	err = w.waiter.Wait()
--- a/api/agent/drivers/docker/docker_client.go
+++ b/api/agent/drivers/docker/docker_client.go
@@ -45,7 +45,6 @@ type dockerClient interface {
 	Stats(opts docker.StatsOptions) error
 	Info(ctx context.Context) (*docker.DockerInfo, error)
 	LoadImages(ctx context.Context, filePath string) error
-	RecordWaitContainerResult(ctx context.Context, exitCode int)
 }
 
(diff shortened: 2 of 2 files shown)


api/agent/drivers/docker/docker_client.go
```diff
@@ -43,6 +45,7 @@ type dockerClient interface {
 	Stats(opts docker.StatsOptions) error
 	Info(ctx context.Context) (*docker.DockerInfo, error)
 	LoadImages(ctx context.Context, filePath string) error
+	RecordWaitContainerResult(ctx context.Context, exitCode int)
```
rdallman: this API method should map directly onto an API from `fsouza/go-dockerclient` (I don't believe that it does?), it should be a method on the specific driver, not the client.

kmjohansen:
Seems like a reasonable start, modulo Reed's concerns.

How difficult would it be to add metrics around the number of containers active, number of containers paused, number of images cached, size of images cached, disk space used (according to docker), and things like that?

skinowski:
@kmjohansen I'm planning to submit more PRs for other areas you mentioned.

skinowski:
I'm going to dismiss @rdallman since I think I've addressed the issue, but pls let me know if followup/fixup is needed.

api/agent/drivers/docker/docker.go
```diff
@@ -368,35 +367,22 @@ func dockerMsg(derr *docker.Error) string {
 // The docker driver will attempt to cast the task to a Auther. If that succeeds, private image support is available. See the Auther interface for how to implement this.
 func (drv *DockerDriver) run(ctx context.Context, container string, task drivers.ContainerTask) (drivers.WaitResult, error) {
 
-	attachSuccess := make(chan struct{})
 	mwOut, mwErr := task.Logger()
-
-	waiter, err := drv.docker.AttachToContainerNonBlocking(ctx, docker.AttachToContainerOptions{
-		Success:      attachSuccess,
+	waiter, err := drv.docker.AttachToContainer(ctx, docker.AttachToContainerOptions{
```
rdallman: this doesn't respect the context afaict https://github.com/fsouza/go-dockerclient/blob/master/container.go#L1441 -- hence the NB attach. for metrics we can probably get them in the waiter, but the waiter dance is critical until this is fixed upstream. have attempted to fix this upstream before https://github.com/moby/moby/pull/36544 (and move to that docker client) but alas, nothing landed. this is a bad bug we need to fix asap, since if the docker client doesn't timeout properly we can leak goroutines and containers... sorry, should have caught this earlier.

rdallman: @skinowski ^ urgent

rdallman: ah, i see we just renamed the thing to try to push it down but it's the same thing. i think still kinda misses the idea of the wrapper around the client (of which i am ashamed in general, but don't see any better options), which was to keep the same behaviors of the underlying fsouza api but add retries and timeouts to the docker calls since docker api calls like to fail frequently... anyway, no bug if underneath uses NB, but this still smells since it's a mismatch

api/agent/drivers/docker/docker_client.go
```diff
@@ -187,15 +252,34 @@ func (d *dockerWrap) Info(ctx context.Context) (info *docker.DockerInfo, err err
...
+
+	// We use non-blocking here since we need the CloseWaiter
+	waiter, err := d.docker.AttachToContainerNonBlocking(opts)
+
+	// Sync up with NB Attacher above before starting the task
+	if err == nil {
+		// WARNING: the I/O below requires docker hijack function to honor
+		// the contract below, specifically if an error is not returned
+		// from AttachToContainerNonBlocking, then max blocking time
+		// here should be what drv.docker dialer/client config was set to.
```
rdallman: i am not sure we are setting this anymore either, not sure this assumption is safe. i can double check through this tomorrow hopefully

skinowski: So, I got this queued up: https://github.com/fnproject/fn/pull/1201
I'm not sure if it's a significant improvement, but at least it honors the context. I think we used to set some timeouts in the dialers in the driver, but now, I think hijack & Dialer/D combo may indeed block without. If that happens, then the dialer go-routine would block/leak, along with our go-routine in #1201. What do you think? The rationale here (AFAIR) was that we couldn't start the task before these streams were attached. (Might be less of a problem with docker-syslog and upcoming socket-IO, but it probably safer to leave it in.)

Issue #1198, opened 2018-09-09
Unable to create async function with the timeout longer than 300 seconds.


denismakogon:
**Description**

Unable to create async function (v2) with timeout longer than 300 seconds:

func.yaml:
```yaml
schema_version: 20180708
name: bucket-cleaner
version: 0.0.7
runtime: go
entrypoint: ./func
format: json
type: async
memory: 258
timeout: 360
triggers:
- name: bucket-cleaner
  type: http
  source: /bucket-cleaner

```

**Steps to reproduce the issue:**
1. init a function
2. set it to async mode with an timeout greater than 300 secs.
3. deploy

**Describe the results you received:**

Deploy log:
```
Deploying bucket-cleaner to app: videbox
Bumped to version 0.0.7
Building image denismakogon/bucket-cleaner:0.0.7 
FN_REGISTRY:  denismakogon
Current Context:  default
Sending build context to Docker daemon  88.75MB
Step 1/8 : FROM fnproject/go:dev as build-stage
 ---> fac877f7d14d
Step 2/8 : WORKDIR /function
 ---> Using cache
 ---> 1980b829219e
Step 3/8 : ADD . /go/src/func/
 ---> 2a4fef34b60d
Step 4/8 : RUN cd /go/src/func/ && go build -o func
 ---> Running in 7c5b674ab042
Removing intermediate container 7c5b674ab042
 ---> 9e8475e2f224
Step 5/8 : FROM fnproject/go
 ---> 76aed4489768
Step 6/8 : WORKDIR /function
 ---> Using cache
 ---> 63179d25b899
Step 7/8 : COPY --from=build-stage /go/src/func/func /function/
 ---> Using cache
 ---> 517ee1e48507
Step 8/8 : ENTRYPOINT ["./func"]
 ---> Using cache
 ---> 8c535700ceff
Successfully built 8c535700ceff
Successfully tagged denismakogon/bucket-cleaner:0.0.7

Updating function bucket-cleaner using image denismakogon/bucket-cleaner:0.0.7...

Fn: timeout value is out of range, must be between 0 and 300


```

**Describe the results you expected:**

Success is expected. The max timeout for an async functions suppose to be 3600.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `fn version` (CLI command):**

```
Client version is latest version: 0.4.155
Server version:  0.3.548
```

denismakogon:
@rdallman @skinowski @ostrain

[...]

PR #1199, opened 2018-09-10
WIP: Remove Routes

mantree:
Removes all signs of Routes from the Fn server.

Fn Call have been changed to use fn_id over the route path. This has
led to many changes in the code base. Note, the Routes table does
remain, as existing migrations modifiy it, a new migration drops it,
but changing migrations is verboten, so it has to stay.


api/datastore/sql/sql.go
```diff
@@ -955,7 +761,7 @@ func (ds *SQLStore) InsertCall(ctx context.Context, call *models.Call) error {
 		completed_at,
 		status,
 		app_id,
-		path,
+    fn_id,
```
riconnon: Super nit: indentation.

api/datastore/sql/sql.go
```diff
@@ -966,7 +772,7 @@ func (ds *SQLStore) InsertCall(ctx context.Context, call *models.Call) error {
 		:completed_at,
 		:status,
 		:app_id,
-		:path,
+    :fn_id,
```
riconnon: Super nit: indentation.

api/server/runner_async_test.go
```diff
@@ -34,72 +29,3 @@ func testRouterAsync(ds models.Datastore, mq models.MessageQueue, rnr agent.Agen
 	s.bindHandlers(ctx)
 	return r
 }
-
-func TestRouteRunnerAsyncExecution(t *testing.T) {
```
riconnon: Is this functionality still available executing Fns? I don't think we test it elsewhere.

api/server/runner_httptrigger_test.go
```diff
@@ -99,40 +98,6 @@ func (mock *errorMQ) Reserve(context.Context) (*models.Call, error)            {
 func (mock *errorMQ) Delete(context.Context, *models.Call) error               { return mock }
 func (mock *errorMQ) Code() int                                                { return mock.code }
 func (mock *errorMQ) Close() error                                             { return nil }
-func TestFailedEnqueue(t *testing.T) {
```
riconnon: Do we need to test this for Fns?

denismakogon:
We need to fix this before we get rid of routes https://github.com/fnproject/fn/issues/1198.

api/datastore/sql/migrations/18_create_new_calls.go
```diff
@@ -0,0 +1,49 @@
+package migrations
+
+import (
+	"context"
+
+	"github.com/fnproject/fn/api/datastore/sql/migratex"
+	"github.com/jmoiron/sqlx"
+)
+
+func up18(ctx context.Context, tx *sqlx.Tx) error {
```
denismakogon: i'm not sure that this migration does exactly what we need.

what if routes table has lots of routes defined, they all must turn into function and triggers, unless we intentionally want to lose data.

api/datastore/sql/migrations/18_create_new_calls.go
```diff
@@ -0,0 +1,49 @@
...
+	stats text,
+	error text,
+	PRIMARY KEY (id)
+);`
+	_, err = tx.ExecContext(ctx, createQuery)
+	return err
+}
+
+func down18(ctx context.Context, tx *sqlx.Tx) error {
+	return nil
```
denismakogon: down must revert a change, so it has to put back a table, ideally with data (not sure how this can be done, but still matters)

mantree:
superseeded by #1210

PR #1201, opened 2018-09-11
fn: check context timeout when waiting for non-blocking attach

skinowski:
With this change, we no longer allow docker client AttachToContainerNonBlocking
to block on Success channel more than our context deadline/timeout.

### DIFF
--- a/api/agent/drivers/docker/docker_client.go
+++ b/api/agent/drivers/docker/docker_client.go
@@ -31,7 +31,7 @@ const (
 type dockerClient interface {
 	// Each of these are github.com/fsouza/go-dockerclient methods
 
-	AttachToContainer(ctx context.Context, opts docker.AttachToContainerOptions) (docker.CloseWaiter, error)
+	AttachToContainerNonBlocking(ctx context.Context, opts docker.AttachToContainerOptions) (docker.CloseWaiter, error)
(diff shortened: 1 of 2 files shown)


api/agent/drivers/docker/docker_client.go
```diff
@@ -256,22 +256,31 @@ func (d *dockerWrap) AttachToContainer(ctx context.Context, opts docker.AttachTo
...
+		go func() {
+			<-opts.Success
+			opts.Success <- struct{}{}
+			close(mon)
+		}()
+
+		select {
+		case <-ctx.Done():
+		case <-mon:
+		}
```
rdallman: honestly, I don't think we should do any of this in the docker client, it is the wrong place to be talking about specific behavior we want around stream attachment and we are changing the fsouza dockerclient contract. if we want anything like this, it needs to be in the caller. for the dockerclient, this is correct:

```
 func (d *dockerWrap) AttachToContainerNonBlocking(ctx context.Context, opts docker.AttachToContainerOptions) (docker.CloseWaiter, error) {
    ctx, span := trace.StartSpan(ctx, "docker_attach_container")
    defer span.End()
    return d.docker.AttachToContainerNonBlocking(opts)
}
```

PR #1202, opened 2018-09-11
http-stream format

rdallman:
this builds on the original POC by @skinowski to get a quasi-MVP thing (stage 1, working-ish, as it is known)

introducing the `http-stream` format support in fn. there are many details for
this, none of which can be linked from github :( -- docs are coming (I could
even try to add some here?). this is kinda MVP-ish level, but does not
implement the remaining spec, ie 'headers' fixing up / invoke fixing up. the
thinking being we can land this to test fdks / cli with and start splitting
work up on top of this. all other formats work the same as previous (no
breakage, only new stuff)

with the cli you can set `format: http-stream` and deploy, and then invoke a
function via the `http-stream` format. this uses unix domain socket (uds) on
the container instead of previous stdin/stdout, and fdks will have to support
this in a new fashion (will see about getting docs on here). fdk-go works,
which is here: fnproject/fdk-go#30 . the output looks
the same as an http format function when invoking a function. wahoo.

there's some amount of stuff we can clean up here, enumerated:

* the cleanup of the sock files is iffy, high pri here

* permissions are a pain in the ass and i punted on dealing with them. you can
run `sudo ./fnserver` if running locally, it may/may not work in dind(?) ootb

* no pipe usage at all (yay), still could reduce buffer usage around the pipe
behavior, we could clean this up potentially before removal (and tests)

* my brain can’t figure out if dispatchOldFormats changes pipe behavior, but
tests work

* i marked XXX to do some clean up which will follow soon… need this to test fdk
tho so meh, any thoughts on those marked would be appreciated however (1 less
decision for me). mostly happy w/ general shape/plumbing tho

* there are no tests atm, this is a tricky dance indeed. attempts were made.
need to futz with the permission stuff before committing to adding any tests
here, which I don't like either. also, need to get the fdk-go based test image
updated according to the fdk-go, and there's a dance there too. rumba time..

* delaying the big big cleanup until we have good enough fdk support to kill
all the other formats.

open to ideas on how to maneuver landing stuff...


rdallman:
looking at tests meow...

rdallman:
i'll have a think about https://github.com/fnproject/fn/pull/1202/commits/90f754527829f1497ceaedc9bffad94b885ed0db -- we do need to limit the output size for http-stream, this does not affect any other formats though.

Makefile
```diff
@@ -8,7 +8,7 @@ dep-up:
 	dep ensure
 
 .PHONY: build
-build: api/agent/grpc/runner.pb.go
```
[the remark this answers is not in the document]
rdallman: i should move it elsewhere, but it shouldn't be part of build at least. thanks, i'll move it

api/agent/agent.go
```diff
@@ -869,6 +981,19 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
...
+			call.slots.queueSlot(&hotSlot{done: make(chan struct{}), fatalErr: err})
+			return
+		}
+	case <-ctx.Done():
+		call.slots.queueSlot(&hotSlot{done: make(chan struct{}), fatalErr: ctx.Err()})
+		return
+	}
+
 	// container is running
 	state.UpdateState(ctx, ContainerStateIdle, call.slots)
```
skinowski: minor nit, we should move this state transition above the uds wait.

rdallman:
i added cursory docs and updated the fdk-test-util image, and was going to start adding tests before doing the header dance to fix those up in that fashion. however, i'd like to run tests without root, so started trying to figure out how to accommodate this today. didn't make a lot of progress, `/dev/shm` can be used on linux but not osx/win to get a tmpfs without having to `mount`, but we have to symlink/chmod it to somewhere for use (to mount into the container with perms). anyway i hate this shit so if anyone is more of a masochist and actually enjoys that, feel free to pick up that bit, i'm gonna stop hating computers for the next 12 hours...

api/agent/call.go
```diff
@@ -311,6 +311,9 @@ func buildConfigWithPath(app *models.App, fn *models.Fn, path string) models.Con
 	}
 
 	conf["FN_FORMAT"] = fn.Format
+	if fn.Format == models.FormatHTTPStream { // TODO should be always soon...
+		conf["FN_LISTENER"] = "unix:/iofs/lsnr.sock" // XXX(reed): hardcoding this is ok right? it's a contract
```
denismakogon: shouldn't this be `unix:///iofs/lsnr.sock` or just `/iofs/lsnr.sock`, this would really simplify FDK life by avoiding necessary URL parsing.

Both Go HTTP server, Python aiohttp server and other just need a file path, so, they need to rip off `unix:/` prefix. Since that's a contract we know that it's gonna be a Unix socket only.

rdallman: having protocol seems like the right thing to do in case we add tcp support. urls are easy to parse and there are parsers available in every major language...

rdallman:
@skinowski see https://github.com/fnproject/fn/pull/1202/commits/a52bd69c54c8605cf8e9532bcefa769913deabdd -- i think this is a sensible default (at least, osx and linux it seems ok, but idk what the ymmv might be, slight concern there). i'm open to ideas if you (or anyone else) has any, i read more man pages than i wanted to in my lifetime today. for further configuration (mount stuff), we can chat about an agent extension, etc.

rdallman:
we're merging in the interest of doing this before https://github.com/fnproject/fn/pull/1210 and for folks to test things such as https://github.com/fnproject/fdk-java/pull/150 -- thanks for review, feel free to comment on latest commit

rdallman:
shaves about 2ms (2/7 -> 30% improvement) off of hot execution time on my laptop for a tiny req/response vs the old http over stdio format. woo. probably quite a bit more for larger input/outputs. didn't rigorously test cold start time, it appears similar and insignificant if any latency improvement/regression.

PR #1203, opened 2018-09-11
Support Fn Calls on V2 api

mantree:
The V2 Calls api endpoints have been added beneath fns:

  /fns/{fnID}/calls
  /fns/{fnID}/calls/{callID}

The S3 implementation forces our hand as we if we want to list Calls
under a Fn, we have to use the FnID as a prefix on the object names,
which mean we need it to look up any Call. It also makes sense in
terms of resource hierarchy.

These endpoints can optionally be disabled (as other endpoints), if a
service provider needs to provide this functionality via other means.

The 'calls' test has been fully migrated to fn calls. This has been
done to reduce the copy pasta a bit, and on balance is ok as the
routes calls will be removed soon.

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -197,6 +197,7 @@ type Server struct {
 	noHTTTPTriggerEndpoint bool
 	noHybridAPI            bool
 	noFnInvokeEndpoint     bool
+	noCallEndpoints        bool
 	appListeners           *appListeners
 	routeListeners         *routeListeners
 	fnListeners            *fnListeners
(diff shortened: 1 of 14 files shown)


mantree:
This has failed twice due to flappy tests:

1) TestWaitGroupBlast
2) Config equality, being fixed in #1204 

Will wait for merge on 2 and try again.

riconnon:
I started a review of this then changed my mind about half of what I was writing. Let's discuss it later.

api/server/server.go
```diff
@@ -1115,6 +1124,11 @@ func (s *Server) bindHandlers(ctx context.Context) {
 			v2.DELETE("/triggers/:triggerID", s.handleTriggerDelete)
 		}
 
+		if !s.noCallEndpoints {
+			v2.GET("/fns/:fnID/calls", s.handleCallList2)
```
riconnon: Should the server provide a helpful "these are disabled" message if we have an API server which is not enabling these endpoints.

PR #1208, opened 2018-09-12
fn: make call.End() blocking to reduce complexity

skinowski:
agent/lb-agent/runner roles execute call.End() in the background
in some cases to reduce latency. With this change, we simplify this
and switch to non-background execution of call.End(). This fixes
hard to detect issues such as non-deterministic calculation of
call.CompletedAt or incomplete Call.Stats in runners.

Downstream projects if impacted by the now blocking call.End()
latency should take steps to handle this according to their requirements.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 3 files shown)


skinowski:
Background: non-blocking call.End() was a recent change in fn, so this PR is a revert of #919 .

rdallman:
merging... i need to rebase this stuff up in agent land

PR #1209, opened 2018-09-13
fn: paused and evicted container stats

skinowski:
With this change, now stats reports paused state
as well as incidents of container exit due to evictions.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -945,12 +945,17 @@ func (a *agent) runHotReq(ctx context.Context, call *call, state ContainerState,
 			return false
 		}
 		isFrozen = true
+		state.UpdateState(ctx, ContainerStatePaused, call.slots)
 	}
 
 	evictor := a.evictor.GetEvictor(call.ID, call.slotHashId, call.Memory+uint64(call.TmpFsSize), uint64(call.CPUs))
 
-	state.UpdateState(ctx, ContainerStateIdle, call.slots)
+	if !isFrozen {
+		state.UpdateState(ctx, ContainerStateIdle, call.slots)
+	}
 	s := call.slots.queueSlot(slot)
+
+	isEvictEvent := false
 
 	for {
 		select {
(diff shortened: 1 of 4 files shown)


skinowski:
In gauge keys, container_done_total was actually a counter and it is not very useful, so it is now removed.

api/agent/agent.go
```diff
@@ -965,10 +970,12 @@ func (a *agent) runHotReq(ctx context.Context, call *call, state ContainerState,
 					return false
 				}
 				isFrozen = true
+				state.UpdateState(ctx, ContainerStatePaused, call.slots)
 			}
 			continue
 		case <-evictor.C:
 			logger.Debug("attempting hot function eject")
+			isEvictEvent = true
```
rdallman: i guess it's a timing thing but uh, `statsContainerEvicted(ctx)` here  will run really fast (it's an atomic add) and then we don't have to drag that boolean variable around.

api/agent/state_trackers.go
```diff
@@ -109,11 +116,15 @@ func (c *containerState) UpdateState(ctx context.Context, newState ContainerStat
...
+	// 1) from busy to idle/paused
+	// 2) from done to waiting
+	// 3) to/from idle from/to paused
+	// otherwise, we can only move forward in states
 	if c.state < newState ||
-		(c.state == ContainerStateBusy && newState == ContainerStateIdle) ||
-		(c.state == ContainerStateDone && newState == ContainerStateIdle) {
+		(c.state != newState && isIdleState(newState) && isIdleState(c.state)) ||
+		(c.state == ContainerStateBusy && isIdleState(newState)) ||
+		(c.state == ContainerStateDone && isIdleState(newState)) {
```
rdallman: i don't think this will allow done -> waiting (as noted, it should).

done < waiting -> false
done != waiting && isIdleState(done) && isIdleState(waiting) -> false, neither are idle states
done == busy && isIdleState(waiting) -> false, waiting is not an idle state
done == done && isIdleState(waiting) -> false, waiting is not an idle state

seemingly, making waiting an idle state would solve this, but didn't analyze the implications of that. my brain hurts now, bravo. at least, i think this needs some more kludging (or a tweak to the comments)

api/agent/agent.go
skinowski: Extra bool is because the eviction event may arrive here, but we have to also check if slot was grabbed a few lines below.

api/agent/state_trackers.go
skinowski: done->waiting might be an old / deprecated requirement.. Checking... Thanks for catching this.

PR #1212, opened 2018-09-14
Add tmpfs IOFS

riconnon:
Looking to make this extensible for third parties wanting to instantiate custom agents.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -1247,6 +1247,7 @@ func newHotContainer(ctx context.Context, call *call, cfg *Config) (*container, 
 			for _, b := range bufs {
 				bufPool.Put(b)
 			}
+			// iofs.Close MUST be called here or we will leak directories and/or tmpfs mounts!
 			if err = iofs.Close(); err != nil {
 				// Note: This is logged with the context of the container creation
 				common.Logger(ctx).WithError(err).Error("Error closing IOFS")
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -1207,8 +1207,12 @@ func newHotContainer(ctx context.Context, call *call, cfg *Config) (*container, 
 	var err error
 	if call.Format == models.FormatHTTPStream {
 		// XXX(reed): we should also point stdout to stderr, and not have stdin
+		if platformSupportsTmpfs() && !cfg.IOFSDisableTmpfs {
+			iofs, err = newTmpfsIOFS(cfg)
+		} else {
+			iofs, err = newDirectoryIOFS(cfg)
+		}
 
-		iofs, err = newTmpfsIOFS(cfg)
 		if err != nil {
 			return nil, err
 		}
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -6,10 +6,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net"
 	"net/http"
-	"os"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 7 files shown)


api/agent/agent.go
```diff
@@ -1191,6 +1157,56 @@ type container struct {
 	stats  *drivers.Stats
 }
 
+// IOFS object encapsulates a temporary path and the method to clean it up
+type IOFS interface {
+	Path() string
```
zootalures: can inherit io.Closer

api/agent/agent.go
```diff
@@ -1191,6 +1157,56 @@ type container struct {
...
+}
+
+func (d *directoryIOFS) Path() string {
+	return d.path
+}
+
+func (d *directoryIOFS) Close() error {
+	err := os.RemoveAll(d.path)
+	if err != nil {
+		logrus.WithError(err).Error("Error removing iofs")
```
rdallman: should do this from caller (we'll have a context logger)

api/agent/agent.go
```diff
@@ -1191,6 +1157,56 @@ type container struct {
...
+		return err
+	}
+	return nil
+}
+
+func newDirectoryIOFS(cfg *Config) (*directoryIOFS, error) {
+	// XXX(reed): need to ensure these are cleaned up if any of these ops in here fail...
```
rdallman: this looks safe now, can also remove the IOFSOpts option since whoever implements this interface can configure that themselves (it was for mount, which is gone)

riconnon:
I've made this switch based on platform (linux vs not linux) and a boolean "disable tmpfs" flag

api/agent/agent.go
```diff
@@ -1302,7 +1247,10 @@ func newHotContainer(ctx context.Context, call *call, cfg *Config) (*container,
 			for _, b := range bufs {
 				bufPool.Put(b)
 			}
-			closer() // XXX(reed): clean up
+			if err = iofs.Close(); err != nil {
+				// Note: This is logged with the context of the container creation
```
rdallman: yea, we could fix this (think you've brought it up). need to plumb the driver.ContainerTask.Close(ctx) method (can add a note, this is parity anyway).

would leave a very loud note here that if anybody decides to shuffle things above, `iofs.Close()` MUST be called or we leak dirs... (or `XXX(reed):` me and i'll get to it this week)

api/agent/iofs.go
```diff
@@ -0,0 +1,81 @@
...
+	// create a tmpdir
+	iofsAgentDir, err := ioutil.TempDir(dir, "iofs")
+	if err != nil {
+		return nil, fmt.Errorf("cannot create tmpdir for iofs: %v", err)
+	}
+
+	if cfg.IOFSMountRoot != "" {
+		iofsRelPath, err := filepath.Rel(dir, iofsAgentDir)
+		if err != nil {
+			return nil, fmt.Errorf("cannot relativise iofs path: %v", err)
```
rdallman: this will leak the dir :/ can leave TODO for me if you want, suspected some of these...

api/agent/iofs_notlinux.go
```diff
@@ -0,0 +1,21 @@
+// +build !linux
```
rdallman: it would be nice if this blew up at build time when a user tried to use this rather than waiting until runtime and blowing up when trying to run a container. this shouldn't even build to me, and agent would have to be configured with this, but yea i guess the bool may be easier to plumb, so as i'm getting used to saying, punt...

riconnon: @rdallman for it to blow up at build time it would have to be configured at build time not start time

rdallman: in any case, i think this use of build flags smells and isn't the proper way to use them... they should be used to block builds or have alternate implementations, not for runtime configuration errors... (at least let it be known...)

rdallman: yep. we can do it...

api/agent/iofs.go
riconnon: I think we just need to call `dirIOFS.Close()` in the event of errors. I can do that

api/agent/iofs_notlinux.go
rdallman: we could leave it as something for 'extensions' to configure, when building their own fn binary with the agent and not using core's main (so not using env vars to configure)

api/agent/iofs.go
rdallman: eh, then it falls on the caller. rather have a simpler caller contract (one where if the err is not nil, there's not a non-nil thing to do something with)

riconnon: I was reading this in the context of the wrong bit of the code

api/agent/agent.go
riconnon: Done

rdallman:
merging to work on top of... thanks

PR #1213, opened 2018-09-14
support configuration-based relative dirs (host and agent) for iofs

zootalures:
I noticed  that http-stream doesn't work on docker4mac - this is because docker mounts volumes relative to the OSX host, not the docker VM. 

This change allows two directory roots for IOFS mounts one with respect to the docker host, one with respect to the agent itself.  The behaviour is conditional on both FN_IOFS_PATH and FN_IOFS_MOUNT_ROOT. 

I've also added the specified headers to the contract (proper HTTP versions rather than the old FN_DEADLINE) as dups , this makes java happy

```
docker run -v $(pwd)/maciofs:/iofs -eFN_LOG_LEVEL=debug  -e FN_IOFS_PATH=/iofs -e FN_IOFS_MOUNT_ROOT=$(pwd)/maciofs  --privileged -v /var/run/docker.sock:/var/run/docker.sock   -p8080:8080  --name fnserver --rm  fnproject/fnserver
```

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -24,6 +24,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"go.opencensus.io/stats"
 	"go.opencensus.io/trace"
+	"path/filepath"
 )
 
 // TODO we should prob store async calls in db immediately since we're returning id (will 404 until post-execution)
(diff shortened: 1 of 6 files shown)


zootalures:
Plan would be for `fn start` to use this behavior along side a new "don't create tmpfs mounts" option which would also (i think) allow us to run Fn unprivileged and make docker4windows work.

api/agent/agent.go
```diff
@@ -726,18 +727,40 @@ func (s *hotSlot) dispatch(ctx context.Context, call *call) chan error {
...
+	// TODO(occ) : fix compatidupes when FDKs are updated
+	req.Header.Set("Fn-Call-Id", call.ID)
 	req.Header.Set("FN_CALL_ID", call.ID)
+	deadline, ok := ctx.Deadline()
+	if ok {
+		deadlineStr := deadline.Format(time.RFC3339)
+		req.Header.Set("Fn-Deadline", deadlineStr)
+		req.Header.Set("FN_DEADLINE", deadlineStr)
+
+	}
```
rdallman: thanks for ^ mods

rdallman:
needs gofmt but LGTM

zootalures:
Yep, oops will fix tomorrow

PR #1215, opened 2018-09-14
fn: new agent resource tracker metrics

skinowski:
New metrics for agent resource tracker: CpuUsed, CpuAvail,
MemUsed, MemAvail.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -311,7 +311,7 @@ func (a *agent) submit(ctx context.Context, call *call) error {
 func (a *agent) handleCallEnd(ctx context.Context, call *call, slot Slot, err error, isStarted bool) error {
 
 	if slot != nil {
-		slot.Close(common.BackgroundContext(ctx))
+		slot.Close()
 	}
 
 	// This means call was routed (executed)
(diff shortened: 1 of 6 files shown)


api/agent/agent.go
```diff
@@ -832,11 +833,21 @@ func (a *agent) prepCold(ctx context.Context, call *call, tok ResourceToken, ch
 
 	call.containerState.UpdateState(ctx, ContainerStateIdle, call.slots)
 
-	slot := &coldSlot{cookie, tok, err}
+	closer := func() {
+		if cookie != nil {
+			cookie.Close(ctx)
```
rdallman: in general this isn't great, we're enclosing a stale context from creation into `cookie.Close` when it's called much later, perhaps with a different enclosing context at the time

rdallman: not just for timeout, but for stats/trace ordering it gets really weird too...

rdallman: it seems like this is cold only after looking, so this is ok really if hot doesn't have this shortcoming

skinowski: yes, cold only and n/a for hot.

PR #1224, opened 2018-09-18
Stream test commence

rdallman:
this assures that Content-Type and Fn-Http-Status are set for an http-stream
function. it took some fixing up of the test utils code for the plumbing to
work, looking forward to deleting most stuff in fn-test-utils.go file around
each format -- had to update fdk-go to latest for http-stream support. this
only adds 1 test, since there's some machinery here, and would like to unblock
working on the http gateway simultaneously while adding a full suite of invoke
tests (this work can be parallelized)...

i added debug logs back to the debugging output. turns out this is useful, but
it can get noisy (only when things fail, hopefully).

### DIFF
--- a/api/server/server_test.go
+++ b/api/server/server_test.go
@@ -19,7 +19,7 @@ import (
 
 func testServer(ds models.Datastore, mq models.MessageQueue, logDB models.LogStore, rnr agent.Agent, nodeType NodeType, opts ...Option) *Server {
 	return New(context.Background(), append(opts,
-		WithLogLevel(getEnv(EnvLogLevel, DefaultLogLevel)),
+		WithLogLevel("debug"),
 		WithDatastore(ds),
 		WithMQ(mq),
 		WithLogstore(logDB),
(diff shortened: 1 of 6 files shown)


api/server/server_test.go
```diff
@@ -19,7 +19,7 @@ import (
 
 func testServer(ds models.Datastore, mq models.MessageQueue, logDB models.LogStore, rnr agent.Agent, nodeType NodeType, opts ...Option) *Server {
 	return New(context.Background(), append(opts,
-		WithLogLevel(getEnv(EnvLogLevel, DefaultLogLevel)),
+		WithLogLevel("debug"),
```
mantree: Not sure about this. Looks like a dev time thing that shouldn't be commited.

mantree: just looks like the kind of thing that someone might be using somewhere, via env vars and this might upset them.

rdallman: just logs yo

PR #1225, opened 2018-09-18
Gateway trigger support

zootalures:


api/agent/call.go
```diff
@@ -107,17 +116,17 @@ func FromHTTPTriggerRequest(app *models.App, fn *models.Fn, trigger *models.Trig
...
+			URL:         requestUrl,
 			Method:      req.Method,
 			AppID:       app.ID,
 			AppName:     app.Name,
 			FnID:        fn.ID,
 			TriggerID:   trigger.ID,
 			SyslogURL:   syslogURL,
 		}
-
+		req.Header = headers
```
mantree: I'd probably do this higher up, as soon as the header rewriting is completed and then just use req.Heade everywhere. Avoid confusion later.

api/server/runner_httptrigger.go
```diff
@@ -62,28 +64,89 @@ func (s *Server) handleTriggerHTTPFunctionCall2(c *gin.Context) error {
...
+	if status != "" {
+		statusInt, err := strconv.Atoi(status)
+		if err == nil {
+			gatewayStatus = statusInt
+		}
+	}
+
+	for k, vs := range trw.headers {
+		if strings.HasPrefix(k, "Fn-Http-H-") {
+			// TODO strip out content-length and stuff here.
```
mantree: Feel like we should act on this TODO now rather than later.

api/agent/call.go
```diff
@@ -53,26 +55,41 @@ const (
...
+			rewriteKey := fmt.Sprintf("Fn-Http-H-%s", k)
+			for _, v := range vs {
+				headers.Add(rewriteKey, v)
+			}
+		}
+		requestUrl := reqURL(req)
+
+		headers.Set("Fn-Http-Method", req.Method)
+		if contentType != "" {
+			headers.Set("Content-Type", contentType)
```
rdallman: think we need to skip Content-Type above, as well, or we'll have `Fn-Http-H-Content-Type` and `Content-Type` (or was this a goal?)... not a deal breaker, anyway

api/agent/call.go
```diff
@@ -83,6 +100,8 @@ func FromHTTPTriggerRequest(app *models.App, fn *models.Fn, trigger *models.Trig
 		// TODO this relies on ordering of opts, but tests make sure it works, probably re-plumb/destroy headers
 		// TODO async should probably supply an http.ResponseWriter that records the logs, to attach response headers to
 		if rw, ok := c.w.(http.ResponseWriter); ok {
+			// TODO deprecate after CLI is updated
+			rw.Header().Add("Fn-Call-ID", id)
```
rdallman: these end up being identical... they both produce `Fn-Call-Id`... do with that what you will

PR #1227, opened 2018-09-19
fn: eviction code simplification

skinowski:
*) Non-blocking intend is now expressed using context passed
to GetResourceToken(). For hot, this is either hot-poll interval
or 0 (non-blocking).
*) checkLaunch, extra timer wait for hot-poll is confusing
and assumed CapacityFull to evict. This is now removed.
*) GetResourceToken() semantics slightly changed where returned
channel always includes a token either valid or CapacityFull.


api/agent/agent.go
```diff
@@ -447,6 +446,17 @@ func (a *agent) checkLaunch(ctx context.Context, call *call, notifyChan chan err
...
 
+	// In blocking mode, we attempt resource allocation in HotPoll intervals
+	// and attempt an evict if we get CapacityFull after HotPoll expiry.
+	// In non-blocking mode, the interval is set to 0.
+	resourceTimeout := a.cfg.HotPoll
+	if a.cfg.EnableNBResourceTracker {
+		resourceTimeout = 0
+	}
+
+	resourceCtx, cancelCtx := context.WithTimeout(ctx, resourceTimeout)
```
rdallman: i think it's safe to do `ctx, cancel = ` if we do it like this:

```go
if !a.cfg.EnableNBResourceTracker {
  var cancel func()
  ctx, cancel = context.WithTimeout(ctx, a.cfg.HotPoll)
  defer cancel()
}
```

this means we don't have to worry about 2 different deadlines in this scope (for `<-ctx.Done()`, the context handles that), don't have 2 contexts in scope, etc. I think this is correct, as `runHot` won't inherit timings and the state updater ignores the deadlines (just used for logging/stats). wdyt ?

api/agent/agent.go
```diff
@@ -465,11 +475,17 @@ func (a *agent) checkLaunch(ctx context.Context, call *call, notifyChan chan err
 	// Non-blocking mode only applies to cpu+mem, and if isNewContainerNeeded decided that we do not
 	// need to start a new container, then waiters will wait.
 	select {
-	case tok := <-a.resources.GetResourceToken(ctx, mem, call.CPUs, isAsync, isNB):
```
rdallman: yay :)

api/agent/agent.go
```diff
@@ -546,20 +558,24 @@ func (a *agent) waitHot(ctx context.Context, call *call) (Slot, error) {
...
-	mem := call.Memory + uint64(call.TmpFsSize)
+	if a.cfg.EnableNBResourceTracker {
+		tmpCtx, tmpCtxCancel := context.WithTimeout(ctx, 0)
+		resourceCtx = tmpCtx
+		defer tmpCtxCancel()
+	}
 
 	select {
-	case tok := <-a.resources.GetResourceToken(ctx, mem, call.CPUs, isAsync, isNB):
+	case tok := <-a.resources.GetResourceToken(resourceCtx, mem, call.CPUs, isAsync):
```
rdallman: 3 :( but cold is dying so maybe it's ok?

api/agent/resource.go
```diff
@@ -292,18 +241,19 @@ func (a *resourceTracker) GetResourceToken(ctx context.Context, memory uint64, c
 		}
 		isWaiting = false
 
-		if ctx.Err() != nil {
-			c.L.Unlock()
-			return
+		// We are here because either context expired or we found a resource
+		var t ResourceToken
+		if a.isResourceAvailableLocked(memory, cpuQuota, isAsync) {
```
rdallman: ok, this is one of the loop invariants (had to pause for this one, but this seems better than checking the ctx, since if we got one before ctx timed out we can try to return it -- slightly nitty, but var assigned in loop wouldn't require double check (i checked for race but we have a lock so it's not, that being said, time under lock...))

api/agent/agent.go
skinowski: yeah, I think we have to pick basically, two contexts and pass resource specific ctx separately to resource.go or pass a timeout argument in order to avoid handling two contexts here. A timeout arg to GetResourceToken() is good/clean too. What do you think? If we use one ctx, then it gets more tricky as we don't want to accidentally bleed this specific time out into runHot() or other function calls in there...

rdallman: right. i'm wondering if there's any shifting of the channel send / receives so that we don't have to worry about this. i think if we only enter the wait loop (and launch goroutine) if we don't immediately have resources, then we can use the same context in the scope?

eg (psuedo)

```go
func GetResourceToken(ctx, params) <-chan Token {
  ch := make(chan Token, 1)
  mu.Lock()
  defer mu.Unlock() // it'll launch goroutine if we don't have resources, no worries
  ok := isResourceAvailableLocked(params)
  if ok {
    ch <- buildToken()
    return ch
  }
  
  // NOW get a chan that is unbuffered to block and launch a goroutine and return
  ch = make(chan Token)
  go awaitResourceToken(ctx, params, ch)
  return ch
}
```

eh?

rdallman: i'll clarify, this fixes the caller if we do it a certain way... 1s

rdallman: in the caller if we do:

```go
ch := GetResourceToken(ctx, params)

select {
case tok := <-ch
  return tok
default:
  // ok, we have to wait now
}

select {
case tok := <-ch:
case <-ctx.Done():
}
```

i think it works... it makes it a little easier to reason about the get resource token contract imo (but maybe not, i'm just a dude)

api/agent/resource.go
```diff
@@ -284,26 +233,31 @@ func (a *resourceTracker) GetResourceToken(ctx context.Context, memory uint64, c
 	go func() {
 		defer span.End()
 		defer cancel()
+
+		var t ResourceToken
+		t = &resourceToken{err: CapacityFull}
```
rdallman: super nit: a comment here stating this is returned when ctx times out would be 🥇

rdallman: well, unless it can get one, of course... see previous comment for an idea about this :)

api/agent/resource.go
```diff
@@ -284,26 +234,32 @@ func (a *resourceTracker) GetResourceToken(ctx context.Context, memory uint64, c
...
-			c.L.Unlock()
-			return
-		}
-
-		t := a.allocResourcesLocked(memory, cpuQuota, isAsync)
 		c.L.Unlock()
 
 		select {
 		case ch <- t:
-		case <-ctx.Done():
```
rdallman: there is a race here where the caller hasn't entered the select yet but this goroutine reaches here. i forget if we added this because we ran into that or not.

skinowski: Hmm, with current agent.go usage, this is OK, but this is fragile... If I add back the ctx, then that's also problematic. I'll figure out and submit a fix..

api/agent/agent.go
```diff
@@ -364,162 +366,146 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
...
-		select {
-		case <-a.shutWg.Closer(): // server shutdown
-			cancel()
-			return
-		case <-ctx.Done(): // timed out
-			cancel()
-			if a.slotMgr.deleteSlotQueue(call.slots) {
-				logger.Info("Hot function launcher timed out")
-				return
+	for {
```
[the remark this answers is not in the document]
skinowski: Moved this stuff into agent_launcher.go, and shuffled things around. I think it's more pleasant to look at now.

api/agent/agent_launcher.go
```diff
@@ -0,0 +1,150 @@
...
+				return s, err
+			}
+		case <-time.After(a.cfg.HotPoll):
+			cancel()
+			if !a.evictor.PerformEviction(call.slotHashId, mem, uint64(call.CPUs)) && a.cfg.EnableNBResourceTracker {
+				return nil, models.ErrCallTimeoutServerBusy
+			}
+		}
+	}
+}
```
rdallman: almost none of these selects (tryGetToken, waitGetToken, this one) seem to respect the ctx timeout. task can timeout. at least, if it does handle it, it's not obvious either

api/agent/agent_launcher.go
```diff
@@ -0,0 +1,150 @@
...
+
+	go func() {
+		dampen <- struct{}{}
+		for {
+			select {
+			case s := <-slotChan:
+				if call.slots.acquireSlot(s) {
+					if s.slot.Error() != nil {
+						s.slot.Close()
+						output <- WaitResult{nil, s.slot.Error()}
```
rdallman: this needs to be in a select, as well. (i don't think this WaitResult makes life any easier, the slot includes an error too already we could use when we need to generate one from this level, but that only seems like for server shutdown, which we really don't need to do anything except bail and never send anything if we want)

rdallman: i think it gives us some sugar for checking if we acquire the slot right?

api/agent/agent_launcher.go
```diff
@@ -0,0 +1,150 @@
...
+			}
+		case <-time.After(a.cfg.HotPoll):
+			cancel()
+			if !a.evictor.PerformEviction(call.slotHashId, mem, uint64(call.CPUs)) && a.cfg.EnableNBResourceTracker {
+				return nil, models.ErrCallTimeoutServerBusy
+			}
+		}
+	}
+}
+
```
rdallman: thanks, this is better... gonna give it another hour of looking it over today, just dense. the obvious thing that sticks out is that we're launching 1:1 request to containers when we get a new herd (during first pull on a machine, it's gonna happen) - i think this was a goal and we can work backwards fine.

api/agent/agent_launcher.go
skinowski: yes, so this is to make slot/acquire thing easier. The way it is right now is that if acquire fails, you have to retry. But this causes, extra for-loop, so I wrapped all this into this WaitResult channel... It includes ctx and shutdown, so caller checkLaunch becomes simpler...

api/agent/agent_launcher.go
```diff
@@ -0,0 +1,150 @@
...
+		for {
+			select {
+			case s := <-slotChan:
+				if call.slots.acquireSlot(s) {
+					if s.slot.Error() != nil {
+						s.slot.Close()
+						output <- WaitResult{nil, s.slot.Error()}
+						return
+					}
+					select {
```
skinowski: here select is needed, because we don't want to leak slot. But for other cases, we pre-close it already in this function so we only need this extra select here.

api/agent/agent_launcher.go
skinowski: ctx and shutdown is rolled into waitChan, which keeps this select simpler.

api/agent/agent_launcher.go
```diff
@@ -103,18 +105,26 @@ func (a *agent) checkLaunch(ctx context.Context, call *call, slotChan chan *slot
...
 			// let's wait for the launch process.
-			s := waitGetToken(waitChan, launchWait)
+			s := waitGetToken(waitChan, launchChans)
 			if s != nil {
 				return s
 			}
 		case <-time.After(a.cfg.HotPoll):
 			cancel()
-			if !a.evictor.PerformEviction(call.slotHashId, mem, uint64(call.CPUs)) && a.cfg.EnableNBResourceTracker {
+			evictChans := a.evictor.PerformEviction(call.slotHashId, mem, uint64(call.CPUs))
```
rdallman: what are the tradeoffs to pushing this down behind the resource tracker?

skinowski:
I'll take another look at this for a bit. I apologize for this super complex PR. Logically, we could potentially split evictor changes from this. Maybe as a subsequent PR on top of this. But I prefer not merging that to the old launcher. I'll spend another day on this and see how things go.

PR #1228, opened 2018-09-19
Use registry auth token from Call extensions to pull images

gkvijay:
- What I did
When pulling docker images to run the function, check if the Call has 'FN_REGISTRY_TOKEN' extension and use that token in the call to registry.

- How I did it
fnLB Call override will add the auth token as extension if the image is from a registry that needs authentication. The token is then sent to runner as Call extension. Runner uses it to pull image for the Call


rdallman:
a test would be great, as well, if possible.

gkvijay:
@rdallman, Added a test as well

rdallman:
sweet, thanks.

PR #1232, opened 2018-09-20
Unwrap log response

mantree:
Currently, the log response is not compatible with the swagger
document.

This change removes the log wrapper that the old version required.


gviedma:
For reference this fixes CLI issue https://github.com/fnproject/cli/issues/430

PR #1233, opened 2018-09-20
fn: freezer/evictor adjustments

skinowski:
*) removed faulty Idle state setter in runHot() since with
UDS wait, we need to wait until we can determine if a container
is idle. This is now moved to runHotReq().
*) evictor now more aggresive and no longer tied to pause
timer/configuration.
*) removed unnecessary optimization on timer=0 case for immediate
pause.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -984,9 +984,6 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
 		return
 	}
 
-	// container is running
-	state.UpdateState(ctx, ContainerStateIdle, call.slots)
-
 	// buffered, in case someone has slot when waiter returns but isn't yet listening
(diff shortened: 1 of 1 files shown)


api/agent/agent.go
```diff
@@ -1161,9 +1137,7 @@ func (a *agent) runHotReq(ctx context.Context, call *call, state ContainerState,
 		break
 	}
 
-	if isEvictable {
-		a.evictor.UnregisterEvictor(evictor)
-	}
+	a.evictor.UnregisterEvictor(evictor)
```
rdallman: this will get called in defer when the function returns, is it ok to call it twice?

skinowski: yes, it's got checks/validations internally.

rdallman:
mr cla_bot is ded ?

PR #1234, opened 2018-09-21
Initial Refactor

mantree:
Removing the repeated logic exposed some problems with the reponse
writers.

Currently, the trigger writer was overlaid on part of the header
writing. The main invoke blog writing into the different levels of the
overlays at different points in the logic.

Instead, by extending the types and embedded structs, the writer is
more transparent. So, at the end of the flow it goes over all the
headers available and removes our prefixes. This lets the invoke logic
just write to the top level.

Going to continue after lunch to try and remove some of the layers and
param passing.

### DIFF
--- a/api/server/runner_fninvoke.go
+++ b/api/server/runner_fninvoke.go
@@ -76,7 +76,6 @@ func (s *Server) handleFnInvokeCall2(c *gin.Context) error {
 }
 
 func (s *Server) ServeFnInvoke(c *gin.Context, app *models.App, fn *models.Fn) error {
-	// TODO: we should combine this logic with trigger, which just wraps this block with some headers wizardry
 	// TODO: we should get rid of the buffers, and stream back (saves memory (+splice), faster (splice), allows streaming, don't have to cap resp size)
 	buf := bufPool.Get().(*bytes.Buffer)
 	buf.Reset()
@@ -86,7 +85,6 @@ func (s *Server) ServeFnInvoke(c *gin.Context, app *models.App, fn *models.Fn) e
 		Buffer:  buf,
 		Headers: c.Writer.Header(),
 	}
-	defer bufPool.Put(buf) // TODO need to ensure this is safe with Dispatch?
 
 	return s.FnInvoke(c, app, fn, writer,
 		agent.WithWriter(writer), // XXX (reed): order matters [for now]
@@ -144,7 +142,6 @@ func (s *Server) FnInvoke(c *gin.Context, app *models.App, fn *models.Fn, writer
 		c.Writer.WriteHeader(writer.Status())
 	}
 
-	c.Writer.Header()
 	io.Copy(c.Writer, writer)
 
 	return nil
--- a/api/server/runner_fninvoke.go
+++ b/api/server/runner_fninvoke.go
@@ -44,10 +44,10 @@ type syncResponseWriter struct {
 func (s *syncResponseWriter) Header() http.Header { return s.headers }
 
 // By storing the status here, we effectively buffer the response
-func (s *syncResponseWriter) WriteHeader(code int)     { s.status = code }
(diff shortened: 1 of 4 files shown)


api/server/runner_fninvoke.go
```diff
@@ -66,31 +78,35 @@ func (s *Server) handleFnInvokeCall2(c *gin.Context) error {
 }
 
 func (s *Server) ServeFnInvoke(c *gin.Context, app *models.App, fn *models.Fn) error {
-	// TODO: we should combine this logic with trigger, which just wraps this block with some headers wizardry
 	// TODO: we should get rid of the buffers, and stream back (saves memory (+splice), faster (splice), allows streaming, don't have to cap resp size)
+	//TODO: Remove this pool, it makes concurrency bugs easy to create, but hard to discover.
```
rdallman: the easy thing to do is make sure that submit returned successfully and the buffer was drained, which will probably be >99% of buffers... we need the pool tho for memory pressure, if we're to keep the buffers it's basically mandatory (fragmentation, gc pressure, etc...) since these will be the largest chunks of memory for this program. eg check for success (sorry, was lazy adding this the first time around, but dispatch with the pipes actually protected us ok):

```
defer func() {
  if buf.Len() == 0 && submitErr != nil {
    bufPool.Put(buf)
  }
}
```

api/server/runner_fninvoke.go
```diff
@@ -66,31 +78,35 @@ func (s *Server) handleFnInvokeCall2(c *gin.Context) error {
...
+		agent.FromHTTPFnRequest(app, fn, c.Request))
+
 	if err != nil {
 		return err
 	}
 
+	return s.FnInvoke(c, app, fn, writer, call)
+}
+
+func (s *Server) FnInvoke(c *gin.Context, app *models.App, fn *models.Fn, writer ResponseBufferingWriter, call agent.Call) error {
```
rdallman: this shouldn't be exported, it has a very contextually specific contract and the parameters don't make sense from an exported API (nor does ResponseBufferingWriter need be exported) - please lower case

api/server/runner_fninvoke.go
```diff
@@ -101,28 +117,29 @@ func (s *Server) ServeFnInvoke(c *gin.Context, app *models.App, fn *models.Fn) e
...
 		jsonPrefix := [1]byte{'{'} // stack allocated
-		if bytes.HasPrefix(buf.Bytes(), jsonPrefix[:]) {
+		if bytes.HasPrefix(writer.GetBuffer().Bytes(), jsonPrefix[:]) {
 			// try to detect json, since DetectContentType isn't a hipster.
 			contentType = "application/json; charset=utf-8"
 		} else {
-			contentType = http.DetectContentType(buf.Bytes())
+			contentType = http.DetectContentType(writer.GetBuffer().Bytes())
 		}
+		fmt.Printf("Setting: %s\n", contentType)
```
rdallman: ;)

test/fn-system-tests/exec_fn_test.go
```diff
@@ -280,6 +282,7 @@ func TestBasicConcurrentExecution(t *testing.T) {
 			results <- nil
 		}()
 	}
+	close(latch)
```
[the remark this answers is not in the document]
mantree: Yeah, this test failed once on the ci, took me ages to get it to fail again, making sure they all set off together seemed to help a little with the repeatability.

api/server/runner_fninvoke.go
mantree: boop

api/server/runner_fninvoke.go
mantree: yeah, but ideally we wouldn't be storing responses in memory, they should be just streamed straight back to the client, as the other note notes.

PR #1239, opened 2018-09-25
Prometheus exporter for open census spans

mantree:
We have added open census span calls in interesting parts of the code,
but without creating a jaeger or zipkin cluster they are not going
anywhere.

Here we add a Prometheus exporter that creates a Histogram for each of
the spans we have added. Every sampled invocation of a Span is
Observed by the Histogram. This commit sets opencensus to sample every
invocation of spans, instead of the default sampler of 1 in
10000(ish).

The Exporter ignores any spans that have a '/' prefix as it's be
obvserved that gin creates spans with names containing IDs. This
endpoints are metric'd via the standard method anyway.

This is a bit rough, but ready.

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -28,6 +28,7 @@ import (
 	"github.com/fnproject/fn/api/models"
 	"github.com/fnproject/fn/api/mqs"
 	pool "github.com/fnproject/fn/api/runnerpool"
+	"github.com/fnproject/fn/api/server/spanexporter"
 	"github.com/fnproject/fn/api/version"
 	"github.com/fnproject/fn/fnext"
 	"github.com/gin-gonic/gin"
@@ -738,6 +739,14 @@ func WithPrometheus() Option {
 		}
 		s.promExporter = exporter
 		view.RegisterExporter(exporter)
+
+		spanExporter, err := spanexporter.NewExporter(spanexporter.Options{
+			Namespace: "fn",
+			Registry:  reg,
+			OnError:   func(err error) { logrus.WithError(err).Error("opencensus prometheus span exporter err") },
+		})
+		trace.RegisterExporter(spanExporter)
+		trace.ApplyConfig(trace.Config{DefaultSampler: trace.AlwaysSample()})
 
 		return nil
 	}
--- a/api/server/spanexporter/prometheus.go
+++ b/api/server/spanexporter/prometheus.go
@@ -1,0 +1,211 @@
+package spanexporter
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+	"unicode"
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	"go.opencensus.io/trace"
+)
+
+// Exporter exports stats to Prometheus, users need
+// to register the exporter as an http.Handler to be
+// able to export.
+type Exporter struct {
+	opts    Options
+	g       prometheus.Gatherer
+	c       *collector
+	handler http.Handler
+}
+
+var _ trace.Exporter = (*Exporter)(nil)
+
+// Options contains options for configuring the exporter.
+type Options struct {
+	Namespace string
+	Registry  *prometheus.Registry
+	OnError   func(err error)
+}
+
+// NewExporter returns an exporter that exports stats to Prometheus.
+func NewExporter(o Options) (*Exporter, error) {
+	if o.Registry == nil {
+		o.Registry = prometheus.NewRegistry()
+	}
+	collector := newCollector(o, o.Registry)
+	e := &Exporter{
+		opts: o,
+		g:    o.Registry,
+		c:    collector,
+	}
+	return e, nil
+}
+
+// ExportSpan exports to the Prometheus
+// Each OpenCensus AggregationData will be converted to
+// corresponding Prometheus Metric: SumData will be converted
+// to Untyped Metric, CountData will be a Counter Metric,
+// DistributionData will be a Histogram Metric.
+func (e *Exporter) ExportSpan(sd *trace.SpanData) {
+	if urlName(sd.Name) {
+		return
+	}
+	histo := e.c.getHistogram(sd)
+
+	spanTimeSpanNanos := sd.EndTime.Sub(sd.StartTime)
+	spanTimeSpanMillis := float64(int64(spanTimeSpanNanos / time.Millisecond))
+
+	histo.Observe(spanTimeSpanMillis)
+}
+
+var _ trace.Exporter = (*Exporter)(nil)
+
+func (c *collector) getHistogram(span *trace.SpanData) prometheus.Histogram {
+	sig := spanName(c.opts.Namespace, span)
+	c.registeredHistosMu.Lock()
+	histogram, ok := c.registeredHistograms[sig]
+	c.registeredHistosMu.Unlock()
+
+	if !ok {
+		histogram = prometheus.NewHistogram(
+			prometheus.HistogramOpts{Namespace: c.opts.Namespace,
+				Name: sanitize(span.Name),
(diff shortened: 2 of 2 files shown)


rdallman:
nice, thanks. will take a look, need something like this. slightly relevant maybe - am generally slightly concerned about 'metrics overload' from this (a lot of these won't be all that useful), curious if you have any thoughts about being able to turn certain 'views' (as opencensus calls them) on/off on a running server.

mantree:
Metric overload is a worry for sure. It's easy to make em, but sometimes hard to track dwon where they come from.

I would say that if this creates a metric that is useless, then the span was going to be useless anyway. The idea here is to answer the "where is all the time spent" question, in particular when we start load testing and want to know the bottle neck.

The other thing we could think about is using span annotations, to create a sort of debug, info, leveling system. Then we cld selectively enable/disable per environment.

If you fire this branch up you can see how many are going to get made after running a few commands.

skinowski:
Distribution buckets should be configurable. eg:
https://github.com/fnproject/fn/blob/master/cmd/fnserver/main.go#L30

api/server/spanconverter.go
```diff
@@ -0,0 +1,129 @@
...
+	stats.Record(context.Background(), m.M(spanTimeMillis))
+}
+
+func (c *SpanConverter) getMeasure(span *trace.SpanData) *stats.Float64Measure {
+	sig := sanitize(span.Name)
+	c.viewsMu.Lock()
+	m, ok := c.measures[sig]
+	c.viewsMu.Unlock()
+
+	if !ok {
```
skinowski: concurrent access without lock. Outside of locks this could register/insert more than once.

skinowski:
I think this should be optional similar to other views (api views, agent views, etc.) Maybe the registration function with distribution buckets? If we are concerned about metrics overload, perhaps the registration function could take an optional list of span names to be excluded/included?

api/server/spanconverter.go
mantree: Yeah it looked wrong to me to, when I lifted it from the main prom exporter. 

In actual fact there is no risk here, only that a view gets created twice, but OC will take care of that with it's deduplication logic.

skinowski:
@mantree Sounds good. The registration function could even have an optional span filter function. Then we can remove url, sanitize, 100 limit, 100 char limit and leave those up to the main.go/components.

api/server/spanconverter.go
```diff
@@ -0,0 +1,129 @@
...
+func (c *SpanConverter) ExportSpan(sd *trace.SpanData) {
+	if c.rejectSpan(sd) {
+		return
+	}
+	m := c.getMeasure(sd)
+
+	spanTimeNanos := sd.EndTime.Sub(sd.StartTime)
+	spanTimeMillis := float64(int64(spanTimeNanos / time.Millisecond))
+
+	stats.Record(context.Background(), m.M(spanTimeMillis))
```
rdallman: we should rip some of the attributes off the span data as well and put them onto the context. this may be slightly painful, but tags are likely important so that we have context for eg function_id, app_id... these seem to map to the fields `SpanData.Attributes` and `tag.Map`, however it's not immediately clear from digging around the docs.

api/server/spanconverter.go
```diff
@@ -0,0 +1,129 @@
...
+
+func (c *SpanConverter) getMeasure(span *trace.SpanData) *stats.Float64Measure {
+	sig := sanitize(span.Name)
+	c.viewsMu.Lock()
+	m, ok := c.measures[sig]
+	c.viewsMu.Unlock()
+
+	if !ok {
+		m = stats.Float64(sig+"_span_time", "The span length in milliseconds", "ms")
+		v := &view.View{
```
rdallman: this seems like the wrong place for view registration. we should maybe/probably `stats.Record` everything and not limit the `stats.Record` bit (it's gated by `rejectSpan` as well as `getMeasure` and an arbitrary 100 limit imposed). at least, `stats.Record` and `view.Measure` are separate processes and should be exclusive. Tolga linked to some pointers for how the other stuff is handled atm, I think initially we're mostly registering views in initialization, we can maybe make this more friendly (ie no reboot required) later but it's relatively nice to work with for the ability to toggle them on/off for eg things that extend fn. then we end up whitelisting things in, which I kind of agree is painful, while at the same time it does make us organize things (like stat/trace names) so it's a net good I think. I think we can remove the limits (100) with that in place.

rdallman: (also won't require a lock - we shouldn't be locking for a stats library, that is on them!)

api/server/spanconverter.go
```diff
@@ -0,0 +1,129 @@
...
+func sanitizeRune(r rune) rune {
+	if unicode.IsLetter(r) || unicode.IsDigit(r) {
+		return r
+	}
+	// Everything else turns into an underscore
+	return '_'
+}
+
+//Gin creates spans for all paths, containing ID values.
+//We can safely discard these, as other histograms are being created for them.
```
rdallman: space between first char always (sorry, am that guy)... `// We`

api/server/spanconverter.go
```diff
@@ -0,0 +1,129 @@
...
+		c.viewsMu.Unlock()
+	}
+
+	return m
+}
+
+const labelKeySizeLimit = 100
+
+// sanitize returns a string that is trunacated to 100 characters if it's too
+// long, and replaces non-alphanumeric characters to underscores.
```
rdallman: this is all stuff we control right? we're the ones making the names up? I don't think we need this (and if we do, we should probably be using tags for some of the stuff in the name?)

mantree: trouble is we don't control them all. See the problem with the Gin generated spans.

api/server/spanconverter.go
mantree: Everything has already been recorded, that's already happened to produce the SpanData. This is copying up the recorded values into the Measure.

api/server/spanconverter.go
mantree: Attributes, MessageEvents and Annotations are all empty for our spans.

mantree:
Ok, I've looked at pulling the Latency config from main, but it creates a cyclic dependency, so it's not going to happen. I've had a look at passing the value down and through the various layers to try and bring it to the Converter, but that ends up being a total mess, it just doesn't feel worth it.

I think to get this closer to the suggestions would really involve replacing the Spans with normal metrics. The idea here is to get *some* value from what we have, if they prove useful then that would strengthen the case to get Jaeger/Zipkin setup and make proper use of them.

rdallman:
> The idea here is to get some value from what we have, if they prove useful then that would strengthen the case to get Jaeger/Zipkin setup and make proper use of them.

hmm, not sure about motivation here if this is the case, I think they're 2 different goals. we lose all causal context by way of this method, and I think Jaeger would be useful in its own right without this (actually, I know it will be useful, because I've used it to fix bugs and performance tweaks already...). we just get a different set of data with Jaeger, especially since this is averaging down in distribution buckets, tail latency samples are lost -- Jaeger is immensely useful in this respect, even when sampling .01% of traffic. for this, it covers up holes in our metrics for things we're already gathering for traces, this is really useful too; docker latencies, for example, are a great example of when both are useful and we have doubled logic to get these things right now (ie before this) - maybe this is ok and what we should be doing instead, I'm not sure I have the answer, we can discuss (it sounds like you've run into some challenges, trust your judgment on this...)

api/server/server.go
```diff
@@ -743,6 +744,14 @@ func WithPrometheus() Option {
 	}
 }
 
+func WithSpanConverter() Option {
```
skinowski: can this take an argument for latency buckets? Not the end of the world, but it would be good to keep buckets consistent with service side.

PR #1241, opened 2018-09-26
HTTP trigger http-stream tests

rdallman:
this both fixes broken behaviors and adds tests to ensure that doesn't happen again easily, in addition to some general shuffling around.

* move headers biz up to front end completely, not in agent at all
* shot ResponseBufferingWriter, we can encapsulate all we need in an http.ResponseWriter
* got rid of cruft from fnInvoke (FXLB-WAIT, unused logger, tucked in evil jason detector)
* got rid of agent.FromHttpTriggerRequest... added agent.WithTrigger to add the id, much smaller
* removed Fn_call_id from output, replaced with Fn-Call-Id
* fixes broken stderr/stdout logging for http-stream functions... logs work now, and include stdout
* fixes broken handling of Content-Length in http triggers (no encap/decap w/ Fn-Http-H- )
* fixes broken Fn-Http-H- header output
* only Content-Length, Fn-Call-Id and Fn-Http-H-* headers are output from http triggers now (actually need to test this limitation more rigorously)
* adds test for Content-Length, Fn-Call-Id and Fn-Http-H- headers to come out properly
* adds test for Fn-Http-Status to be the actual http status on the response
* adds test for headers that hit a function are encapsulated properly (and then stripped, for fdk-go), and that Fn-Http-Request-Url, Fn-Http-Method, and Content-Type make it
* updates fdk-go so that the tests pass

... think that covers it. in some parts (fnInvoke) the diff is rough (lots o red) but the code itself reads pretty nice, might recommend the side by side diff/final for reviewing. tests give reasonable degree of confidence anyway...

### DIFF
--- a/api/server/runner_httptrigger.go
+++ b/api/server/runner_httptrigger.go
@@ -82,32 +82,43 @@ func (trw *triggerResponseWriter) WriteHeader(statusCode int) {
 	}
 	trw.committed = true
 
-	for k, vs := range trw.Header() {
-		if strings.HasPrefix(k, "Fn-Http-H-") {
-			// TODO strip out content-length and stuff here.
-			realHeader := strings.TrimPrefix(k, "Fn-Http-H-")
-			if realHeader != "" { // case where header is exactly the prefix
-				for _, v := range vs {
-					trw.Header().Del(k)
-					trw.Header().Add(realHeader, v)
+	var fnStatus int
+	realHeaders := trw.Header()
+	gwHeaders := make(http.Header, len(realHeaders))
+	for k, vs := range realHeaders {
+		switch {
+		case strings.HasPrefix(k, "Fn-Http-H-"):
+			gwHeader := strings.TrimPrefix(k, "Fn-Http-H-")
+			if gwHeader != "" { // case where header is exactly the prefix
+				gwHeaders[gwHeader] = vs
+			}
+		case k == "Fn-Http-Status":
+			if len(vs) > 0 {
+				statusInt, err := strconv.Atoi(vs[0])
+				if err == nil {
+					fnStatus = statusInt
 				}
 			}
+		case k == "Content-Type":
+			gwHeaders[k] = vs
 		}
 	}
 
-	gatewayStatus := 200
+	// XXX(reed): this is O(3n)... yes sorry for making it work without making it perfect first
+	for k := range realHeaders {
+		realHeaders.Del(k)
+	}
+	for k, vs := range gwHeaders {
+		realHeaders[k] = vs
+	}
 
+	// XXX(reed): simplify / add tests for these behaviors...
+	gatewayStatus := 200
 	if statusCode >= 400 {
 		gatewayStatus = 502
 	}
-
-	status := trw.Header().Get("Fn-Http-Status")
-	if status != "" {
-		statusInt, err := strconv.Atoi(status)
-		if err == nil {
-			gatewayStatus = statusInt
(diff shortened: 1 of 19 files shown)


api/server/runner_httptrigger.go
```diff
@@ -83,56 +82,99 @@ func (trw *triggerResponseWriter) WriteHeader(statusCode int) {
...
+	}
 
+	// XXX(reed): simplify / add tests for these behaviors...
+	gatewayStatus := 200
 	if statusCode >= 400 {
 		gatewayStatus = 502
 	}
+	if fnStatus > 0 {
+		gatewayStatus = fnStatus
+	}
```
mantree: Is there a gap here whereby a fn could set the fn-http-status code to 200, but then a failure occur resulting in a 500 response, which this block will then set back to 200?

To some degree that is a user problem, "don't set fn-http-status until your sure", but i'm thinking of ways fdk's could fail that might be outside their control that might lead to this scenario.

zootalures: Fixed - made this else if

api/server/runner_httptrigger_test.go
```diff
@@ -307,55 +309,86 @@ func TestTriggerRunnerExecution(t *testing.T) {
...
+	expFooHeadersBody := `{"echoContent": "_TRX_ID_",
+		"expectHeaders": {
+			"Content-Type":["application/hateson"],
+			"Fn-Http-H-Test-Header":["foo"],
+			"Fn-Http-Method":["POST"],
+			"Fn-Http-Request-Url":["http://127.0.0.1:8080/t/myapp/httpstream"]
+		},
+		"returnHeaders": {
+			"Test-Header":["foo"]
+		},
```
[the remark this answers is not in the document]
zootalures: added

api/server/runner_httptrigger_test.go
```diff
@@ -307,55 +309,86 @@ func TestTriggerRunnerExecution(t *testing.T) {
...
+	expFooHeaders := map[string][]string{"Content-Type": {"application/hateson"}, "Test-Header": {"foo"}}
+	expFooHeadersBody := `{"echoContent": "_TRX_ID_",
+		"expectHeaders": {
+			"Content-Type":["application/hateson"],
+			"Fn-Http-H-Test-Header":["foo"],
+			"Fn-Http-Method":["POST"],
+			"Fn-Http-Request-Url":["http://127.0.0.1:8080/t/myapp/httpstream"]
+		},
+		"returnHeaders": {
+			"Test-Header":["foo"]
```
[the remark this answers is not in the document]
zootalures: agree, fixed

api/server/runner_httptrigger.go
rdallman: i don't think we really pegged down this behavior well. honestly if the user manages to get an http status out of the function I see no reason to overwrite it with a 5xx for any reason from fn (this was the case without else if)

rdallman:
thanks for review and fixes... the gateway status thing I think needs fleshing out from an API perspective (and then solidified with tests). also the O(3n) in the header rewriter disturbs me but it's not pressing (n is probably never on the order of thousands).

rdallman:
from an end-consumer (i.e. I hit a web page and expect a web page back -- ie not a programmer) perspective, that makes the most sense (ie we shouldn't plumb out functions errors to end-consumers), and the function then presumably gets a log and error on the backend for that for the programmers to deal with.

PR #1246, opened 2018-09-27
Add APIErrorWrapper so that underlying errors can be logged

ostrain:
Loggable errors have a separate message that is logged only but is not part of the response.

### DIFF
--- a/api/models/error.go
+++ b/api/models/error.go
@@ -194,3 +194,26 @@ type ErrorWrapper struct {
 func (m *ErrorWrapper) Validate() error {
 	return nil
 }
+
+// APIErrorWrapper wraps an error with an APIError such that the APIError
+// governs the HTTP response but the root error remains accessible.
+type APIErrorWrapper interface {
+	APIError
+	RootError() error
+}
+
+type apiErrorWrapper struct {
+	APIError
+	root error
+}
+
+func (w apiErrorWrapper) RootError() error {
+	return w.root
+}
+
+func NewAPIErrorWrapper(apiErr APIError, rootErr error) APIErrorWrapper {
+	return &apiErrorWrapper{
+		APIError: apiErr,
+		root:     rootErr,
+	}
+}
--- a/api/server/error_response.go
+++ b/api/server/error_response.go
@@ -27,6 +27,9 @@ func handleErrorResponse(c *gin.Context, err error) {
 // HandleErrorResponse used to handle response errors in the same way.
 func HandleErrorResponse(ctx context.Context, w http.ResponseWriter, err error) {
 	log := common.Logger(ctx)
+	if w, ok := err.(models.APIErrorWrapper); ok {
+		log = log.WithField("root_error", w.RootError())
+	}
 
 	if ctx.Err() == context.Canceled {
 		log.Info("client context cancelled")
--- a/api/runnerpool/ch_placer.go
+++ b/api/runnerpool/ch_placer.go
@@ -34,12 +34,10 @@ func (p *chPlacer) PlaceCall(rp RunnerPool, ctx context.Context, call RunnerCall
 	key := call.Model().FnID
 	sum64 := siphash.Hash(0, 0x4c617279426f6174, []byte(key))
 
+	var runnerPoolErr error
 	for {
-		runners, err := rp.Runners(call)
-		if err != nil {
-			state.HandleFindRunnersFailure(err)
-			return err
-		}
+		var runners []Runner
+		runners, runnerPoolErr = rp.Runners(call)
 
 		i := int(jumpConsistentHash(sum64, int32(len(runners))))
 		for j := 0; j < len(runners) && !state.IsDone(); j++ {
@@ -59,6 +57,14 @@ func (p *chPlacer) PlaceCall(rp RunnerPool, ctx context.Context, call RunnerCall
 		}
 	}
 
+	if runnerPoolErr != nil {
+		// If we haven't been able to place the function and we got an error
+		// from the runner pool, return that error (since we don't have
+		// enough runners to handle the current load and the runner pool is
+		// having trouble).
+		state.HandleFindRunnersFailure(runnerPoolErr)
+		return runnerPoolErr
+	}
 	return models.ErrCallTimeoutServerBusy
 }
 
--- a/api/runnerpool/naive_placer.go
+++ b/api/runnerpool/naive_placer.go
@@ -28,12 +28,10 @@ func (sp *naivePlacer) PlaceCall(rp RunnerPool, ctx context.Context, call Runner
 	state := NewPlacerTracker(ctx, &sp.cfg)
 	defer state.HandleDone()
 
+	var runnerPoolErr error
 	for {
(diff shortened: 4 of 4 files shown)


api/server/error_response.go
```diff
@@ -46,7 +50,11 @@ func HandleErrorResponse(ctx context.Context, w http.ResponseWriter, err error)
...
 	} else {
-		log.WithError(err).WithFields(logrus.Fields{"stack": string(debug.Stack())}).Error("internal server error")
+		if le, ok := err.(models.LoggableError); ok {
+			log.WithError(err).WithFields(logrus.Fields{"stack": string(debug.Stack())}).Error(le.Message())
+		} else {
+			log.WithError(err).WithFields(logrus.Fields{"stack": string(debug.Stack())}).Error("internal server error")
+		}
 		statuscode = http.StatusInternalServerError
 		err = ErrInternalServerError
 	}
```
rdallman: after discussion and thinking on this, I've made the following table of assumptions, we have 3 (not distinct) kinds of errors presented here:

a) a `models.APIError`, whose error message should be both logged as well as output & presents statusCode
b) a `models.LoggableError`, whose error message should be logged but not output & 500 (in the case its <500, it's not logged)
c) a plain-ole-err, whose error message should be logged but not output & 500

in this respect, I don't find much of a difference between b & c, aside from an additional message that we are adding to errors (could this be encoded in the error? why or why not?), I am not sure of the value exactly there and can't imagine it's anything more descriptive than 'internal server error' without us wanting to use the more heavy-handed models.APIError to output something useful. I do see the fact that when using a models.APIError, it is both logged and output, however passing any non-models.APIError seems to give the value of logging the error but not outputting it. this begs the question, are there models.APIErrors where we want to encode a additional information to the user but not output that error? in any event, after drawing these out, i'm less sure of the utility provided here (all this just to change the log message line when we just need to pass a non-models.APIError?) and am working off the 'keep things as simple as possible' principle. could we clarify the value proposition here? what am I missing :) ?

api/agent/pure_runner.go
```diff
@@ -669,7 +669,7 @@ func (pr *pureRunner) Engage(engagement runner.RunnerProtocol_EngageServer) erro
 
 	// if we didn't respond with TooBusy, then this means the request
 	// was processed.
-	if err != models.ErrCallTimeoutServerBusy {
+	if err != nil {
```
rdallman: this will give us something more along the lines of 'requests received' vs 'requests handled' -- we probably want both, but at least this change makes requestsHandled misleading from its original definition (I am not sure of the existence of the other)

api/server/error_response.go
ostrain: My intent was to allow for cases where we want the HTTP request to return an error code other than 500 and we want to log some specifics about what happened (but not include all of those specifics in the HTTP response since it's nothing that's actionable by the user). The first use case I had in mind is a case where we currently return a 503 -- so I'd like to continue to return a 503, but log some additional information associated with the request.

Also, I hadn't noticed that we'll only log the message if the status is >=500. Is there any reason that couldn't be >= 400 instead?

api/agent/pure_runner.go
ostrain: Hmm, I'm not sure what the right thing to do is in that case. I was trying to account for the case where we get an error from the RunnerPool (other than models.ErrCallTimeoutServerBusy) -- in which case we never get as far as handling the request. Any suggestions of how to account for that without skewing this for other cases?

api/server/error_response.go
rdallman: > The first use case I had in mind is a case where we currently return a 503 -- so I'd like to continue to return a 503, but log some additional information associated with the request.

gotcha, from above I wasn't clear on how this would be different than c) - or passing an error that is _not_ of models.APIError, however, it seems like the case of primary interest _is_ a models.APIError and we want to decorate it with some additional information (to log, not to the user). I think that puts me on the same page? please correct if not. in any event, I'm not sure the message section is the best spot for this, it seems like we want the 'real error' and not the models.APIError masking it?

in the past I've seen this done this way https://github.com/iron-io/runner/blob/master/common/errors.go#L24 - I can't admit to being a fan (the bool is unnecessary, for one, either implement it or don't), but maybe it gets the idea juices flowing. the models.APIError thing does suppress errors, especially in the 5xx case, so I kind of buy this 2 different kinds of error thing.

for our 5xx errors, would something like this make sense or am I missing things?

```go
// LogErr attaches an additional error onto a models.APIError
type LoggableErr interface {
  models.APIError
  LogError() error
}

type loggableErr struct {
  models.APIError
  root error
}

func (le loggableErr) LogError() error { return le.root }

func NewLoggableAPIError(apiErr, rootErr error) LoggableErr {
  return &loggableErr{
    APIError: apiErr,
    root: rootErr,
  }
}

// where we're interested in handleErrResponse...
log := common.Logger(ctx)
if le, ok := (models.LoggableError); ok {
  log = log.WithField("root_error": le.LogError())
}

// leave the rest as is with the models.APIError dance ?...
```

this would prevent 'error suppression' from models.APIErrors (we could get at the root error), we wouldn't need to use the message field (confusing) with eg `err=xyz root_err=abc`, and most errors could use the old error trap for 503 (`c)` from above). I'm not sure it's the best solution or the one you're going for, if I'm still missing the point please let me know, and worst case maybe that offers some ideas.

> Also, I hadn't noticed that we'll only log the message if the status is >=500. Is there any reason that couldn't be >= 400 instead?

noise would be the primary argument against doing so, we could push muting things to the logging backend though (ie every `not found` gets logged, and that's on the user / not very informative to us usually). for OSS, it would make some sense though to log every >= 400.

rdallman: the inimitable dave cheney has done it this way https://github.com/pkg/errors which I think turned out pretty slick, adopting something in this vein would make sense (previous efforts to adopt the pkg/errors package wholeheartedly have not worked out well, but for our api error suppression, something along these lines makes sense)

api/agent/pure_runner.go
rdallman: ah, I see. it may be wise to put this specific statistic in call.End for pureRunner, but I'm not 100%. at least, call.End is only reached in pure runner if the task calls Submit, at which point we either ran it and it blew up or it worked. may want to consult @skinowski

api/server/error_response.go
ostrain: Nice, I like this better. Implemented something along the lines of what you suggested.

api/models/error.go
```diff
@@ -195,6 +195,25 @@ func (m *ErrorWrapper) Validate() error {
 	return nil
 }
 
-type LoggableError interface {
-	Message() string
+// APIErrorWrapper wraps an error with an APIError such that the APIError
+// governs the HTTP response but the root error remains accessible.
```
rdallman: 🥇

rdallman:
thanks for your patience w/ me, this LGTM now with the exception of https://github.com/fnproject/fn/pull/1246/files#r221094370 on which I'm still not sure, ping @skinowski

PR #1251, opened 2018-09-28
Fix missing context on request sent through UDS

hhexo:
Fixed the fact that the http.Request sent through UDS in http-stream does not have a Context.

Old dispatch methods use CallInfo and provide the original request (which has a Context), but the new dispatch method for http-stream through UDS creates a new request and does not set the context there. I have merely added a `WithContext` call.


rdallman:
thanks!

PR #1253, opened 2018-09-28
fn: disable keepalives for UDS http-stream client

skinowski:

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -965,6 +965,7 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
 		udsClient = http.Client{
 			Transport: &http.Transport{
 				// XXX(reed): other settings ?
+				IdleConnTimeout: 1 * time.Second,
 				DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
 					var d net.Dialer
 					return d.DialContext(ctx, "unix", filepath.Join(container.UDSAgentPath(), udsFilename))


api/agent/agent.go
```diff
@@ -965,6 +965,7 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
 		udsClient = http.Client{
 			Transport: &http.Transport{
 				// XXX(reed): other settings ?
+				DisableKeepAlives: true, // single request per connection.
```
rdallman: is this to avoid inode issue? any details? it seems useful to use keep alives intuitively, but inode thing kind of makes sense, if it's that, I'm not sure doing this fixes the inode issue (the first conn probably hangs around in TIME_WAIT/CLOSE_WAIT etc)

kojustin: @rdallman this is to track down a possible file descriptor leak. Some people have been seeing FD leaks in some of the runner agents. There is a hypothesis that it is related to the unix domain socket.

skinowski: Swapped this with idle timeout now... I thought we wanted to restrict these to 1 conn per req originally (despite overhead), but idle time is good enough to keep conn counts at sane levels for now.

rdallman: lsof should give a good answer if it's just a theory... I think this would exacerbate the issue, not help it...

rdallman: @skinowski should we add some other settings here too (can do elsewhere)? https://github.com/fnproject/fn/blob/master/api/agent/hybrid/client.go#L60 has a few... max idle conns per host at least, probably not tls stuff, probably don't need timeout either (we have dial context / req context) ?

PR #1257, opened 2018-10-01
fn: adding hot launcher eviction waiting

skinowski:
If checkLaunch triggers evictions, it must wait
for these eviction to complete before returning.
Premature returning from checkLaunch will cause
checkLaunch to be called again by hot launcher.
This causes checkLaunch to receive an out of
capacity error and causes a 503.

The evictor is also improved with this PR and it
provides a slice of channels to wait on if evictions
are taking place.

Eviction token deletion is performed *after*
resource token close to ensure that once an
eviction is done, resource token is also free.

### DIFF
--- a/api/agent/evictor.go
+++ b/api/agent/evictor.go
@@ -2,15 +2,19 @@ package agent
 
 import (
 	"sync"
+	"sync/atomic"
+
+	"github.com/fnproject/fn/api/id"
+
+	"github.com/sirupsen/logrus"
 )
 
 // Evictor For Agent
-// Agent hot containers can register themselves as evictable using
-// Register/Unregister calls. If a hot container registers itself,
-// a starved request can call PerformEviction() to scan the eligible
+// Agent hot containers register themselves to the evictor system.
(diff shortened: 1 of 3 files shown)


api/agent/evictor.go
```diff
@@ -71,39 +85,42 @@ func (tok *EvictToken) isEligible() bool {
...
 
 	e.lock.Lock()
 
-	// be paranoid, do not register if it's already there
 	_, ok := e.tokens[token.key.id]
-	if !ok {
-		e.tokens[token.key.id] = token
-		e.slots = append(e.slots, token.key)
+	if ok {
+		logrus.Fatalf("id collusion key=%+v", key)
```
rdallman: panic("") may be better... which will halt the goroutine and 503 instead of shut the server down

rdallman: there are a few other Fatalf I'd say we should probably do the same (or better, plumb the error, but if you're c++ programmer, used to making assertions, is ok life gets better...)

PR #1262, opened 2018-10-05
fn: adding docker events to stats

skinowski:
Streaming docker events is useful as we can record/capture some
asynchronous containers events such as out-of-memory. For now,
we record these in opencensus/prometheus stats.

A sample section of prometheus metrics is:
```
# HELP fn_docker_events docker events
# TYPE fn_docker_events counter
fn_docker_events{event_action="attach",event_type="container"} 1
fn_docker_events{event_action="connect",event_type="network"} 1
fn_docker_events{event_action="create",event_type="container"} 1
fn_docker_events{event_action="destroy",event_type="container"} 1
fn_docker_events{event_action="die",event_type="container"} 1
fn_docker_events{event_action="disconnect",event_type="network"} 1
fn_docker_events{event_action="kill",event_type="container"} 1
fn_docker_events{event_action="oom",event_type="container"} 1
fn_docker_events{event_action="pause",event_type="container"} 12
fn_docker_events{event_action="start",event_type="container"} 1
fn_docker_events{event_action="unpause",event_type="container"} 12
```

### DIFF
--- a/api/agent/drivers/docker/docker.go
+++ b/api/agent/drivers/docker/docker.go
@@ -53,6 +53,7 @@ func (r *runResult) Error() error   { return r.err }
 func (r *runResult) Status() string { return r.status }
 
 type DockerDriver struct {
+	cancel   func()
 	conf     drivers.Config
 	docker   dockerClient // retries on *docker.Client, restricts ad hoc *docker.Client usage / retries
 	hostname string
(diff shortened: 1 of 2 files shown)


skinowski:
If this looks OK, then I'll remove this case block in switch since it's now dead code:

https://github.com/fnproject/fn/blob/master/api/agent/drivers/docker/docker.go#L638

rdallman:
> If this looks OK, then I'll remove this case block in switch since it's now dead code:

not dead right? just... challenging. i think at the docker driver layer it's still useful, ie if someone is just running containers the docker driver is for that really and it's still useful info (for cold it's more obviously useful), but anyway it seems to offend people since it's not perfect, idk, computers suck in general, do as you please...

skinowski:
@rdallman The general default in the switch catches exit code already. We can leave it as is, but I think it just gives the impression that we explicitly process oom there. It's just misleading (sort of dead code) since it doesn't work as it looks. (Think of someone not familiar with the project looking at the code.) I'll leave it as is, we can sweep/clean later if we choose to.

PR #1265, opened 2018-10-05
fn: fixup possible go-routine leak

skinowski:
Make sure we unlock a blocked io.Copy()

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -784,10 +784,17 @@ func (s *hotSlot) dispatch(ctx context.Context, call *call) chan error {
 		}
 		common.Logger(ctx).WithField("resp", resp).Debug("Got resp from UDS socket")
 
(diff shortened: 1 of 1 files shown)


api/agent/agent.go
```diff
@@ -784,10 +784,17 @@ func (s *hotSlot) dispatch(ctx context.Context, call *call) chan error {
...
 		defer resp.Body.Close()
 
+		ioErrChan := make(chan error, 1)
+		go func() {
+			ioErrChan <- writeResp(s.cfg.MaxResponseSize, resp, call.w)
+		}()
+
 		select {
-		case errApp <- writeResp(s.cfg.MaxResponseSize, resp, call.w):
+		case ioErr := <-ioErrChan:
```
rdallman: up to you: you could do this with 1 channel that already exists ;) ping pong channel...

Issue #1266, opened 2018-10-08
AckSync with FN configured in LB mode


reclaro:
It would be nice to have a kind of fire-and-forget scenario available, a user wants to have a response back from the server as soon as the server has placed the call on a runner.
In the first version of this change we want to implement an mechanism to send a 202 response from the server as soon as the execution of the function starts.

In this version the user doesn't have a way to check if the function has terminated, if the function failed or succeeded.
In this version we are going to allow those kind of calls only if the server is configured in the LB mode (using the lb_agent.go).

Please note that this is NOT an async call, in this scenario we don't put a request of execution in any kind of queue, the server will try to execute it as soon as it can and it will respect the function timeout.

The new contract can define a specific timeout for the server to send an ack back, if the timeout expires it means that the server was unable to place a call in time and the execution never started.

abhirockzz:
Is there supporting documentation for this ?

reclaro:
@abhirockzz not yet, good point we need to document this feature.

rdallman:
closing, need to doc maybe

PR #1267, opened 2018-10-08
WIP: Feature/acksync response writer

reclaro:
This change implements a mechanism to send an ack back to the caller as soon as the function starts the execution, more details are reported in #1266


denismakogon:
One of the best things we have with an async model is the timeout. In `acksync` I'm not seeing anything related to an execution time, except the placement timeout. So, my point is: ok `async` approach wasn't scalable enough, I get that, however, there has to be a way to extend an execution timeout up to an async timeout constraint (AFAIR 1 hour).

Moreover, I don't think that sending whatever a process spits out to the `/dev/null` is a good way, I'm okay not having a chance to check call status, but as a developer, I can accept only `fire-and-forget-but-get-logs-in-syslog-at-least`. Is that something that's going to happen?

zootalures:
@denismakogan yeah agree about timeouts but think we should come back to that in a different context/PR, when we look at async itself there are a lot of other considerations to take into account there

denismakogon:
@zootalures i see, if `async` is still going to happen, that's fine. The only thing that bothers me with ths PR is an invocation method name, can it be `detached` instead of `acksync`?

reclaro:
@denismakogon I like detached, you know how hard is naming stuff ;-)

api/agent/pure_runner.go
```diff
@@ -559,9 +583,28 @@ func (pr *pureRunner) AddCallListener(cl fnext.CallListener) {
...
+
 func (pr *pureRunner) spawnSubmit(state *callHandle) {
 	go func() {
+		isAcksync := state.c.Type == models.TypeAcksync
+		// we keep track of the callHandle which will be used in the process to send the ack back
+		if isAcksync {
+			pr.saveCallHandle(state)
+		}
 		err := pr.a.Submit(state.c)
+
```
skinowski: we might as well call removeCallHandle() here too, but really not important. :-)

api/agent/pure_runner.go
```diff
@@ -864,6 +907,31 @@ func (pr *pureRunner) Status(ctx context.Context, _ *empty.Empty) (*runner.Runne
...
+	pr.callHandleLock.Unlock()
+	if ch == nil {
+		err = models.ErrCallHandlerNotFound
+	}
+	ch.enqueueAckSync(err)
+	return nil
+}
+
+// AfterCall called after a funcion is executed
+func (pr *pureRunner) AfterCall(ctx context.Context, call *models.Call) error {
```
skinowski: I think it would be safer to move the removeCallHandle to right after Submit() above. It's just easier to follow and we won't be at the mercy of the handler code... And simply turn this callback into a noop.

api/agent/pure_runner.go
```diff
@@ -864,6 +907,31 @@ func (pr *pureRunner) Status(ctx context.Context, _ *empty.Empty) (*runner.Runne
...
+func (pr *pureRunner) BeforeCall(ctx context.Context, call *models.Call) error {
+	if call.Type != models.TypeAcksync {
+		return nil
+	}
+	var err error
+	// it is an ack sync we send ResultStart message back
+	pr.callHandleLock.Lock()
+	ch := pr.callHandleMap[call.ID]
+	pr.callHandleLock.Unlock()
+	if ch == nil {
```
skinowski: This can be a panic/logrus.Fatal IMHO. I wouldn't add error codes to try to relay this error.

api/runnerpool/ch_placer.go
```diff
@@ -27,7 +27,9 @@ func NewCHPlacer(cfg *PlacerConfig) Placer {
 // Because we ask a runner to accept load (queuing on the LB rather than on the nodes), we don't use
 // the LB_WAIT to drive placement decisions: runners only accept work if they have the capacity for it.
 func (p *chPlacer) PlaceCall(rp RunnerPool, ctx context.Context, call RunnerCall) error {
-
+	if call.Model().Type == models.TypeAcksync {
```
skinowski: I still think this should be a new field in the config (eg. PlayerAsyncTimeout). We override the cfg here and that's violates the placer constructors. For example, for an unsuspecting developer, it would come out as a surprise where we override the original config (eg. NewNaivePlacer()) with a new one here. In OCI side, we construct these configs from env vars. (see lb main.go) We do not customize it yet, but we can anytime in case of emergencies. Because of that, if this becomes an extra field in the cfg, it's much easier to deal with from OCI side as well as OSS.

api/runnerpool/placer_config.go
```diff
@@ -13,9 +13,9 @@ type PlacerConfig struct {
 	PlacerTimeout time.Duration `json:"placer_timeout"`
 }
 
-func NewPlacerConfig() PlacerConfig {
+func NewPlacerConfig(timeout time.Duration) PlacerConfig {
```
skinowski: See oci lb, main.go to see how this would work there.. NewPlacerConfig() provides reasonable defaults. For tests, these fields are already public and you can easily override them.

api/agent/lb_agent.go
```diff
@@ -177,8 +185,41 @@ func (a *lbAgent) Submit(callI Call) error {
...
+		}
+	}
+}
+
+func (a *lbAgent) spawnPlaceCall(ctx context.Context, call *call, errCh chan error, isAckSync bool) {
+	if isAckSync {
+		var cancel func()
+		ctx = common.BackgroundContext(ctx)
+		// We don't want this to run indefinetely we need to guard this context
+		ctx, cancel = context.WithTimeout(ctx, time.Duration(60+call.Timeout)*time.Second)
```
skinowski: So, this 60 secs is related with the 30 secs we assume for async-ack placer config. Hardcoding this for now is OK, but let's add a comment about the math/rationale here. I think we said something like:

30 secs PlacerTimeout for AsyncAck + call.Timeout (inside container) + 30 sec headroom

to make sure we do not wait indefinitely, but also wait enough time for worst case.

skinowski:
@denismakogon Execution time is not changed in this PR, it is *call.Timeout* (amount of time in a container). I agree that this may need to be increased for async functions as longer runtimes are desired. But we can address that in a future PR as @zootalures commented.

api/agent/pure_runner.go
reclaro: are you sure we want to panic here? I see that we shouldn't be in this scenario, if we can't find a call handle something really bad happened, but is it worth to tear down the service for an acksync error?

api/runnerpool/ch_placer.go
reclaro: I agree, I am going to change it even if that was done by @JadeRedworth . ATM she is super busy with something else so I'll go ahead and modify it and make sure I'll inform her about it

api/runnerpool/placer_config.go
reclaro: I added a new public field so we don't change the signature for NewPlacerConfig

api/agent/lb_agent.go
skinowski: So, here the math is as follows:
a) placer could wait up to 30 secs, which means, it could wait 29.999 secs, then proceed to 'b'
b) container exec timeout
c) headroom: to account for delays in scheduling/network as well as docker-pull.

Now that I think about it, I think we do need to increase this to account for docker pull. I didn't think of docker pull duration before. We could keep it simple and set it to 360 secs + call.Timeout to make sure get close to 100% of cases as much as possible.

api/agent/pure_runner.go
reclaro: fixed the bug, thanks

zootalures:
@JadeRedworth  @reclaro  - looks OK in principal - feels like the value of Fn-Invoke-Type should be hard-validated somewhere - and also it worries me that `async` could get in there and do something unexpected (the removal of older async was mostly implemented by hard coding the `type` field of calls, which this undoes)

reclaro:
@zootalures good point about the risk to get the old async sneaking in

api/agent/lb_agent.go
```diff
@@ -258,5 +300,18 @@ func (a *lbAgent) handleCallEnd(ctx context.Context, call *call, err error, isFo
...
+	return w.origin.Header()
+}
+
+func (w *DetachedResponseWriter) Write(data []byte) (int, error) {
+	return w.origin.Write(data)
+}
+
+func (w *DetachedResponseWriter) WriteHeader(statusCode int) {
+	w.origin.WriteHeader(statusCode)
+	w.acked <- struct{}{}
```
rdallman: is this correct? we want to ack immediately after placement but before the task actually runs, won't this only be acked once the task has begun to write the response (in theory, near the end of the task)?

reclaro: this change adds a new call listener and if the call is detached in the BeforeCall method we send a RunnerMsg_ResultStart which will trigger the writing of the header on the lb_agent via the runner_client, so in detached mode we send an ack during the BeforeCall

api/agent/call.go
```diff
@@ -95,7 +97,7 @@ func FromHTTPFnRequest(app *models.App, fn *models.Fn, req *http.Request) CallOp
 			ID:    id,
 			Image: fn.Image,
 			// Delay: 0,
-			Type:   "sync",
+			Type:   invokeType,
```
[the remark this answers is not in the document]
reclaro: we added a function to validate the type a user can pass which address the concern raised about unchecked string passed by clients. In my understanding the Type is the right place to put it what is the alternative?

api/agent/lb_agent.go
```diff
@@ -177,8 +185,42 @@ func (a *lbAgent) Submit(callI Call) error {
...
+	call.w = &DetachedResponseWriter{
+		origin: call.ResponseWriter(),
+		acked:  make(chan struct{}, 1),
+	}
+	isDetached := call.Type == models.TypeDetached
+	rw := call.w.(*DetachedResponseWriter)
+
+	go a.spawnPlaceCall(ctx, call, errPlace, isDetached)
+
+	for {
```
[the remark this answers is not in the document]
reclaro: the idea is to ack all type of calls so as we can decide to use this information in the future for other purposes even during a regular sync call, I was thinking to record some metrics, for example.
Thinking more about it we can still achieve the same thing adding metrics in other places of the code so my use case is not a solid one and I don't have any other use-cases at the moment for justifying the send of the acks for all calls, so i am going to remove this assumption and going to simplify the code

api/agent/lb_agent.go
```diff
@@ -177,8 +185,42 @@ func (a *lbAgent) Submit(callI Call) error {
...
+	errPlace := make(chan error, 1)
+
+	call.w = &DetachedResponseWriter{
+		origin: call.ResponseWriter(),
+		acked:  make(chan struct{}, 1),
+	}
+	isDetached := call.Type == models.TypeDetached
+	rw := call.w.(*DetachedResponseWriter)
+
+	go a.spawnPlaceCall(ctx, call, errPlace, isDetached)
```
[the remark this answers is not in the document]
reclaro: with the assumption that we take care of the acks only for detached code (see my comment below) I like your idea, thanks.

test/fn-system-tests/system_test.go
```diff
@@ -286,6 +286,7 @@ func SetUpPureRunnerNode(ctx context.Context, nodeNum int) (*server.Server, erro
 	pureRunner, err := agent.NewPureRunner(cancel, grpcAddr,
 		agent.PureRunnerWithAgent(innerAgent),
 		agent.PureRunnerWithStatusImage(StatusImage),
+		agent.PureRunnerWithDetached(),
```
skinowski: I'd say let's just enable it.

api/agent/lb_agent.go
```diff
@@ -177,10 +208,40 @@ func (a *lbAgent) Submit(callI Call) error {
...
+	err := a.placer.PlaceCall(a.rp, ctx, call)
 	return a.handleCallEnd(ctx, call, err, true)
 }
 
+func (a *lbAgent) spawnPlaceCall(ctx context.Context, call *call, errCh chan error) {
+	var cancel func()
+	ctx = common.BackgroundContext(ctx)
+	// 30 secs PlacerTimeout for Detached + call.Timeout (inside container) + 360 sec headroom
+	// to make sure we do not wait indefinitely, but also wait enough time for worst case.
+	ctx, cancel = context.WithTimeout(ctx, time.Duration(360+call.Timeout)*time.Second)
```
rdallman: 360 is kinda a magic number here. this doesn't respect the cancellation of the calling request context, which may cancel or have a timeout associated with it itself (remove `common.BackgroundContext`). this seems like the right place to impose `DetachedPlacerTimeout`, really, and let everything underneath consume that. this goes back to an earlier comment about plumbing, if we make the Submit contract have an additional surface area for returning early on 202, stuff like this kind of falls out (PlaceCall would return after getting the call detached message, but before the task is finished, rather than us having to guess how long it might take). my real question would be for PlaceCall for detached can we make it just return earlier for detached rather than having 2 different calling paths? this would make it 'just fall out' and return from placeCall to the http handler after the call was 'placed' but not executed in 'detached' mode?

api/server/runner_fninvokedetached.go
```diff
@@ -0,0 +1,40 @@
...
+	if err != nil {
+		return err
+	}
+
+	if rw.Status > 0 {
+		resp.WriteHeader(rw.Status)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
reclaro: the only reason was to avoid an if path inside the normal invoke path. To me passing a different writer is the "big" thing here and having a different function makes it clearer the fact that we are doing something different. As it is more a question of personal flavo(u)r I am happy to go for the single method with an "if"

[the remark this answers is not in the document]
reclaro: that file is gone now

api/agent/lb_agent.go
```diff
@@ -26,6 +28,35 @@ type lbAgent struct {
...
+	return w.Headers
+}
+
+func (w *DetachedResponseWriter) Write(data []byte) (int, error) {
+	return len(data), nil
+}
+
+func (w *DetachedResponseWriter) WriteHeader(statusCode int) {
+	w.Status = statusCode
+	w.acked <- struct{}{}
```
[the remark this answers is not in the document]
reclaro: Because the new feature will be somethng not enabled by default and that it is likely that the all-in-one installation will go away, we agreed to have the Detached mode available only (if enabled) with a LB configuration, because of that I think we can leave the ack here. I might thing again to see if there is a better place for the ack channel but for now I leave it here

api/agent/pure_runner.go
```diff
@@ -864,6 +920,29 @@ func (pr *pureRunner) Status(ctx context.Context, _ *empty.Empty) (*runner.Runne
...
 
+// BeforeCall called before a function is executed
+func (pr *pureRunner) BeforeCall(ctx context.Context, call *models.Call) error {
+	if call.Type != models.TypeDetached {
+		return nil
+	}
+	var err error
+	// it is an ack sync we send ResultStart message back
+	pr.callHandleLock.Lock()
+	ch := pr.callHandleMap[call.ID]
```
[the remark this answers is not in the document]
reclaro: I don't think we can put the channel in the context as we can't pass it via gRPC protocol. This side of the code is the pure runner so leaving on a different machine and to communicate via the runner_client it needs to go via gRPC, the communication can't go via channel, am I right?
We could decide to change the runner protocol to allow a special message to signal the ack but we have discarded this solution at the beginning when we started to reason about this change. I think @skinowski suggested to not mess with the runner protocol as it is not necessary if we use the call listener.

api/agent/pure_runner.go
```diff
@@ -559,13 +584,35 @@ func (pr *pureRunner) AddCallListener(cl fnext.CallListener) {
...
 		state.enqueueCallResponse(err)
 	}()
 }
 
+func (pr *pureRunner) spawnDetachSubmit(state *callHandle) {
+	go func() {
+		pr.saveCallHandle(state)
+		err := pr.a.Submit(state.c)
+		pr.removeCallHandle(state.c.Model().ID)
+		state.enqueueCallResponse(err)
```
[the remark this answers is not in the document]
reclaro: i don't think we need, the lb_agent will retrun to the client with a 202 or a 50x error, then the actual execution of the function is executed on a separate routine and once it will complete we shut down everything as it was a regular call. I can draw a graph of what I think this works so you can see if it makes sense to you and if I miss something important

api/agent/lb_agent.go
reclaro: I refactored the code here, please havea look and see if it makes sense. 
In my understanding at this point  we have to create a new context and we can't reuse the original context coming with the request as the original request will be closed as soon as we return the 202 and that will cause the abort of the running function. Is my understanding correct?

@skinowski as regards the new timeout of the new context I think it should be call.Timeout + DetachPlacerTimeout, again I don't see why we should add extra room here and we don't do the same for the regular sync call

api/agent/lb_agent.go
```diff
@@ -227,18 +227,19 @@ func (a *lbAgent) placeDetachCall(ctx context.Context, call *call) error {
 }
 
 func (a *lbAgent) placeCall(ctx context.Context, call *call) error {
-	err := a.placer.PlaceCall(a.rp, ctx, call)
+	err := a.placer.PlaceCall(a.rp, ctx, call, a.placer.Config().PlacerTimeout)
```
rdallman: we don't need to add the parameter to the placer interface - we can pass a context that has placerTimeout+callTimeout from right here to encode that info when applicable (when detached)

api/agent/lb_agent.go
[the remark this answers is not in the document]
rdallman: ack seems like it should be in BeforeCall in order to be 'obviously correct' ? regardless of which config (lb/full) we are running in.

api/server/runner_fninvoke.go
```diff
@@ -74,19 +78,20 @@ func (s *Server) fnInvoke(resp http.ResponseWriter, req *http.Request, app *mode
...
-
-	opts := []agent.CallOpt{
-		agent.WithWriter(&writer), // XXX (reed): order matters [for now]
-		agent.FromHTTPFnRequest(app, fn, req),
-	}
-	if trig != nil {
-		opts = append(opts, agent.WithTrigger(trig))
+	isDetached := req.Header.Get("Fn-Invoke-Type") == models.TypeDetached
+
+	var writer ResponseBuffer
```
rdallman: slick 👍

api/server/runner_fninvoke.go
```diff
@@ -103,11 +108,36 @@ func (s *Server) fnInvoke(resp http.ResponseWriter, req *http.Request, app *mode
...
+			agent.WithWriter(rw), // XXX (reed): order matters [for now]
+			agent.FromHTTPFnRequest(app, fn, req),
+			agent.InvokeDetached(),
+		}
+	} else {
+		opts = []agent.CallOpt{
+			agent.WithWriter(rw), // XXX (reed): order matters [for now]
+			agent.FromHTTPFnRequest(app, fn, req),
+		}
+	}
```
rdallman: nit: this could be 5 loc instead of 11, using append when detached, and less risk of dupe code mistakes

api/agent/pure_runner.go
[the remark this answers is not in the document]
skinowski: @reclaro In this case, you'd be adding the channel inside the ctx in the runner already. So, it does not need to cross gRPC.

api/agent/lb_agent.go
skinowski: As @reclaro says, we should not respect caller ctx timeout here. This is "background" detached mode in LB. It means, the request is being run as if client is still present, hence common.BackgroundContext(). We no longer care about client ctx basically. 

Here the rationale is this:

* 30 sec placer timeout means, LB might wait up to 30 secs waiting to push the call to a runner
* call.Timeout means time inside a container
* 360 headroom means anything else (docker-pull, runner scheduler delay, gRPC/network retransmit, etc. etc.)

New changes are coming in runner soon to bound the docker-pull and other latencies. So, this 360 needs to be in coordination of that. We should move the constant to a config/env var.

The code however is not in sync with the comments. Math there seems like "360 + call.Timeout" which omits placer timeout.

rdallman: placer timeout from the update makes sense, thanks. 

i'm not sure that's the case wrt ctx cancel with lb_agent/pure_runner, they have a protocol to handle moving the call state forward that is more 'stateless' in nature. in lb_agent using the request context won't cancel a task running in the pure runner, as far as I know, I only see one handle of cancellation in lb_agent and it doesn't shut down the task (my understanding is, it would have to handle it explicitly to do this, and I don't see it's doing that). this just means the placer timeout is imposing a server side configuring client timeout for detached tasks, basically. while we do need to bound it, it's interesting to think about just allowing the http request to specify this timeout (it's not a hard req or blocking here, what's here with placer timeout looks ok wrt timings)

rdallman: lb_agent doesn't appear have timeouts around each state machine for each call, so i'm not sure the assertions about ignoring request context are true. in any case, the 360 number really what we want is 30 here? and lb_agent shouldn't be coordinating timeouts with the runner at all because the protocol doesn't time out or cancel any individual tasks? not that the protocol is well defined in that respect

api/agent/lb_agent.go
skinowski: @rdallman I'm ok with reducing arguments here, but if you actually meant setting a deadline for that ctx, then it's a bit tricky to do that using that ctx. Here ctx is a detached client that's kept alive by LB using some magic timeout numbers (see above) to set an upper bound of total wait time. Placer timeout however is a subset (special) timer that's only active when searching for a runner. This timer  gets ignored if a runner is found.

rdallman: just a few lines later https://github.com/fnproject/fn/pull/1267/files#diff-a4aff9524fa41710d2ab6943f2bbc713R243 it ends up doing this

api/agent/lb_agent.go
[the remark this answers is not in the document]
skinowski: I don't think we can ack in BeforeCall() in LB. The call has to be forwarded to a runner and runner has to respond with 202.

api/agent/lb_agent.go
skinowski: Math in timeout calc in there is wrong.

api/agent/lb_agent.go
```diff
@@ -177,10 +212,41 @@ func (a *lbAgent) Submit(callI Call) error {
...
+func (a *lbAgent) placeCall(ctx context.Context, call *call) error {
+	err := a.placer.PlaceCall(a.rp, ctx, call, a.placer.Config().PlacerTimeout)
 	return a.handleCallEnd(ctx, call, err, true)
 }
 
+func (a *lbAgent) spawnPlaceCall(ctx context.Context, call *call, errCh chan error) {
+	var cancel func()
+	ctx = common.BackgroundContext(ctx)
+	placerTimeout := a.placer.Config().DetachedPlacerTimeout
+	// PlacerTimeout for Detached + call.Timeout (inside container)
```
skinowski: Here, I think the math needs to include docker-pull and other network/scheduler latencies. New upcoming changes to Fn will put upper bounds to this, so we should leave this headroom as a configurable, where ctx timeout needs to be placerTimeout + call.Timeout + headroom.

skinowski: Remember here we are simulating a connected client, but placer timeout is a specific shorter timeout inside placers (eg. 30 secs). This means, we can only tolerate 30 secs of searching for a runner. On the other hand, clients should be able to wait for potentially long docker-pull and other scheduler/network latencies in addition to placer timeout and call.Timeout.

rdallman: we can leave headroom up to client timeouts? placer timeout as well is pretty much just a client timeout for detached specifically? at least, wrt placer timeout seems like we can zoom out a little, shouldn't be thinking about call timeout even really should we? if a client sets a 2s timeout on their http request, doesn't do us a lot of good to set 30s call timeout to enclose that - they seem like separate things.

api/agent/pure_runner.go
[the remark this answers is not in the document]
reclaro: hold on I am confused now, which channel are we talking about?  the only channel I have is on the LB side to return from the Submit method and this channel can't cross gRPC.
Are here we talking about putting the callHandler in the context? is the misunderstanding coming from the name of the variable called ch? if so then yes I agree that the callHandler is just a thing for runner and agent so it is not crossing the gRPC pond. I am always a bit reluctant to abuse the context but as Reed said the callHandler is thread local so it could be ok to add it here.

api/agent/lb_agent.go
reclaro: as pointed out by @skinowski we can't pass just via the context as we need to differentiate between the detach placement timeout and the request timeout. I don't feel terrible to add the placertimeout to the PlaceCall interface, this change is all about set different timeouts so the single placeTimeout is not enough anymore. I can add a method to the placer to set the placerTimeout and use just that timeout in the PlacerTracker but to me that makes the code less readable.

api/server/runner_fninvoke.go
reclaro: done

api/runnerpool/runner_pool.go
```diff
@@ -13,7 +14,8 @@ import (
 // Placer implements a placement strategy for calls that are load-balanced
 // across runners in a pool
 type Placer interface {
-	PlaceCall(rp RunnerPool, ctx context.Context, call RunnerCall) error
+	PlaceCall(rp RunnerPool, ctx context.Context, call RunnerCall, placerTimeout time.Duration) error
+	Config() PlacerConfig
```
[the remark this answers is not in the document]
reclaro: done

api/agent/pure_runner.go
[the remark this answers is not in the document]
reclaro: mmh I looked at it, the pure_runner is where we create the callHandler then the pure_runner calls the agent.Submit passing just the call. Adding the callhandler to the context is not easy, we can pollute the call.req.Context() adding the callHandler or adding a new context field to the call just to store the callHandler? I don't think the latter is a good solution at that point I'd add a callHandler field to the call but that even more dirty as the callHandler has a field for the call so we have a kind of loop dependency which seems wrong. The other solution is to leave the lock in place here, I heard your voice as regards lock nightmare, is there anything we can do to make it more solid?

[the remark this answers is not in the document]
rdallman: that's okay, it was just an idea anyway. this code appears to work and we can improve later.

PR #1272, opened 2018-10-16
Update promclient and its usage

hibooboo2:
- What I did
Update prom client to version 9

Update to the new version of promethus client


api/server/server.go
```diff
@@ -731,13 +731,20 @@ func New(ctx context.Context, opts ...Option) *Server {
 func WithPrometheus() Option {
 	return func(ctx context.Context, s *Server) error {
 		reg := promclient.NewRegistry()
-		reg.MustRegister(promclient.NewProcessCollector(os.Getpid(), "fn"),
+		reg.MustRegister(promclient.NewProcessCollector(promclient.ProcessCollectorOpts{
+			PidFn:     func() (int, error) { return os.Getpid(), nil },
```
rdallman: kinda nitty but our pid for our own process won't change so we don't need the func version for this one (I did not look if it went away), saves a syscall every collect interval or whatever (which is not much, so whatever)

hibooboo2: It went away :(

PR #1275, opened 2018-10-18
Favour user fn response codes for triggers, with caveats

mantree:
Currently, this is a little unclear and there is a problem with user
fns returning error responses which we can not differentiate from our
service errors.

This changes breaks the status selection logic out and adds some tests that spell out
how we want this fiddly bit to work. The offering here is: user fn
status returned, if set, but modify 5xx to 502. If no response code
from user fn, use one the service has deemed appropriate. Default to
200 if neither of them have commented.

### DIFF
--- a/api/server/runner_httptrigger.go
+++ b/api/server/runner_httptrigger.go
@@ -112,15 +112,23 @@ func (trw *triggerResponseWriter) WriteHeader(statusCode int) {
 		realHeaders[k] = vs
 	}
 
-	// XXX(reed): simplify / add tests for these behaviors...
-	gatewayStatus := 200
-	if statusCode >= 400 {
-		gatewayStatus = 502
-	} else if fnStatus > 0 {
-		gatewayStatus = fnStatus
+	trw.inner.WriteHeader(determineFinalStatus(statusCode, fnStatus))
+}
+
+func determineFinalStatus(serviceStatus, userFnStatus int) int {
+	finalStatus := 200
+
+	if userFnStatus > 0 {
+		if userFnStatus >= 500 {
+			finalStatus = 502
+		} else {
+			finalStatus = userFnStatus
+		}
+	} else if serviceStatus > 0 {
+		finalStatus = serviceStatus
 	}
 
-	trw.inner.WriteHeader(gatewayStatus)
+	return finalStatus
 }
 
 var skipTriggerHeaders = map[string]bool{
--- a/api/server/runner_httptrigger_test.go
+++ b/api/server/runner_httptrigger_test.go
@@ -11,12 +11,16 @@ import (
 	"context"
 	"os"
 
+	"reflect"
+
 	"github.com/fnproject/fn/api/agent"
 	"github.com/fnproject/fn/api/datastore"
 	"github.com/fnproject/fn/api/logs"
 	"github.com/fnproject/fn/api/models"
 	"github.com/fnproject/fn/api/mqs"
-	"reflect"
+	"github.com/leanovate/gopter"
+	"github.com/leanovate/gopter/gen"
+	"github.com/leanovate/gopter/prop"
 )
 
 func envTweaker(name, value string) func() {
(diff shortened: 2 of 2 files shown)


hhexo:
Seems good to me.

mantree:
@rdallman I think you should take a glance at this change, you to @zootalures.

This seems like the right way to go, but is one of those changes that modifies mental models.

api/server/runner_httptrigger.go
```diff
@@ -112,15 +112,23 @@ func (trw *triggerResponseWriter) WriteHeader(statusCode int) {
...
+	finalStatus := 200
+
+	if userFnStatus > 0 {
+		if userFnStatus >= 500 {
+			finalStatus = 502
+		} else {
+			finalStatus = userFnStatus
+		}
+	} else if serviceStatus > 0 {
+		finalStatus = serviceStatus
```
rdallman: thanks for extracting func on this. big readability improvement for me would be to return statuses instead of falling through a bunch of branches (always feels like a newspaper maze puzzle to me anyway), also I just wanted to try the new suggestion thing, here goes: 

```suggestion
func determineFinalStatus(serviceStatus, userFnStatus int) int {
                switch {
                case userFnStatus > 400:
                  return 502
                case userFnStatus > 0:
                  return userFnStatus
                case serviceStatus > 0:
                  return serviceStatus
                default:
                  return 200
                }
}
```

rdallman: hmm, need some way to suggest over a range of lines, cool idea though something I didn't know I wanted (I suggest a lot)

api/server/runner_httptrigger.go
```diff
@@ -112,15 +112,23 @@ func (trw *triggerResponseWriter) WriteHeader(statusCode int) {
...
-	} else if fnStatus > 0 {
-		gatewayStatus = fnStatus
+	trw.inner.WriteHeader(determineFinalStatus(statusCode, fnStatus))
+}
+
+func determineFinalStatus(serviceStatus, userFnStatus int) int {
+	finalStatus := 200
+
+	if userFnStatus > 0 {
+		if userFnStatus >= 500 {
```
rdallman: I guess 400 made sense to me, since we're kind of acting as a gateway in this case on the function's behalf, if the function gives us a 400 back on the actual function response (note: NOT `Fn-Http-Status`), it means fn gave it a botched request of some kind and it would translate to a server side error I think? it's really confusing anyway and I don't think we have a real case for it at the moment anyway, either way.

rdallman: actually, I take this all back. I don't think we should rewrite user status codes at all, it kind of defeats the whole point. I was thinking this was the function status code, not the user status code, which makes sense to futz with. I think this was a mistranslation from previous code, user codes should not be rewritten, and in the previous case the container returning a status code overrode the function status code (this part makes less sense to me, if the user manages to get a status out of there we should probably write it, but it may be the case that they didn't get a chance to write the body as they wanted, so maybe not? ugh)

zootalures:
I think I’d rather trigger status codes didn’t propagate if the function was in error. I see this as nesting of “what did the service say?” takes precedence over “what did the FDK say” takes precedence over “what did the user code say”

To a certain extent I’m wondering if propagating any headers (possibly including partial ones that the function set when it was still on the happy path) is right

api/server/runner_httptrigger.go
```diff
@@ -112,15 +112,23 @@ func (trw *triggerResponseWriter) WriteHeader(statusCode int) {
...
+			finalStatus = 502
```
zootalures: Hmm - not super hot on this  

My 2c : 
  - if the function failed to execute correctly we should always return a 502 irrespective of the user status (possibly with exception of 400 which we might honour) 
 -  if the user gave us a status and the function wasn't in error we should honour it 

if we're depending on HTTP status for much downstream then we should fix this.

api/server/runner_httptrigger_test.go
```diff
@@ -611,3 +615,61 @@ func TestTriggerRunnerMinimalConcurrentHotSync(t *testing.T) {
 		}
 	}
 }
+
+//Generally we want to return the user function status, only making an adjustment to
+//502 for 5xx codes. Otherwise, we return the code that our service has selected.
+//Failing any information, we just give it a 200.
+func TestFinalStatusSetting(t *testing.T) {
```
zootalures: Kind of feel these tests would be easier to read if they were case based. 

Given user error of X and function error of Y : expect endpoint error of Z

skinowski:
As @zootalures pointed out, we've already seen an example of this happening where FDK emits a 500, but depending on the failure type & timing of events, that could easily end up being a 502 due to IO failure detected by the agent. We should also keep in mind **both** invokes and trigger cases where we freely read and propagate the http status code from FDK (user-side). The status codes in general I think need a context, that's why I was suggesting inserting a header to show origin. For example:

- User errors due to I/O/parsing(*) are flagged as 502 => origin=function (502 is punched in by agent.go)
- If  we ever read a http status code from FDK => origin=function
- everything else => origin=service

Then we can use this for stats as a tag, etc. Or even include it in http headers in responses to clarify where the http code originated from.

(*) We'll need to carefully cherry pick some of these, as some cases may be due to service issue, eg. docker failure. Today, UDS I/O failure or docker-exit is considered a 502 by agent code.

api/server/runner_httptrigger.go
mantree: Ooo i like that. Yeah this was well fugly, was more interested in the what over how. I way prefer your how.

api/server/runner_httptrigger.go
mantree: So you'd like to see

if userFnStatus >= 400 {
  finalStatus = 502
}

I dunno, I guess i'm just not that bothered about 4xx, let them send them if they want.

api/server/runner_httptrigger_test.go
mantree: I disagree on this one. Writing out the properties we want from the code is way clearer than the case based style where you have to translate it all and all the check logic is smudged together.

This style has already proved it's worth to my mind (with the config equality bug), and even here they helped steer the code nicely.

api/server/runner_httptrigger.go
zootalures: hmm actually I take it back I think the above is OK -

zootalures: I think one of the things i want to be able to implement in triggers is REST apis - so any <500 codes should be honoured  - agree on angling towards coalescing all >=500 codes to 502 though

zootalures: I think one of the things i want to be able to implement in triggers is REST apis - so any <500 codes should be honoured  - agree on angling towards coalescing all >=500 codes to 502 though

rdallman: that would maybe be okay as a default policy on an API gateway that a user could modify but for users that want to run a website off of fn http triggers with their own 500 codes that their services use to interact with one another, this seems like way too low of a level to be rewriting status codes, especially on the basis that the only reason we're staring at these things is because we don't want our stats blowing up when users return their own codes - this seems a selfish and lazy solution to take. say, they retry on 503, now they suddenly get every 501 and 503 as a 502 and they have to encode the actual code in their error message (what are status codes for anyway?) - this is _at best_ frustrating for users to realize that they wish they had found this so that they didn't waste so much time trying to integrate to realize they have to go change all their error messages to migrate over to functions and it's going to be a larger migration than they originally thought. without any way for users to return their own codes in some capacity (whether it's via an api gateway or just straight from their function) I think this solution of rewriting >500 to 502 is not justifiable - and alternative solutions exist to solve the problem we are having currently.

skinowski:
related #1314

api/server/runner_httptrigger.go
skinowski: I agree with @zootalures regarding precedence here. I think the logic here simply should be:

```
finalCode := serviceCode
if serviceCode < 400 {
    // copy over Fn-Http-Status or 502
}
```

Which I think doesn't change the outcome, but it's better to be defensive and guard against LB&Runner that sets a service error despite the existence of Fn-Http-Status header.

skinowski:
FYI: #1337

mantree:
#1337  has superseded these ideas.

PR #1278, opened 2018-10-19
Update dependencies

hibooboo2:
- What I did
Update dependencies
Update dependencies


Gopkg.toml
```diff
@@ -24,13 +24,18 @@ ignored = ["github.com/fnproject/fn/cli",
...
+  non-go = true
+
 [[constraint]]
   name = "github.com/boltdb/bolt"
   revision = "fa5367d20c994db73282594be0146ab221657943"
 
 [[constraint]]
   name = "github.com/gin-gonic/gin"
-  version = "v1.2"
+  version = "v1.3.0"
```
rdallman: we need 1.3.1 for this bug fix https://github.com/fsouza/go-dockerclient/releases/tag/v1.3.1

Gopkg.toml
```diff
@@ -24,13 +24,18 @@ ignored = ["github.com/fnproject/fn/cli",
            "github.com/Microsoft/go-winio*",
            "github.com/Azure/go-ansiterm*"]
 
+[prune]
+  go-tests = true
+  unused-packages = true
+  non-go = true
```
rdallman: the other 2 seems fine but `non-go` we may be missing some assembler/c files (sqlite, at least?) we might like to have the code from for optimization, license files for legal reasons. while it's probably a good idea to run tests on all of our dependencies especially for patches like this one, it's not something we have been doing...

Gopkg.toml
```diff
@@ -42,8 +47,7 @@ ignored = ["github.com/fnproject/fn/cli",
 
 [[constraint]]
   name = "github.com/sirupsen/logrus"
-  revision = "89742aefa4b206dcf400792f3bd35b542998eb3b"
-
+  revision= "v1.0.6"
```
rdallman: since we're here https://github.com/sirupsen/logrus/blob/master/CHANGELOG.md v1.1.0 has some good looking bugs, like fixing a race for formatting

Gopkg.toml
rdallman: sorry, wrong dep. moving comment...

Gopkg.toml
```diff
@@ -56,11 +60,11 @@ ignored = ["github.com/fnproject/fn/cli",
 
 [[constraint]]
   name = "github.com/fsouza/go-dockerclient"
-  revision = "1b33efcb40bef3494af8bc6a1f73600f37a3c9d1"
+  revision = "v1.3.0"
```
rdallman: we need 1.3.1 for this bug fix https://github.com/fsouza/go-dockerclient/releases/tag/v1.3.1

Gopkg.toml
hibooboo2: License files are not removed...
87 are still in vendor dir. 👍

Gopkg.toml
```diff
@@ -83,5 +87,9 @@ ignored = ["github.com/fnproject/fn/cli",
...
 
 [[constraint]]
-    name = "github.com/prometheus/client_golang"
-    version = "0.9.0-pre1"
+  name = "github.com/prometheus/client_golang"
+  version = ">=0.9.0"
+
+[[constraint]]
+  name = "golang.org/x/net"
+  branch = "master"
```
rdallman: this one wasn't getting updated when running `dep ensure -update` ?

hibooboo2: I just wanted to be explicit. Since thats what we use.

Gopkg.toml
```diff
@@ -56,11 +60,11 @@ ignored = ["github.com/fnproject/fn/cli",
 
 [[constraint]]
   name = "github.com/fsouza/go-dockerclient"
-  revision = "1b33efcb40bef3494af8bc6a1f73600f37a3c9d1"
+  revision = "v1.3.0"
 
 [[constraint]]
   name = "github.com/coreos/go-semver"
-  version = "0.2.0"
+  revision = "e214231b295a8ea9479f11b70b35d5acf3556d9b"
```
rdallman: do we need the constraint on this dep? (we don't need constraints on every dep) - am not sure why this version but this dep is really minimal and just used for checking docker's version...

hibooboo2: I was told lock to this one.

Gopkg.toml
rdallman: can you remove please? as few constraints as possible helps, dep will handle it until it's an issue

hibooboo2: K

Gopkg.toml
rdallman: thanks

PR #1279, opened 2018-10-23
fn: uds timeout and cookie run ordering

skinowski:
1) UDS timeout is now set with request/call timeout. UDS wait
is moved to hot slot execution. This means, UDS startup time
is part of user function call execution. Alternatively, we
could set specific timeouts for this, but bundling it to
function call timeout seemed simpler.

2) Cookie Run in case of HTTP Stream format is very similar
to UDS timeout. Reordered this so that UDS wait accomplishes
the same barrier. This is a significant change as it incorporates
function/container start time into function execution time. This
means a cold start function needs to take into account this duration
as well as UDS startup time *and* function execution time.

This means, if a Java container with significant startup time of
10 seconds, with UDS startup delay of another 1 second, should
allocate at least 12 seconds of timeout for the function. In this
example, function execution is left with 1 seconds if cold, 12
seconds if hot execution.

Previously, both of these UDS startup and Cookie Run did not
have associated timeouts.


rdallman:
thanks, yea, I hit this one too... I have a patch I need to clean up but I think it's a bit simpler, there's less moving parts (just moved the context around, instead of moving the parts): https://github.com/fnproject/fn/compare/fdk2-test-image?expand=1#diff-02ea3c2166e8abe7a3c4684f68d78db5R781

skinowski:
Prefer to merge this after #1281 to remove branchy code.

PR #1282, opened 2018-10-25
fn: eviction resource correction

skinowski:
Previously evictor did not perform an eviction
if total cpu/mem of evictable containers was less
than requested cpu/mem. With this change, we
try to perform evictions based on actual needed cpu & mem
reported by resource tracker.


hhexo:
Seems good. I considered whether the "needed capacity" in the resource tokens could get out of date / stale, but in the worst case we just evict one more container because "at some point the capacity was needed", and I think that's acceptable.

rdallman:
i'm gonna move this into the resource thing as soon as I'm free :)

PR #1284, opened 2018-10-26
go modifyed

ericfode:
- Link to issue this resolves
So many

- What I did
Used Go Mod

- How I did it
Quickly

- How to verify it
go mod download; go mod vendor -v

Burned go dep. In a fire.

packing up those mods
Packing up those mods.


rdallman:
Thanks, this is great. I'm generally on board with the direction here. I see a few concerns at the moment:

* certain transitive dependencies seem to have been removed by mod, a quick glance I find 2 (valid) transitive deps: `golang.org/x/text` `github.com/petar/GoLLRB`. even if the following point is possible, this is a little concerning just for the status of `go mod` but I'm optimistic that it's just part of the migration or something we can fix?
* is it possible to swap out the `Gopkg.*` files for the `go.sum` without any update of the dependencies here? whereby running `go mod tidy` or `go mod vendor` or `go mod verify` would not cause any `vendor` files to change (pardon my unfamiliarity with the tool, I believe I'm mostly talking about `verify` or `vendor` though, where the `go.sum` file would ensure the `vendor/` has all matching deps in the same manner as `dep ensure` did)?  it would make this PR really easy to green light if this is possible. ie it's not clear whether this patch moves to `go mod` or moves to `go mod` and uses `go mod` to update the dependencies (perhaps this could just be clarified). Any delta of deps would be nice as it's not easy to see from the diff, if leaving `vendor/` alone isn't possible.
* will packages that import `fn` but still use `dep` handle this okay? I'm optimistic, I just honestly do not know. I know we have at least 2.

ericfode:
Current status on each question.
certain transitive dependencies seem to have been removed by mod, a quick glance I find 2 (valid) transitive deps: golang.org/x/text github.com/petar/GoLLRB. even if the following point is possible, this is a little concerning just for the status of go mod but I'm optimistic that it's just part of the migration or something we can fix?
I can explicitly include that. The indication I got from the docs is that it will prune modules that are not reachable. So I'm not sure what happened there. 

is it possible to swap out the Gopkg.* files for the go.sum without any update of the dependencies here? whereby running go mod tidy or go mod vendor or go mod verify would not cause any vendor files to change (pardon my unfamiliarity with the tool, I believe I'm mostly talking about verify or vendor though, where the go.sum file would ensure the vendor/ has all matching deps in the same manner as dep ensure did)? it would make this PR really easy to green light if this is possible. ie it's not clear whether this patch moves to go mod or moves to go mod and uses go mod to update the dependencies (perhaps this could just be clarified). Any delta of deps would be nice as it's not easy to see from the diff, if leaving vendor/ alone isn't possible.
It uses the Gopkg.* files to generate it's initial list of dependencies. You have to do some replace commands to get everything to work. `go mod vendor` causes it to use the go.mod versions to populate the vendor directly, no updates are caused as long as you run your go commands with the `-mod=vendor`. If you do not include this flag it will implicitly to minor version updates.

will packages that import fn but still use dep handle this okay? I'm optimistic, I just honestly do not know. I know we have at least 2.
My understanding is that it should still work... But I am still experimenting. No claims yet.

rdallman:
> got rid of the vendor directory

we learned this lesson a long time ago, checking in the vendor directory was worth it to prevent users from having dependency issues when getting started (maybe dampened now to new devs), as well as not having dependency issues among the active team, as well as bringing down CI times. please add it back :)

ericfode:
Haha ok :) I was ambivalent about it. Im glad you have a opinion about it.

Gopkg.lock
```diff
@@ -123,8 +123,11 @@
 [[projects]]
```
rdallman: this file should be removed?

rdallman:
> github.com/fnproject/fn/vendor/github.com/Azure/go-ansiterm
> github.com/fnproject/fn/vendor/github.com/Microsoft/go-winio

these were previously excluded (as they're only needed to build on WIN), am not sure if they fit compliance offhand or if there is a way for go mod to ignore these once again.

> github.com/fnproject/fn/vendor/github.com/petar/GoLLRB/llrb

this one is curiously gone, and btree needs it, and we vendor btree. can you investigate? there may be others I missed, too.

I think everything else is okay, the text package just looks shuffled. a diff of the Gopkg.lock list and the go.sum list may be more useful (few shell commands of wizardry?), I had to do some digging around with the provided diff anyway.

ericfode:
Redid it but locking to the revisions that were in the gopkg.lock

ericfode:
Regarding
> This PR introduces a go module at the root of the repository and then also new modules for images/fn-test-utils and images/hello and images/fn-status-checker. Do these sub-directories need to be their own modules? Can they just be sub-packages of the one top-level module?
They can be, I was experimenting and forgot to delete those.

ericfode:
> github.com/fnproject/fn/vendor/github.com/Azure/go-ansiterm
> github.com/fnproject/fn/vendor/github.com/Microsoft/go-winio
> 
> these were previously excluded (as they're only needed to build on WIN), am not sure if they fit compliance offhand or if there is a way for go mod to ignore these once again.
I added exclude statements for both of these.
> 
> github.com/fnproject/fn/vendor/github.com/petar/GoLLRB/llrb
> 
> this one is curiously gone, and btree needs it, and we vendor btree. can you investigate? there may be others I missed, too.
So it looks like btree replaces that package
https://github.com/google/btree/blob/master/btree.go#L43
So we don't need it anymore. Also packages with capital letters are the worst.

> 
> I think everything else is okay, the text package just looks shuffled. a diff of the Gopkg.lock list and the go.sum list may be more useful (few shell commands of wizardry?), I had to do some digging around with the provided diff anyway.

rdallman:
> I added exclude statements for both of these.

the files are still checked into `vendor/` - possible to remove them or mod bug?

rdallman:
I'd do a clear out of `vendor/` and grab them again with the latest mod file, I think with that this LGTM.

ping @zootalures for 2nd eyes

ericfode:
Will do

ericfode:
It looks like you can exclude a package that is needed for any builds with go mod like you can with go dep. And we don't want to check the Microsoft code into our repo (for the windows build) so my current proposal is to just add the paths in the vendor directory that hold these deps to the .gitignore. Thoughts @rpallman, other folks with thoughts?

a second option is creating a package called null (or something like it) and use the replace directive to replace the packages with an empty one.

ericfode:
Created a noop module and replaced.

rdallman:
> It looks like you can exclude a package that is needed for any builds with go mod like you can with go dep. 

assuming this meant to say 'can't'  - that's ok. I am unfamiliar with the compliance process and I imagine that if we do check them in, they will simply need to be vetted. if that's easy, that may be the right thing to do instead of making the dependency process our own special incantation to work around go mod deficiencies we were previously accustomed to having - I do appreciate the effort. do you mind looking into that or do we think this is the best path forward? as far as I'm aware, we're not in a huge rush either here (please correct me if this isn't the case)

ericfode:
I found a less hacky answer. I just made a folder in this repo called noop with empty module and replaced the offending modules with it.

rdallman:
> I found a less hacky answer. I just made a folder in this repo called noop with an empty module and replaced the offending modules with it.

ok. any details? this sounds like rewriting some import paths or mod files on some of our dependencies?

ericfode:
Nothing quite that drastic :)
I added this 
```
replace (
        github.com/Azure/go-ansiterm => ./noop
        github.com/Microsoft/go-winio => ./noop
)
```
to our mod file.

rdallman:
cool, that seems easy to follow and not need any maintenance when we update 👍

ericfode:
Anything else we need to do @rdallman @zootalures

rdallman:
LGTM here.

> My understanding is that it should still work... But I am still experimenting. No claims yet.

any claims re: how our packages that import this one will hold up? as long as that is smooth sailing, I'm ready to pull the trigger anyway

ericfode:
If you are using mod it's smooth sailing, it looks like it works for dep too. As a test I have tried a new project with go dep, and added a dependency to this branch of fn. Everything seemed to work.

PR #1288, opened 2018-10-31
fn: container initialization monitoring

skinowski:
Container initialization phase consumes resource tracker
resources (token), during lengthy operations.
In order for agent stability/liveness, this phase has
to be evictable/cancelable and time bounded.

With this change, introducing two new system wide environment settings
to bound the time spent in container initialization phase loosely separated
as docker-pull and UDS-initialize. If during initialization, the caller is
disconnected or serviced by another hot container, then initializing
hot container becomes eligible for eviction.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -679,7 +679,7 @@ func (s *hotSlot) dispatch(ctx context.Context, call *call) chan error {
 		resp, err := s.container.udsClient.Do(req)
 		if err != nil {
 			common.Logger(ctx).WithError(err).Error("Got error from UDS socket")
-			errApp <- models.NewAPIError(http.StatusBadGateway, errors.New("error receiving function response"))
+			errApp <- models.ErrFunctionResponse
 			return
 		}
 		common.Logger(ctx).WithField("resp", resp).Debug("Got resp from UDS socket")
@@ -780,9 +780,9 @@ InitPhase:
 			return
 		case <-timer.C: // init timeout
 			if pulled != nil {
-				tryQueueErr(models.ErrDockerPullTimeoutServerBusy, errQueue)
+				tryQueueErr(models.ErrDockerPullTimeout, errQueue)
 			} else {
-				tryQueueErr(models.ErrContainerInitFail, errQueue)
+				tryQueueErr(models.ErrContainerInitTimeout, errQueue)
 			}
 			return
 		}
@@ -851,6 +851,12 @@ func (a *agent) runHot(ctx context.Context, caller slotCaller, call *call, tok R
 		// in cases of overcapacity. If an accidental eviction occurs, then the client
 		// will have to wait hot poll period to spawn a new container.
 		if !evictor.isEvicted() {
+			select {
+			case <-initialized:
+			default:
+				tryQueueErr(models.ErrContainerInitFail, errQueue)
+			}
+
 			if err := getQueueErr(errQueue); err != nil {
 				call.slots.queueSlot(&hotSlot{done: make(chan struct{}), fatalErr: err})
 			}
@@ -1013,6 +1019,9 @@ func inotifyAwait(ctx context.Context, iofsDir string, udsWait chan error) error
 			case <-ctx.Done():
 				return
 			case err := <-fsWatcher.Errors:
+				// TODO: We do not know if these cases would be due to customer container/FDK
+				// fault or some kind of service/runner issue. As conservative choice,
+				// we reflect back a non API error, which means a 500 back to user.
 				logger.WithError(err).Error("error watching for iofs")
 				udsWait <- err
 				return
@@ -1021,6 +1030,8 @@ func inotifyAwait(ctx context.Context, iofsDir string, udsWait chan error) error
 				if event.Op&fsnotify.Create == fsnotify.Create && event.Name == filepath.Join(iofsDir, udsFilename) {
 					err := checkSocketDestination(filepath.Join(iofsDir, udsFilename))
 					if err != nil {
+						// This case is more like a bad FDK/container, so let's reflect this back to
+						// clients as container init fail.
 						logger.WithError(err).Error("Failed to check socket destination")
 						udsWait <- models.ErrContainerInitFail
 					} else {
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -737,6 +737,29 @@ func newSizerRespWriter(max uint64, rw http.ResponseWriter) http.ResponseWriter 
 
 func (s *sizerRespWriter) Write(b []byte) (int, error) { return s.w.Write(b) }
 
+// nannyHotContainer is a helper function to monitor initialization channel, container context as well as
+// agent shutdown during the lifetime of the container.
+func (a *agent) nannyHotContainer(ctx context.Context, cancel context.CancelFunc, evictor *EvictToken, initialized chan struct{}) {
+	defer cancel()
+
+	// Start monitoring initialization phase
+	select {
+	case <-initialized: // good, container is ready
+	case <-ctx.Done():
+	case <-a.shutWg.Closer():
+	case <-evictor.C: // eviction
+		return
+	case <-time.After(a.cfg.HotStartTimeout): // init timeout
+		return
+	}
+
+	// Start monitoring agent shutdown
+	select {
+	case <-a.shutWg.Closer():
+	case <-ctx.Done():
+	}
+}
+
 func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state ContainerState) {
 	// IMPORTANT: get a context that has a child span / logger but NO timeout
 	// TODO this is a 'FollowsFrom'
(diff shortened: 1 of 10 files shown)


api/agent/agent.go
```diff
@@ -734,117 +733,141 @@ func newSizerRespWriter(max uint64, rw http.ResponseWriter) http.ResponseWriter
...
+	container, queueErr = newHotContainer(ctx, call, &a.cfg, id, logger, initialized)
+	if queueErr != nil {
 		return
 	}
-
-	ctx, shutdownContainer := context.WithCancel(ctx)
-	defer shutdownContainer() // close this if our waiter returns, to call off slots, needs to follow cookie.Close so the cookie crumbles
-
-	udsAwait := make(chan error)
-	// start our listener before starting the container, so we don't miss the pretty things whispered in our ears
```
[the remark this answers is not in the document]
rdallman: code ordering wise it's important if people shuffle stuff around to launch it before start, feel free to clarify...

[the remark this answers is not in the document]
rdallman: i suppose there is a tiny race, we may need to create the fs watcher in this goroutine and add the dir we're watching before launching the goroutine to watch it and then return the channel. oops.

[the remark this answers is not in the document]
skinowski: this PR fixes it.

skinowski:
This is an alternative PR to original proposed https://github.com/fnproject/fn/pull/1279.

api/agent/agent.go
```diff
@@ -737,100 +745,155 @@ func newSizerRespWriter(max uint64, rw http.ResponseWriter) http.ResponseWriter
 
 func (s *sizerRespWriter) Write(b []byte) (int, error) { return s.w.Write(b) }
 
-func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state ContainerState) {
+// nannyHotContainer is a helper function to monitor initialization channel, container context as well as
+// agent shutdown during the lifetime of the container.
+func (a *agent) nannyHotContainer(ctx context.Context, callerCtx context.Context, cancel context.CancelFunc, evictor *EvictToken, initialized chan struct{}, pulled chan struct{}) {
```
rdallman: we should be able to reduce the contexts here by having the caller of nanny handle the cancel, I don't see any usage other than the first defer. since we only need to listen on the done channel for caller, it may read easier to just pass the channel from `callerCtx.Done()` than the whole context (not 2 contexts in scope) - sorry, as you know this is a trigger for me...

api/agent/agent.go
```diff
@@ -737,100 +745,155 @@ func newSizerRespWriter(max uint64, rw http.ResponseWriter) http.ResponseWriter
...
+		}
+	}
+
+	timer.Stop()
+
+	// Start monitoring agent shutdown
+	select {
+	case <-a.shutWg.Closer():
+	case <-ctx.Done():
+	}
```
rdallman: why do we need this goroutine to be live at this point?

api/agent/agent.go
```diff
@@ -887,52 +947,57 @@ func checkSocketDestination(filename string) error {
...
-		err = checkSocketDestination(filepath.Join(iofsDir, udsFilename))
-	}
-	select {
-	case awaitUDS <- err:
-	case <-ctx.Done():
-	}
-}
 
-func inotifyAwait(ctx context.Context, iofsDir string) error {
+func inotifyAwait(ctx context.Context, iofsDir string, logger logrus.FieldLogger, initialized chan struct{}) error {
```
rdallman: logger in func sig is a smell. use `ctx = common.WithLogger(ctx, logger)` and yank back out in callees please

api/agent/agent.go
```diff
@@ -887,52 +947,57 @@ func checkSocketDestination(filename string) error {
...
-			// XXX(reed): damn it would sure be nice to tell users they didn't make a uds and that's why it timed out
-			return ctx.Err()
-		case err := <-fsWatcher.Errors:
-			return fmt.Errorf("error watching for iofs: %v", err)
-		case event := <-fsWatcher.Events:
-			common.Logger(ctx).WithField("event", event).Debug("fsnotify event")
-			if event.Op&fsnotify.Create == fsnotify.Create && event.Name == filepath.Join(iofsDir, udsFilename) {
-				// wait until the socket file is created by the container
-				return nil
+	go func() {
```
rdallman: thanks. a note about why we add and then launch go routine may prove useful here one day as well so that we don't have this bug again

skinowski:
@rdallman Yeah, removed the extra ctx, we actually only need a done channel there. Nanny is indeed the heart and the most complex piece there. It monitors two phases: initialization versus shutdown-wait. Shutdown-wait is the simple case, it just collapses shutWg + ctx into ctx to make it easier for rest of the code.

Initialization phase is complex and it's a small state machine. It's split into before/during docker-pull and after docker-pull phases, where HotPullTimeout and HotStartTimeout is applied accordingly. During init phase, caller disconnect/serviced event makes the container evictable. Once initialized is set, then this phase moves on to shutdown-wait state.

cancel() calls are a bit unsavory. But compared to previous code where we had it in 2 places, we now have it in 3 places. I could split this and perhaps introduce more specific channels (eg. runHotReq exit, waiter wait exit, etc.) but I'm not sure if it'll look simpler.

api/agent/agent.go
```diff
@@ -768,9 +768,9 @@ InitPhase:
 			return
 		case <-a.shutWg.Closer(): // agent shutdown
 			return
-		case <-callerCtx.Done(): // original caller disconnected or serviced by another container?
+		case <-caller.done: // original caller disconnected or serviced by another container?
 			evictor.SetEvictable(true)
-			callerCtx = context.Background() // block 'callerCtx' after this point
+			caller.done = make(chan struct{}) // block 'doneChan' after this point
```
rdallman: nil would work and not allocate as well, not that this is a very hot path, just for future reference (it's not widely known that receive on nil channel doesn't panic and can be useful sometimes, like here)

api/agent/agent.go
rdallman: i see, because the cancel collapses the ctx done. kinda tricky, but yea makes other code cleaner

skinowski: Honestly this only saves 4-5 lines of code, and for-loop below can get away with if ctx.Err() check, but if you think this is too tricky and not worth it, I could add shutWg back to old spots.

rdallman:
thanks for explanation, that helps clarify i mostly solved the puzzle, woo. gives me some intuition that the nanny is much more straightforward if it's solely responsible for setting the container to evictable state if initialization is ongoing when the initiating thread returns? i do see how it's a convenient place to run a state machine since it's there, there are just 2 of them basically, one of them actually doing the things (cookie.Prepare, cookie.Run, etc), another watching for those things to happen in a separate place - given the cookie apis aren't async, nanny just seems maybe more complicated than it needs to be - like i said earlier though, it may lead to context jungle, anyway trust your judgment, it's just an idea.

api/agent/agent.go
```diff
@@ -359,17 +359,30 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
 
 	call.slots, isNew = a.slotMgr.getSlotQueue(call.slotHashId)
 	call.requestState.UpdateState(ctx, RequestStateWait, call.slots)
+
+	// setup slot caller with a ctx that gets cancelled once waitHot() is completed.
+	// This allows runHot() to detect if original caller has been serviced by
+	// another container or if original caller was disconnected.
```
rdallman: 👍

api/agent/agent.go
```diff
@@ -737,114 +745,183 @@ func newSizerRespWriter(max uint64, rw http.ResponseWriter) http.ResponseWriter
...
+	var cookie drivers.Cookie
+	var waiter drivers.WaitResult
+	var runRes drivers.RunResult
+	var err error
+
+	id := id.New().String()
+	logger := logrus.WithFields(logrus.Fields{"id": id, "app_id": call.AppID, "fn_id": call.FnID, "image": call.Image, "memory": call.Memory, "cpus": call.CPUs, "idle_timeout": call.IdleTimeout})
+	ctx, cancel := context.WithCancel(common.WithLogger(ctx, logger))
+
+	initialized := make(chan struct{}) // initialized means (docker pulled, container is created/started, uds is ready)
```
rdallman: would you specify the contract here please that this is only ever closed and no values are sent? (there are multiple to untangle here, it's nice to know imo and for future devs to not try to send anything)

api/agent/agent.go
```diff
@@ -737,114 +745,183 @@ func newSizerRespWriter(max uint64, rw http.ResponseWriter) http.ResponseWriter
...
 	ctx, span := trace.StartSpan(ctx, "agent_run_hot")
 	defer span.End()
 
-	// IMPORTANT: evict token is deleted *after* resource token in defer statements below.
-	// This ordering allows resource token to be freed first, which means once evict token
-	// is deleted, eviction is considered to be completed.
+	var container *container
+	var cookie drivers.Cookie
+	var waiter drivers.WaitResult
+	var runRes drivers.RunResult
```
rdallman: i think `runRes` and `waiter` can be eliminated from this list of upfront declarations, the only use appears to be for a log line in a separate goroutine which can go in the same thread and to not use `:=` in the other (at least as someone reading this i have to go back and forth over usage to be able to reason about scope these are being used in to figure out what's going on, and it's a lot of cognitive overhead for something that maybe doesn't need to be). reducing these in general would be nice, if possible. tracking back and forth between this and goroutine isn't very fun. a lot of the defers could be on their own, at least `container.Close` and `cookie.Close`, after checking the err on the method that produces them. think that covers them all.

api/agent/agent.go
```diff
@@ -737,114 +745,183 @@ func newSizerRespWriter(max uint64, rw http.ResponseWriter) http.ResponseWriter
...
-		statsUtilization(ctx, a.resources.GetUtilization())
-	}()
+		// IMPORTANT: we do not emit eviction errors to slot queue. Remember evictions occur
+		// in cases of overcapacity. If an accidental eviction occurs, then the client
+		// will have to wait hot poll period to spawn a new container.
+		if !evictor.isEvicted() {
+			select {
+			case <-initialized:
+			default:
+				tryQueueErr(models.ErrContainerInitFail, errQueue)
```
rdallman: this appears to queue an eviction error to the error queue just before dequeueing it to send to the slot queue, which the notice says shouldn't happen. not sure if comment or code or my understanding should change

api/agent/agent.go
rdallman: i kinda agree that seeing the ordering inside of the defer is nice instead of having to read defers backwards, so remove 1/2 this comment I guess. the defer itself sticks out (ie read about what happens at the end before doing the things) but overall seems ok

api/agent/agent.go
```diff
@@ -737,114 +745,183 @@ func newSizerRespWriter(max uint64, rw http.ResponseWriter) http.ResponseWriter
...
+	if err != nil {
+		select {
+		case ch <- err:
+		default:
+		}
+	}
+	return err
+}
+
+func (a *agent) runHot(ctx context.Context, caller slotCaller, call *call, tok ResourceToken, state ContainerState) {
```
rdallman: would it help to return an `err` from `runHot` and have the calling thread handle queueing the error to the slot queue? idk, idea, there's a lot of queueing errors so we'd go to more natural return err and it would be one less thing in the defer. i guess i see 2 uses of tryQueueError from a child goroutine. leaving idea, but feel free to ignore

api/agent/agent.go
```diff
@@ -737,114 +745,183 @@ func newSizerRespWriter(max uint64, rw http.ResponseWriter) http.ResponseWriter
...
+		isInitialized, err := a.waitInitialize(ctx, caller, evictor, udsWait)
+		if err != nil {
+			tryQueueErr(err, errQueue)
+		}
+		if isInitialized {
+			close(initialized)
+		} else {
+			cancel()
+		}
+	}()
```
rdallman: initialized seems like largely a dupe of udsWait. this thread forwards the waiting on udsWait to receive to initialized. i think waitInitialize needs to exist to wait on any eviction events but the addition of udsWait into that seems only to happen so that we can turn the send of 1 error into udsWait into a 'closed channel' that can be received on multiple times, at least it seems slightly unnecessary. am not sure it's exactly the constraint, trying to deduce, anyway it'd be possible to get a similar effect with something like https://play.golang.org/p/SSZLKc-PxwN - where udsWait wouldn't return anything (ie hang on channel receive) if the container never ended up getting prepared for any other reason or the uds itself not being ready, but after that we could definitely say 'the container was initialized' to eg avoid queueing an error that initialization failed. just trying to simplify here if possible.

api/agent/agent.go
```diff
@@ -737,114 +745,183 @@ func newSizerRespWriter(max uint64, rw http.ResponseWriter) http.ResponseWriter
...
-	if err != nil {
-		call.slots.queueSlot(&hotSlot{done: make(chan struct{}), fatalErr: err})
+	waiter, err = cookie.Run(ctx)
+	if tryQueueErr(err, errQueue) != nil {
 		return
 	}
 
 	go func() {
-		defer shutdownContainer() // also close if we get an agent shutdown / idle timeout
+		defer cancel() // also close if we get an agent shutdown / idle timeout
```
rdallman: def found the last name more obvious

api/agent/agent.go
skinowski: I agree, the queue error logic needs rework. But saving that for another PR.

api/agent/agent.go
skinowski: Only if no eviction occurs, then we try to process errors. Adjusted the comments to make this a bit clearer.

api/agent/agent.go
skinowski: maybe, but it was kind of misleading. Which container? "container" from newHotContainer? docker container in cookie? Here cancel is generic, and it means "we cancel all operations" inside this function. This PR moves the cancel to the beginning of runHot(), which covers a lot more than it did before.

api/agent/agent.go
skinowski: removed initialized now.

api/agent/agent.go
skinowski: yeah, it's left over from previous logic, etc. With timeouts in nanny go-routine this made more sense. It's gone now.

api/agent/agent.go
skinowski: err handling needs a rework, but I'm deferring that to a future PR.

api/agent/agent.go
skinowski: This got slightly better now I think.

skinowski: Our tear down ordering is tricky, so I really felt the need to spell it out there and avoid any bugs if we shuffle/reorder things in runHot() body.

api/agent/agent.go
```diff
@@ -855,20 +825,32 @@ func (a *agent) runHot(ctx context.Context, caller slotCaller, call *call, tok R
...
+			select {
+			case err := <-udsWait:
+				if tryQueueErr(err, errQueue) != nil {
+					cancel()
+				}
+				return
+			case <-ctx.Done(): // container shutdown
+				return
+			case <-a.shutWg.Closer(): // agent shutdown
+				cancel()
```
rdallman: is this going to shut down the container when a call has a hold on a slot and is active? should we let the main slot dispense goroutine handle this one?

rdallman: at least i can reason about a timing where the main slot dispense goroutine gets udsWait and gets past initialization, hands out a slot, and then this goroutine is scheduled to run and hits shutWg. In practice it may never happen but ostensibly... ?

api/agent/agent.go
```diff
@@ -855,20 +825,32 @@ func (a *agent) runHot(ctx context.Context, caller slotCaller, call *call, tok R
...
+				return
+			case <-caller.done: // original caller disconnected or serviced by another container?
+				evictor.SetEvictable(true)
+				caller.done = nil // block 'caller.done' after this point
+			case <-evictor.C: // eviction
+				cancel()
```
rdallman: am curious about https://github.com/fnproject/fn/pull/1288/commits/af5236241d131d58d3e20c0b2478353a90f5ee79#r234048169 for this and the udsWait too - i'm not 100% sure, it may be okay, worth thinking about maybe, i'll think too...

api/agent/agent.go
skinowski: I'm not too worried about this. The flow might not even make it to main slot go-routine in a few cases, so we have to poll it here. Similar conditions with evictions also exist.

rdallman: yea, it seems slim but possible. i guess the 'easy' thing might be to run this thing until just before we wait on initialization in the main slot go routine (over all the cookie stuff, etc) and then shut it down before the init barrier, might not add too much, though yea kinda unfortunate

skinowski: That's kind why I had the initialized channel earlier, which is easier to follow barrier, then multi readers with err/close semantics. Let me think about this.

PR #1292, opened 2018-10-31
fn: cleanup of docker timeouts and docker health check

skinowski:
Moving the timeout management of various docker operations
    to agent. This allows for finer control over what operation
    should use. For instance, for pause/unpause our tolerance
    is very low to avoid resource issues. For docker remove,
    the consequences of failure will lead to potential agent
    failure and therefore we wait up to 10 minute.
    For cookie create/prepare (which includes docker-pull)
    we cap this at 10 minutes by default.
    
    With new UDS/FDK contract, health check is now obsoleted
    as container advertise health using UDS availibility.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -775,7 +775,7 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
 		return
 	}
 
-	defer cookie.Close(ctx)
+	defer cookie.Close(common.BackgroundContext(ctx))
 
 	err = a.driver.PrepareCookie(ctx, cookie)
 	if err != nil {
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -28,8 +28,7 @@ import (
 )
 
 const (
-	removeTimeout = 10 * time.Minute // docker remove
-	pauseTimeout  = 5 * time.Second  // docker pause/unpause
+	pauseTimeout = 5 * time.Second // docker pause/unpause
 )
 
 // TODO we should prob store async calls in db immediately since we're returning id (will 404 until post-execution)
@@ -781,11 +780,8 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
 		return
 	}
 
-	defer func() {
-		ctx, cancel := context.WithTimeout(common.BackgroundContext(ctx), removeTimeout)
-		cookie.Close(ctx)
-		cancel()
-	}()
+	// WARNING: we wait forever.
+	defer cookie.Close(common.BackgroundContext(ctx))
 
 	err = a.driver.PrepareCookie(ctx, cookie)
 	if err != nil {
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -29,7 +29,6 @@ import (
 
 const (
 	removeTimeout = 10 * time.Minute // docker remove
-	prepTimeout   = 10 * time.Minute // docker create+pull
 	pauseTimeout  = 5 * time.Second  // docker pause/unpause
 )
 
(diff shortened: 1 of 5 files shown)


skinowski:
This is a prep-work for https://github.com/fnproject/fn/pull/1288

api/agent/drivers/docker/docker.go
```diff
@@ -577,36 +577,7 @@ func (drv *DockerDriver) startTask(ctx context.Context, container string) error
...
-
-		// if no health check for this image (""), or it's healthy, then stop waiting.
-		// state machine is "starting" -> "healthy" | "unhealthy"
-		if cont.State.Health.Status == "" || cont.State.Health.Status == "healthy" {
-			break
-		}
-
-		time.Sleep(100 * time.Millisecond) // avoid spin loop in case docker is actually fast
-	}
-	return nil
```
rdallman: 👍 this makes sense with uds I think, good idea

api/agent/agent.go
```diff
@@ -764,26 +770,32 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
...
 	}
 
-	defer cookie.Close(ctx)
+	defer func() {
+		ctx, cancel := context.WithTimeout(common.BackgroundContext(ctx), removeTimeout)
+		cookie.Close(ctx)
+		cancel()
+	}()
 
 	err = a.driver.PrepareCookie(ctx, cookie)
```
rdallman: i forgot why create and prepare cookie got split up, maybe it was for cold, seems like we can put humpty dumpty back together again now (not implying that needs to get done here, just observing...)

api/agent/drivers/docker/docker.go
```diff
@@ -276,6 +275,7 @@ func (drv *DockerDriver) PrepareCookie(ctx context.Context, c drivers.Cookie) er
 		return err
 	}
 
+	cookie.opts.Context = ctx
```
skinowski: switching to correct ctx here, ctx passed in CreateCookie() should not be used in CreateContainer().

rdallman:
yea, makes sense to move these to caller with these changes, couple notes

api/agent/drivers/docker/docker_client.go
```diff
@@ -386,15 +383,9 @@ func (d *dockerWrap) PullImage(opts docker.PullImageOptions, auth docker.AuthCon
 }
 
 func (d *dockerWrap) RemoveContainer(opts docker.RemoveContainerOptions) (err error) {
-	// extract the span, but do not keep the context, since the enclosing context
-	// may be timed out, and we still want to remove the container. TODO in caller? who cares?
-	ctx := common.BackgroundContext(opts.Context)
-	ctx, closer := makeTracker(ctx, "docker_remove_container")
+	ctx, closer := makeTracker(opts.Context, "docker_remove_container")
```
skinowski: moved this to caller as how to long to wait on this depends on higher level code.

api/agent/agent.go
skinowski: cookie Iface doesn't make sense right now, also wait result. Another PR...  I've modified them here: https://github.com/fnproject/fn/pull/1279, which we can pick and carry into a separate PR later.

api/agent/agent.go
```diff
@@ -27,6 +27,12 @@ import (
 	"os"
 )
 
+const (
+	removeTimeout = 1 * time.Minute // docker remove
+	prepTimeout   = 2 * time.Minute // docker create+pull
+	pauseTimeout  = 5 * time.Second // docker pause/unpause
```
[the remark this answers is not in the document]
skinowski: Increased that to 10 mins now. I'm keeping somewhat similar values as before.

skinowski:
I think I should add another timeout here, docker-pull vs UDS/start. Blanket timeout for all of it seems too broad, and leaves too much time for UDS/start phase which will eat cpu/mem outside of call.Start/End.

api/agent/agent.go
```diff
@@ -764,26 +768,29 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
...
+	defer cancel()
+
 	cookie, err := a.driver.CreateCookie(ctx, container)
 	if err != nil {
 		call.slots.queueSlot(&hotSlot{done: make(chan struct{}), fatalErr: err})
 		return
 	}
 
-	defer cookie.Close(ctx)
+	// WARNING: we wait forever.
```
rdallman: retry in docker_client will give up at 10 tries and this should bail at that point, at least, currently I don't think this is true. so long as every call to the docker_client has a timeout in the context (in applicable cases, ie everything but this one?), we could remove the 10 retry cap, as the retry loop respects the context timeout/cancel as it is.

rdallman:
> I think I should add another timeout here, docker-pull vs UDS/start. Blanket timeout for all of it seems too broad, and leaves too much time for UDS/start phase which will eat cpu/mem outside of call.Start/End.

that seems fine. original uds proposal had 15s or something, yea, they're on the clock for the uds thing.

PR #1295, opened 2018-11-01
Fix filtering by name when getting list of funcs

### LINKED ISSUE #1294: Filtering by name when when getting function list doesn't work
**Description**

Using the 'name' parameter to filter by function name on the /v2/fns endpoint fails to filter by name. Instead, the full list of functions for the given app ID is always returned.

1. Create an app and multiple functions associated with it.
2. `curl "localhost:8080/v2/fns?app_id=MYAPPID&name=MYFUNC1"`

A response including all functions associated with MYAPPID, as if the 'name' paramater was not present.

A response consisting of only MYFUNC1.

harryjsmith:
- Link to issue this resolves
#1294 

- What I did
Fixed buildFilterFnQuery in local datastore code to respect 'Name' filter.

- How to verify it
Use 'name' parameter on v2/fns GET endpoint to filter function list by name. E.g. `curl "localhost:8080/v2/fns?app_id=MYAPPID&name=MYFUNC1"` 

Fixed filtering by name when getting function list

### DIFF
--- a/api/datastore/datastoretest/test.go
+++ b/api/datastore/datastoretest/test.go
@@ -734,9 +734,8 @@ func RunFnsTest(t *testing.T, dsf DataStoreFunc, rp ResourceProvider) {
 			}
 			if len(fns.Items) != 1 {
 				t.Fatalf("expected result count to be 1, got %d", len(fns.Items))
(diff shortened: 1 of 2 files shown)


rdallman:
> Can you add the missing test to datastore tests

+1 - good find, seems like we only have cursor tests, no name filter tests :(

PR #1300, opened 2018-11-06
WIP: Launch 3-component deployment using docker-compose.yml

riconnon:
I thought I'd experiment with launching a 3-component deployment as a local environment using docker-compose

Interested to know what people think.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,10 +10,6 @@ services:
   logstore:
     hostname: logstore
     image: minio/minio
-    networks:
-      - fn-network
-    ports:
-      - "9091:9000"
     environment:
       - MINIO_ACCESS_KEY=admin
       - MINIO_SECRET_KEY=password
@@ -23,10 +19,6 @@ services:
   db:
     image: "mysql:5.7.22"
     restart: always
-    networks:
-      - fn-network
-    ports:
-      - "3306:3306"
     environment:
       - "MYSQL_DATABASE=funcs"
       - "MYSQL_ROOT_PASSWORD=root"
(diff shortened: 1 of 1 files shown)


rikgibson:
I think it would be good to maintain the single-container model for fnserver if possible, even if internally the container was running three processes - it keeps the black box simple and consistent for the user.

rdallman:
it should be possible to do this all in a single process, too, there's some... factoring issues... to solve.

docker-compose.yml
```diff
@@ -35,68 +27,74 @@ services:
...
-    ports:
-      - "9090:9090"
-    links:
-      - fnserver
+      - api
+      - lb
+      - runner
     volumes:
       - ./examples/grafana/prometheus.yml:/etc/prometheus/prometheus.yml
   fnserver-ui:
```
denismakogon: let's exclude the UI component, because it doesn't work =(

docker-compose.yml
```diff
@@ -35,68 +27,74 @@ services:
...
-      - "logstore"
     environment:
       FN_DB_URL: "mysql://root:root@tcp(db:3306)/funcs"
       FN_MQ_URL: "redis://mq:6379/"
       FN_LOGSTORE_URL: "s3://admin:password@logstore:9000/us-east-1/fnlogs"
+      FN_PUBLIC_LB_URL: "http://127.0.0.1:8081"
+  lb:
+    restart: always
+    depends_on:
+      - db
```
denismakogon: should also depend on api node.

docker-compose.yml
```diff
@@ -35,68 +27,74 @@ services:
...
+  api:
     restart: always
     depends_on:
       - mq
       - db
       - logstore
-    build: .
-    networks:
-      - fn-network
+    build:
```
denismakogon: Network attribute was necessary to let functions be within the Fn network, and the idea was to have a flat network where each component has a binding.

denismakogon:
As for testing, i'm getting `Failed during call placement` while calling hello-world samples.

denismakogon:
Let's do the following, I'd rather get rid of compose file and use OSS helm charts. @riconnon thoughts?

PR #1307, opened 2018-11-12
Enable tagging to vx.y.z tags

zootalures:
This should enable go mod to accurately resolve Fn as a dependency


riconnon:
LGTM
Do we actually need to keep the old version tags as well?

zootalures:
Don't know if it will break anything  - bunch of stuff  like CLI builds are still  using dep so might depend on them

PR #1308, opened 2018-11-13
initial addition of the diskfree op

ericfode:
- Link to issue this resolves
Cutting this PR up into two steps https://github.com/fnproject/fn/pull/1175/files
- What I did
Just did the changes to the docker client.
- How I did it
Added DiskFree, ListImages, RemoveImage
- How to verify it
Compile and test
Add DiskFree, ListImages, and RemoveImage DockerAPI calls.


api/agent/drivers/docker/docker_client.go
```diff
@@ -308,6 +311,19 @@ func (d *dockerWrap) LoadImages(ctx context.Context, filePath string) error {
 	})
 }
 
+func (d *dockerWrap) ListImages(ctx context.Context) (imgs []docker.APIImages, err error) {
+	ctx, closer := makeTracker(ctx, "list_docker_images")
```
rdallman: can we do `docker_list_images` to be consistent with the other stats in this file please? this makes eg summing over `docker_*` ops easy, aside from consistency benefits for being easy to find

api/agent/drivers/docker/docker_client.go
```diff
@@ -308,6 +311,19 @@ func (d *dockerWrap) LoadImages(ctx context.Context, filePath string) error {
...
+	defer closer()
+
+	logger := common.Logger(ctx).WithField("docker_cmd", "ListImages")
+	err = d.retry(ctx, logger, func() error {
+		imgs, err = d.docker.ListImages(docker.ListImagesOptions{All: false})
```
rdallman: most of the operations in this file allowing specifying the docker options as well (like this one). I forget how well articulated the intent of this package was but it was intended to mirror the parts of the fsouza API we need but just add retries, stats, etc. could you make the signature for the `dockerClient` interface `ListImages(docker.ListImagesOptions)` as well, please? I assume there's a context in the options somewhere or the fsouza client should have a `ListImageWithContext` method as well. we need to thread the context into the underlying operation, as well, to avoid leaking goroutines / connections / etc.

PR #1309, opened 2018-11-14
Add support for counting kdumps in pure-runner.

kmjohansen:
- What I did
Added support for counting the number of kdumps that are on a runner at the time the hostagent is started.
- How I did it
This is accomplished by adding a field to the runner status call.  Pure runner consumers who want to set this value may do so by passing the value in at the time the pure runner is created.
- How to verify it
Retrieve the status message and check for the presence of the field. 
Add support for counting kdumps in runner status message.


api/agent/pure_runner.go
```diff
@@ -995,6 +997,19 @@ func PureRunnerWithStatusImage(imgName string) PureRunnerOption {
 	}
 }
 
+// PureRunnerWithKdumpsOnDisk returns a PureRunnerOption that indicates that
+// kdumps have been found on disk.  The argument numKdump is a counter that
+// indicates how many dumps were on disk at the time the runner was created.
+func PureRunnerWithKdumpsOnDisk(numKdumps uint64) PureRunnerOption {
+	return func(pr *pureRunner) error {
+		if pr.status.kdumpsOnDisk != 0 {
+			return fmt.Errorf("Duplicate kdump count configuration! old=%u new=%u", pr.status.kdumpsOnDisk, numKdumps)
```
skinowski: compiler complains about %d instead of %u here.

kmjohansen: Thanks for catching this; fixed.

PR #1312, opened 2018-11-14
fn: cookie and driver api changes

skinowski:
Now obsoleted driver.PrepareCookie() call handled image and
container creation. In agent, going forward we will need finer
grained control over the timeouts implied by the contexts.
For this reason, with this change, we split PrepareCookie()
into Validate/Pull/Create calls under Cookie interface.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -782,7 +782,21 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
 	// WARNING: we wait forever.
 	defer cookie.Close(common.BackgroundContext(ctx))
 
-	err = a.driver.PrepareCookie(ctx, cookie)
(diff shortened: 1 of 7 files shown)


api/agent/agent.go
```diff
@@ -782,7 +782,21 @@ func (a *agent) runHot(ctx context.Context, call *call, tok ResourceToken, state
 	// WARNING: we wait forever.
 	defer cookie.Close(common.BackgroundContext(ctx))
 
-	err = a.driver.PrepareCookie(ctx, cookie)
+	shouldPull, err := cookie.ValidateImage(ctx)
```
skinowski: I will add timeouts and better error handling with future PRs here. Agent.go changes in this PR as very basic and intentionally kept bare minimum.

PR #1313, opened 2018-11-15
disable pure runner logging

rdallman:
there's a racey bug where the logger is being written to when it's closing,
but this led to figuring out that we don't need the logger at all in pure
runner really, the syslog thing isn't an in process fn thing and we don't need
the logs from attach for anything further in pure runner. so this disables the
logger at the docker level, to save sending the bytes back over the wire, this
could be a nice little performance bump too. of course, with this, it means
agents can be configured to not log debug or have logs to store at all, and
not a lot of guards have been put on this for 'full' agent mode while it hangs
on a cross feeling the breeze awaiting its demise - the default configuration
remains the same, and no behavior changes in 'full' agent are here.

it was a lot smoother to make the noop than to try to plumb in 'nil' for
stdout/stderr, this has a lot lower risk of nil panic issues for the same
effect, though it's not perfect relying on type casting, plumbing in an
interface to check has the same issues (loss of interface adherence for any
decorator), so this seems ok. defaulting to not having a logger was similarly
painful, and ended up with this. but open to ideas

### DIFF
--- a/api/agent/call.go
+++ b/api/agent/call.go
@@ -172,6 +172,14 @@ func WithWriter(w io.Writer) CallOpt {
 	}
 }
 
+// WithLogger sets stderr to the provided one
+func WithLogger(w io.ReadWriteCloser) CallOpt {
+	return func(c *call) error {
+		c.stderr = w
+		return nil
+	}
+}
+
 // InvokeDetached mark a call to be a detached call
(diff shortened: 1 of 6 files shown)


rdallman:
thought that might happen... 1s

PR #1314, opened 2018-11-15
fn: enforce container/FDK contract in dispatch

skinowski:
FDK must respond with 200/502/504 to http requests. Any other status is invalid and container
must be terminated if that occurs. HTTP timeouts are also considered a shutdown reason
for the container.

With this PR, we trust http client to use the context timeout and therefore avoid spawning
a go-routine. This simplifies the flow and also avoids case of call.stderr.Close() executing before swapBack.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -606,20 +606,11 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 
 	call.req = call.req.WithContext(ctx) // TODO this is funny biz reed is bad
 
-	errApp := s.dispatch(ctx, call)
-
-	select {
-	case err := <-errApp: // from dispatch
-		if err != nil {
-			if models.IsAPIError(err) {
-				s.trySetError(err)
-			}
-		}
-		return err
-	case <-ctx.Done(): // call timeout
-		s.trySetError(ctx.Err())
-		return ctx.Err()
+	err := s.dispatch(ctx, call)
+	if err != nil {
+		s.trySetError(err)
 	}
+	return err
 }
 
 var removeHeaders = map[string]bool{
@@ -632,7 +623,7 @@ var removeHeaders = map[string]bool{
 	"authorization":     true,
 }
 
-func callToHTTPRequest(ctx context.Context, call *call) *http.Request {
+func createUDSRequest(ctx context.Context, call *call) *http.Request {
 	req, err := http.NewRequest("POST", "http://localhost/call", call.req.Body)
 	if err != nil {
 		common.Logger(ctx).WithError(err).Error("somebody put a bad url in the call http request. 10 lashes.")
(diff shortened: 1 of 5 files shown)


api/agent/agent.go
```diff
@@ -661,70 +647,91 @@ func callToHTTPRequest(ctx context.Context, call *call) *http.Request {
...
-		select {
-		case ioErr := <-ioErrChan:
-			errApp <- ioErr
-		case <-ctx.Done():
-			errApp <- ctx.Err()
+	select {
+	case ioErr := <-ioErrChan:
+		return ioErr
+	case <-ctx.Done():
+		if ctx.Err() == context.DeadlineExceeded {
```
skinowski: This is a bit naive for slow clients, but we buffer or run with LB at front.

api/agent/agent.go
```diff
@@ -661,70 +647,91 @@ func callToHTTPRequest(ctx context.Context, call *call) *http.Request {
...
+		// IMPORTANT: Container contract: If http-uds errors/timeout, container cannot continue
+		s.trySetError(err)
+		// first filter out timeouts
+		if ctx.Err() == context.DeadlineExceeded {
+			return context.DeadlineExceeded
 		}
-		common.Logger(ctx).WithField("resp", resp).Debug("Got resp from UDS socket")
+		return models.ErrFunctionResponse
+	}
+	defer resp.Body.Close()
```
rdallman: leaving note to double check - used to have to do `io.Copy(ioutil.Discard, resp.Body)`  and if we get a 504 from fdk i'm assuming we're going to return that code without the body, i forget if they added resp.Body.Close discarding the body, it was at least discussed... just leaving note to double check

api/agent/agent.go
```diff
@@ -661,70 +647,91 @@ func callToHTTPRequest(ctx context.Context, call *call) *http.Request {
...
+		// Any other code. Possible FDK failure. We shutdown the container
+		s.trySetError(fmt.Errorf("FDK Error, invalid status code %d", resp.StatusCode))
+		fdkErr = models.ErrFunctionInvalidResponse
 	}
-	_, err := io.Copy(rw, resp.Body)
-	return err
+
+	// despite errors above, we try to read the body, which could contain information why the request
+	// has failed, etc.
+	_, ioErr := io.Copy(rw, resp.Body)
```
rdallman: this is odd and maybe unnecessary, we should probably discard body for 504/502? (error handling here is odd to chase with ioErr and fdkErr)

api/agent/agent.go
skinowski: I think this is safe as per http.Client Do documentation here:
https://golang.org/pkg/net/http/#Client.Do

api/agent/agent.go
rdallman: at least if we do this we buffer it into the thing at the moment - in the future this probably gets streamed back to client somewhere - should we define a response message for 502/504? at least it seems weird to write to the output and return an error from a contract level in dispatch

rdallman: also it'd make me more comfy (again, forgot about ioutil.Discard being done in resp.Body.Close) if we had a `defer io.Copy(ioutil.Discard, resp.Body)` above this thing in case we hit the clamp writer limit or something - at least, with go, have had fun times with conns being left open from forgetting to do this but i think recent versions they may have fixed so you might just ignore me now

api/agent/agent.go
```diff
@@ -661,70 +647,91 @@ func callToHTTPRequest(ctx context.Context, call *call) *http.Request {
...
 	}
-	if resp.StatusCode > 0 {
-		rw.WriteHeader(resp.StatusCode)
+
+	var fdkErr error
+
+	// IMPORTANT: Container contract: Enforce 200/502/504 expections
+	switch resp.StatusCode {
+	case 200:
+		// FDK processed the request OK
```
rdallman: this relies on the default behavior of writing a 200 status code when writing bytes to a responsewriter's write method (maybe better to be explicit here to make it clear for anybody changing this in future?)

api/agent/agent.go
```diff
@@ -661,70 +647,91 @@ func callToHTTPRequest(ctx context.Context, call *call) *http.Request {
...
+	ctx, span := trace.StartSpan(ctx, "agent_dispatch_httpstream")
+	defer span.End()
 
-		// TODO it's possible we can get rid of this (after getting rid of logs API) - may need for call id/debug mode still
-		// TODO there's a timeout race for swapping this back if the container doesn't get killed for timing out, and don't you forget it
-		swapBack := s.container.swap(call.stderr, call.stderr, &call.Stats)
-		defer swapBack()
+	// TODO it's possible we can get rid of this (after getting rid of logs API) - may need for call id/debug mode still
+	swapBack := s.container.swap(call.stderr, call.stderr, &call.Stats)
+	defer swapBack()
```
rdallman: yea, need to land this fix, thanks - thought we had done this last week for some reason

Issue #1310, opened 2018-11-14
Invalid Invoke path HTTP method error should be more explicit


denismakogon:
**Description**
Invoke annotation path allows to call a function with HTTP POST only, when any other method used Fn returns HTTP 404 Not Found.
The thing is the path is there, but it doesn't allow to use any other HTTP methods except POST.

**Steps to reproduce the issue:**
1. Deploy a function
2. Inspect a function
3. `curl -v $(fn inspect fn <app> <fn> | grep "invokeEndpoint" | awk '{ print $2 }' | xargs)`
4. See the result
**Describe the results you received:**
```
 curl -v http://localhost:8080/invoke/01CW79DNHMNG8G00GZJ0000002
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 8080 (#0)
> GET /invoke/01CW79DNHMNG8G00GZJ0000002 HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.0
> Accept: */*
> 
< HTTP/1.1 404 Not Found
< Content-Type: application/json; charset=utf-8
< Date: Wed, 14 Nov 2018 10:36:06 GMT
< Content-Length: 65
< 
{"message":"Path not found: /invoke/01CW79DNHMNG8G00GZJ0000002"}
* Connection #0 to host localhost left intact

```

**Describe the results you expected:**
In HTTP 1.1 spec there's HTTP 405 Not Allowed.
This particular code fits better than any other.
Also error message should be more explicit (i.e., only POST allowed, etc.)

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `fn version` (CLI command):**

```
Client version: 0.5.29
Server version:  0.3.613
```

PR #1315, opened 2018-11-15
Fix response message on invoke handler

denismakogon:
Closes: #1310

- Link to issue this resolves: #1310 

- What I did: changed Gin handler definition with some HTTP method validation for the sake of valid error message.

- How I did it: with no problem ;)

- How to verify it: call invoke handler with any http method except POST.

- One line description for the changelog: The following commit improves error status code and message for the cases when invoke path called with any HTTP method except HTTP POST

robot

### DIFF
--- a/api/server/runner_fninvoke.go
+++ b/api/server/runner_fninvoke.go
@@ -44,10 +44,6 @@ func (s *syncResponseWriter) Status() int          { return s.status }
 
 // handleFnInvokeCall executes the function, for router handlers
 func (s *Server) handleFnInvokeCall(c *gin.Context) {
-	if c.Request.Method != http.MethodPost {
-		handleErrorResponse(c, models.ErrInvokePostOnly)
-		return
-	}
 	fnID := c.Param(api.ParamFnID)
 	ctx, _ := common.LoggerWithFields(c.Request.Context(), logrus.Fields{"fnID": fnID})
 	c.Request = c.Request.WithContext(ctx)
--- a/api/server/runner_fninvoke.go
+++ b/api/server/runner_fninvoke.go
@@ -2,7 +2,6 @@ package server
 
 import (
 	"bytes"
-	"errors"
 	"io"
 	"net/http"
 	"strconv"
@@ -46,9 +45,7 @@ func (s *syncResponseWriter) Status() int          { return s.status }
 // handleFnInvokeCall executes the function, for router handlers
 func (s *Server) handleFnInvokeCall(c *gin.Context) {
 	if c.Request.Method != http.MethodPost {
-		handleErrorResponse(c, models.NewAPIError(http.StatusMethodNotAllowed,
-			errors.New("Method not allowed, only HTTP POST supported.")),
-		)
+		handleErrorResponse(c, models.ErrInvokePostOnly)
 		return
 	}
 	fnID := c.Param(api.ParamFnID)
--- a/api/server/runner_fninvoke.go
+++ b/api/server/runner_fninvoke.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"bytes"
+	"errors"
 	"io"
 	"net/http"
 	"strconv"
@@ -44,6 +45,12 @@ func (s *syncResponseWriter) Status() int          { return s.status }
 
 // handleFnInvokeCall executes the function, for router handlers
 func (s *Server) handleFnInvokeCall(c *gin.Context) {
+	if c.Request.Method != http.MethodPost {
+		handleErrorResponse(c, models.NewAPIError(http.StatusMethodNotAllowed,
+			errors.New("Method not allowed, only HTTP POST supported.")),
+		)
+		return
+	}
 	fnID := c.Param(api.ParamFnID)
 	ctx, _ := common.LoggerWithFields(c.Request.Context(), logrus.Fields{"fnID": fnID})
 	c.Request = c.Request.WithContext(ctx)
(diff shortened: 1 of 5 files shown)


api/server/runner_fninvoke_test.go
```diff
@@ -374,7 +374,8 @@ func TestInvokeRunnerMinimalConcurrentHotSync(t *testing.T) {
 		expectedCode    int
 		expectedHeaders map[string][]string
 	}{
-		{"/invoke/fn_id", `{"sleepTime": 100, "isDebug": true}`, "POST", http.StatusOK, nil},
+		{"/invoke/fn_id", `{"sleepTime": 100, "isDebug": true}`, http.MethodPost, http.StatusOK, nil},
+		{"/invoke/fn_id", `{"sleepTime": 100, "isDebug": true}`, http.MethodPut, http.StatusMethodNotAllowed, nil},
```
rdallman: could you move this to the TestFnInvokeRunnerExecution block please? this test is for concurrency and testing the http method has nothing to do with testing concurrency, and could test the actual error message from there too.

api/server/runner_fninvoke.go
```diff
@@ -44,6 +45,12 @@ func (s *syncResponseWriter) Status() int          { return s.status }
 
 // handleFnInvokeCall executes the function, for router handlers
 func (s *Server) handleFnInvokeCall(c *gin.Context) {
+	if c.Request.Method != http.MethodPost {
+		handleErrorResponse(c, models.NewAPIError(http.StatusMethodNotAllowed,
+			errors.New("Method not allowed, only HTTP POST supported.")),
+		)
```
rdallman: wanna make `models.ErrInvokePostOnly` in `api/models/error.go` ? could test against the error string too, which would be nice

api/server/runner_fninvoke_test.go
denismakogon: Yup, thanks, addressed.

api/server/runner_fninvoke.go
denismakogon: Done.

rdallman:
> i dug around gin a little and didn't find much (an old open issue that says it's fixed, but it's not? maybe because we use custom no route handler and need to do it ourselves?) any chance you looked?

any word on this? i'll take another look today, would rather implement for all endpoints though if we can and it's relatively simple.

denismakogon:
That’s the reason why I did the way I did it.
Gin doesn’t have a functionality as we need.

rdallman:
@denismakogon I'm not entirely sure that's true (maybe it is, I didn't test this), it seems like we can do something like this:

```go
engine.NoRoute(func(c *gin.Context) {
    // replace all the values of the path with the original `:variable` to match them
    url := c.Request.URL.String()
    for _, p := range c.Params {
        url = strings.Replace(url, p.Value, ":"+p.Key, 1)
    }
    
    // if we find the route, but the method doesn't match, return 405
    var err models.APIError
    for _, route := range engine.Routes {
        if route.Path == url {
            err = models.ErrMethodNotAllowed
        }
    }

    // default to 404
    if err == nil {
        err = models.ErrPathNotFound
    }
    err = models.NewAPIError(e.Code(), fmt.Errorf("%v: %s %s", e.Error(), c.Request.Method, c.Request.URL.Path))
    handleErrorResponse(c, err)
})
```

denismakogon:
Okay, if the reason is to have a more-or-less versatile solution,`NoRoute` is a definitely a better option, despite the thing that we've faced with the only one case where we actually need to check HTTP methods.
PR updated, waiting for CI to become green

rdallman:
yea, our swagger docs (among others) anyway pretty clearly indicate it's POST, I think 'method not allowed' should hopefully trigger the not 100% lazy ones to go look at the docs and see POST without having to ask us. thanks. i think it's useful for other methods too (PATCH/PUT, eg)

rdallman:
ci didn't pass - yea, it might need a little fiddling with test-wise locally to get it working - i think the pieces are there, though?

denismakogon:
I'm working on it right now.

denismakogon:
The thing is `c.Params` is empty for `/invoke` path.

denismakogon:
Here's what works:
```go
	engine.NoRoute(func(c *gin.Context) {
		var err models.APIError
		fmt.Println(c.Keys)
		parts := strings.Split(c.Request.URL.Path, "/")
		if len(parts) > 1 {
			// redundant because this is not a real path parameters
			for i, v := range parts[2:] {
				parts[i+2] = ":"+v
			}
			domain := parts[1]
			u := strings.Join(parts, "/")
			fmt.Println("URL:", u)
			engine.Routes()
			for _, route := range engine.Routes() {
				if strings.HasPrefix(route.Path, domain) {
					err = models.ErrMethodNotAllowed
				}
			}
		}

		// default to 404
		if err == nil {
			err = models.ErrPathNotFound
		}
		err = models.NewAPIError(err.Code(), fmt.Errorf("%v: %s %s", err.Error(), c.Request.Method, c.Request.URL.Path))
		handleErrorResponse(c, err)
	})

```

rdallman:
that should be fine. why aren't the routes matching? can you print what we're looking for and what we're testing against?

rdallman:
can you just split off the leading `/` ? that seems weird. happy to look at it if it would help

denismakogon:
Okay, that seems to work, API returns 405 when HTTP not-POST to `/invoke/:fnID` happens.

rdallman:
you fit the solution to your test case, the solution won't work for > 1 dimensionality eg `/invoke/:thing/otherthing` it'll match the first one incorrectly

rdallman:
found 2 issues - 1) need to use `c.Request.URL.Path` and 2) `c.Params` doesn't seem to work as outlined. investigating 2, but 1 is a lot easier to fix than using strings.Split...

rdallman:
c.Params isn't filled in because the route didn't match anything - that makes sense now. thinking

rdallman:
i come bearing good news: https://github.com/gin-gonic/gin/blob/master/gin.go#L222 - wanna switch to this?

denismakogon:
thanks, looks better than mine.

PR #1318, opened 2018-11-19
fn: error handling and handler.Start semantics

skinowski:
1) call.Start should always call handler.Start, which
can decide to delete from MQ if async itself.
2) call.End errors are ignored, updated iface to
clarify


api/agent/call.go
```diff
@@ -325,27 +324,15 @@ func (c *call) Start(ctx context.Context) error {
...
-		// XXX (reed): should we store the updated started_at + status? we could
-		// use this so that if we pick up a call from mq and find its status is
-		// running to avoid running the call twice and potentially mark it as
-		// errored (built in long running task detector, so to speak...)
-
-		err := c.handler.Start(ctx, c.Model())
-		if err != nil {
-			return err // let another thread try this
-		}
+	err := c.handler.Start(ctx, c.Model())
```
rdallman: this will still make an api call if the hybrid api is being used, not sure it's what we want

rdallman: can we gate the hybrid client not to make this request if async? we also need to change direct data access to not delete the mq message here if it's sync, that call will fail too (oddly tests don't fail, this worries me)

api/agent/call.go
```diff
@@ -34,9 +34,8 @@ type Call interface {
 
 	// End will be called immediately after attempting a call execution,
 	// regardless of whether the execution failed or not. An error will be passed
-	// to End, which if nil indicates a successful execution. Any error returned
-	// from End will be returned as the error from Submit.
-	End(ctx context.Context, err error) error
+	// to End, which if nil indicates a successful execution.
+	End(ctx context.Context, err error)
```
rdallman: we need a way to communicate errors back from things here?

api/agent/call.go
rdallman: https://github.com/fnproject/fn/blob/master/api/agent/data_access.go#L157 think we need a similar gate here around if async

api/agent/call.go
skinowski: I don't think so, because two sources of error in call.End(), which are c.handler.Finish() or c.ct.fireAfterCall(). In the old code, we log c.handler.Finish() while return c.ct.fireAfterCall. But in the code base these are all ignored anyway. I felt like it's clearer to just acknowledge it and fixup the interface, but I can also take this change out of this PR if this looks problematic. (Also going forward I think we should also take error returns from fireAfterCalls() since this can cause leaks /issues since due to errors some of fireAfterCalls might not fire especially when multiple of these are registered.)

rdallman: yea, bug definitely snuck in here then. it should get the error from fire after call and hand it to end (for example, if fire after call errors, async may not want to delete the message in finish to retry it) - we don't have great testing around async and it's effectively dead now, yea. i agree about fire after call contract we should probably call all of them and hand off preceding errors, in general it's a bit of a mess with start/beforecall/finish/aftercall

PR #1321, opened 2018-11-21
actually disable stdout/stderr. stdout>stderr

rdallman:
* for pure runner this turns it off for real this time.
* this also just makes the agent container type send stdout to stderr, since
we're not using stdout for function output anymore this is pretty
straightforward hopefully.
* I added a panic and some type checking printlns to ensure this is true for
pure_runner, both stdout and stderr are off, also added a unit test from agent
to ensure this behavior from its container type, which pure_runner utilizes
(no integration test though)

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -652,7 +652,8 @@ func (s *hotSlot) dispatch(ctx context.Context, call *call) error {
 	defer span.End()
 
 	// TODO it's possible we can get rid of this (after getting rid of logs API) - may need for call id/debug mode still
-	swapBack := s.container.swap(call.stderr, call.stderr, &call.Stats)
+	// TODO there's a timeout race for swapping this back if the container doesn't get killed for timing out, and don't you forget it
+	swapBack := s.container.swap(call.stderr, &call.Stats)
 	defer swapBack()
 
 	resp, err := s.container.udsClient.Do(createUDSRequest(ctx, call))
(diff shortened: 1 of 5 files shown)


rdallman:
will fix up ci... can't run all tests locally these days woo

api/agent/agent.go
```diff
@@ -1239,29 +1227,30 @@ func newHotContainer(ctx context.Context, call *call, cfg *Config, id string, ud
...
+	c.stderr = stderr
 	ocs := c.stats
 	c.stats = cs
 	c.swapMu.Unlock()
 
 	return func() {
-		c.stdout.(common.GhostWriter).Swap(ostdout)
-		c.stderr.(common.GhostWriter).Swap(ostderr)
 		c.swapMu.Lock()
+		c.stderr = stderr
```
jan-g: Is this supposed to put the old stderr back?

rdallman: that's a bingo

api/agent/agent.go
```diff
@@ -1161,39 +1161,29 @@ func newHotContainer(ctx context.Context, call *call, cfg *Config, id string, ud
...
-		logrus.WithFields(logrus.Fields{"tag": "stderr", "app_id": call.AppID, "fn_id": call.FnID, "image": call.Image, "container_id": id}),
-	}}
+	// disable container logs if they're disabled on the call (pure_runner) -
+	// users may use syslog to get container logs, unrelated to this writer.
+	// otherwise, make a line writer and allow logrus DEBUG logs to host stderr
+	// between function invocations from the container.
 
-	stdout.Swap(newLineWriterWithBuffer(buf1, soc))
-	stderr.Swap(newLineWriterWithBuffer(buf2, sec))
+	var bufs []*bytes.Buffer
```
skinowski: we no longer need an array anymore, but this is fine too. :-)

skinowski:
@rdallman feel free to remove unused ghost code from common.

rdallman:
figured out why the ghost writer was a thing - need to swap out underlying writers handed off to the docker driver's stderr io.Writer - d'oh - so we keep them I guess.

PR #1323, opened 2018-11-26
Add a way to disable admin profiler

reclaro:
This change adds a way to disable the admin profiler when the admin
server is enabled.

I added a new env variable that can be used to disable the admin profiler
To verify it you can:
- Enable the admin api
- set the "FN_ENABLE_ADMIN_PROFILER" to "false"
-  GET the admin_api:admin_port/debug/vars, you should get an error

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -119,6 +119,9 @@ const (
 	// EnvMaxRequestSize sets the limit in bytes for any API request's length.
 	EnvMaxRequestSize = "FN_MAX_REQUEST_SIZE"
 
+	// EnvEnableAdminProfiler is used to enable the profiler for the admin API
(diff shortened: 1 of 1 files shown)


reclaro:
Hey @rdallman thanks for your comment, this is not a big deal i am happy to abandon this change if we think there is no value, the reason because I am proposing this is that we can have some FN implementations where different profiling mechanism are plugged into the system and therefore the net/http/pprof won't be necessary and ppl want a way to switch it off.

reclaro:
working on this thing I noticed that we don't call the wrapper for injecting middleware in the Admin API, I opened a PR for this https://github.com/fnproject/fn/pull/1327

rdallman:
@reclaro i think that #1327 is the right path forward here, thanks for following up

reclaro:
Ok I close this ticket in favor of #1327

PR #1332, opened 2018-11-30
fn: latency metrics for various call states

skinowski:
This complements the API latency metrics available
on LB agent. In this case, we would like to measure
calls that have finished with the following status:
     "completed"
     "canceled"
     "timeouts"
     "errors"
     "server_busy"
and while measuring this latency, we subtract the
amount of time actual function execution took. This
is not precise, but an approximation mostly suitable
for trending.

Going forward, we could also subtract UDS wait time and/or
docker pull latency from this latency as an enhancement
to this PR.

### DIFF
--- a/api/agent/lb_agent.go
+++ b/api/agent/lb_agent.go
@@ -353,14 +353,16 @@ func recordCallLatency(ctx context.Context, call *call, status string) {
 	// request processing latency as observed by runner agent.
 	execLatency := call.GetUserExecutionTime()
 
-	// some sanity check before
-	if execLatency != nil && *execLatency > callLatency {
-		common.Logger(ctx).Errorf("invalid latency callLatency=%v execLatency=%v", callLatency, execLatency)
-		return
-	}
+	// some sanity check before. If sanity checks flags something, then
+	// this is likely that runners are sending malicious/suspicious data.
 	if execLatency != nil {
+		if *execLatency >= callLatency {
+			common.Logger(ctx).Errorf("invalid latency callLatency=%v execLatency=%v", callLatency, execLatency)
+			return
+		}
 		callLatency -= *execLatency
 	}
+
 	statsCallLatency(ctx, callLatency, status)
 }
 
--- a/api/agent/lb_agent.go
+++ b/api/agent/lb_agent.go
@@ -303,27 +303,65 @@ func (a *lbAgent) handleCallEnd(ctx context.Context, call *call, err error, isFo
 		statsStopRun(ctx)
 		if err == nil {
 			statsComplete(ctx)
+			recordCallLatency(ctx, call, completedMetricName)
 		}
 	} else {
 		statsDequeue(ctx)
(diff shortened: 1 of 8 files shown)


api/agent/lb_agent.go
```diff
@@ -303,28 +303,66 @@ func (a *lbAgent) handleCallEnd(ctx context.Context, call *call, err error, isFo
...
+		common.Logger(ctx).Error("cannot determine call start time")
+		return
+	}
+
+	// We want to exclude time spent in user-code. Today, this is container
+	// request processing latency as observed by runner agent.
+	execLatency := call.GetUserExecutionTime()
+
+	// some sanity check before
+	if execLatency != nil && *execLatency > callLatency {
```
kmjohansen: Wonder if we could re-factor 355-363 to something like:
```
if execLatency != nil {
    if *execLatency >= callLatency {
        common.Logger(ctx).Warnf("invalid latency callLatency=%v execLatency=%v", callLatency, execLatency)
        return
    }
    callLatency -= *execLatency
}
```

Aside from reducing the number of `execLatency` pointer checks, this would also a) ensure that we don't generate a callLatency of zero, which would presumably also be bogus; and b) prevent any low severity pages from logged errors.

kmjohansen:
Naming nit: using `UserExecutionTime` sounds a bit like we're collecting user, sys, and idle times.  I was going to suggest `ContainerExecutionTime`, but I realize that doesn't differentiate between the FDK / user cases, so maybe this is fine as is.  If you have other ideas for a name that doesn't share the potential for confusion with tracking time spent other cpu states, though, I'm certainly open to revision.

rdallman:
@skinowski did call.CompletedAt-call.StartedAt diverge from meaning actual user execution time?

skinowski:
@rdallman This is actually completedAt-StartedAt coming from runners but this is recorded as UserExecutionTime on LB side. On LB side, StartedAt timestamp has a different meaning. StartedAt is "placer started searching". We probably could shuffle things and make these right on LB, but since now handlers/callbacks are executed *before* placer algorithm kicks in, it is difficult to change with dependencies. Another approach is set call.StartedAt on LB side again after the call is executed using 'StartedAt' from runners, but either way it's a hack.

rdallman:
oh boy. thanks for explanation. it seems to have lost its meaning in that case (and without storing them, why not I guess) - where it wouldn't be set until execution began on the runner. I guess if that were the case most of this wouldn't have been necessary though (new interface methods, methods on call, etc)

> Another approach is set call.StartedAt on LB side again after the call is executed using 'StartedAt' from runners, but either way it's a hack.

this makes sense to me and is the intended usage anyway (as it's documented on the field itself), with the exception that LB probably shouldn't set it to begin with and should read it from the runner's call object (intuitively, maybe the runner call object should really replace the lb call object?). i guess ideally the LB just uses CreatedAt as there isn't a meaningful amount of time between the call object getting created in the lb and starting to place things? idk

Issue #1333, opened 2018-11-30
Default Container Initialization Timeout is Too Short for Some FDK environments


gviedma:
Depending on the amount of CPU/memory available to a function container, it is possible that the current timeout of 5 seconds is insufficient to load all of the customer's imports. In interpreted languages like Python and Node, the problem is exacerbated by the fact that there is mechanism to enforce that the UDS socket listener is started before loading any customer code or doing other initialization. I propose that we raise the default to 30 seconds, which should cover most customer initialization use-cases.

PR #1334, opened 2018-11-30
#1333: Raise default container start timeout to 30 seconds

gviedma:
Addresses https://github.com/fnproject/fn/issues/1333 by raising the default container start timeout to 30 seconds.


kmjohansen:
I'd also prefer that we not implement a solution that increases the timeout unless it's absolutely necessary.  We're fixing the specific problem in the FDK.  I'd vote to close this out.

PR #1336, opened 2018-11-30
fn: uds init wait latency metric in prometheus

skinowski:
Adding tracker for UDS initialization during container start. This
complements our existing container state latency trackers and
docker-api latency trackers.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -895,24 +895,26 @@ func (a *agent) runHot(ctx context.Context, caller slotCaller, call *call, tok R
 	go func() {
 		defer cancel() // also close if we get an agent shutdown / idle timeout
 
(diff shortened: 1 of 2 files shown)


api/agent/agent.go
```diff
@@ -895,16 +895,24 @@ func (a *agent) runHot(ctx context.Context, caller slotCaller, call *call, tok R
 	go func() {
 		defer cancel() // also close if we get an agent shutdown / idle timeout
 
+		// We record init wait for two cases below: I. when it succeeds, II. when it times out,
+		// other cases are unusual early exits (eg. agent shutdown, eviction, errors, etc)
+		// and we assume that those cases are already tracked separately and should not pollute
```
kmjohansen: OOC, is this assumption true?

skinowski: updated with tags to avoid this assumption.

PR #1337, opened 2018-12-01
fn: error handling updates

skinowski:
* docker-pull timeout is now a 504 which classifies it as a
    service error. Avoid using 503 to make sure LB does not retry.

    * Only applicable to detached mode, a timeout on LB is
    now a ErrServiceReservationFailure (500). In detached mode,
    this is unlikely to make it back to a client and it is mostly
    for documentation/metrics purposes.
      
    * For Triggers, avoid scrubbing service code.

### DIFF
--- a/api/agent/drivers/docker/cookie.go
+++ b/api/agent/drivers/docker/cookie.go
@@ -306,13 +306,24 @@ func (c *cookie) PullImage(ctx context.Context) error {
 		// TODO need to inspect for hub or network errors and pick; for now, assume
 		// 500 if not a docker error
 		msg := err.Error()
-		code := http.StatusInternalServerError
+		dockerCode := 0
 		if dErr, ok := err.(*docker.Error); ok {
 			msg = dockerMsg(dErr)
-			code = dErr.Status // 401/404
+			dockerCode = dErr.Status // 401/404
 		}
 
-		return models.NewAPIError(code, fmt.Errorf("Failed to pull image '%s': %s", c.task.Image(), msg))
+		finalCode := 0
+
+		switch {
+		case dockerCode == http.StatusGatewayTimeout:
+			finalCode = http.StatusGatewayTimeout
+		case dockerCode >= 400 && dockerCode < 500:
+			finalCode = dockerCode
+		default:
+			finalCode = http.StatusBadGateway
+		}
+
+		return models.NewAPIError(finalCode, fmt.Errorf("Failed to pull image '%s': %s UpstreamStatus=%v", c.task.Image(), msg, dockerCode))
 	}
 
 	return nil
--- a/api/agent/drivers/docker/cookie.go
+++ b/api/agent/drivers/docker/cookie.go
@@ -301,23 +301,17 @@ func (c *cookie) PullImage(ctx context.Context) error {
 
 	err := c.drv.docker.PullImage(docker.PullImageOptions{Repository: repo, Tag: c.imgTag, Context: ctx}, *cfg)
 	if err != nil {
+		log.WithError(err).Error("Failed to pull image")
+
 		// TODO need to inspect for hub or network errors and pick; for now, assume
 		// 500 if not a docker error
 		msg := err.Error()
 		code := http.StatusInternalServerError
-		origin := 0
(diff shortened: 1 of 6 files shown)


skinowski:
WIP: I will add tests based on discussion outcome.

api/agent/drivers/docker/cookie.go
```diff
@@ -306,13 +306,24 @@ func (c *cookie) PullImage(ctx context.Context) error {
...
+		switch {
+		case dockerCode == http.StatusGatewayTimeout:
+			finalCode = http.StatusGatewayTimeout
+		case dockerCode >= 400 && dockerCode < 500:
+			finalCode = dockerCode
+		default:
+			finalCode = http.StatusBadGateway
+		}
+
+		return models.NewAPIError(finalCode, fmt.Errorf("Failed to pull image '%s': %s UpstreamStatus=%v", c.task.Image(), msg, dockerCode))
```
mantree: I don't like this determining of the "final code" this deep in the app. For me, the final code needs to be decided at one point at the front. At this level, we should just be communicating what the problem is, not making decisions about how it should be represented.

skinowski: Updated the code as per the discussion. We only cherry pick 4xx range here.

rdallman:
thanks x10 for the really clear description of behavior. a few small things that stick out to me just as discussion items really:

* it's not clearly defined that this is the case only for the invoke api, and not clear whether user containers are attempting to return status codes from the container. our contract defines that containers may only return 502 and 504, maybe it's early but it seems suggestive that containers can return other codes and we'll sweep them into 502? maybe it's something not to worry about and i'm reaching.
* docker pull timeout seems like a 504? honestly don't think these should get bubbled up to user really even (from a service install perspective the lb should timeout and retry elsewhere long before this anyway), so agree about it being service level error just not returned to user I guess (user would get 504 that request timed out, not docker)
*
>        * service http code takes priority
>        * if service http code is 2xx, only then we decap user-http-code
>        * user-http-code can return 502 and 504, but other 5xx codes are scrubbed into a 502.  
guess i'd flip these, if user manages to get a code out then i think it should get returned regardless of eg if container timed out or conn got broken, etc. it makes sense in the current world where we buffer things, but if we defined this strictly as a contract then there's an obvious breakage if/once we allow streaming response out of container and client can handle unexpected EOF / fucked up chunk / fucked up content length / etc - I guess a simpler case would be 200 still implies the thing happened on the server side, like eg inserting a db row, and while apps need to tolerate 500s and ofc can't know for sure the thing happened unless they got a 200, in this case they'd get that user 200 code out and short of above assumption being incorrect I don't see reason not to return this other than the convenience of trying to plumb the service error out for our metrics (we should have lower level metrics around most of these failure cases, most likely, though?). anyway, I do worry about the api breakage here around streaming especially.
* 
>        * user-http-code can return 502 and 504, but other 5xx codes are scrubbed into a 502.
still don't really agree with this, I think it's a misinterpretation of the gateway contract altogether. in a service setting (the reason we're here...) really, this should be returned from invoke itself with the user-http-code in the headers intact, and the gateway should unwind this. it's what is happening afaik, however there's the little direct http endpoint also in fn, but let's ignore that (it's a local version of the gateway handler, basically, so semantics are same). anyway, i maintain that users should be able to get out whatever codes they want when using http triggers for all manner of reasons - namely, the whole serverless use case around running your API with a serverless backend becomes mooted for many because we're adding an incompatibility for most API implementations for our own metrics convenience, seemingly. I appreciate we need a way to analyze whether the codes returned from this endpoint are service errors or not, I don't agree about limiting the output range for users to accomplish that I guess. to substantiate health checking of lb i'd hope we have explicitly defined health check endpoint and we could likely scrape up 'system errors' vs 'user errors' there, for things like logs the trigger endpoints are clearly marked with /t/ to filter, etc etc. at least, i see alternative solutions that don't limit user's ability that exist anyway if we examine why we're doing this.

mantree:
Can we get the status determining code broken out like in the other PR and use 'userCode' and 'serviceCode' or something similar. Make it really clear where these things have come from and how we're going to decide between them.

PR #1340, opened 2018-12-05
fn: cherry pick 4xx errors from docker-pull and map rest to 502

skinowski:

### DIFF
--- a/api/agent/drivers/docker/cookie.go
+++ b/api/agent/drivers/docker/cookie.go
@@ -306,10 +306,12 @@ func (c *cookie) PullImage(ctx context.Context) error {
 		// TODO need to inspect for hub or network errors and pick; for now, assume
 		// 500 if not a docker error
 		msg := err.Error()
-		code := http.StatusInternalServerError
+		code := http.StatusBadGateway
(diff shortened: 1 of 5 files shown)


api/agent/drivers/docker/cookie.go
```diff
@@ -309,7 +309,12 @@ func (c *cookie) PullImage(ctx context.Context) error {
 		code := http.StatusInternalServerError
 		if dErr, ok := err.(*docker.Error); ok {
 			msg = dockerMsg(dErr)
-			code = dErr.Status // 401/404
+			code = dErr.Status // 500 or 404 according to docker-api
+
+			// This is unfortunate, docker-api returns "API error (500): unauthorized: authentication required" not 401.
+			if dErr.Status == http.StatusInternalServerError && strings.Contains(msg, "unauthorized: authentication required") {
```
hhexo: I think different docker registries (that is, different implementations of the docker api) might return different strings here. I'm not sure the api specification absolutely requires the message to be like that.

rdallman: it does not, and returns various codes for different modes of saying 'not authenticated' even, good catch

PR #1344, opened 2018-12-07
adds tests for Fn-Call-Id, and adds for timeouts

rdallman:
back when we streamed responses out from the container, we set this header
regardless of whether the task ran or not. now, we're setting this header at
the end since we're doing buffering, only if the task runs successfully atm.
we can change to the old policy pretty readily if we want to, up for
discussion. as this patch presents it, it's just adding a test that it exists
in the normal case and the timeout case, and adding the timeout case. however,
other relevant cases exist, like fdk protocol errors, and we can whitelist
those each or just cover all errors, even if the call was never executed, up
to us. when we add streaming, the plumbing goes back to the way it was where
we just add this at some point before the container writes any headers out of
it, and we can pick what level to do that if we want special behavior around
this header for whether the call executed or not.

### DIFF
--- a/api/server/runner_fninvoke.go
+++ b/api/server/runner_fninvoke.go
@@ -97,18 +97,16 @@ func (s *Server) fnInvoke(resp http.ResponseWriter, req *http.Request, app *mode
 		return err
 	}
 
+	// add this before submit, always tie a call id to the response at this point
+	writer.Header().Add("Fn-Call-Id", call.Model().ID)
+
 	err = s.agent.Submit(call)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


rdallman:
I feel inclined to add it for all error cases as well, will make that change here. thanks for feedback

rdallman:
@skinowski https://github.com/fnproject/fn/pull/1344/commits/6449dfc22e6cd88fb532bb96fe7668377165bc89

skinowski:
LGTM

rdallman:
ty

PR #1345, opened 2018-12-07
fn: logging/context improvements for runner status calls

skinowski:
Avoid blocking calls to runStatusCall() to make sure gRPC
context can be cancelled/timedout. This is unlikely an
issue, but blocked runStatusCall() while gRPC is cancelled
is a hard to follow case mentally. New flow is a bit
easier to follow.

Log all error cases in Status() gRPC entry point including
client side cancellations.

### DIFF
--- a/api/agent/pure_runner.go
+++ b/api/agent/pure_runner.go
@@ -746,15 +746,10 @@ func (pr *pureRunner) Engage(engagement runner.RunnerProtocol_EngageServer) erro
 
 // Runs a status call using status image with baked in parameters.
 func (pr *pureRunner) runStatusCall(ctx context.Context) *runner.RunnerStatus {
-
-	// IMPORTANT: We have to use our own context with a set timeout in order
-	// to ignore client timeouts. Original 'ctx' here carries client side deadlines.
-	// Since these deadlines can vary, in order to make sure Status runs predictably we
-	// use a hardcoded preset timeout 'ctxTimeout' instead.
-	// TODO: It would be good to copy original ctx key/value pairs into our new
-	// context for tracking, etc. But go-lang context today does not seem to allow this.
-	execCtx, execCtxCancel := context.WithTimeout(context.Background(), StatusCtxTimeout)
-	defer execCtxCancel()
+	// IMPORTANT: apply an upper bound timeout
+	var cancel context.CancelFunc
+	ctx, cancel = context.WithTimeout(ctx, StatusCtxTimeout)
+	defer cancel()
 
 	result := &runner.RunnerStatus{}
 	log := common.Logger(ctx)
@@ -790,7 +785,7 @@ func (pr *pureRunner) runStatusCall(ctx context.Context) *runner.RunnerStatus {
 	agentCall, err := pr.a.GetCall(FromModelAndInput(&c, player),
 		WithLogger(common.NoopReadWriteCloser{}),
 		WithWriter(recorder),
-		WithContext(execCtx),
+		WithContext(ctx),
 	)
 
 	if err == nil {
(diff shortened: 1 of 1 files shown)


api/agent/pure_runner.go
```diff
@@ -836,87 +831,78 @@ func (pr *pureRunner) runStatusCall(ctx context.Context) *runner.RunnerStatus {
...
-		// Wait channel is not present, we install a new one.
-		myChan = make(chan struct{})
-		pr.status.wait = myChan
+	isCached = now.Before(pr.status.expiry)
+	if !isCached {
+		if pr.status.wait == nil {
+			pr.status.wait = make(chan struct{})
+			isSpawner = true
+		}
+		waitChan = pr.status.wait
```
rdallman: maybe nitty but it would make the rest of the function a good deal simpler I think if we had a little func that did 895-903 and loaded the cache object and returned it, so that we could do something like:

```go
if isCached {
  return loadCacheObj(pr.status)
}

// now handle the uncached case
```

there's pretty much 3 blocks below here that handle whether it's a spawner or not a cached object, and it can't be a spawner if it's cached either, so I think it will clean up a lot of the conditional blocks and they could just be done at the top level for the !isCached at least, and the spawner case is slightly more straightforward of a condition. it will read more simply... is it cached? yes, okay return or no, do we need to spawn? okay, maybe, then we wait for that result and load it and return it. et voila...

that being said, the code seems correct as i've understood it as is.

api/agent/pure_runner.go
```diff
@@ -836,87 +831,78 @@ func (pr *pureRunner) runStatusCall(ctx context.Context) *runner.RunnerStatus {
...
 	pr.status.lock.Unlock()
 
-	// signal waiters
-	close(myChan)
-	return cachePtr, nil
+	cacheObj.Cached = isCached
+	cacheObj.Active = atomic.LoadInt32(&pr.status.inflight)
+	cacheObj.RequestsReceived = atomic.LoadUint64(&pr.status.requestsReceived)
+	cacheObj.RequestsHandled = atomic.LoadUint64(&pr.status.requestsHandled)
+	cacheObj.KdumpsOnDisk = atomic.LoadUint64(&pr.status.kdumpsOnDisk)
```
rdallman: it maybe makes sense to do this only from the spawner goroutine and maybe before spawning the actual status call (as the context here itself could be indicative of why the status call fails), and return only that cached object instead of loading these fields at the current state for any cached object, then the status object returned is identical to the status when the status call itself ran for any calls to this where it's cached in a window. for example, the status call itself could run when there are 10000 outstanding requests, this could cause the status call to fail. while we're returning that failed state when returning a cached object, we're also going to reload the number of outstanding requests, which could now be 10. it makes sense when the status call fails with 10000 outstanding requests when maybe it makes a lot less sense when there are only 10 and also the server is in a different state wrt these than when the status call actually ran - at least, we've lost all the context of the outstanding status of the server when the status call failed (ie these fields) outside of what we get back from the status call itself. that was a bit rambly, if that's not coherent press 1 to try again...

rdallman:
added @hhexo and @gviedma for double check if they wish

api/agent/pure_runner.go
skinowski: These fields are used by PM and they need to be up-to-date. Logically your comment makes sense as it's a bit weird here we keep some section of data up-to-date, but rest comes from cache. I'll add some comments to clarify why we are doing this odd data copy.

api/agent/pure_runner.go
skinowski: I've resisted so far to split functions off of this big function to keep/isolate lock code in same block. But that's probably OK, let me see if that version will look simpler.

PR #1346, opened 2018-12-10
Make syslog-url nullable in the swagger so it can be explicitly unset

hhexo:
- What I did
Fixed the swagger spec to allow syslog-url in apps to be nullable, therefore enabling the possibility to distinguish between it being not set and it being explicitly set to an empty string.
The Fn server already handles that distinction, but the swagger did not allow clients to take advantage of that.

- How I did it
The swagger spec is changed to include the `x-nullable` notation for `syslog-url` in the `App` definition.

- How to verify it
Regenerate clients, see that the generated code changes according to the language of choice.

Fixed the swagger spec to allow syslog-url in apps to be nullable, therefore enabling the possibility to distinguish between it being not set and it being explicitly set to an empty string.

### DIFF
--- a/docs/swagger_v2.yml
+++ b/docs/swagger_v2.yml
@@ -504,6 +504,7 @@ definitions:
           type: object
       syslog_url:
         type: string
+        x-nullable: true
         description: "A comma separated list of syslog urls to send all function logs to. supports tls, udp or tcp. e.g. tls://logs.papertrailapp.com:1"
       created_at:
         type: string


hhexo:
Note that while this change is not a breaking change for the Fn server (as it was already handling the distinction between "set to empty" and "not set"), it _will_ cause a backwards-incompatible change for the generated code on some clients (e.g. if `fn_go` is regenerated it will now require a `*string` instead of a `string` for that field).

PR #1348, opened 2018-12-11
adds user blame tag for trigger/invoke response metric

rdallman:
we are attempting to ascertain whether a status code is from a function or
from the platform for invoke and http triggers. this proposes that some errors
(models.APIError) are at the same level of fault to the user as a user
returning a given status code from their trigger or from the container,
examples of such errors are: docker 401, timeout 504, bad function response
502. the thinking being, these are equally the fault of the user and not the
platform, whereas things like panics, or rpc errors, or something we're not
expecting is 'the platform's fault'.

this adds a tag for blame=[service|user] which defaults to 'service' unless
set by invoke or http trigger, by the above standard. the plumbing in invoke
and http trigger sucks because we don't have the gin context thing to mutate
down in the invoke function we have (to make it easier to work with our fancy
response writers), so this is a bit blah, esp if we start whitelisting certain
errors, but it's a start.


api/server/gin_middlewares.go
```diff
@@ -31,6 +31,11 @@ var (
 	apiResponseCountMeasure = common.MakeMeasure("api/response_count", "API response count", stats.UnitDimensionless)
 	apiLatencyMeasure       = common.MakeMeasure("api/latency", "Latency distribution of API requests", stats.UnitMilliseconds)
 
+	// ambiguously named, both of these measure responses from post function invocation (or, if the function
+	// cannot be placed, after that) - critically, only expected errors or function responses.
+	invokeResponseCountMeasure      = common.MakeMeasure("api/invoke_count", "Invoke response count", stats.UnitDimensionless)
+	httpTriggerResponseCountMeasure = common.MakeMeasure("api/http_trigger_count", "HTTP trigger response count", stats.UnitDimensionless)
```
rdallman: i'm speechless for names, have at it... describing what this is exactly is not intuitive imo either (and makes me think it's all wrong, but meh)

api/server/runner_httptrigger.go
```diff
@@ -57,7 +58,20 @@ func (s *Server) handleTriggerHTTPFunctionCall2(c *gin.Context) error {
 	// gin sets this to 404 on NoRoute, so we'll just ensure it's 200 by default.
 	c.Status(200) // this doesn't write the header yet
 
-	return s.ServeHTTPTrigger(c, app, fn, trigger)
+	err = s.ServeHTTPTrigger(c, app, fn, trigger)
+	if _, ok := err.(models.APIError); ok || err == nil {
+		// report all user-directed errors and function responses from here, after submit has run.
+		// this is our never ending attempt to distinguish user and platform errors.
+		status := strconv.Itoa(c.Writer.Status())
```
rdallman: d'oh... 1s

api/server/runner_fninvoke.go
```diff
@@ -66,7 +67,19 @@ func (s *Server) handleFnInvokeCall2(c *gin.Context) error {
...
+		// this is our never ending attempt to distinguish user and platform errors.
+		ctx, err := tag.New(c.Request.Context(),
+			tag.Upsert(whodunitKey, "user"),
+		)
+		if err != nil {
+			panic(err)
+		}
+		c.Request = c.Request.WithContext(ctx)
+	}
+	return err
```
skinowski: I think we need to move these insertion code to runner. And apply it to cases where we know it's user error. Here, runner could respond with APIError which may or may not be user errors.

rdallman: i'd prefer to do error handling at the highest point possible and feed things into it, especially since we're modifying the context in this specific case in order to do this already. if we need a 'UserAPIError' interface to do that, then so be it (it can implement the `APIError` interface, as well, just with additional method/separate factory method). the thinking in this patch is, we'll start with all APIErrors as being user errors, and any errors we're not expecting (grpc, weird docker error, eg; anything that would 503 normally) are 'service errors'. from that point, we get good coverage (a lot better than now), and can work to blacklist whatever might be 'service' errors; at present, there's no clear definition or list of these. so, we can add facilities to do `UserAPIError` but this will make this patch additionally change probably all errors in `models/` package to `UserAPIError` with maybe a few exceptions (I am not familiar enough with all of our errors to say, but no cases come to mind really) -- up to discuss, intuitively it seems like a bit of a wash without us clearly defining what the difference is.

rdallman: ok. I'm going to propose something relatively similar but for the agent package specifically. here's what I'm thinking:

```go
// in agent package (for use by agent, pure_runner, lb_agent, and transmitted over grpc)
// if this should all be exported to be used from other extenders of fn, we can do that (tell me),
// i didn't think too much about it but it would make sense

// funcError is an error that is the function's fault, that uses the models.APIError but distinguishes fault to function specific errors
type funcError interface {
  models.APIError

  // hidden method needed for duck typing
  userError()
}

type concFuncError models.APIError // maybe embedded, I didn't compile...

func (c concFuncError) userError() { }

func newFuncError(err models.APIError) error { return concFuncError(err) }

func isFuncError(err error) bool { _, ok := err.(funcError); return ok }

// add field 'userError bool' to CallFinished grpc which pure_runner can use isFuncError(err)
// to set the userError field bit in order to distinguish user and service errors over grpc. 
// lb_agent can receive these and recreate the function error (or just normal apierror if not) by 
// doing newFuncError(models.NewAPIError(callproto.status, callproto.msg)) which can then be
// bubbled up to the front end to be consumed as shown in this patch already, which we need
// to do in order to add this to higher level metrics used in middleware (the relevant status code
// metrics)
```

rdallman:
small sample of metrics:

```
fn_api_latency_sum{blame="user",method="POST",path="/invoke/:fn_id",status="200"} 63173
fn_api_latency_count{blame="user",method="POST",path="/invoke/:fn_id",status="200"} 20
fn_api_response_count{blame="service",method="GET",path="/metrics",status="200"} 2
fn_api_response_count{blame="service",method="GET",path="/v2/apps",status="200"} 20
fn_api_response_count{blame="service",method="GET",path="/v2/fns",status="200"} 20
fn_api_response_count{blame="user",method="POST",path="/invoke/:fn_id",status="200"} 20
```

this applies these to every endpoint as can be seen (for cardinality reasons...), where only invoke and http trigger will have 'blame="user"' in practice.

i left details on commit message https://github.com/fnproject/fn/pull/1348/commits/eeefd2df97a4cca1809e33f392adb7b8fb59aa09 but it should be easy to track the errors this now covers from the api/models/error.go - this patch is intentionally not exhaustive, we'll have to cull these kinds of errors over time from data, but this gives us pretty good coverage to start hopefully.

still fiddling with ci it looks...

PR #1349, opened 2018-12-11
add gosec scanning to ci

### LINKED ISSUE #1303: add gosec to CI build
https://github.com/securego/gosec tells you useful things about how your code can be extorted. I attempted to run it against our codebase and it hung for over an hour, need to debug, I don't know if our code is so insecure that it would take hours to analyze everything wrong with it or if the tool is stalling out from misuse or bugs (cpu would indicate the latter). 

we can add this to CI if it runs in a reasonable amount of time and it would be nice to catch some of these things automatically.

rdallman:
gosec severity=medium passes, all severity=low errors are from unhandled
errors, we have 107 of them. tbh it doesn't look worth it to me, but maybe
there are a few assholes even itchier than mine out there. medium has some
good stuff in it, and of course high makes sense if we're gonna do this at
all.

this adds some nosec annotations for some things like sql sprintfs where we
know it's clean (we're constructing the strings with variables in them). fixed
up other spots where we were sprinting without need.

some stuff like filepath.Clean when opening a file from a variable, and file
permissions, easy stuff...

I can't get the CI build to shut up, but I can locally get it to be pretty
quiet about imports and it just outputs the gosec output. fortunately, it
still works as expected even when it's noisy. I got it to shut up by unsetting
some of the go mod flags locally, but that doesn't seem to quite do it in
circle, printed the env out and don't see them, so idk... i give up, this
works

closes #1303

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ jobs:
     working_directory: ~/go/src/github.com/fnproject/fn
     environment: # apparently expansion doesn't work here yet: https://discuss.circleci.com/t/environment-variable-expansion-in-working-directory/11322
       - GOPATH=/home/circleci/go
-      - GOVERSION=1.11
+      - GOVERSION=1.11.2
       - OS=linux
       - ARCH=amd64
       - FN_LOG_LEVEL=debug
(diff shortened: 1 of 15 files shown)


rdallman:
ty

PR #1351, opened 2018-12-12
added change for 'fn_id' to propagate through to logs

nishalad95:


api/server/server.go
```diff
@@ -974,11 +975,28 @@ func whoAmI() net.IP {
...
 	}
 	return fields
 }
 
+var camel = regexp.MustCompile("(^[^A-Z0-9]*|[A-Z0-9]*)([A-Z0-9][^A-Z]+|$)")
+
+// converts a camelCase string to lowercase_and_underscore
+func camelCaseToUnderscore(s string) string {
+	var newField []string
+	for _, sub := range camel.FindAllStringSubmatch(s, -1) {
```
rdallman: this is going to run on every request (multiple times, for certain paths) so I'd like to make sure that this function runs with minimal overhead. regex is going to be very slow to match this, go's regex isn't very cheap generally and this query is really simple to replace without using regex. i'd say that we should run through these and find the indices to replace and then create one string with the replacements via concatenation, but honestly we should just change all the url params located in this file to be snake case eg `fn_id` instead of `fnID` -- the params are only used by our code, we should just make them what we need and not do any string replacement - we probably should have made them all snake case a long time ago, anyway, so that seems like the best plan of action to me...

nishalad95: @rdallman I agree, there was some confusion over this and so I am reverting back to a previous commit.

api/server/server.go
```diff
@@ -1148,25 +1148,25 @@ func (s *Server) bindHandlers(ctx context.Context) {
 			runner.POST("/finish", s.handleRunnerFinish)
 
 			runnerAppAPI := runner.Group(
-				"/apps/:appID")
+				"/apps/:app_id")
 			runnerAppAPI.Use(setAppIDInCtx)
 			// Both of these are somewhat odd -
 			// Deprecate, remove with routes
-			runnerAppAPI.GET("/triggerBySource/:triggerType/*triggerSource", s.handleRunnerGetTriggerBySource)
+			runnerAppAPI.GET("/trigger_by_source/:trigger_type/*trigger_source", s.handleRunnerGetTriggerBySource)
```
skinowski: NOTE: This changes the API, so it's a breaking change.

rdallman:
LGTM after fixing @skinowski 's note, that was the only line i saw any api breakage. thanks

PR #1353, opened 2018-12-14
Format error string and return do not log error reduce log spamming

hibooboo2:
- What I did
Format error string in grpc dial function to reduce log spam.
- How I did it
Changed the error that is returned.
- How to verify it
Read the logs.
Reduce log spam from grpc dials

### DIFF
--- a/grpcutil/dial.go
+++ b/grpcutil/dial.go
@@ -28,7 +28,7 @@ func dial(ctx context.Context, address string, creds credentials.TransportCreden
 		defer cancel()
 		conn, err := (&net.Dialer{Cancel: ctx.Done(), Timeout: timeoutDialer}).Dial("tcp", address)
 		if err != nil {
-			log.WithError(err).Warn("Failed to dial grpc connection")
+			log.WithError(err).Debug("Failed to dial grpc connection")
 			return nil, err
 		}
 		if creds == nil {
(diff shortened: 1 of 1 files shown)


skinowski:
Does this still log the IP:port of the target? Before this PR, logger was set to include gRPC address.

grpcutil/dial.go
```diff
@@ -28,8 +29,7 @@ func dial(ctx context.Context, address string, creds credentials.TransportCreden
 		defer cancel()
 		conn, err := (&net.Dialer{Cancel: ctx.Done(), Timeout: timeoutDialer}).Dial("tcp", address)
 		if err != nil {
-			log.WithError(err).Warn("Failed to dial grpc connection")
-			return nil, err
+			return nil, fmt.Errorf("Failed to dial grpc connection: %s", err.Error())
```
rdallman: the types on the connections returned from the net package are useful - eg they implement the `Temporary()`, `Timeout()` methods, see https://golang.org/pkg/net/#AddrError.Temporary - would prefer to preserve these and more generally not use fmt.Errorf to wrap errors anywhere (it's bit us already)  - i know and i know you know that go errors suck, sigh, logging may be best but i'm guessing this is being proposed because there wasn't enough context in the logs?

hibooboo2:
Just changed logs to debug

PR #1363, opened 2019-01-04
fn: introducing startup/restart container cleanup

skinowski:
If runner/agent is killed or crashes, containers may be
left running and they may not properly get cleaned up.
A subsequent start of the agent should scan containers and
if these belong to previous runner/agent instances, they
should be cleaned up. Since the runner/agent tracks
cpu/memory resources, the cleanup is essential.

This PR introduces a simple background task that
executes once at docker driver start. The docker
driver compares container labels to determine if
containers belong to current or previous instances
of the process and initiates a cleanup accordingly.

### DIFF
--- a/api/agent/drivers/docker/docker_client.go
+++ b/api/agent/drivers/docker/docker_client.go
@@ -300,8 +300,8 @@ func (d *dockerWrap) ListContainers(opts docker.ListContainersOptions) (containe
 	ctx, closer := makeTracker(opts.Context, "docker_list_containers")
 	defer closer()
 
-	logger := common.Logger(ctx).WithField("docker_cmd", "ListContainers")
-	err = d.retry(ctx, logger, func() error {
+	ctx, _ = common.LoggerWithFields(ctx, logrus.Fields{"docker_cmd": "ListContainers"})
+	err = d.retry(ctx, func() error {
 		containers, err = d.docker.ListContainers(opts)
 		return err
 	})


api/agent/drivers/docker/docker.go
```diff
@@ -19,9 +19,15 @@ import (
 	"github.com/fnproject/fn/api/common"
 	"github.com/fnproject/fn/api/models"
 	"github.com/fsouza/go-dockerclient"
+	"github.com/rs/xid"
```
rdallman: is this necessary? we have `github.com/fnproject/fn/api/id` which generates ids based on machine id, time and a monotonic (no process id but similar to xid, but are we worried about sub millisecond restarts?) - adding dependencies is a whole thing now, and this one just seems unnecessary without any explanation

rdallman: in any event, if we are, it seems like pid + timestamp would work for this purpose fine without needing the dependency (worst case we say that machine id and monotonic don't add any entropy from xid anyway)

api/agent/drivers/docker/docker.go
```diff
@@ -102,18 +105,88 @@ func NewDocker(conf drivers.Config) *DockerDriver {
...
+	}
+
+	containers, err := driver.docker.ListContainers(opts)
+	if err != nil {
+		logrus.WithError(err).Errorf("cannot list docker containers with filter %s", filter)
+		return
+	}
+
+	for _, item := range containers {
+		logrus.Errorf("checking %+v", item)
```
rdallman: nit: can we log this with debug level? it's just alarming I guess for someone not familiar to see a bunch of errors fly out at the beginning that aren't really errors and we're logging terminations below already

api/agent/drivers/docker/docker.go
skinowski: Swapped this with crypto/rand implementation.

PR #1369, opened 2019-01-09
fix build, add pointers for forking

rdallman:
added some pointers with commands for forking fn and making contributions to
it. i also did a test of using go1.11 and cloning the repo to somewhere
outside of my $GOPATH and then doing `make build` and it required making a
small change to the noop stuff in order for the build to work. pretty sweet
that the gopath thing died, woo, updated the contribution guide to reflect
this, as well.


rdallman:
ty!

PR #1370, opened 2019-01-09
update generated proto to use correct library

rdallman:
the protobuf generated go files we had in tow were not generated by the
version of protobuf that we had vendored. if one did attempt to delete the
runner.pb.go file and generate from the proto file we had with the protobuf
version we had vendored the build would error with a compile time check for
IsProtobufVersion3 which did not exist in the protobuf version we had
vendored. If one changed the runner proto file to use proto version 2,
generation would error as our proto is in format for 3.  there is no tagged
release of golang/protobuf to vendor that contains a protobuf 3 generator,
this vendors tip until that happens.

this also updates the protobuf generated go file to use the version that we
now have in tow, which made some small changes. I have no clue how we had
protobufs that were generated with a proto version 3 set at the top by a
protobuf 2 generator to begin with, but I don't need another existential
question to come to grips with, really.

I did not test, but from the looks of it this shouldn't break anything as the
removed methods can just use the normal marshal path (I think they were just
optimizations?)


rdallman:
true. i have a feeling this was lost in the change from dep to mod, since mod marks every dep as versioned and uses minimal version, i'm less worried about the dependency changing (maybe something else is dependant on the protobuf package, i can't think of what). testing changes unfortunately requires deleting the file and then regenerating it, which seems odd to do but maybe is ok

PR #1372, opened 2019-01-10
Log latency error if the difference is >= 2ms

### LINKED ISSUE #1371: recordCallLatency in lb_agent log error check leads to false positive because it is too agressive
**Description**
In the lb_agent there is a check to verify that the call latency is greater than exec latency, if that is not true we log an error reporting that we have an invalid latency.
This check is too aggressive and can lead to false positive which will end up in logging errors which are not real errors.
We should make the check a bit more robust to verify that the difference between the two latency is above a min threshold, at the moment even a sub-millisecond difference will be logged as error.
That kind of difference can be related to at least a couple of reasons, first a rounding problem call latency uses 9 decimal digits, exec latency only 3. Plus the exec latency, if I get it right, comes from the runner the call latency is calculated in the load balancer, a small difference between the 2 clocks can lead to false positive.

In my implementation I hit the issue several times, below the results I got:

invalid latency callLatency=9.915378156s execLatency=9.916s diff = 0.000621844s
invalid latency callLatency=9.908625844s execLatency=9.909s diff = 0.000374156s
invalid latency callLatency=9.807090754s execLatency=9.808s diff = 0.000909246s
invalid latency callLatency=9.258049138s execLatency=9.259s diff = 0.000950862s
invalid latency callLatency=9.23659439s execLatency=9.237s  diff = 0.00040561ss

I expect that a sub-millisecond difference will be ignored, I think we can set a threshold equals to 2 ms.

reclaro:
The lb_agent logs an error if the difference between exec and call
latency is >=0, this change increase the value to reduce some potential
false positive.

resolve #1371

### DIFF
--- a/api/agent/lb_agent.go
+++ b/api/agent/lb_agent.go
@@ -364,7 +364,7 @@ func recordCallLatency(ctx context.Context, call *call, status string) {
 	// some sanity check before. If sanity checks flags something, then
 	// this is likely that runners are sending malicious/suspicious data.
 	if execLatency != nil {
-		if *execLatency >= callLatency {
+		if *execLatency-callLatency >= 2*time.Millisecond {
 			common.Logger(ctx).Errorf("invalid latency callLatency=%v execLatency=%v", callLatency, execLatency)
 			return
 		}


api/agent/lb_agent.go
```diff
@@ -364,7 +364,7 @@ func recordCallLatency(ctx context.Context, call *call, status string) {
 	// some sanity check before. If sanity checks flags something, then
 	// this is likely that runners are sending malicious/suspicious data.
 	if execLatency != nil {
-		if *execLatency >= callLatency {
+		if *execLatency-callLatency >= 2*time.Millisecond {
```
rdallman: a comment here about why 2ms might be nice, out of context this comparison doesn't make any sense. flooring to the nearest millisecond would perhaps read more clearly and not require a comment. go has decent facilities for this (https://golang.org/pkg/time/#Duration.Truncate) eg:

```go
if (*execLatency).Truncate(time.Millisecond) >= callLatency.Truncate(time.Millisecond)
```

skinowski: I think Truncate() strips monotonic clock, so I don't know if we want that here.

rdallman: it's a simple floor: https://golang.org/src/time/time.go?s=25865:25912#L801 -- we should do `>` and not `>=` though, good catch

skinowski:
Thanks for spotting the issue. I think the problem lies deeper and truncation/clock math is wrong in the original code. I suggest instead of trying to correct the result above, we should record the deltas in runner and lb properly and send deltas (based on monotonic clock) instead of wall-clock readings which are subject to clock skew/adjustments. In other words:

*) in gRPC proto for CallFinished, add the following two new fields:
int64 schedulerDuration
int64 executionDuration

*) in pure_runner, use Submit/BeforeCall/AfterCall functions to perform the math
to populate the above two durations. eg. start := time.Now(), delta := time.Now().Sub(start)

*) in runner_client.::recordFinishStats, directly use the new gRPC fields, instead of performing delta math on Created/Started/Completed.

*) in lb_agent:recordCallLatency, use the correct math to calculate the delta. StartedAt/CreatedAt are formatted strings (wall clocks), and math based on these will not be accurate. This means you'll have to record x:= time.Now() in Submit(), then do a time.Now().Sub(x) in recordCallLatency().

rdallman:
@reclaro closing since #1374 is merged - we can try this again if it isn't fixed by that, thanks!

Issue #1365, opened 2019-01-07
Adding currently hot function information (appId, functionId, image name) to prometheus


SamTS:
Currently `/metrics` only shows the count of current paused/started/waiting/etc functions. It would be useful if it also showed which specific functions were paused/started/waiting/etc, for purposes such as load balancing and or just getting better stats about exactly what functions are hot.

Sample output:
fn_container_paused_function{app_id="id1", function_id="id2", image_name="sampleImage:0.1.2"}
fn_container_paused_function{app_id="id3", function_id="id4", image_name="sampleImage2:0.2.3"}

rdallman:
we're using opencensus for this tagging, the docs for going about this are https://godoc.org/go.opencensus.io/tag#example-New and there are a lot of examples in our code if you search `tag.New` - since the container path and the request path split off, this should be easy to do near the top of checkLaunch or runHot in agent.go without polluting the request metrics at all, i'd expect app_id and fn_id to already be present so I'm surprised they aren't. we may want to just do the tagging of the context inside of UpdateState to minimize the number of metrics that get the image name tag, though I think we maybe should only have app_id and fn_id since image can be deduced from the function id (with the exception that it can change and we don't have versions, but this may be okay to start)

SamTS:
Made the changes [here](https://github.com/fnproject/fn/pull/1373). It's waiting for Oracle CLA.

PR #1373, opened 2019-01-10
Added change that exposes what functions are currently hot to metrics

SamTS:
* Added to UpdateState  so as a container changes status to starts and done it notes that a given app/function/image tuple is hot or not in open census
* Changed UpdateState to take in the entire call for increased flexibility

- Link to issue this resolves
[1365](https://github.com/fnproject/fn/issues/1365)

- What I did
* Added to UpdateState  so as a container changes status to starts and done it notes that a given app/function/image tuple is hot or not in open census
* Changed UpdateState to take in the entire call for increased flexibility

- How I did it
In `agent.go`  I switched UpdatedState to take in `call` instead of `call.slots` so it can be more powerful.
In `stats.go` I added a `hotFunctionMeasure` and tags for it
in `state_trackers.go` I added the functionality such that when a container starts it increases the hot function measure by one for a given appId/FunctionId/Image and when that container closes it decreases the hot function measure by one

- How to verify it
Call a function on FN, then, while during it's timeout period, call 
`curl localhost:8080/metrics  | grep -a5 -b10 Function` and you should see a line that looks like 
`fn_Hot_Functions{app_id="01CVRYP3EWNG8G00GZJ0000001",function_id="01D0JFCNH8NG8G00GZJ000000Z",image_name="imagename:0.0.1} 1` After the function times out (which is 30 seconds by defualt) and is not longer hot make that call again and you should see something like `fn_Hot_Functions{app_id="01CVRYP3EWNG8G00GZJ0000001",function_id="01D0JFCNH8NG8G00GZJ000000Z",image_name="imagename:0.0.1} 1`` 

Added change that exposes what functions are currently hot to metrics

### DIFF
--- a/api/agent/state_trackers.go
+++ b/api/agent/state_trackers.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"context"
+	"go.opencensus.io/tag"
 	"sync"
 	"time"
 
@@ -12,9 +13,11 @@ type RequestStateType int
 type ContainerStateType int
 
 type containerState struct {
-	lock  sync.Mutex
-	state ContainerStateType
-	start time.Time
+	lock                  sync.Mutex
+	state                 ContainerStateType
+	start                 time.Time
+	cfg                   *Config
+	containerStateMetrics bool
 }
 
 type requestState struct {
@@ -24,7 +27,7 @@ type requestState struct {
 }
 
 type ContainerState interface {
-	UpdateState(ctx context.Context, newState ContainerStateType, slots *slotQueue)
+	UpdateState(ctx context.Context, newState ContainerStateType, call *call)
 	GetState() string
 }
 type RequestState interface {
@@ -35,8 +38,10 @@ func NewRequestState() RequestState {
 	return &requestState{}
 }
 
-func NewContainerState() ContainerState {
-	return &containerState{}
+func NewContainerState(containerStateMetrics bool) ContainerState {
+	cs := &containerState{}
+	cs.containerStateMetrics = containerStateMetrics
+	return cs
 }
 
 const (
@@ -129,11 +134,19 @@ func (c *containerState) GetState() string {
 	return containerStateKeys[res]
 }
 
-func (c *containerState) UpdateState(ctx context.Context, newState ContainerStateType, slots *slotQueue) {
+func (c *containerState) UpdateState(ctx context.Context, newState ContainerStateType, call *call) {
+	var slots = call.slots
 
 	var now time.Time
 	var oldState ContainerStateType
 	var before time.Time
+
+	if c.containerStateMetrics {
+		ctx, _ = tag.New(ctx,
+			tag.Upsert(appIdKey, call.AppID),
+			tag.Upsert(functionIdKey, call.FnID),
+			tag.Upsert(imageNameKey, call.Image))
+	}
 
 	c.lock.Lock()
 
@@ -158,6 +171,7 @@ func (c *containerState) UpdateState(ctx context.Context, newState ContainerStat
 		return
 	}
 
+	//call.AppID, call.FnID, call.Image
 	// reflect this change to slot mgr if defined (AKA hot)
 	if slots != nil {
 		slots.enterContainerState(newState)
--- a/api/agent/stats.go
+++ b/api/agent/stats.go
@@ -22,6 +22,11 @@ var (
 	statusCallCacheKey    = common.MakeKey("cached")
 	statusCallSuccessKey  = common.MakeKey("success")
 	statusCallNetReadyKey = common.MakeKey("network")
+
+	appIdKey      = common.MakeKey("app_id")
+	functionIdKey = common.MakeKey("function_id")
+	imageNameKey  = common.MakeKey("image_name")
+	containerKeys = []tag.Key{appIdKey, functionIdKey, imageNameKey}
 )
 
 func statsCalls(ctx context.Context) {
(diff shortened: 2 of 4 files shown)


api/agent/state_trackers.go
```diff
@@ -129,7 +131,31 @@ func (c *containerState) GetState() string {
...
+func setHot(ctx context.Context, appId string, fnId string, imageName string, state ContainerStateType) {
+	if state != ContainerStateStart && state != ContainerStateDone {
+		//if the container isn't being made or closed then we don't want to change the note of if it's hot or not
+		return
+	}
+
+	ctx, _ = tag.New(ctx,
+		tag.Upsert(appIdKey, appId),
+		tag.Upsert(functionIdKey, fnId),
+		tag.Upsert(imageNameKey, imageName))
```
rdallman: might we do the tag insertion in `UpdateState`? this would apply it for all the container state metrics, which seems useful (i am curious why they seem missing to begin with, but alas). image name seems gratuitous, as fn id points to image name, however in case of function updates it may well be useful to see functions of same id with different images, to see if they hang around, so I think we should leave it (just bringing it up). 

it also seems like this metric is a dupe of a sum of `fn_container_start` + `fn_container_idle` + `fn_container_paused` + `fn_container_busy` - is this equation prohibitive or incorrect for your purposes? I agree that the proposal reads much better out of the box, I'm wondering if we can somehow surface this better using existing metrics, though?

SamTS: Moving the tag insertion to the beginning of UpdateState sounds great.

The problem I ran into with having it work like `fn_container_start` and `fn_container_pause` is that once a metric is added to opencensus I saw no way to remove it, only modify it. So, having start, idle, paused, and busy would require 4 lines, 3 of which would be meaningless at any given point. The other option would be to have `fn_container_status` and have the metric 0-3 for the lifecycle of a container. That isn't prefect either as I believe, and please correct me if I'm wrong, that multiple containers can exist for a given function under high loads. All that said I'm very flexible to changing it if I'm missing something or if you think adding more lines is worth it for clarity purposes.

skinowski:
Thanks for the PR. Is it possible to make the new metric or tags optional? The concern here is that additional 3 tags will increase the cardinality of the metric. In a federated prometheus environment this might end up with too much cardinality.

SamTS:
Sure, I can make the logging optional if performance is a large concern.

SamTS:
Made changes to the commit based on feedback.
Now uses env FN_FUNCTION_STATE_METRICS to turn on metrics
Metrics now are unique per container state, eg: `container_paused_function`

api/agent/state_trackers.go
```diff
@@ -86,6 +90,15 @@ var containerTimeKeys = [ContainerStateMax]string{
...
 }
 
+var functionStateMeasureKeys = [ContainerStateMax]string{
+	"",
+	"container_wait_function",
+	"container_start_function",
+	"container_idle_function",
+	"container_paused_function",
+	"container_busy_function",
+}
```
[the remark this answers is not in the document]
SamTS: That is correct, these are all of container states. These exist to mirror `containerTimeKeys` on line 84 ([here](https://github.com/fnproject/fn/blob/master/api/agent/state_trackers.go#L80)). The philosophy I tried to follow for naming the metrics was what I saw in the rest of the metrics. The two existing ones being `fn_container_busy_duration_seconds_bucket` and `fn_container_busy_total` which I didn't see filling the role of keeping track of the function information in the container.

api/agent/state_trackers.go
```diff
@@ -15,6 +16,7 @@ type containerState struct {
 	lock  sync.Mutex
 	state ContainerStateType
 	start time.Time
+	cfg   *Config
```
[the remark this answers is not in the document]
SamTS: I can modify this to only pass in the necessary configs.

api/agent/stats.go
```diff
@@ -310,11 +314,36 @@ func RegisterContainerViews(tagKeys []string, latencyDist []float64) {
...
+	for i, key := range containerGaugeKeys {
+		if key == "" {
+			continue
+		}
+		v := common.CreateView(containerGaugeMeasures[i], view.Sum(), tagKeys)
+
+		if err := view.Register(v); err != nil {
+			logrus.WithError(err).Fatal("cannot register view")
+		}
+	}
```
rdallman: this block is a duplicate of https://github.com/fnproject/fn/pull/1373/files#diff-4eaa01cd93e12bd2b4e3fb22e40fc312R316 - we can remove it. I can't recall the semantics here exactly it looks like we're registering 2 views of the container gauges with different tags, does opencensus merge these for us? can we simply join the tags together and create the view one time for behavior clarity? I'm also curious if this works when we make metrics that don't have all the tags (as we would if we don't enable the new tags), will the view have the tags but empty or are the tags not there (this would possibly create upstream prometheus issues)? if the tags only show up in prometheus when the metrics fill them in, then I think we can add these tags to all the container metrics in here, as well (since this will only happen when configured)?

SamTS: The short answer is that a registered view is only applicable for the metric with the tags registered for the view. I did some testing with this and if the view with tags is registered (and the view without tags isn't) then the metrics will only show up if all the tags are turned on. It's pretty annoying but this seemed like the best solution.

rdallman:
@SamTS I've checked this out and ran it with `FN_CONTAINER_STATE_METRICS=true ./fnserver` and in the metrics field I'm not seeing any tags on the metrics I was expecting, eg:

from `$ curl -sSL localhost:8080/metrics`:

```
# TYPE fn_container_paused_total untyped                                                          
fn_container_paused_total 1    
```

is my testing methodology off? any ideas here?

p.s. if you want to rebase and push up that would be great, thanks!

rdallman:
@SamTS bad news I think. when I start this with the option off, I see:

```
✗: curl -sSL localhost:8080/metrics | grep "paused"
# HELP fn_container_paused_total containers in state container_paused_total
# TYPE fn_container_paused_total untyped
fn_container_paused_total{app_id="",function_id="",image_name=""} 1.0
```

though, I'm glad to see updating opencensus fixed this, I think we need to move the configuration setting to be around the views not registering certain tags rather than the metrics? that is, we can throw metrics with any tags at the views, but unless the view registers those tags then they won't appear in prometheus. at present, we're registering all the tags but not offering them from the metrics, meaning the tags will be the zero value (not what we want). is this plausible?

rdallman:
@SamTS I did not get to this today, sorry, I will fiddle with how to configure the tags tomorrow and post whatever the results are here... async brain processing hasn't given me too many ideas since Monday :(

SamTS:
sounds good! let me know how it goes.

rdallman:
@SamTS this is what I'm thinking https://github.com/fnproject/fn/compare/SamTS-hotmetrics?expand=1 -- this allows anybody building fn to build with whatever tags they want, and on the backend we can wire in whatever tags we want, just letting the main file add which tags they want to expose -- without additional configuration for various tags across the backend. I'd like to cohesive plan for advertising and bunching what those tags are eventually for people to use, but this at least is a start and is in line with what I'm thinking - I think this is the way the metrics/views are supposed to be split really: add whatever tags you want to metrics, turn on with views. 

let me know what you make of that, it will meet requirements by just having them on in OSS by default, but won't bug our version of fn where we don't want to turn these tags on [yet]. I can PR this with your commit here or to your fork, let me know which you'd prefer. we do need to get the CLA signed to merge any commits though, let me know if you have signed it and it's just our bot acting up (it happens) and we'll get that taken care of.

SamTS:
@rdallman The one bit I am concerned about is under the current construction 
`fn_container_{container_status}_duration_seconds_bucket`  populates by the cross product of latency and each app/fn/image tuple. If you're good with it, I'm fine, but it's something I noticed that could make the information in that section more unclear.

rdallman:
@SamTS yea, I think they will be useful for the latency buckets as well - if i'm understanding correctly, I see them there. it's relatively easy to filter out tags in grafana to get eg the duration across all functions, I forget how prometheus handles the tags in queries (ie if the query is expensive or not to ignore tags), I guess that brings up a good point that we may want the metrics raw separately from the fully tagged versions, just to quickly query the former, if that's an issue (honestly idk, I guess I should figure that out). 

cool, i can PR this and tag another for review if that works for you? any luck with CLA bot https://github.com/fnproject/fn/pull/1373#issuecomment-453208872 ? let me know if I can help, would be nice to close this out, sorry it's dragged on so long

SamTS:
Yeah that works great, CLA should be in progress, sent off the form so however long that takes.

delabassee:
@rdallman FYI @SamTS 's OCA is now approved!

rdallman:
@SamTS sorry for delay again here, I've opened https://github.com/fnproject/fn/pull/1438 to close this out I hope

rdallman:
closed via #1438 - thanks!

PR #1374, opened 2019-01-11
fn: move timestamps to monotonic clocks

skinowski:
In Model, CreatedAt/StartedAt/CompletedAt should
be used for documenting/logging current time.
For measuring time durations, math that uses monotonic
clock should be used.
See: https://golang.org/pkg/time/#hdr-Monotonic_Clocks

Based on this, added monotonic timestamps in call
for startTime, createTime and completeTime to be used in
measurement calculations.

RunnerStatus and CallFinished gRPC messages are extended
with two new duration fields. These latencies are
calculated on the same server using monotonic
clocks for transmission over the network.

Fixes: #1371


api/agent/call.go
```diff
@@ -284,6 +286,11 @@ type call struct {
 
 	// LB & Pure Runner Extra Config
 	extensions map[string]string
+
+	// time.Time equivalents of model Call CreatedAt/StartedAt/CompletedAt
+	createTime   time.Time
+	startTime    time.Time
+	completeTime time.Time
```
rdallman: CreatedAt, et al, are all common.DateTime which is an alias for time.Time, the only difference being the marshal/string methods to use our preferred format, this should not strip monotonic time which would make these unnecessary, no? the main problem currently being we have to set eg CreatedAt on a runner and not on an lb?

api/agent/call.go
```diff
@@ -247,6 +247,8 @@ func (a *agent) GetCall(opts ...CallOpt) (Call, error) {
 
 	c.handler = a.da
 	c.ct = a
+	c.createTime = time.Now()
```
rdallman: this is odd as it doesn't match CreatedAt ?

PR #1378, opened 2019-01-16
Prevent in-built docker VOLUME commands

kppk:
- What I did
Added validation to check function image for VOLUME definitions, reject to run if VOLUME is present.

- How to verify it
Try to run an image with VOLUME defined.

Function image can't contain VOLUME definition

### DIFF
--- a/api/agent/drivers/docker/cookie.go
+++ b/api/agent/drivers/docker/cookie.go
@@ -3,6 +3,7 @@ package docker
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"path"
@@ -16,6 +17,8 @@ import (
 	"github.com/sirupsen/logrus"
 	"go.opencensus.io/trace"
 )
+
+var ErrImageWithVolume = models.NewAPIError(http.StatusBadRequest, errors.New("image has Volume definition"))
 
 // A cookie identifies a unique request to run a task.
 type cookie struct {
@@ -368,6 +371,11 @@ func (c *cookie) ValidateImage(ctx context.Context) (bool, error) {
 		return false, err
 	}
 
+	// check image doesn't have Volumes
+	if !c.drv.conf.ImageEnableVolume && img.Config != nil && len(img.Config.Volumes) > 0 {
+		err = ErrImageWithVolume
+	}
+
 	c.image = &CachedImage{
 		ID:       img.ID,
 		ParentID: img.Parent,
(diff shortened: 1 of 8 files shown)


api/agent/drivers/docker/cookie.go
```diff
@@ -298,19 +301,31 @@ func (c *cookie) ValidateImage(ctx context.Context) (bool, error) {
...
+		if err == nil {
+			c.cachedImg = &CachedImage{
+				ID:       img.ID,
+				ParentID: img.Parent,
+				RepoTags: img.RepoTags,
+				Size:     uint64(img.Size),
+			}
+			c.drv.imgCache.MarkBusy(c.cachedImg)
+		}
+		if err == ErrImageWithVolume {
```
rdallman: assuming we delete images from disk that we pull only to find volumes, shouldn't we just never add this to the image cache? the previous error condition seems fine in this respect, but perhaps i'm missing the ordering of pulling and adding to image cache

api/agent/drivers/docker/cookie.go
```diff
@@ -346,14 +361,22 @@ func (c *cookie) PullImage(ctx context.Context) error {
...
+
 	if c.drv.imgCache != nil && c.cachedImg == nil {
-		img, err := c.drv.docker.InspectImage(ctx, c.task.Image())
-		// Log and ignore these errors. Hard failure will be caught in CreateContainer()
 		if err == docker.ErrNoSuchImage {
-			log.WithError(err).Errorf("InspectImage image likely removed due to disk pressure %s", c.task.Image())
-		} else if err != nil {
-			log.WithError(err).Errorf("Failed to inspect image %s", c.task.Image())
-		} else {
+			return models.ErrCallTimeoutServerBusy
```
rdallman: this makes err no such image a call timeout now? this error seems misleading/incorrect in this case, previously it would result in a 503 internal server error (which is true, because it means we pulled an image and then it was gone - that's on us)

rdallman: well, says 'server busy' - i guess this is saying that we're low on resources, which is the case we use for ram and cpu being at capacity, which makes sense even if slightly nuanced... think you can ignore me and my latent understanding

api/agent/drivers/docker/cookie.go
skinowski: Volume checker here leaves the cleanup to image cacher. If a volume is in the image, then it is inserted into LRU via Update() and if the need to cleanup ever arises then we assume LRU would kick in and clean this up. (Note that LRU Update() is not perfect here as this code path might keep this image at the head of the cache. I might provide a special image cacher function instead of Update() in the near future as an improvement.)

rdallman:
@kppk merge whenever you're ready in this repo, updates LGTM

PR #1380, opened 2019-01-17
fn: driver cookie interface changes for finer control

skinowski:
Users of driver cookie, as seen in agent.go require specific
timeouts in contexts passed to the different operations. For
example, cookie.PullImage() context shows intent to apply
timeout/duration for the pull operation. To clarify this
and avoid tricky timeout related issues, removing implicit
docker-inspect in cookie.PullImage(). In addition, separating
the Authentication step as Auther interface calls unknown
code.

Additional changes to cookie implementation to clarify
inspected image, created container and authentication
configuration, where each AuthImage(), ValidateImage()
and CreateContainer() steps initialize the corresponding
pointer object in the cookie.

Changes now require ValidateImage() call if a PullImage()
is attempted. In all cases, AuthImage() call is required
before ValidateImage().

An example usage is as follows:

        cookie, err := driver.CreateCookie(ctx, task)
        err = cookie.AuthImage(ctx)
        pull, err := cookie.ValidateImage()
        if pull {
                err = cookie.PullImage(ctx)
                pull, err = cookie.ValidateImage()
        }

        err = cookie.CreateContainer(ctx)
        waiter := cookie.Run(ctx)
        waiter.Wait(ctx)

error handling is omitted for clarity in the above example.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -867,21 +867,32 @@ func (a *agent) runHot(ctx context.Context, caller slotCaller, call *call, tok R
 		return
 	}
 
-	needsPull, err := cookie.ValidateImage(ctx)
+	err = cookie.AuthImage(ctx)
 	if tryQueueErr(err, errQueue) != nil {
 		return
 	}
 
+	needsPull, err := cookie.ValidateImage(ctx)
 	if needsPull {
-		ctx, cancel := context.WithTimeout(ctx, a.cfg.HotPullTimeout)
-		err = cookie.PullImage(ctx)
-		cancel()
-		if ctx.Err() == context.DeadlineExceeded {
-			err = models.ErrDockerPullTimeout
+		// pull image ctx scope with HotPullTimeout
+		{
+			ctx, cancel := context.WithTimeout(ctx, a.cfg.HotPullTimeout)
+			err = cookie.PullImage(ctx)
+			cancel()
+			if err != nil && ctx.Err() == context.DeadlineExceeded {
+				err = models.ErrDockerPullTimeout
+			}
 		}
-		if tryQueueErr(err, errQueue) != nil {
-			return
+		if err == nil {
+			needsPull, err = cookie.ValidateImage(ctx)
+			if needsPull && err == nil {
+				// Image must have removed by image cleaner, manual intervention, etc.
+				err = models.ErrCallTimeoutServerBusy
+			}
 		}
+	}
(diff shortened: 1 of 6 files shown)


api/agent/agent.go
```diff
@@ -867,22 +867,33 @@ func (a *agent) runHot(ctx context.Context, caller slotCaller, call *call, tok R
...
 
+	needsPull, err := cookie.ValidateImage(ctx)
 	if needsPull {
-		ctx, cancel := context.WithTimeout(ctx, a.cfg.HotPullTimeout)
-		err = cookie.PullImage(ctx)
-		cancel()
-		if ctx.Err() == context.DeadlineExceeded {
-			err = models.ErrDockerPullTimeout
+		// pull image ctx scope with HotPullTimeout
+		{
```
kppk: do we need the extra brackets?

api/agent/agent.go
```diff
@@ -867,22 +867,33 @@ func (a *agent) runHot(ctx context.Context, caller slotCaller, call *call, tok R
...
+			ctx, cancel := context.WithTimeout(ctx, a.cfg.HotPullTimeout)
+			err = cookie.PullImage(ctx)
+			cancel()
+			if err != nil && ctx.Err() == context.DeadlineExceeded {
+				err = models.ErrDockerPullTimeout
+			}
 		}
-		if tryQueueErr(err, errQueue) != nil {
-			return
+		if err == nil {
```
kppk: Why don't we use the same check as in any other place:
```
if tryQueueErr(err, errQueue) != nil {
 	return
 }
```

api/agent/agent.go
```diff
@@ -867,22 +867,33 @@ func (a *agent) runHot(ctx context.Context, caller slotCaller, call *call, tok R
...
+			needsPull, err = cookie.ValidateImage(ctx)
+			if needsPull && err == nil {
```
kppk: Should we use `if needsPull` only here?

api/agent/agent.go
skinowski: yes, it's for the new 'ctx' in that scope with specific timeout for the pull.

skinowski: subsequent ValidateImage() uses outer scope ctx while PullImage() uses inner ctx with hot pull timeout.

kppk: right, which is hard to read, I would use something like `pullCtx` to make this behavior explicit, but it is a super nit... 😈

PR #1381, opened 2019-01-17
Rework call placement/finish logs to take advantage of user blame tag

gviedma:
If a customer function fails due to an issue with the function itself (e.g. customer-side timeout or usage of an old FDK), the error is logged at the error level. We should log at warning since it is not a platform error and it only affects the function in question.


riconnon:
Reworked. The new logic removes the "Failed during call placement" log line and makes the "Call finished" line conditional on whether there was an error and the blame

PR #1382, opened 2019-01-18
fn: startup container cleanup timeout

skinowski:
ListContainers call which is asynchronously spawned
during docker driver start needs to have a reasonable
timeout and should be retried if timeout expires.


rdallman:
there are 2 layers of retries going on here fwiw, since we want to retry any error for this one it's a little different than just temporary errors like we do in the docker client, anyway it sticks out as something to think about but I think this is absolutely fine, can't think of any issues or way to merge the two cleanly really

PR #1383, opened 2019-01-19
fn: track network availability for runner status checks

skinowski:
Runner Status Call is extended to support the network readiness state. This is configured via runner option that provide the runner a file path to check for existence.


api/agent/drivers/docker/docker_client.go
```diff
@@ -120,32 +122,25 @@ func RecordImageCleanerStats(ctx context.Context, sample *ImageCacherStats) {
 }
 
 // listenEventLoop listens for docker events and reconnects if necessary
-func listenEventLoop(ctx context.Context, client *docker.Client) {
-	for ctx.Err() == nil {
-		err := listenEvents(ctx, client)
+func (d *dockerWrap) listenEventLoop(ctx context.Context) {
+	limiter := rate.NewLimiter(2.0, 1)
```
kmjohansen: This is pedantic, I know, but IMO it would be preferable to define constants for these limiters instead of hard-coding them to a particular value.

api/agent/drivers/driver.go
```diff
@@ -178,6 +181,9 @@ type ContainerTask interface {
 	// UDSDockerDest is the destination mount point for uds path. it is the path
 	// of the directory where the sock file resides inside of the container.
 	UDSDockerDest() string
+
+	// Returns true if network is disabled.
+	DisableNet() bool
```
kmjohansen: Change to `IsNetworkDisabled()` to be congruent with `IsNetworkReady()`, perhaps?

skinowski: So, here, DisableNet() is orthogonal to driver.IsNetworkReady(). The agent can set this to false regardless of driver.IsNetworkReady(). Currently, agent only sets DisableNet() for status checks, if driver.IsNetworkReady() is false.

rdallman:
maybe out of band but why is the agent starting before the networks are created? is agent startup so slow that we have to do it concurrently (and, uh, how? thing can start super fast, worried about funny biz, if any)? it would save us quite a bit of code, anyway

api/agent/drivers/docker/docker_networks.go
```diff
@@ -0,0 +1,194 @@
...
+
+func (n *DockerNetworks) isDockerNetworkReady() bool {
+	if len(n.waitList) > 0 {
+		select {
+		case <-n.waitListReady:
+		default:
+			return false
+		}
+	}
+	return true
```
rdallman: hmm, i figured part of the use of this patch was to wait for multiple networks. not saying to do that, just seems odd to need all this machinery in that case. maybe the wrong place to thread, but I find:

> In addition, if driver.IsNetworkReady() is false, then status call is executed using network disabled. This ensures that Status call runs successfully while waiting for docker networks.

this seems questionable. we want status calls to run successfully for some other info presumably (no reason listed, $jumping_to_conclusions)? but aside from seeming odd it seems like a pretty good stand in for exactly what this patch does in a simpler way, that is, if the docker networks aren't ready then status calls will fail, right? at least intuitively, this seems desirable. is there information from a status call that the lb or another component would use to ever route traffic to this node before the networks are ready? is there another reason (to satisfy some health checker condition that would potentially take out a node that is in fact starting up? something else?)?

rdallman: additionally quoted condition seems troublesome for reasoning about. i'm sure things are too far gone as it is (for easily reasoning about... health), but it is a bit hairy to go flipping 'oh the status call succeeds when networking is off, then gets flipped on and the networking starts failing so the status call starts failing' -- seems odd to change the status call in place and let it succeed in the first place, makes sense to just fail i guess. one more thing i guess

skinowski: I agree, we can simply let the status fail. This PR lets status succeed to exercise more code paths in docker and agent before the networks are available, but at the cost of this additional complexity.

api/agent/grpc/runner.pb.go
```diff
@@ -594,6 +594,7 @@ type RunnerStatus struct {
 	KdumpsOnDisk         uint64   `protobuf:"varint,14,opt,name=kdumpsOnDisk,proto3" json:"kdumpsOnDisk,omitempty"`
 	SchedulerDuration    int64    `protobuf:"varint,15,opt,name=schedulerDuration,proto3" json:"schedulerDuration,omitempty"`
 	ExecutionDuration    int64    `protobuf:"varint,16,opt,name=executionDuration,proto3" json:"executionDuration,omitempty"`
+	IsNetworkDisabled    bool     `protobuf:"varint,17,opt,name=isNetworkDisabled,proto3" json:"isNetworkDisabled,omitempty"`
```
rdallman: i see this becoming something like `IsUserReady` or or `IsSpecialized` or something eventually, but protos aren't that hard to migrate so we can probably punt

PR #1384, opened 2019-01-21
fn: retriable error handling in runner client

skinowski:
During gRPC communication, previously we assumed that
Send(Try) message failure in TryExec(), can always
be retried. However, this is not robust as we cannot assume
no data was written to wire. With this change, before we
can conclude that the call can be retried, we also check
Unavailable error code.

### DIFF
--- a/api/agent/runner_client.go
+++ b/api/agent/runner_client.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"google.golang.org/grpc"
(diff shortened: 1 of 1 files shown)


skinowski:
Hmm, I'll dig around the grpc code to verify their claim, but in the spec the litmus test seems strict. (See the description in FailedPrecondition section in https://godoc.org/google.golang.org/grpc/codes#Code)

PR #1391, opened 2019-01-29
fn: docker-pull errors are user errors

skinowski:
Classify docker pull errors as user errors and adjust log level
accordingly.

### DIFF
--- a/api/agent/drivers/docker/cookie.go
+++ b/api/agent/drivers/docker/cookie.go
@@ -356,7 +356,7 @@ func (c *cookie) PullImage(ctx context.Context) error {
 
 	err := c.drv.docker.PullImage(docker.PullImageOptions{Repository: repo, Tag: c.imgTag, Context: ctx}, *cfg)
 	if err != nil {
-		log.WithError(err).Error("Failed to pull image")
+		log.WithError(err).Info("Failed to pull image")
 
 		// TODO need to inspect for hub or network errors and pick; for now, assume
 		// 500 if not a docker error
@@ -369,7 +369,8 @@ func (c *cookie) PullImage(ctx context.Context) error {
 			}
 		}
 
-		return models.NewAPIError(code, fmt.Errorf("Failed to pull image '%s': %s", c.task.Image(), msg))
+		err := models.NewAPIError(code, fmt.Errorf("Failed to pull image '%s': %s", c.task.Image(), msg))
+		return models.NewFuncError(err)
 	}
 	return nil
 }


api/agent/drivers/docker/cookie.go
```diff
@@ -369,7 +368,16 @@ func (c *cookie) PullImage(ctx context.Context) error {
 			}
 		}
 
-		return models.NewAPIError(code, fmt.Errorf("Failed to pull image '%s': %s", c.task.Image(), msg))
+		result := models.NewAPIError(code, fmt.Errorf("Failed to pull image '%s': %s", c.task.Image(), msg))
+
+		// repository does not exist or no read access (see: https://docs.docker.com/engine/api/v1.39/#operation/ImageCreate)
+		if code == http.StatusNotFound {
+			log.WithError(err).Debug("Failed to pull image")
```
kmjohansen: Can we mark this one as an Info instead of a Debug?

api/agent/drivers/docker/cookie.go
```diff
@@ -369,7 +368,16 @@ func (c *cookie) PullImage(ctx context.Context) error {
...
+			return models.NewFuncError(result)
+		}
+
+		log.WithError(err).Error("Failed to pull image")
```
kmjohansen: Should this one be a Warn instead of an Error?

api/agent/drivers/docker/cookie.go
```diff
@@ -369,7 +368,16 @@ func (c *cookie) PullImage(ctx context.Context) error {
 			}
 		}
 
-		return models.NewAPIError(code, fmt.Errorf("Failed to pull image '%s': %s", c.task.Image(), msg))
+		result := models.NewAPIError(code, fmt.Errorf("Failed to pull image '%s': %s", c.task.Image(), msg))
+
+		// repository does not exist or no read access (see: https://docs.docker.com/engine/api/v1.39/#operation/ImageCreate)
+		if code == http.StatusNotFound {
+			log.WithError(err).Debug("Failed to pull image")
+			return models.NewFuncError(result)
```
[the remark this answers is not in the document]
skinowski: We have 500, 404 and 200 error codes possible here. There are 500 cases which can also be user errors, but this PR conservatively marks the known 404 classes only.

api/agent/drivers/docker/cookie.go
```diff
@@ -356,7 +356,7 @@ func (c *cookie) PullImage(ctx context.Context) error {
 
 	err := c.drv.docker.PullImage(docker.PullImageOptions{Repository: repo, Tag: c.imgTag, Context: ctx}, *cfg)
 	if err != nil {
-		log.WithError(err).Error("Failed to pull image")
+		log.WithError(err).Info("Failed to pull image")
```
riconnon:
```suggestion
		log.WithError(err).Debug("Failed to pull image")
```
I'm not sure this even wants to be an Info level.

riconnon: Actually I see this was already discussed.

PR #1396, opened 2019-02-06
plumb docker auther to the depths

rdallman:
this patch accomplishes a few things wrt the docker auth interface used for
configuring credentials to be used for pull:

* adds context for timing out the docker auth call
* no longer call docker auth before validating an image, we were using this to
validate that a user still has access to an image that we have locally. this
policy itself could be TBD but we're being flagged for hitting registries too
hard, and runners have bounded lifetimes on the order of hours, so we decided
this is ok. now we only call docker auth function when pulling an image, and
it's time used is included in the pull timeout we've so delicately plumbed
* adds ability for agents to add call configurations to every call created via
getcall, this ought to help allow fn extenders configure calls and puts us on
the path of getting rid of the call overrider (and is enough to be able to
replace our current one that is causing issues)
* adds an option to configure calls to have a docker auth function they
provide, so that fn extenders can plumb this thing into here without having to
fork the driver (which would be absurd!)
* some lints...

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -1321,10 +1321,13 @@ func (c *container) WriteStat(ctx context.Context, stat drivers.Stat) {
 	c.swapMu.Unlock()
 }
 
+// assert we implement this at compile time
+var _ dockerdriver.Auther = new(container)
+
 // DockerAuth implements the docker.AuthConfiguration interface.
-func (c *container) DockerAuth(ctx context.Context) (*docker.AuthConfiguration, error) {
+func (c *container) DockerAuth(ctx context.Context, image string) (*docker.AuthConfiguration, error) {
 	if c.dockerAuth != nil {
-		return c.dockerAuth.DockerAuth(ctx)
+		return c.dockerAuth.DockerAuth(ctx, image)
 	}
 
 	// TODO(reed): kill this after using that
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -15,6 +15,7 @@ import (
 	"path/filepath"
 
 	"github.com/fnproject/fn/api/agent/drivers"
+	dockerdriver "github.com/fnproject/fn/api/agent/drivers/docker"
 	"github.com/fnproject/fn/api/common"
 	"github.com/fnproject/fn/api/id"
 	"github.com/fnproject/fn/api/models"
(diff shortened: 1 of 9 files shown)


rdallman:
tested the real deal... i'll fix up the mockers (oops) for ci

api/agent/drivers/docker/docker.go
```diff
@@ -39,7 +38,7 @@ type Auther interface {
 	// certain restrictions on images or if credentials must be acquired right
 	// before runtime and there's an error doing so. If these credentials don't
 	// work, the docker pull will fail and the task will be set to error status.
-	DockerAuth() (*docker.AuthConfiguration, error)
+	DockerAuth(ctx context.Context) (*docker.AuthConfiguration, error)
```
reclaro: This is a breaking change as it is changing an existing Interface am I right? The right process would be to create a new interface and have a kind of deprecation path? I know it is a pain I am just interested in your opinion and understand if we have a process to deal with these kind of changes.

api/agent/drivers/driver.go
```diff
@@ -41,9 +41,6 @@ type Cookie interface {
 	// Unfreeze a frozen container to unpause frozen processes
 	Unfreeze(ctx context.Context) error
 
-	// Authenticate image. Returns non-nil error if auth fails.
-	AuthImage(ctx context.Context) error
-
```
reclaro: same here

api/agent/drivers/docker/docker.go
rdallman: i was not able to find any third party use of this package and all other required changes lie here. thanks for catching this - I should've made a note.

api/agent/drivers/driver.go
rdallman: same comment as the other

rdallman:
actually need to pass in the image name to `DockerAuth` as well - I'm going to add this here, too

rdallman:
need to wire into lb agent too.. 1 sec

api/agent/agent.go
```diff
@@ -109,7 +110,12 @@ type agent struct {
 	shutWg   *common.WaitGroup
 	shutonce sync.Once
 
+	// TODO(reed): shoot this fucking thing
```
riconnon: Yes

api/agent/call.go
```diff
@@ -39,13 +41,15 @@ type Call interface {
 	End(ctx context.Context, err error) error
 }
 
-// Interceptor in GetCall
+// CallOverrider should die. Interceptor in GetCall
```
riconnon: Yes

api/agent/agent.go
skinowski: let's not add profanity to the code base.

PR #1397, opened 2019-02-07
WIP:  POC to drop caps and set user

zootalures:
This is what I did when I tested all the FDK's hello worlds  obviously needs more config etc.


rdallman:
#1442

PR #1400, opened 2019-02-11
Allow configuration of invoke URL pattern via environment variables

riconnon:
- Link to issue this resolves

- What I did

This allows for an Fn server to customize the path of the invoke URLs it handles

- How I did it

There are two new environment variables for configuration:
FN_INVOKE_GROUP represents the static prefix of the invoke endpoints
FN_INVOKE_TEMPLATE represents the URL path and should include `:fn_id` to represent the function ID

- How to verify it

Run an Fn server with `FN_INVOKE_GROUP=/runthefunction` and `FN_INVOKE_TEMPLATE=/:fn_id/prettyplease`
Deploy a function.
Run `curl http://127.0.0.1:8080/runthefunction/<function_id>/prettyplease`

Allow configuration of invoke URL pattern via environment variables


rdallman:
hmm. we have methods to add endpoints to the server itself and we have handlers (the invoke handler which uses the same var), so while this solves the problem I'm not sure we need the additional configurations and code and this could be a one liner in the configuration of fn?

PR #1401, opened 2019-02-13
fn: clean docker retry logic and legacy code

skinowski:
Cleanup of docker retry error handling logic. The code
is likely due to older less stable docker versions. Instead,
we would like to expose any underlying issues and fix them
as necessary.

Examination of logs/events from last 3-4 weeks do not
show any occurrence of retries/errors and on the contrary
causing more issues (eg. retry of docker-pull if 5xx)

### DIFF
--- a/api/agent/drivers/docker/docker.go
+++ b/api/agent/drivers/docker/docker.go
@@ -94,7 +94,7 @@ func NewDocker(conf drivers.Config) *DockerDriver {
 	driver := &DockerDriver{
 		cancel:     cancel,
 		conf:       conf,
-		docker:     newClient(ctx, conf.MaxRetries),
+		docker:     newClient(ctx),
 		hostname:   hostname,
 		auths:      auths,
 		network:    NewDockerNetworks(conf),
@@ -412,20 +412,11 @@ func (drv *DockerDriver) CreateCookie(ctx context.Context, task drivers.Containe
 	return cookie, nil
 }
 
-func (drv *DockerDriver) removeContainer(ctx context.Context, container string) error {
-	err := drv.docker.RemoveContainer(docker.RemoveContainerOptions{
-		ID: container, Force: true, RemoveVolumes: true, Context: ctx})
-
-	if err != nil {
-		logrus.WithError(err).WithFields(logrus.Fields{"container": container}).Error("error removing container")
-	}
-	return nil
-}
-
 // Run executes the docker container. If task runs, drivers.RunResult will be returned. If something fails outside the task (ie: Docker), it will return error.
 // The docker driver will attempt to cast the task to a Auther. If that succeeds, private image support is available. See the Auther interface for how to implement this.
 func (drv *DockerDriver) run(ctx context.Context, container string, task drivers.ContainerTask) (drivers.WaitResult, error) {
 
+	log := common.Logger(ctx)
 	stdout, stderr := task.Logger()
 	successChan := make(chan struct{})
 
@@ -471,6 +462,7 @@ func (drv *DockerDriver) run(ctx context.Context, container string, task drivers
 
 	if err != nil && ctx.Err() == nil {
 		// ignore if ctx has errored, rewrite status lay below
+		log.WithError(err).WithFields(logrus.Fields{"container": container, "call_id": task.Id()}).Error("error attaching to container")
 		return nil, err
 	}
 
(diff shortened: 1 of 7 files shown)


kmjohansen:
Thanks for making these changes.  I'm inclined to take them, since I think it lets us remove a number of cases that were causing us to send a lot of traffic to registry servers when we encountered an error.  I realize that this may not be entirely appealing to @rdallman, however it appears that docker has become a bit more reliable.  We spent a bunch of time digging through our own logs and only saw 500s and timeouts in response to other upstream errors.  IOW, not from dockerd and not from problems on the host.  This does mean that we have the potential to re-introduce problems that were fixed by the retry logic.  That said, I think @skinowski and I would volunteer to chase these down individually as they re-appear.

skinowski:
Risk here is we might need to reintroduce some of this code back. But I think it's good to start from blank slate especially if we are agreeing to chase issues down operationally.

rdallman:
i found the opposite approach to be more appealing, specifically since it ends up in user errors, for when errors that shouldn't be retried come up to whitelist them to not - cognitively, the exact same burden commitment. wholesale getting rid of this seems not ideal and making it more conservative seems reasonable, but hey, what do i know, i only dealt with customers after running billions of containers with this code... we're only in LA, we're playing a probability game with docker errors and when customers actually go prod the probabilities will show... y'all have fun

rdallman:
we'll work on 'we' usage meanwhile

kmjohansen:
@rdallman The probabilities is precisely why it makes sense to track down the errors that necessitated retries in the past.  As a general rule, adding retries to the code just causes it to run slower and paper over existing problems.  If engineers can find the cases that shouldn't be occurring, and get them fixed then both our product and the upstream container technology become more reliable.  If instead the code continues to simply retry, then there's no forward progress on making docker or the product more stable.  If in this process, analysis turns up a case where a retry is warranted then you can have the satisfaction of saying that you told me so.

api/agent/drivers/docker/docker_client.go
```diff
@@ -366,187 +287,113 @@ func (d *dockerWrap) LoadImages(ctx context.Context, filePath string) error {
...
 }
 
 func (d *dockerWrap) StartContainerWithContext(id string, hostConfig *docker.HostConfig, ctx context.Context) (err error) {
 	ctx, closer := makeTracker(ctx, "docker_start_container")
-	defer closer()
-
-	ctx, _ = common.LoggerWithFields(ctx, logrus.Fields{"docker_cmd": "StartContainer"})
-	err = d.retry(ctx, func() error {
-		err = d.docker.StartContainerWithContext(id, hostConfig, ctx)
-		if _, ok := err.(*docker.NoSuchContainer); ok {
```
skinowski: seems like old docker bug: https://github.com/moby/moby/issues/1319

rdallman:
2 years of logs man, anything in the code that has an exception is from that. it's not about you and me buddy. i know i'm not up to your human standards, and i'm deeply sorry that you have to interact with me.

api/agent/drivers/docker/docker_client.go
```diff
@@ -246,117 +249,35 @@ func RegisterViews(tagKeys []string, latencyDist []float64) {
...
-// some 500s are totally cool
-func filter(ctx context.Context, err error) error {
-	log := common.Logger(ctx)
-	// "API error (500): {\"message\":\"service endpoint with name task-57d722ecdecb9e7be16aff17 already exists\"}\n" -> ok since container exists
-	switch {
-	default:
-		return err
-	case err == nil:
-		return err
-	case strings.Contains(err.Error(), "service endpoint with name"):
```
skinowski: This error string is no longer present in moby project.

api/agent/drivers/docker/docker_client.go
```diff
@@ -246,117 +249,35 @@ func RegisterViews(tagKeys []string, latencyDist []float64) {
...
-	}
-	log.WithError(err).Warn("filtering error")
-	return nil
-}
-
-func filterNoSuchContainer(ctx context.Context, err error) error {
```
skinowski: Dead code. Caller of WaitContainer ignore any errors anyway.

api/agent/drivers/docker/docker_client.go
```diff
@@ -246,117 +249,35 @@ func RegisterViews(tagKeys []string, latencyDist []float64) {
...
-	return err // TODO could return context.DeadlineExceeded which ~makes sense
-}
-
-func isDocker50x(err error) bool {
-	derr, ok := err.(*docker.Error)
-	return ok && derr.Status >= 500
-}
-
-// implement common.Temporary()
-type temporary struct {
```
skinowski: Dead code, only used by StartContainer if check based on old Docker bug.

api/agent/drivers/docker/docker.go
```diff
@@ -615,27 +608,13 @@ func cherryPick(ds *docker.Stats) drivers.Stat {
 	}
 }
 
-func (drv *DockerDriver) startTask(ctx context.Context, container string) error {
-	log := common.Logger(ctx)
-	log.WithFields(logrus.Fields{"container": container}).Debug("Starting container execution")
-	err := drv.docker.StartContainerWithContext(container, nil, ctx)
-	if err != nil {
-		dockerErr, ok := err.(*docker.Error)
```
skinowski: Extra complication due to retries, based on https://github.com/iron-io/runner/commit/22a867ac0a575880d15bd1ea3768decabe417652

skinowski:
As reference, although the workloads are different, k8s or docker-client do not have retries.

PR #1402, opened 2019-02-13
Handle syslog error

kppk:
syslog error is treated as an user error, it is still returned as 500, but with "Syslog Unavailable" message.

### DIFF
--- a/api/agent/drivers/docker/docker_test.go
+++ b/api/agent/drivers/docker/docker_test.go
@@ -21,6 +21,7 @@ type taskDockerTest struct {
 	input      io.Reader
 	output     io.Writer
 	errors     io.Writer
+	logURL     string
 }
 
 func (f *taskDockerTest) Command() string                         { return f.cmd }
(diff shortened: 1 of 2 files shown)


api/agent/drivers/docker/docker_test.go
```diff
@@ -537,3 +540,33 @@ func TestRunnerDockerNoNetwork(t *testing.T) {
...
+
+func TestRunnerDockerInvalidSyslog(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(30)*time.Second)
+	defer cancel()
+
+	dkr := NewDocker(drivers.Config{})
+	defer dkr.Close()
+
+	task := createTask("test-docker-no-net")
+	task.logURL = "tcp://invalid:9999"
```
skinowski: I think we can use: 192.0.2.0 (https://tools.ietf.org/html/rfc5737 TEST-NET1) for this.. Minor nit...

api/agent/drivers/docker/docker_client.go
```diff
@@ -263,6 +264,12 @@ func (d *dockerWrap) retry(ctx context.Context, f func() error) error {
 		default:
 		}
 
+		if isSyslogError(err) {
+			// syslog error is an user error
+			e := models.NewAPIError(500, errors.New("Internal server error (SYSLOG)"))
```
skinowski: Minor nit, but I think it would be good to hoist this to models/error.go, see ferr{} class of errors there... (eg. ErrRequestContentTooBig as example.) Maybe we could clarify the intent too, "failed to connect to syslog" or something like that perhaps?

kppk: Yeah, the issue is that it is the function consumer who is getting the error, not the function owner.

PR #1414, opened 2019-02-22
Send RunnerMsg_ResultStart message

### LINKED ISSUE #1413: Invoke call doesn't return Fn Http Headers if the fn doesn't return a body. (LB configration only)
**Description**
When Fn is configured in LB configuration there is a bug related to custom HTTP Header not returned in case the function doesn't return a body.
The problem here is in the writeResp method in the agent.go, precisely the problem is due to the io.Copy:
https://github.com/fnproject/fn/blob/73778de0ae919268d2fa293001ea6546269d7c82/api/agent/agent.go#L746

the code for io.Copy doesn't call the <dst>.Write if the source is empty:
https://golang.org/src/io/io.go?s=14042:14087#L403
In our case the dst is a http.ResponseWriter and if we don't call the write we skip the send of the "RunnerMsg_ResultStart" which is where we include the custom Headers defined by the function:
https://github.com/fnproject/fn/blob/73778de0ae919268d2fa293001ea6546269d7c82/api/agent/pure_runner.go#L441

As a result of that we won't see custom headers sent back to the caller if the function return an empty body or doesn't return a body at all.

reclaro:
This change adds a call to a function to send the RunnerMsg_ResultStart
message during the enqueueCallResponse function.
The RunnerMsg_ResultStart contains any headers set by the function and
the status code.
This fixes the case where we don't send custom headers if a function
doesn't return a body.

Fixes #1413


api/agent/pure_runner.go
```diff
@@ -425,31 +432,10 @@ func (ch *callHandle) Write(data []byte) (int, error) {
...
-					},
-				},
-			},
-		})
-	})
 
-	if err != nil {
-		return 0, err
-	}
+	ch.WriteHeader(ch.status)
```
rdallman: >  In the code we have in master if we see any errors during the send of RunnerMsg_ResultStart during the Write call, the Write return that error now that is not the case anymore.

pursuant to this, since we're using enqueueMsgStrict in WriteHeader then this will still bail, right? if not, it may be worth saving any errors from WriteHeader on the call handle and then checking that error here after calling WriteHeader, to block any further Write.

api/agent/pure_runner.go
```diff
@@ -399,6 +380,32 @@ func (ch *callHandle) Header() http.Header {
...
+							StatusCode: int32(status),
+						},
+					},
+				},
+			},
+		})
+	})
+
+	if err != nil {
+		logrus.WithError(err).Error("Error in WriteHeader, unable to send RunnerMsg_ResultStart, shutting down callHandler")
```
skinowski: This should be info level to avoid tripping alerts.

api/agent/pure_runner.go
skinowski: probably, we need this block here (or call a new function that does this):
`
select {
	case <-ch.ctx.Done():
	        return io.EOF
	case <-ch.doneQueue:
        	return io.EOF
        default:
	}
`

api/agent/pure_runner.go
```diff
@@ -992,7 +978,15 @@ func (pr *pureRunner) BeforeCall(ctx context.Context, call *models.Call) error {
 		err = models.ErrCallHandlerNotFound
 		return err
 	}
-	ch.enqueueDetached(err)
+	statusCode := http.StatusAccepted
+	if err != nil {
```
skinowski: here, error is always nil.

skinowski:
@reclaro Thanks for making these changes, LGTM.

PR #1421, opened 2019-02-26
make ch shed load under 503/error

rdallman:
previously we were consistently hashing to the same node even if we were full
on that node we were trying it with 100% probability and when it's loaded it
would with a high probability fail, this is a somewhat easy addition to shed
load with some probability when requests start failing against any given node,
using ewma, which will bounce up or down based on requests succeeding or
failing over time. when load goes down, probability will go up, otherwise it
will get shed across nodes.

this is not intended to be a very perfect solution and most of the values
themselves should be futzed with via simulation, however it should do better
than the naive ch algorithm if it works as intended, I haven't tested, this
was mostly for illustration and hopefully inspiration... one point of interest
is that this will loop back around on the ring and when all nodes are at
capacity this is likely to shed load somewhat evenly across nodes but it may
take a lot of retries, didn't optimize that case, but also not sure if we need
to as nodes should be scaling up when this happens?

i tried to leave comments to make things clear, but let me know if anything is unclear or if there's any obvious tweaks. i'm not sure this is in use, opening PR just for comments, we can merge I guess since this isn't really in use if it's easier to test that way and if that's desirable. anyway, opening discussion to tweaking this algorithm at a minimum because testing it in its current form is not going to perform well and it may well be more suitable in some shape...

### DIFF
--- a/api/runnerpool/ch_placer.go
+++ b/api/runnerpool/ch_placer.go
@@ -5,21 +5,30 @@ package runnerpool
 
 import (
 	"context"
-
-	"github.com/fnproject/fn/api/models"
+	"math/rand"
+	"sync"
+	"sync/atomic"
+	"time"
 
 	"github.com/dchest/siphash"
+	"github.com/fnproject/fn/api/models"
 	"github.com/sirupsen/logrus"
 )
 
 type chPlacer struct {
 	cfg PlacerConfig
+
+	load   map[string]*int32
+	loadMu sync.RWMutex
+	rng    *rand.Rand
 }
 
 func NewCHPlacer(cfg *PlacerConfig) Placer {
 	logrus.Infof("Creating new CH runnerpool placer with config=%+v", cfg)
 	return &chPlacer{
-		cfg: *cfg,
+		cfg:  *cfg,
+		rng:  rand.New(&lockedSource{src: rand.NewSource(time.Now().Unix()).(rand.Source64)}),
+		load: make(map[string]*int32),
 	}
 }
 
@@ -44,10 +53,18 @@ func (p *chPlacer) PlaceCall(ctx context.Context, rp RunnerPool, call RunnerCall
 
 		i := int(jumpConsistentHash(sum64, int32(len(runners))))
 		for j := 0; j < len(runners) && !state.IsDone(); j++ {
-
 			r := runners[i]
+			if !p.checkLoad(key, r.Address()) {
+				// try to shed load, this simulates a probabilistic coin toss, see checkLoad for details
+				// NOTE: this is probabalistic and should converge... it shouldn't take forever, but noting so you know
+				continue
+			}
 
 			placed, err := state.TryRunner(r, call)
+
+			// set this based on error free placement
+			// TODO some errors are irredeemable, I didn't get that far in reasoning but this is ok?
+			p.setLoad(key, r.Address(), placed && err == nil)
 			if placed {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


skinowski:
I'm in general favor of introducing some kind of weight to these algorithms and I know the code above is based on the old fnlb, but I think we can simplify this further for our newer env.

I'd recommend making the 'load' metric explicit, in our case, 503, which should be set in 'err' returned from TryExec(). Other failures should not impact CH IMHO, (eg. cross-function 503/error.) Cross-function load is something to watch out as well. Large FnA may trigger 503, while small FnB might not. FnC may just fail for unrelated reasons; docker-auth...

Maybe opt for counting 503 only since the proposal attempts to mix in two different events into a single decayed variable? The proposed decay will likely get fed false samples quickly, but true samples will be much slower esp function exec/pull takes a long time. (And CAS there is already lossy, runner silence is an implicit true, etc.)

Perhaps also exp decay based on time instead of ewma? It sounds simpler if we counted 503 and expo decayed that. (Especially considering IPs can be reused and our metrics are passively acquired.)

Some of these imperfections could be OK since this is a probabilistic method, but I think we can simplify it slightly to make it more predictable/stable/simpler.

As a side note, this will also have a positive side effect of refreshing runner list quicker due to more skips. (This is an existing bug in CH/Naive, not refreshing runner list in some cases until full list is scanned.)

rdallman:
great feedback, thanks, agree with every suggestion... a few can probably be fixed up pretty quickly here. 

> Maybe opt for counting 503 only since the proposal attempts to mix in two different events into a single decayed variable? The proposed decay will likely get fed false samples quickly, but true samples will be much slower esp function exec/pull takes a long time. (And CAS there is already lossy, runner silence is an implicit true, etc.)

clarifying here. as it stands it's effectively the percent of successful requests in the last 100 requests. I agree we will get more falses than trues, which will just end up spreading across nodes. If we flip this on its head and count only failures, we want to map that onto some probability to send to the current node or to try another node - but I think what's being gotten at is that once things start succeeding we should wipe out failures faster and increase our P of using a 'good' node? one option here is just reduce the decay (could only count last 10, then when things go green they flip really fast). asking for clarity here + any suggestions on what to do with the 503 count in mind, policy wise, if you had something in mind.

> Perhaps also exp decay based on time instead of ewma? It sounds simpler if we counted 503 and expo decayed that. (Especially considering IPs can be reused and our metrics are passively acquired.)

yea, it would be ideal, the ewma is a bit of a hack for functions that run very infrequently (maybe performance wise it's meh in that case anyway - other cases?). in any event, it will correct relatively quickly for pathological cases, but we should smooth, agree - reducing decay factor would be a very 'cheap' solution here, as well. data structure wise, this would imply we need to keep a period of data I think, right? I don't think we're too concerned about memory overhead, ewma is very cheap in this respect, off-the-cuff if we ever get 100k rps on an lb, keep a 60 second period of int32, we're talking 200MB, honestly talking worst case for exp decay there that's not really concerning imho. ewma is just average of that, but yea, expiring at time window is impossible without keeping the entire period of interest stored to compute it so when calls are disparate this number is rather stale - in the success case, this is okay of course, with mix or fails it's worth analyzing at a minimum.

> Some of these imperfections could be OK since this is a probabilistic method, but I think we can simplify it slightly to make it more predictable/stable/simpler.

yea, totally, main point is we should tweak this thing and sit with the simulator to see if we can start converging on a digestible solution with it (or maybe we find out that we're not converging) - your suggestions will certainly help, thank you

rdallman:
> As a side note, this will also have a positive side effect of refreshing runner list quicker due to more skips. (This is an existing bug in CH/Naive, not refreshing runner list in some cases until full list is scanned.)

fixed this https://github.com/fnproject/fn/pull/1421/commits/2a0bfc9b7be30332d5667e50503316f9d1c780a5 - appears we're caching underneath, so this'll be good

I didn't get to extracting the 503 out, for one I'm not sure whether it's an actual API error and the cast will work, leaving as is until I/somebody tests that we can suss out the type of the error and do this check correctly (hopefully soon). slightly related, after staring at it, it seems a little odd not to ding for most kinds of errors, I guess at a minimum it means that if we get a strong offered load that consistently fails in the non-503 way, we'll bog down one machine with these possibly - I agree with the principle of only counting 503 and basing policy there, just need to sort out what to do with other errors too I guess. the other thing that sticks out is that we are not backing off at all between failed invocations in a single run through the runner list, there's an obvious tradeoff here of course but it seems a bit dangerous anyway to have 0 (I have function X that fails consistently [not in expected fashion] and 50 runners to go fail on...)

api/runnerpool/ch_placer.go
```diff
@@ -49,11 +49,17 @@ func (p *chPlacer) PlaceCall(ctx context.Context, rp RunnerPool, call RunnerCall
...
-		for j := 0; j < len(runners) && !state.IsDone(); j++ {
+		for j := 0; !state.IsDone(); j++ {
+			// refresh each iteration (should be cached...). detect if runner list changed, and reset j if so (?)
+			oldRunners := runners
+			runners, runnerPoolErr = rp.Runners(ctx, call)
+			if !equal(runners, oldRunners) {
+				j = 0
+			}
+			i := int(jumpConsistentHash(sum64, int32(len(runners))))
+			i = (i + j) % len(runners)
```
rdallman: @jan-g @JadeRedworth i think this will aid the test results quite a bit, I could see why things were egregious... previously we did this:

```go
runners, runnerPoolErr = rp.Runners(ctx, call)
i := int(jumpConsistentHash(sum64, int32(len(runners))))
for j := 0; j < len(runners) && !state.IsDone(); j++ {
  r := runners[i]
  placed, err := state.TryRunner(r, call)
  // ...
}
```

meaning we were trying the same `runner=runners[i]` for `n=len(runners)` times, ouchy

skinowski:
For decay, this seems simple https://en.wikipedia.org/wiki/Half-life
where we can perhaps decay number of 503s. (eg. half life 30 secs)

rdallman:
gonna close, actually think we could just switch to https://github.com/lafikl/consistent and instead of having a separate centralized node with 2 layers of consistent hashing to manage the hashes for a given lb set, if we rr to each lb in an lb set we could probably just get away with running chwbl independently on each lb as there will in theory be an even distribution anyway across lbs...

PR #1422, opened 2019-02-28
fn: make evictor less aggressive

skinowski:
Experiments in load testing revealed that eviction of
containers that are starting is detrimental since this
increases churn and acts as a feedback mechanism
increasing docker API rates. Starting and deleting
containers are one of the most expensive docker API
calls and traffic arriving on a busy/loaded server
can trigger flood of container create actions that
get canceled with evictions. Previous code tried to prevent
this by tracking the presence of the original request that
caused the container start, but based on the load test
data this does not seem good enough to avoid flurry
of evictions. In other words, in a busy system, likelihood
of the original request getting quickly serviced by another
container is not that rare.

With this change, we restrict evict-ability of containers
to Idle and Paused states exclusively. This increases the
likelihood of back pressure (503) since it allows starting
containers to initialize.

A race condition that occasionally causes a busy
container eviction is also fixed. The fix proposed
here is that we unblock the listener as if busy-container
really got evicted, but in the busy container we simply
refresh the eviction token and get rid of the evicted
old token. Unblocked listener checkLaunch rechecks
and attempts another launch after this.

In future, based on empirical data, we may consider
introducing evictions for slow docker pulls.

If the above changes still result in aggressive evictions,
a possible next step can be to restrict evictions to only
paused containers.


api/agent/agent.go
```diff
@@ -1221,6 +1198,7 @@ func newHotContainer(ctx context.Context, call *call, cfg *Config, id string, ud
 	}
 
 	return &container{
+		ctx:        ctx,
```
rdallman: nit: this is an anti-pattern (I know we do it in callHandle but we should fix that too), via https://golang.org/pkg/context/ : 

```
Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx
```

here I only see one usage for IsDone(), small change

api/agent/agent.go
```diff
@@ -1115,21 +1087,21 @@ func (a *agent) runHotReq(ctx context.Context, call *call, state ContainerState,
 				state.UpdateState(ctx, ContainerStatePaused, call.slots)
 			}
 			continue
-		case <-evictor.C:
+		case <-c.GetEvictChan():
```
kmjohansen: I don't think this can happen right now, since `SetEvictable` always creates a token.  However, should `GetEvictChan()` return `nil` because an `evictToken` is absent, it looks like this will block forever.

api/agent/agent.go
```diff
@@ -1115,21 +1087,21 @@ func (a *agent) runHotReq(ctx context.Context, call *call, state ContainerState,
...
 	// if we can acquire token, that means we are here due to
-	// abort/shutdown/timeout, attempt to acquire and terminate,
+	// abort/shutdown/timeout/evict, attempt to acquire and terminate,
 	// otherwise continue processing the request
 	if call.slots.acquireSlot(s) {
 		slot.Close()
 		return false
 	}
 
+	c.SetEvictable(call, false)
```
kmjohansen: Is it worth adding a note here to remind people that the `SetEvictable(false)` needs to happen after slot acquisition so we re-create a token if the evictor already fired?

api/agent/agent.go
```diff
@@ -1321,6 +1299,81 @@ func (c *container) WriteStat(ctx context.Context, stat drivers.Stat) {
...
+	// are we too late?
+	select {
+	case <-c.evictToken.C:
+	default:
+		return
+	}
+
+	// delete this already evicted token and fetch a new one
+	c.evictor.DeleteEvictToken(c.evictToken)
+	c.evictToken = c.evictor.CreateEvictToken(call.slotHashId, call.Memory+uint64(call.TmpFsSize), uint64(call.CPUs))
```
kmjohansen: Was about to remark that it would be worthwhile to `SetEvictable(false)` after creation, but then I realized that this is go.

api/agent/agent.go
skinowski: I documented this as much as possible now, GetEvictChan() returned channel must not be held beyond Enable/DisableEviction() calls (since these can alloc/realloc the token.)

api/agent/agent.go
skinowski: done

api/agent/agent.go
skinowski: By default eviction tokens are not in evictable state. This allows for more flexible usage avoiding some race conditions.

skinowski:
I've slightly modified the code to address the review feedback. I've also split SetEvictable() into enable/disable, which is more readable/simpler and in this state I think existing tests are sufficient
to cover it. (Except for creating the race condition, which is difficult to reproduce.)

For locking, despite having only 2 go-routines, there were some race conditions without locks particularly cases of leaks, double deletes, etc. I also noticed even with locks where's a correctness issue is IsEvicted() check in parent go-routine. I also noticed that I still left over some old logic in there. Since we no longer evict during initialization, I was able to clean some more code. Now, evictor related code is almost 100% isolated into child go-routine and no locks are needed except for calling container.Close().
Parent now waits for child go-routine to exist before calling container.Close().

api/agent/agent.go
```diff
@@ -1321,6 +1308,57 @@ func (c *container) WriteStat(ctx context.Context, stat drivers.Stat) {
...
+	default:
+		return
+	}
+
+	c.evictor.DeleteEvictToken(c.evictToken)
+	c.evictToken = c.evictor.CreateEvictToken(call.slotHashId, call.Memory+uint64(call.TmpFsSize), uint64(call.CPUs))
+}
+
+// Close closes container and releases resources associated with it
+func (c *container) Close() {
```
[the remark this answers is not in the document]
skinowski: I'll add the safe guards back. Looks like we explicitly documented this too: https://github.com/fnproject/fn/blob/master/api/agent/drivers/driver.go#L159

api/agent/agent.go
```diff
@@ -1321,6 +1308,57 @@ func (c *container) WriteStat(ctx context.Context, stat drivers.Stat) {
...
+	}
+}
+
+// GetEvictChan returns a channel that closes if an eviction occurs. Do not
+// refer to the same channel after calls to EnableEviction/DisableEviction flags or Close().
+func (c *container) GetEvictChan() chan struct{} {
+	if c.evictToken != nil {
+		return c.evictToken.C
+	}
+	return nil
```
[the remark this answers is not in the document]
skinowski: Discussed internally, but for the record: here the channel is designed to block if no evictions take place. runHotReq() expects this behavior and waits on a few channels that rarely or never close.

PR #1424, opened 2019-03-01
Don't log stack trace at the top handler error if we have gRPG errors

reclaro:
In the top error handler we log the stacktrace if we get back a non-API errors. Under that category we will have gRPC errors as well, there is no point to log a stacktrace for a gRPC errors where those could happen on the server side or due to some temporary network blips.
In this change we skip the stack trace log in presence of gRPC errors, plus we change some log level from Error down to Info for some retryable errors.


api/models/error.go
```diff
@@ -161,6 +161,11 @@ var (
 		error: errors.New("Unable to service the request for the reservation period"),
 	}
 
+	ErrRunnerUnavailable = err{
+		code:  http.StatusBadGateway,
+		error: errors.New("Received a gRPC error Unavailable from a runner."),
+	}
```
rdallman: should we include some idea of Retry-After headers for errors like these, where the intention is to retry on the client? the list here is mostly intended to be user-facing errors, now there's lb-facing and user-facing errors, these seem like two separate things (for example, this error is not intended to show to the user and is very much more of an 'internal' server error for us, that shouldn't get surfaced but we should log). without retry headers (afaik the preferred way to signal to client to retry) it's unclear that this error should be retried on the client side without parsing the string message out, and it seems easier to add retry-after for any relevent error rather than build up a set of error messages to retry. also seems like in that case this should be 503 instead of 502, RFC: https://tools.ietf.org/html/rfc7231#section-6.6.4

so i'm kinda thinking for errors like these we want to tack on the ability to add a retry header without displaying the message in the response. if this is completely off base, then just let me know. here's what I'm thinking:

```go
type RetryableError interface {
  error
  RetryAfter() int // seconds 
}

type retryableError struct {
  error
  delay int
}

// no need to override Error() we don't want to change the error on these
func (r *retryableError) RetryAfter() int { return r.delay }

func NewRetryableError(e error) RetryableError {
  return &retryableError{e, 1*time.Second} // some default, pick one...
}

func NewRetryableErrorDelay(e error, delay int) RetryableError {
  return &retryableError{e, delay}
}
```

and https://github.com/fnproject/fn/blob/master/api/server/error_response.go would after checking for models.APIError (which the idea here with the retryable error is that it would _not_ be one of these, ie we just hand off the original error to NewRetryableError and log it just the same as an internal server error), it could check for retryable, eg:

```go
switch x := err.(type); x {
case models.APIError:
  // same
case models.RetryableError:
  status = 503
  w.Header().Set("Retry-After", strconv.Itoa(x.RetryAfter())
default:
}

// TODO log it and then write response (log retryable same as internal server error)
```

we're already doing the retry-after thing https://github.com/fnproject/fn/blob/master/api/server/error_response.go#L48 but we can clean it up and stop using the api errors as a stand in for them, I think. what does anyone make of all that? I'd be willing to write this up if things are unclear to clarify them, it should not be too much code. the main thing is not using api errors for retryable internal errors at the lb level, so that the api errors only serve one purpose...

thanks for proposing this, just gave me the above idea after looking around. let me know what y'all think

reclaro:
In the last commit I added the concept of retryable error. The new interface contains an APIErrorWrapper that allows to define a specific root error which gets logged for debugging and a more general error which is returned back to the caller. Before this change we were setting the Retry-After header in case we get a ErrCallTimeoutServerBusy, now we intercept in the agent.Submit the error returned and if it is an ErrCallTimeoutServerBusy we transform it inot a retryable error.

rdallman:
> Before this change we were setting the Retry-After header in case we get a ErrCallTimeoutServerBusy, now we intercept in the agent.Submit the error returned and if it is an ErrCallTimeoutServerBusy we transform it inot a retryable error.

when we declare ErrCallTimeoutServerBusy can we just declare it as a Retryable error? then we don't have to worry about catching it anywhere else. that seems to be the intention for it, at least, do you think it makes sense?

reclaro:
I'll add and fix existing tests once we have an agreement that Retryable error is a good idea.

skinowski:
"likely a transient error and may be corrected by retrying" does not really mean we can retry all these gRPC cases. We already handle this. We can *only* retry in these cases:

1) TryExec() runner closed case.
2) TryExec() Engage() failure case
3) TryExec() Send ClientMsg_Try case of gRPC code is codes.Unavailable
4) If Runner tells LB that it can be retried.

All other cases, we cannot retry. We cannot intercept Recv() since Send() might have pushed data already to the container.

We already handle 4 cases above. Can we have an internal discussion how this issue exhibited itself in platform?

skinowski:
In other words, if Recv() gRPC tells us "codes.Unavailable" *may* be retried. In our case, except for 4 cases above we might have already pushed the data to the container and it is too late...

api/models/error.go
```diff
@@ -293,3 +298,32 @@ func NewAPIErrorWrapper(apiErr APIError, rootErr error) APIErrorWrapper {
 		root:     rootErr,
 	}
 }
+
+type RetryableError interface {
+	APIErrorWrapper() APIErrorWrapper
```
rdallman: my nit with this is that this interface does too much. this interface only needs to implement the ability to set an error as retryable. an error can then still implement APIErrorWrapper to get at a root error if it wants to in addition to being a RetryableError, but it needn't be an APIErrorWrapper itself. then we can compose the legos

reclaro:
In this version the gRPC error is not a retryable error anymore as we don't know if the function run for a while on the runner so we don't return a response with the Retry-After header but we return a 500 error, it is up to the caller to decide to try again the same call.
There is a simplified version of the Retryable error,  if we don't have an agreement on the retryable error thing it is probably worth just fixing the gRPC issue where the fix will be just transform the generic error in an APIError. 
Let me know what do you think, system_tests are still failing but I haven't looked at them yet.

reclaro:
I changed my mind, as we agree that this gRPC error won't be a retryable error i am going to change this PR so as it manages just the better management of the gRPC error and remove all the Retryable idea here. If we want to implement the Retryable errors that can be done on a separate PR.

skinowski:
@reclaro Thanks for your patience with this. 

Instead of adding inspect function into receiveFromRunner() function, is it better to add error wrapping code in top level error handler on service side? Because there's a few more errors that this PR doesn't catch eg. ErrorRunnerClosed, io.EOF, io.ErrShortWrite, ErrorPureRunnerNoEOF, grpc errs from Engage() and Send_tryMsg, json.Marshall() errors, etc. etc.

Top level handler can translate these to 500 and choose to log them or not.

Alternatively, maybe try to handle them in TryExec() function itself?

reclaro:
This is a super simplified version where in the top error handler we just verify if we have any gRPC error back and if so we don't log the stacktrace with the error.

as @skinowski pointed out in some other discussion, is the stacktrace helpful at all at that level?

PR #1425, opened 2019-03-05
Move and expand community section in prep of first community call

carimura:
Please see matching PR on the docs side https://github.com/fnproject/docs/pull/53


carimura:
sweet. and from the other PR just making sure this is captured:

the cleanup continues. @rdallman brought up the good point that there are a lot of repo-specific things that still need to be with the repo (like specific developer-facing contributing stuff ie checking out, committing, etc.). I'd definitely like to make sure each repo still has these things. Maybe as part of their own CONTRIBUTING.md or DEVELOPER.md (as I did with the fn server repo)

PR #1429, opened 2019-03-07
added comment to update kube shutdown grace period with updates to pl…

shreyagarge:


api/runnerpool/placer_config.go
```diff
@@ -18,7 +18,8 @@ type PlacerConfig struct {
 
 func NewPlacerConfig() PlacerConfig {
 	return PlacerConfig{
-		RetryAllDelay:         10 * time.Millisecond,
+		RetryAllDelay: 10 * time.Millisecond,
+		// kubernetes terminationGracePeriod needs to be updated when PlacerTimeout is updated.
```
reclaro: nit: Please make this comment more generic

api/runnerpool/placer_config.go
```diff
@@ -19,7 +19,8 @@ type PlacerConfig struct {
 func NewPlacerConfig() PlacerConfig {
 	return PlacerConfig{
 		RetryAllDelay: 10 * time.Millisecond,
-		// kubernetes terminationGracePeriod needs to be updated when PlacerTimeout is updated.
+		// If running on a container orchestration platform, container termination period needs to be updated when PlacerTimeout is changed.
```
reclaro: sorry i should have caught this before, can you explain how it should be updated? I guess it has to be aligned in some way to the container termination period but it is not clear from the comment above.

PR #1433, opened 2019-03-13
fn: introducing a new image pull layer

skinowski:
Due to unnecessary http traffic caused by docker internally, with
this PR a new image pull layer is introduced. This allows us to
serialize same image docker pulls using a simple active transfer
with list of listeners model.

The timeout behavior is slightly different when multiple listeners
are waiting. The timeout from first listener is emitted to all
listeners. However since the docker-pull timeout is globally
configured, overall timeout behavior is essentially the same.

### DIFF
--- a/api/agent/drivers/docker/cookie.go
+++ b/api/agent/drivers/docker/cookie.go
@@ -2,7 +2,6 @@ package docker
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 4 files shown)


rdallman:
was there a reason not to use https://github.com/fnproject/fn/blob/master/api/common/singleflight/singleflight.go to cover this? it's hiding out in common, but I'm not sure if it was analyzed and ruled out - would get rid of most of this code, from what I can tell to just do `singleflight.Do(imageKey(call), func() { dockerPull() })` inline where it was and the timeout behavior would be the same, as well.

PR #1435, opened 2019-03-14
To Enable configuring http server settings through env

shreyagarge:
currently the fn server uses default http server settings such as timeouts and limits on header sizes.

This change enables setting them using environment variables, but if not set, default values will be used.

### DIFF
--- a/api/server/init.go
+++ b/api/server/init.go
@@ -62,7 +62,7 @@ func getEnvDuration(key string, fallback time.Duration) time.Duration {
 	var err error
 	res := fallback
 	if tmp := os.Getenv(key); tmp != "" {
-		res, err = time.ParseDuration(tmp)
+		res, err = time.ParseDuration(tmp + "s")
 		if err != nil {
 			res = fallback
 			err = fmt.Errorf("%s invalid %s=%s", err.Error(), key, tmp)
--- a/api/server/init.go
+++ b/api/server/init.go
@@ -2,7 +2,6 @@ package server
 
 import (
 	"context"
-	"fmt"
 	"io/ioutil"
 	"os"
 	"os/signal"
@@ -41,7 +40,7 @@ func getEnvInt(key string, fallback int) int {
 		var err error
 		var i int
 		if i, err = strconv.Atoi(value); err != nil {
-			panic(err) // not sure how to handle this
+			logrus.WithError(err).WithFields(logrus.Fields{"string": value, "environment_key": key}).Fatal("Failed to convert string to int")
 		}
 		return i
 	} else if value, ok := os.LookupEnv(key + "_FILE"); ok {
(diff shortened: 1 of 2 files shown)


api/server/init.go
```diff
@@ -56,6 +58,20 @@ func getEnvInt(key string, fallback int) int {
 	return fallback
 }
 
+func getEnvDuration(key string, fallback time.Duration) time.Duration {
+	var err error
+	res := fallback
+	if tmp := os.Getenv(key); tmp != "" {
+		res, err = time.ParseDuration(tmp + "s")
```
skinowski: Do we need to append "s"? If we expect a duration string, then we can just try to parse it?

api/server/init.go
```diff
@@ -56,6 +58,20 @@ func getEnvInt(key string, fallback int) int {
...
+		if err != nil {
+			res = fallback
```
skinowski: setting res to fallback right before panic() is slightly odd. We can remove this line?

api/server/init.go
shreyagarge: I added that because it would be nice going forward to be able to automatically compute the write time in helm templates using the read times and placement times. To be able to do arithmetic in helm, I removed the "s" at the from there. So we don't expect duration strings here, just the integral part in seconds.

api/server/init.go
shreyagarge: true! I wrongly added it in two places

api/server/init.go
```diff
@@ -56,6 +58,19 @@ func getEnvInt(key string, fallback int) int {
 	return fallback
 }
 
+func getEnvDuration(key string, fallback time.Duration) time.Duration {
+	var err error
+	res := fallback
+	if tmp := os.Getenv(key); tmp != "" {
+		res, err = time.ParseDuration(tmp + "s")
+		if err != nil {
+			err = fmt.Errorf("%s invalid %s=%s", err.Error(), key, tmp)
```
riconnon: I'm not sure I'm following this format string. The error is invalid?

I would probably use `logrus.WithFields().Fatal()` here, but that would be slightly different to the `getEnvInt` behaviour.

shreyagarge: It would give a message something like this - panic: time: invalid duration abcd invalid ENV_VAR=abcd if we had passed abcd as the duration be parsed.. just to make the panic message more understandable

api/server/init.go
skinowski: I see. In that case, may I suggest that we use parse int? We can get rid of this function and just use getEnvInt? (or add GetEnvUint?)

api/server/server.go
```diff
@@ -116,9 +116,24 @@ const (
...
 
-	// EnvMaxRequestSize sets the limit in bytes for any API request's length.
+	// EnvMaxRequestSize sets the limit in bytes for any API request body's length.
 	EnvMaxRequestSize = "FN_MAX_REQUEST_SIZE"
 
+	// EnvMaxHeaderSize sets the limit in bytes for any API request body's length.
+	EnvMaxHeaderSize = "FN_MAX_REQUEST_HEADER_SIZE"
+
+	// EnvReadTimeout sets the timeout limit for reading a http request body.
+	EnvReadTimeout = "FN_REQUEST_BODY_READ_TIMEOUT"
```
skinowski: if these timeouts have to be in seconds, can you please rename or document that?

api/server/init.go
skinowski: or rename this as getEnvSeconds()?

PR #1438, opened 2019-03-21
add app/fn id and image to container metrics

rdallman:
in main we wire in view tags, so for OSS we can wire these tags in by default
and anybody that builds their own version of fn that doesn't want them can
easily turn them off (I did this, and it works as expected, the tags don't
appear in prometheus). I think we probably want to add these tags in more
places, but for purpose of discussion and closing out for the metrics we want,
I'm okay with starting out with a limited set of metrics.

I'll say that the symmetry of wiring them into the agent to optionally turn
them on in main seems somewhat appealing though slightly odd specifically for
cases where certain sets of metrics may have different sets of tags. it seems
like we should bundle metrics by what tags they offer, and allow configuring
views that way. I have not gotten into the weeds on strictly defining this
across each set of metrics, as noted earlier, this handles the metrics of
interest to add these tags of interest, but I think the general intent
expressed here seems appealing, we just need to define all the bundles well.

works as intended:

```
✗: curl -sSL localhost:8080/metrics | grep "paused"
fn_container_paused_total{app_id="01D2RHGQ15180043RZJ0000001",fn_id="01D2SN4A7Z180043RZJ000000C",image_name="hello:0.0.41"} 1.0
```

again, anyone that registers the views without these tags won't see these.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -455,7 +455,7 @@ func (a *agent) checkLaunch(ctx context.Context, call *call, caller slotCaller) 
 		return
 	}
 
-	state := NewContainerState(a.cfg.EnableContainerMetrics)
+	state := NewContainerState()
 	state.UpdateState(ctx, ContainerStateWait, call)
 
 	mem := call.Memory + uint64(call.TmpFsSize)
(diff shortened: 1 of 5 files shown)


rdallman:
closes #1373 - CLA is signed for the first commit

rdallman:
bump @skinowski when you get a moment

rdallman:
thanks!

PR #1441, opened 2019-03-26
fn: limit function header size if configured

skinowski:
Allow limiting function response header size.


api/agent/agent.go
```diff
@@ -682,6 +682,9 @@ func (s *hotSlot) dispatch(ctx context.Context, call *call) error {
 		if ctx.Err() == context.DeadlineExceeded {
 			return context.DeadlineExceeded
 		}
+		if strings.Contains(err.Error(), "server response headers exceeded ") {
```
reclaro: We have a white space at the end of this string is that intentional or a bug? I am not a great fan of string matching in general, seems very fragile.

PR #1442, opened 2019-03-26
run function container as unprivileged user

kppk:
Runs the function container as user (id:1000, gid:1000) in docker. This requires corresponding changes in base runtime images (add user/group) that are used by the fdks.

It also drops all the security capabilities (by using `CapDrop=all`).

Note that there is an opt-out environment variable `FN_DISABLE_DOCKER_SECURITY` which can be set to true to disable this features.

### DIFF
--- a/api/agent/config.go
+++ b/api/agent/config.go
@@ -11,39 +11,40 @@ import (
 
 // Config specifies various settings for an agent
 type Config struct {
-	MinDockerVersion        string        `json:"min_docker_version"`
-	ContainerLabelTag       string        `json:"container_label_tag"`
-	DockerNetworks          string        `json:"docker_networks"`
-	DockerLoadFile          string        `json:"docker_load_file"`
-	FreezeIdle              time.Duration `json:"freeze_idle_msecs"`
-	HotPoll                 time.Duration `json:"hot_poll_msecs"`
-	HotLauncherTimeout      time.Duration `json:"hot_launcher_timeout_msecs"`
-	HotPullTimeout          time.Duration `json:"hot_pull_timeout_msecs"`
-	HotStartTimeout         time.Duration `json:"hot_start_timeout_msecs"`
-	DetachedHeadRoom        time.Duration `json:"detached_head_room_msecs"`
-	MaxResponseSize         uint64        `json:"max_response_size_bytes"`
-	MaxHdrResponseSize      uint64        `json:"max_hdr_response_size_bytes"`
-	MaxLogSize              uint64        `json:"max_log_size_bytes"`
-	MaxTotalCPU             uint64        `json:"max_total_cpu_mcpus"`
-	MaxTotalMemory          uint64        `json:"max_total_memory_bytes"`
-	MaxFsSize               uint64        `json:"max_fs_size_mb"`
-	MaxPIDs                 uint64        `json:"max_pids"`
-	PreForkPoolSize         uint64        `json:"pre_fork_pool_size"`
-	PreForkImage            string        `json:"pre_fork_image"`
-	PreForkCmd              string        `json:"pre_fork_pool_cmd"`
-	PreForkUseOnce          uint64        `json:"pre_fork_use_once"`
-	PreForkNetworks         string        `json:"pre_fork_networks"`
-	EnableNBResourceTracker bool          `json:"enable_nb_resource_tracker"`
-	MaxTmpFsInodes          uint64        `json:"max_tmpfs_inodes"`
-	DisableReadOnlyRootFs   bool          `json:"disable_readonly_rootfs"`
-	DisableDebugUserLogs    bool          `json:"disable_debug_user_logs"`
-	IOFSEnableTmpfs         bool          `json:"iofs_enable_tmpfs"`
-	IOFSAgentPath           string        `json:"iofs_path"`
-	IOFSMountRoot           string        `json:"iofs_mount_root"`
-	IOFSOpts                string        `json:"iofs_opts"`
-	ImageCleanMaxSize       uint64        `json:"image_clean_max_size"`
-	ImageCleanExemptTags    string        `json:"image_clean_exempt_tags"`
-	ImageEnableVolume       bool          `json:"image_enable_volume"`
+	MinDockerVersion              string        `json:"min_docker_version"`
+	ContainerLabelTag             string        `json:"container_label_tag"`
+	DockerNetworks                string        `json:"docker_networks"`
+	DockerLoadFile                string        `json:"docker_load_file"`
+	DisableUnprivilegedContainers bool          `json:"disable_unprivileged_containers"`
+	FreezeIdle                    time.Duration `json:"freeze_idle_msecs"`
+	HotPoll                       time.Duration `json:"hot_poll_msecs"`
+	HotLauncherTimeout            time.Duration `json:"hot_launcher_timeout_msecs"`
+	HotPullTimeout                time.Duration `json:"hot_pull_timeout_msecs"`
+	HotStartTimeout               time.Duration `json:"hot_start_timeout_msecs"`
+	DetachedHeadRoom              time.Duration `json:"detached_head_room_msecs"`
+	MaxResponseSize               uint64        `json:"max_response_size_bytes"`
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       # Rebuild fnserver if necessary
       - run: |
           if [[ -n "$FN_NEEDED" ]]; then
-            make test -j $(nproc)
+            sudo PATH=/usr/local/go/bin:$PATH make test -j $(nproc)
```
rdallman: do we need sudo to run tests now? is this line still necessary?

api/agent/iofs.go
```diff
@@ -3,6 +3,7 @@ package agent
 import (
 	"context"
 	"fmt"
+	"github.com/fnproject/fn/api/agent/drivers/docker"
```
rdallman: could you run goimports on this file please? https://godoc.org/golang.org/x/tools/cmd/goimports (it's easy to set up to auto-run on save/check-in/etc in your environment)

api/agent/iofs.go
```diff
@@ -64,6 +65,13 @@ func newDirectoryIOFS(ctx context.Context, cfg *Config) (*directoryIOFS, error)
 		}
 		return nil, fmt.Errorf("cannot create tmpdir for iofs: %v", err)
 	}
+	if !cfg.DockerDisableSecurity {
+		err = os.Chown(iofsAgentDir, docker.FnUserId, docker.FnGroupId)
```
rdallman: pedantic: not sure if it would be better to do at create time (reduce syscall load, latency, etc) -- fully realize the horror that is the iofs stuff, it's maybe not worth it.

api/agent/iofs.go
kppk: yeah, this is necessary because we have to change the owner of the iofs directory to uid 1000 (so the function container can write to it), the change of ownership will fail if we run tests locally (not as root).

we could add make target to run the `make test` in docker as root, that container would need to have dind and golang, afaik.

anyway, to make sure, I will try PR without the sudo, so we can see where is fails...

rdallman: hmmm, gotcha. concerned about the local experience running fn too. I understand we want to have similar behavior b/w service and oss defaults here, I'm wondering what our options are; by default, map a host user pid we have perms on onto fn pid for the container (can allow configuring this or something)? something else? I'm not as familiar here

kppk: omg, I put my previous comment to wrong place :-)) sorry

.circleci/config.yml
kppk: running fn server should not cause any troubles, cause we are it in dind container as root, the iofs ownership change will work.

I think this is more about fn dev experience, especially when running tests locally :-(

rdallman: yea, I agree. i'm curious what our options are. it may be interesting to explore what docker themselves do [for local dev/test], they have a similar thing to us in a few ways between the unix socket and the docker user (e.g. docker run requires sudo w/o adding the docker user to your group)

kppk: without sudo we get:

`msg="failed to change owner" app_id=01D8SBQ92T1A70320ZJ0000002 cpus= error="chown /tmp/iofs310208425: operation not permitted"`

see https://github.com/fnproject/fn/pull/1478

we could also disable the security feature all together, but I would probably disable it by default and enable it in circleci only

api/agent/iofs.go
kppk: we use `ioutil.TempDir` to create the dir, which doesn't allow to set user, but yeah, I can add similar function

images/fn-test-utils/Dockerfile
```diff
@@ -10,6 +10,7 @@ RUN cd $D && go build -o fn-test-utils && cp fn-test-utils /tmp/
 
 # final stage
 FROM alpine
+RUN addgroup -g 1000 -S fn && adduser -S -u 1000 -G fn fn
```
skinowski: we should make the same change in fn-status-checker, fn-test-volume images.

rdallman:
after looking at the docker namespace mapping stuff, it seems not very fun when dealing with volumes like we have to for uds, not sure we're going to get an easy win there but it may be possible anyway (check uid of process being run, use it to create uds volume and map it to container, but the volume fiddling may not even work with this)

re: local dev experience and testing, i'd rather require users to make the fn user (1000) to be able to run locally with these settings turned on than to force the usage of sudo to run the tests (for which, we just need to update CONTRIBUTING.md and such). though, this still seems less than ideal, should try to make it as easy as possible (otherwise, slippery slope). we could make looser perms on the uds directory so that any user can write a file inside of it, and inside the container then the fn user would be able to make a file in there (we could make the perms configurable, too) no problem, I guess we'd have to make fdks make the uds file with the same loose perms so that the host being a different user could read the file for the uds read. with the file perms under control though, i think we can pretty much map the uid of the user running fn into the container as 1000 instead of needing the fn user on the host, I guess the biggest issue is the uds permissions. 

i tried to run the docker cli integration tests (after removing my docker user) and learned that they don't really care about making this easy for users to run, either, seemingly, so no ideas from doing that.

I guess tmpfs is being attempted?

kppk:
We could also chown the iofs directory to GID 1000 and make it 770, for running the tests locally one would need to be member of GID 1000, I'm gonna try if that works...

Yes, I'm attempting tmpfs too, not straightforward.

.circleci/config.yml
```diff
@@ -56,6 +56,8 @@ jobs:
       # Rebuild fnserver if necessary
       - run: |
           if [[ -n "$FN_NEEDED" ]]; then
+            # sudo addgroup fn --gid 1000
```
kmjohansen: Nit: delete the commented out line?

api/agent/config.go
```diff
@@ -58,6 +59,8 @@ const (
 	EnvDockerNetworks = "FN_DOCKER_NETWORKS"
 	// EnvDockerLoadFile is a file location for a file that contains a tarball of a docker image to load on startup
 	EnvDockerLoadFile = "FN_DOCKER_LOAD_FILE"
+	// EnvDockerDisableSecurity disables docker security features like user name, cap drop etc.
```
kmjohansen: Nit: is the intent that this will also be a switch for other security features that are eventually integrated?  If not, it might be worthwhile to pick a more precise and descriptive name.

api/agent/drivers/docker/cookie.go
```diff
@@ -16,7 +16,15 @@ import (
 	"go.opencensus.io/trace"
 )
 
-var ErrImageWithVolume = models.NewAPIError(http.StatusBadRequest, errors.New("image has Volume definition"))
+const (
```
kmjohansen: Two thoughts here:

- Seems like this would benefit from a comment that explains the interdependence between the uid/gid and the runtime in the container.  IOW, if somebody comes along and changes this code for their own purpose, it would be nice to give them a hint that they also need to potentially change what's going on inside their container.
- Is it worthwhile to set this as a default to 1000 but then allow users to override?  I'm fine with the answer that this is a "patches welcome" case if somebody else wants it, but figured it would be worth asking up front.

api/agent/config.go
kppk: yeah, good question, we currently set docker's User, CapDrop and SecurityOpt, so I couldn't find more descriptive name, did you have some in mind?

kmjohansen: Not sure that I have any brilliant suggestions.  Does `FN_DISABLE_UNPRIVILEGED_CONTAINERS` sound more descriptive, or just more confusing?

kppk: sounds good to me :-)
one possible inconvenience: docker has `--privileged` option and we don't use it with this setting, so it might be little confusing...

CONTRIBUTING.md
```diff
@@ -67,6 +67,17 @@ This builds and runs the `fn` binary. It will start Fn using an embedded `sqlite
...
+
+```sh
+# create group fn with gid 1000
+sudo addgroup fn --gid 1000
+
+# add current user to it
+sudo usermod -a -G 1000 $USER
+```
+
+Run tests:
```
rdallman: this is stale after most recent commit I think? (woo)

PR #1450, opened 2019-04-01
Add an extention point that can deal with container creation

jan-g:
Sometimes, CallListener doesn't quite cut it. There are some use cases where
we'll want to have access to invocation extensions metadata - and possibly,
interact with the container creation process.

Take two: add the ability to wrap Close, plus an extensible BeforeCall/AfterCall.

These can be used by a custom driver to add listeners to a ContainerTask - this also offers a way to get per-call behaviour into a Before/After pair, plus an avenue (from the custom driver modifying the Cookie's definition before returning it) to intercept container creation.

I'm not 100% convinced this is neater than an explicit extension inferface, but I offer it in contrast. FWIW, the code here can be used to do the same kind of mount-mangling and population I had in mind for the original.


jan-g:
Up for a review of the basic shape. If this should instead be in fnext, lemme know.

Can wrap a couple of tests around this once there's some agreement on where it should live.

api/agent/agent.go
```diff
@@ -113,13 +113,39 @@ type agent struct {
...
 
+type ExtensionInterceptor interface {
+	// BeforeCreate is triggered immediately preceding container creation, after all other
+	// configuration is in place.
+	// ContainerTask.Id() can be used to provide a stable per-container key.
+	BeforeCreate(context.Context, drivers.ContainerTask, drivers.Cookie, Call) error
+
+	// AfterDestroy is triggered immediately after container shutdown.
+	// ContainerTask.Id() can be used to provide a stable per-container key.
+	AfterDestroy(context.Context, drivers.ContainerTask) error
```
reclaro: at this point we have already returned data back to LB? is that the reason because we don't have the Call object here?

api/agent/agent.go
```diff
@@ -113,13 +113,39 @@ type agent struct {
...
+	// BeforeCall is triggered immediately preceding call invocation.
+	// ContainerTask.Id() can be used to provide a stable per-container key that corresponds
+	// to the Id seen by BeforeCreate/AfterDestroy.
+	BeforeCall(context.Context, drivers.ContainerTask, Call, CallExtensions) error
+
+	// AfterCall is triggered immediately after an invocation, to tidy up any per-invocation state.
+	// ContainerTask.Id() can be used to provide a stable per-container key that corresponds
+	// to the Id seen by BeforeCreate/AfterDestroy.
+	AfterCall(context.Context, drivers.ContainerTask, Call, CallExtensions) error
+}
```
rdallman: the rest of these kinds of things are defined in the `fnext` package, could we move these there as well or is there a reason to keep them in agent? they are basically stubs, so they are easy to move around I think. to fit in with the other kids at school, perhaps `ContainerListener` would be more fitting than `ExtensionInterceptor` after moving to that package, with some more filled in method names to match, too, most likely, e.g. `BeforeContainerCreate`, etc.? feel free to take more artistic license with the names..

api/agent/agent.go
```diff
@@ -113,13 +113,39 @@ type agent struct {
...
 
 	// deferred actions to call at end of initialisation
 	onStartup []func()
 }
 
+type ExtensionInterceptor interface {
+	// BeforeCreate is triggered immediately preceding container creation, after all other
+	// configuration is in place.
+	// ContainerTask.Id() can be used to provide a stable per-container key.
+	BeforeCreate(context.Context, drivers.ContainerTask, drivers.Cookie, Call) error
```
rdallman: it would be nice to eliminate the cookie here, if we can - usage is unclear from this PR alone, just floating the thought out there. I could see a case for wanting to manipulate it though (with how it is now, which is a bit odd) it'd be ideal to be able to modify the containertask or call probably instead, similar to the other listener like things, which would hopefully cover 'configure the container this way' and 'add this info on the call', respectively (which are the only things i can think of wanting here) - oddly, all of these things are interfaces and not easily mutated from packages outside of their origin, which is unlike the other listeners: again, usage is just unclear here.

api/agent/agent.go
jan-g: Sure.

api/agent/agent.go
jan-g: The basic thrust of this is to be able to do things like paralleling the iofs-style additional mountpoint. the gist here is to be able to do two things:

1. mount a filesystem (tmpfs or something) into the container (maybe set an env var to point to the mount)
2. Per-invocation, shove some extension metadata into that filesystem for access from user code.

The point about this is to bolt on "automagic" delivery of creds (amongst other things - there are perhaps some other uses of this around debugging) into an invocation.

(If credentials live for less than the runtime of an instance, it may be the case in the future that we want to bolt on an "UpdateMetadata" message to the Exchange protocol - which could refresh creds and their ilk through this.)

If there's a decent way at the container def via ContainerTask I'll look to do that instead.

api/agent/agent.go
jan-g: Yeah, this is afterwards, on eviction or other forms of container shutdown.

api/agent/agent.go
jan-g: (The lack of context here is a bit of a PITA, I agree. Credential management is pretty service-specific; there's not a huge amount that can be done around that I think :-( )

skinowski:
FYI, this can't replace the call overrider since that runs (and it needs to run) much earlier than the proposed callbacks here.

rdallman:
thanks for digging into this. I think I understand the crux of the issue here (though, perhaps not), primarily that currently we're calling Close() [here](https://github.com/fnproject/fn/blob/master/api/agent/agent.go#L798) on the concrete implementation of the container task so we need something to expose a way for drivers to add things to the container close method. this does seem like a general flaw in interface design, namely, that we've created an interface to hand over that will have most of its methods called by another package, yet we still require calling one of them from the caller (Close). 

I did some marinating and had a thought that maybe we should move this onto the cookie. If we made part of the contract for close [here](https://github.com/fnproject/fn/blob/master/api/agent/drivers/driver.go#L24) to be responsible for closing the container task, we get part of the way there. we'd also need to make CreateCookie responsible for calling close on the container if it fails to return a cookie (otherwise we leak resources). the thinking being that we need to be able to decorate the driver and container and have a clear contract there; handing off all container responsibilities to the driver via the cookie. any thoughts here? is it sufficient for this use case?

BeforeCall and AfterCall seem out of place being on the driver itself as the driver is simply about container execution (not call execution), but I can't think of alternative ways to access the underlying container before and after each call on a container. what is the need for the WrapXxxCall methods? since the agent package is calling cookie.BeforeCall/AfterCall it should be possible to simply implement the BeforeCall/AfterCall methods on the cookie without the need for wrap by having the decorated driver return the decorated cookie with the BeforeCall/AfterCall methods implemented?

jan-g:
So the gist is this: the WrapBeforeCall and WrapAfterCall were there to add lifecycle interception just either side of the invocation running - where they have access to that request's call extensions.

It's true, the cookie could be the carrier for these.

The reason that the original version - with an explicit interface for plugins that behave like this - came about was that it seemed to me to be cleaner to expose an interface that simply had the requisite lifecycle extension points on it, rather than requiring hooks on Task or Cookie. The last in particular seems like an odd place to put things; Task is arguably a better fit in terms of what it represents. What v1 bundled together was the set of behaviours (a pre-creation hook, a post-deletion one, and per-call before and after). The main issue with the interceptor extension was that - since it typically requires per-container state (even if that's just a random id) - the ContainerTask winds up being a part of its method signatures.

Having this behaviour associated with a Task makes some sense, perhaps, in as much as if you're doing something before and after container creation that's closely related to things you do either side of a call, "Task" seems to be the decriptor of that: you're mutating the container/environment that a call executes in.

It's not clear to me what the cookie really represents, rather than a packaged up frozen request to a driver to make a container.

jan-g:
As per conversation with Reed last night -

WrapClose, WrapBeforeCall and WrapAfterCall in place. They're still on ContainerTask (for the moment) - it'd be nice to have the Before/After on a call, but with hot containers we need to intercept to wrap for each call to a container. (Agent might be the other place to put this behaviour, but for the moment it's on the CT as that's got the context available for the required decorations to be able to do their thing).

Dropped the Set/GetCallID stuff.

api/agent/agent.go
```diff
@@ -1302,22 +1311,8 @@ func (c *container) UDSDockerPath() string              { return c.iofs.DockerPa
...
-	var callId string
-	c.callIdLck.Lock()
-	callId = c.callId
-	c.callIdLck.Unlock()
-	return callId
-}
-
 // WriteStat publishes each metric in the specified Stats structure as a histogram metric
-func (c *container) WriteStat(ctx context.Context, stat drivers.Stat) {
+func (c *container) WriteStat(ctx context.Context, stat driver_stats.Stat) {
```
rdallman: not blocking, just curious, what was the motivation for this? I'm inclined to move it back [later] as it conflicts with opencensus and named imports are a smell of poorly named packages, via [the guide](https://github.com/golang/go/wiki/CodeReviewComments#imports)

api/agent/drivers/docker/docker_pool.go
```diff
@@ -54,27 +56,27 @@ type poolTask struct {
...
+func (c *poolTask) FsSize() uint64                                 { return 0 }
+func (c *poolTask) TmpFsSize() uint64                              { return 0 }
+func (c *poolTask) Extensions() map[string]string                  { return nil }
+func (c *poolTask) LoggerConfig() drivers.LoggerConfig             { return drivers.LoggerConfig{} }
+func (c *poolTask) WriteStat(ctx context.Context, stat stats.Stat) {}
+func (c *poolTask) UDSAgentPath() string                           { return "" }
+func (c *poolTask) UDSDockerPath() string                          { return "" }
+func (c *poolTask) UDSDockerDest() string                          { return "" }
+func (c *poolTask) GetCallId() string                              { return "" }
+func (c *poolTask) SetCallId(string)                               {}
```
rdallman: we can clean these up while we're here

api/agent/agent.go
jan-g: Circular imports, alas - driver stats are embedded into models.Call, but the before/after stuff wants to talk about a models.Call also.

(This is probably an indication that this *does* belong on the agent.)

api/agent/drivers/docker/docker_pool.go
jan-g: D'oh! I'll get those too.

api/agent/agent.go
rdallman: ah, thanks. yea. maybe move to models or something.

PR #1451, opened 2019-04-01
Extract name validation from model validators

vzDevelopment:
Extracting the name validation functionality from the models' validators means we can use them else where. In particular this means we can use them to validate user input in the CLI without duplicating the logic c.f. fnproject/cli#474

### Link to issue this resolves
This is a prerequisite for my changes in fnproject/cli#474

### What I did
The full explaination can be found [here](https://github.com/fnproject/cli/issues/474#issuecomment-478509980). In particular this commit seperates the name validation from `models.App`, `models.Fn`, and `models.Trigger` `Validate` functions so that they can be called independently. 

### How to verify it
```
cd $GOPATH/src/github.com/fnproject/fn/api/models/
go test
```

```
fn := models.Fn{Name: name}
fn.ValidateName()
fn.Validate()
```

```
app := models.App{Name: name}
app.ValidateName()
app.Validate()
```

```
trigger := models.Trigger{Name: name}
trigger.ValidateName()
trigger.Validate()
```

### One line description for the changelog

> Extract name validation from model validators

### One moving picture involving robots (not mandatory but encouraged)
```
<!DOCTYPE html>
<html>
<head>
<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.3.1/jquery.min.js"></script>
<script> 
$(document).ready(
    setInterval(
        function() {
            $("img").animate({
                    left: '100px'
                },
                500,
                function() {
                    $(this).animate({
                        left: 0
                    }, 500);
                }
            )
        },
        100
    )
);
</script> 
</head>
<body>

<img style="position:absolute" src="https://i.gifer.com/XmuP.gif" /></div>

</body>
</html>
```

### DIFF
--- a/api/models/trigger.go
+++ b/api/models/trigger.go
@@ -140,21 +140,12 @@ var (
 
 //Validate checks that trigger has valid data for inserting into a store
 func (t *Trigger) Validate() error {
-	if t.Name == "" {
-		return ErrTriggerMissingName
-	}
-
 	if t.AppID == "" {
 		return ErrTriggerMissingAppID
 	}
 
-	if len(t.Name) > MaxTriggerName {
-		return ErrTriggerTooLongName
-	}
-	for _, c := range t.Name {
-		if !(unicode.IsLetter(c) || unicode.IsNumber(c) || c == '_' || c == '-') {
-			return ErrTriggerInvalidName
-		}
+	if err := t.ValidateName(); err != nil {
+		return err
 	}
 
 	if t.FnID == "" {
@@ -176,6 +167,24 @@ func (t *Trigger) Validate() error {
 	err := t.Annotations.Validate()
 	if err != nil {
 		return err
+	}
+
+	return nil
+}
+
+func (t *Trigger) ValidateName() error {
+	if t.Name == "" {
+		return ErrTriggerMissingName
+	}
+
+	if len(t.Name) > MaxTriggerName {
+		return ErrTriggerTooLongName
+	}
+
+	for _, c := range t.Name {
+		if !(unicode.IsLetter(c) || unicode.IsNumber(c) || c == '_' || c == '-') {
+			return ErrTriggerInvalidName
+		}
 	}
 
 	return nil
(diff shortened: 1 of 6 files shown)


api/models/trigger.go
```diff
@@ -181,6 +176,20 @@ func (t *Trigger) Validate() error {
 	return nil
 }
 
+func (t *Trigger) ValidateName() error {
```
vzDevelopment: Need to also move the check for`ErrTriggerMissingName` to `ValidateName`

vzDevelopment:
I have added unit tests and fixed the issue of the missing fn name check in `Fn.ValidateName()`

vzDevelopment:
>  happy to chat on slack or here about the plan, it would be nice to lock down a plan for validating as there are a few bugs/PRs open for this now.

Humm these are fair points and I can't think of any decent solutions for it. The only options I can think of are:

* To have a shared libraries repo which everything used. However, this still has most of the issues you've pointed out above - it'd get huge, you'd have to recompile everything that used them, version mismatches etc.

* To have an API endpoint on the server which we could use for validation - this is how I'd get around the issue with websites - I'd have functions on the server side that can do the validation and then use ajax on the client side to make use of them. I don't like the idea of requiring a server to be able to call `fn init` though. I also fear this will pollute the functions API.

Which leaves me with what you suggested and that's keeping the validation server side and passing it through to the client.

In which case I'm going to leave things as they are. You've sorted the immediate issue with `-trigger` being appended to the function names.

vzDevelopment:
Something else I was wondering - I've never used Swagger, but I noticed the fn_go models had some validation that looked auto generated using it. Is there a way we could centralise it using Swagger and then have the functions server and the client use the Swagger generated models?

rdallman:
yep, we used to use the same models for everything (fn_go are generated as it is, fn were a long time ago, but have been manually updated since). it would be possible to go back to that but i'm not sure how much undoing might have to get done to do that in fn server in particular. I forgot if there was a reason that happened or not, API updates have been painful regardless, but I have a feeling it was from that... say we use fn_go models in fn, we have to update them at fn_go and vendor them in fn before using them in order to make an API update. that's just a vague recollection though. 

I think that moving most of the validations to be all server side (well, removing the cli validations) will make things a bit smoother here, too.

PR #1453, opened 2019-04-01
fn-dataplane: adding ack type in LogRequestMsg/Ack

skinowski:
Ack messages in LogRequestMsg can now specify specific ack type.


api/agent/grpc/runner.proto
```diff
@@ -100,6 +100,11 @@ message LogRequestMsg {
 
     // Acknowledgement and re-enable stream message sent by Collectors to Runners
     message Ack {
+        enum Type {
+            PUSH = 0;
```
ostrain: Minor, but what if these were called "received" and "committed" or something, just to make it a little more obvious what they mean?

PR #1458, opened 2019-04-04
remove logs/calls apis, async, and most of hybrid

rdallman:
\* I've left the comments splayed out in case someone prefers that for larger PRs, I will clean up later.

# details

the goal here was to do raw removals without doing any shuffling. there are some shuffling opportunities, no doubt, but this shaves a lot of things off to hopefully make that easier should we get the time. most of the changes are just to the tests, really, as the mq/log stuff was all over the place and this just removes it.

**on logs and calls:**

the service can't support the logs api so this removes it. there was basically no discussion when it was brought up to discuss log removal, specifically, and with these changes, we will end up in the stone age, so to speak, where users can get logs from fn either by starting it in debug mode or by hooking up a syslog url to their apps. wrt the service, that is about parity at time of writing, though this is changing soon. we've been recommending syslog mostly for some time anyway since it was impossible to debug uds bugs with the logs API anyway, and that's the source of seemingly all of our support requests now (let's not talk about it here). i'm ears (and mouth) on making a better getting started experience for logs in particular, however, this removes quite a bit of cruft which may make it easier to pave forward on that. I think in particular some kind of extension that hands the logs off to a function that users can configure to do whatever may be useful, and we can probably use some of the s3 stuff deleted here to make an example of that...

the calls api was also not easily supported by a service and so this removes it. we have extensions for this one, unlike the logs, so there's slightly less damage. there are some updates to add things like call ids into the logs that will further dampen the blow here. without async, this was not exactly necessary either.

both of these have a migration and are removed from the datastore tables, the service code never used this stuff so there's nothing to do there.

**on hybrid/data access:**

there is one use that is not part of the '/v2' normal APIs for things like get app, fn, and that is the GetTriggerBySource method. this prevents full scale removal of the whole thing without some rewiring. the hybrid client is basically a wrapper around the v2 API now, with the previously noted exception. the service code uses this interface (agent.DataAccess), as well, in the lb. there are no longer any uses of this interface in the agent, which not only speeds up function calls by default (15%ish for me locally) but will make the agent easier to munge around too. again, tried to keep shuffling to a minimum here to make this easier to follow. it would be nice to get rid of this interface wholesale soon enough.

FN_NODE_TYPE=runner has gone away (in theory, completely unused, since we have pure runner)

**on hard interface breakages:**

agent.New() has changed, and agent.NewLBAgent() has changed, simply to remove the first arg (a data access thing, from the hybrid stuff). afaik, there are no other hard breakages for still to be used interfaces. these are easy to update, only updating the factory method call for each.

CallHandler has also died, this is easy to remove too.

**on API changes:**

we need to update SDKs, I have the CLI update ready, I've left 410 stubs for dead APIs during the transition period. the swagger has been updated. there may be docs to update as well, this can be done pretty quickly, starting the process here and don't think it's blocking. 

**on async:**

mostly this is raw removal, and like the log thing we can probably make some examples of using a queue to fire functions off with this code, but it's dead code atm. there are some changes to the resource manager that are worth eye balling to double check, this is mostly just removal of the async watermark and the special method for async to wait on resources to free up.

# user facing breakage issues

if this eats into any current user's use case, i'm reluctant to think we shouldn't proceed here as this is really a lot of cruft removal we need to do here to align things, but we should figure out a path forward and fill in any holes for things like extensions. considering the pain involved in making this change, it'd be nice to rip the band aid off and work quickly to fill in the holes with this landed, even if users need to skip a few versions and freeze on a current version of fn before this. TBD, please pipe up if this will shatter anyone's world, would like to keep the users happy.


rikgibson:
This rather makes me sad. I've always been on the side of keeping calls and logs in OSS, at least until good-experience alternatives are in place.

denismakogon:
After some reading and considerations, I’d remove logs API but will keep calls API until better alternative. We’ve seen people using calls API to connect it with prometheus.

So, that’s said, logs must go in favor of syslog, but calls should remain until better option.

rdallman:
> logs API must go away, but calls should remain until better option.

can we add the prometheus metrics of interest to the particular case noted earlier? any info here? I'm guessing call start/end and status, anything else? am not sure if we're already logging metrics for those, as it is. also, we have extensions for calls, so if a user wants to send the call as a json blob somewhere or anything else (metrics, too, when it starts or ends), they can do that with the code that exists with this patch. I maintain it's a good idea to move forward here to tidy up the API to be consistent with the service, for tooling and ux purposes, but let's talk about how to fill these gaps.

> One question: if we no longer have async calls (for now, of course), should we also get rid of MQ dependencies?

thanks, i'll add this. probably aws bindings will go away too.

api/server/server.go
```diff
@@ -276,9 +262,6 @@ func NewFromEnv(ctx context.Context, opts ...Option) *Server {
 	opts = append(opts, WithJaeger(getEnv(EnvJaegerURL, "")))
 	opts = append(opts, WithPrometheus()) // TODO option to turn this off?
 	opts = append(opts, WithDBURL(getEnv(EnvDBURL, defaultDB)))
-	opts = append(opts, WithMQURL(getEnv(EnvMQURL, defaultMQ)))
-	opts = append(opts, WithLogURL(getEnv(EnvLogDBURL, "")))
-	opts = append(opts, WithRunnerURL(getEnv(EnvRunnerURL, "")))
```
skinowski: should we remove the EnvMQURL and EnvLogDBURL and EnvRunnerURL as well? It'll keep dependent repos cleaner.

rdallman: yes - thanks

api/server/server.go
```diff
@@ -1121,35 +985,23 @@ func (s *Server) bindHandlers(ctx context.Context) {
...
-			v2.GET("/fns/:fn_id/calls", s.goneResponse)
-			v2.GET("/fns/:fn_id/calls/:call_id", s.goneResponse)
-			v2.GET("/fns/:fn_id/calls/:call_id/log", s.goneResponse)
-		}
+		// TODO remove these in 30 days or something
+		v2.GET("/fns/:fn_id/calls", s.goneResponse)
+		v2.GET("/fns/:fn_id/calls/:call_id", s.goneResponse)
+		v2.GET("/fns/:fn_id/calls/:call_id/log", s.goneResponse)
 
 		if !s.noHybridAPI { // Hybrid API - this should only be enabled on API servers
```
skinowski: is noHybridAPI still used? Can we remove it?

denismakogon:
> can we add the prometheus metrics of interest to the particular case noted earlier? any info here

@rdallman can we please wait for 1 more day, I’ll try to reach our users I’ve been talking about. I’ll figure out their use cases and will go back to you.

rdallman:
@denismakogon sure, no problem, thanks

denismakogon:
@rdallman as promised, I'll try to translate you what I've got from a customer:
```
In fact, we need CRD (call detail record) - which functions failed, which ones were successful. We got used to collecting them (calls) in ELK* and then doing aggregation for the statistics, i.e., % of successful VS % of failed. But we'll be ok with removing calls API as long as you'd provide an alternative (middleware/plugin) that will allow us to do the same but with no Fn API polling on our side.

```
*  - no technical details on how they do that, but I asked how to provide some details

rdallman:
@denismakogon thanks! It sounds like https://github.com/fnproject/fn/blob/master/fnext/listeners.go#L77 will be sufficient for them. It's also possible to get this from prometheus, which is doing aggregations already. any way we can help them move over to unblock here? they can freeze fn version, too, if they need a little time.

denismakogon:
> It's also possible to get this from prometheus

the one thing that bothers me that `syslog` feature is accessible through the API but when calls API will go away which alternative we'd offer to users (not operators). which there was an option to let Fn an endpoint where to post calls objects (or whatever).

rdallman:
using the call extensions stuff, it should be relatively easy for a user to connect to their syslog url and post a log with the call to it. we had code to connect to syslog not so long ago, it was maybe 30 LOC (dial, form body, copy). we could have an example extension showing that but it's probably not a part of core.

with the way things are at present, without async, the calls api had relatively low utility to begin with, as the client will get the info back about errors or success in the response. with async this api was more needed, to show errors for calls that fail, but extension based would be okay too.

rdallman:
the problem with supporting webhooks on function invokes in core API would be that we would need to support it in the service, and supporting webhook requests from the runner nodes presents networking issues as well as scalability considerations - it's better to leave as extension I guess. the docker logs driver ends up making this easier to do for logs than webhooks, but in any case, we're talking about new features now.

denismakogon:
> using the call extensions stuff, it should be relatively easy for a user to connect to their syslog url and post a log with the call to it

that's fine, would be nice to provide an example with, I guess.

rdallman:
will plan to merge this monday and dance around with CLI/SDKs/other deps unless any objections come in over the weekend... thanks all

rdallman:
doing it live.

PR #1461, opened 2019-04-05
add httptrace to span, increase idle timeout

rdallman:
this adds https://golang.org/pkg/net/http/httptrace/ support to our opentracing spans using their built in goodies for it https://godoc.org/go.opencensus.io/plugin/ochttp for our uds requests from fn to function containers. these were pretty handy. learned some things: 1) the unfreeze latency actually shows up in latency inside of the function to wake up, not just in the unfreeze docker call, figured this out from when bytes are received in the function. 2) most of the time on my laptop anyway, even pipelining function invokes, I would not get an idle conn to the container and the uds client would have to make a new one, on my laptop this takes about 5ms, which is about 20-25% of the function exec time (with the freeze latency... without the freeze latency, my function runs in 1ms and this adds 500% latency increase to my function, which sucks...)

so this patch changes the idle timeout to 120s, I think we can turn it off considering other additions here, just add your +1 for support, until then leaving 120s, totally out of a hat - since we only have 1 conn per container, and we don't really expect on the order of over hundreds of funcs on a server, it's hard to imagine a scenario this causes a problem anyway (considering the new close logic), but maybe someone else is capable of thinking of one. this patch also closes all idle connections when shutting down a container. this was going to happen after 1s presumably, anyway, but for 'correctness' and to allow us to fiddle with things, the way proposed here seems ideal. I think this is an easy win, anyway.

trying to make small patches :)

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -8,11 +8,11 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"os"
+	"path/filepath"
 	"strings"
 	"sync"
 	"time"
-
-	"path/filepath"
 
 	"github.com/fnproject/fn/api/agent/drivers"
 	dockerdriver "github.com/fnproject/fn/api/agent/drivers/docker"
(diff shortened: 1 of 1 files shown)


rdallman:
i guess an added bonus is this will reduce connection churn, esp if we turn off the idle conn timeout altogether... the perf result on tip is negligible because things are kinda sad on tip but it's a decent speedup with some other changes that are coming (about 10-15% speedup here with everything else underneath, the other stuff is like a 100% speedup and a 50% speedup)

rdallman:
looks like this: 
Screen Shot 2019-04-04 at 5 07 46 PM

rdallman:
ty

PR #1463, opened 2019-04-08
Change individual placement failure to warning

riconnon:
We are currently logging at error level for individual placement
failures which could be retried and could still succeed. This makes the
maximum log level for a per-placement-attempt failure a warning, and
reduces runner 503s to info level (since they are expected if the system
is under any kind of load)

### DIFF
--- a/api/agent/runner_client.go
+++ b/api/agent/runner_client.go
@@ -434,22 +434,21 @@ DataLoop:
 }
 
 func logCallFinish(log logrus.FieldLogger, msg *pb.RunnerMsg_Finished, headers http.Header, httpStatus int32) {
+	errorCode := msg.Finished.GetErrorCode()
 	errorUser := msg.Finished.GetErrorUser()
 	runnerSuccess := msg.Finished.GetSuccess()
 	logger := log.WithFields(logrus.Fields{
 		"function_error":     msg.Finished.GetErrorStr(),
 		"runner_success":     runnerSuccess,
-		"runner_error_code":  msg.Finished.GetErrorCode(),
+		"runner_error_code":  errorCode,
 		"runner_error_user":  errorUser,
 		"runner_http_status": httpStatus,
 		"fn_http_status":     headers.Get("Fn-Http-Status"),
 	})
-	if runnerSuccess {
-		logger.Info("Call finished successfully")
-	} else if errorUser {
-		logger.Info("Call finished with user error")
+	if !runnerSuccess && !errorUser && errorCode != http.StatusServiceUnavailable {
+		logger.Warn("Call finished")
 	} else {
-		logger.Error("Call finished with platform error")
+		logger.Info("Call finished")
 	}
 }
 


api/agent/runner_client.go
```diff
@@ -436,21 +436,14 @@ DataLoop:
...
 		"runner_error_user":  errorUser,
 		"runner_http_status": httpStatus,
 		"fn_http_status":     headers.Get("Fn-Http-Status"),
-	})
-	if runnerSuccess {
-		logger.Info("Call finished successfully")
-	} else if errorUser {
-		logger.Info("Call finished with user error")
-	} else {
-		logger.Error("Call finished with platform error")
```
rdallman: this one, at least, seems like something we want. we have error codes but there's a catch all for errors we aren't expecting, and it seems useful to surface those so that we can debug and fix, otherwise am not sure how we would go about debugging without these. if the issue is too many of a particular error we aren't interested in, it seems wise to suppress in the logging system instead of suppressing all errors? are we logging this twice, somewhere else?

it's also useful to see all landings of a function, fwiw, for example if the function runs twice and succeeds twice because of retries, this can be useful (need the id logged in order to see this). not sure if i missed some context to understand why we want to do this.

riconnon: OK I'm sold. My initial thought was that we could fall back on metrics for occurences of specific errors and codes, but happy to be conservative at least for now. Main thing I wanted was not to log at error level for something that might still succeed, especially for conditions we expect, such as runner 503.
Please see the updated version of this change.

rdallman: ok. this has 503 at info, I think 503 has more layers than a teenager with daddy issues though. 503 are going to be the code for all errors we aren't expecting and probably want to investigate and also the source of some seemingly expected errors we're not interested in. is there any way to distinguish these? 

we have models.FuncError, models.ErrCallTimeoutBusy (I think this is the one we want to filter that's coming up a lot? can we just filter that one to debug/warn instead of the code for anything 503? does logstash not allow filtering this on their side (filtering here is fine, too)?)

riconnon: We don't have access to the actual `models.APIError` object on this side of the runner protocol afaik, only the code and a string (and I don't want to get into string matching errors)
Seems that runner error codes should be unambiguous. This isn't a public facing service so why are we tying ourselves to the range and semantics of HTTP statuses? Even if we are using HTTP statuses 503 should not be the catch-all "something went wrong" bucket. Surely that should be 500.

riconnon: Additionally, `models.ErrCallTimeoutServerBusy` is the only error in `api/models` which has `code: http.StatusServiceUnavailable` so I think we should be OK matching 503.

riconnon: Re: filtering here vs later in the logging pipeline. I think we should have a clear idea what our definitions of severity are.

I work on something like:
debug: something happened, it's probably not interesting most of the time
info: something happened that might be interesting
warn: something _bad_ happened that might be interesting but should not affect user experienace
error: something bad happened which will have produced user facing failure

I'm sure that other people have conflicting ideas of what these shoudl be, so maybe we should flesh out some definitions for our project and write them down.

If, after applying some consistent definition of severity to log levelling we still have cases where our logging pipeline is causing us trouble then it's time to look at filtering downstream, but I'd prefer to have a really solid definition of log level to base this on first.

PR #1464, opened 2019-04-08
go mod tidy

rdallman:
was getting checksum mismatches, this fixes

i still don't really understand mod, there's all kinds of crap in here we don't have any trace of in vendor, but it builds so?


rdallman:
i updated go to 1.12.1 in circleci, assuming that this was the root cause of my mismatch to begin with, so that we can build the thing after merging this thing.

riconnon:
While we're on the topic, we should probably build with go mod enabled in CI
a la https://github.com/fnproject/fn/commit/354223e3a8a3ff34caae0c8b48a8a65072e9a256

rdallman:
@riconnon the makefile ought to be doing this?

PR #1465, opened 2019-04-09
increase app/fn/trigger length caps

### LINKED ISSUE #1452: Make app and fn name length configurable
**Description**

We have a need to create app and fn names that are longer than the current 30 character limit. The associated database fields are `varchar(256)`, yet looking through the code, we see a const declaration of 30 characters for each and these cannot be overridden:

https://github.com/fnproject/fn/blob/master/api/models/error.go#L10

I would like to propose that we make 30 characters the default, but make them configurable at runtime.

1. Try to create a function or app with more than 30 characters

We get an error saying that neither field can't be over 30 characters

We expect to be able to set an app and/or function name that is greater than 30 characters without failing

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `fn version` (CLI command):**

curl to fnserver directly:
```
$ curl http://localhost:8080/v2/apps -XPOST -d "{\"name\":\"thisislongerthanthirtycharactersthisislongerthanthirtycharactersthisislongerthanthirtycharactersthisislongerthanthirtycharactersthisislongerthanthirtycharacters\"}"
{"message":"App name must be 30 characters or less"}
```

**Additional environment details (OSX, Linux, flags, etc.):**
OSX

rdallman:
was 30, now 255. the service validates the same value, via a different method,
for each of these already. sql uses varchar 256 so we're good. can't think of
any other ways this might break, so going for it.

closes #1452

### DIFF
--- a/api/models/app.go
+++ b/api/models/app.go
@@ -31,7 +31,7 @@ var (
 	}
 	ErrAppsTooLongName = err{
 		code:  http.StatusBadRequest,
-		error: fmt.Errorf("App name must be %v characters or less", maxAppName),
+		error: fmt.Errorf("App name must be %v characters or less", MaxLengthAppName),
 	}
 	ErrAppsInvalidName = err{
 		code:  http.StatusBadRequest,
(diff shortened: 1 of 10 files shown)


denismakogon:
@rdallman it's good to go. Big green "Merge" button waiting for you.

rdallman:
thanks @denismakogon

PR #1471, opened 2019-04-12
lb read cache everything under it

rdallman:
currently, the lb read cache only covers apps. however, the API itself can be [and is] used for fns, app ids and triggers, as well. without using the read cache, this means that we were hitting the datastore every invoke request to get the function and the app id, which resulted in added latency due to contention and costs there. get about a 20-30% throughput bump by doing this, and it seems like a pretty obvious thing to do - my load test has functions that run in about 1ms themselves in a not under duress fn environment, making the 5s window coverage a considerable reduction in db calls (like 10k less, in tests, since i'm running 10k in about 10s now it's about 2 db calls instead of 10k)

note that this does affect the service upstream, as the service uses this as well. this means that functions will have an additional worst case 5 second update latency from the sync update. from what I can tell, it's using bolt, so it's decent savings (at least, with sqlite3 it's about 20% for my load test) - if this is undesirable, it is one update in service code to stop wrapping the read access store with the cache (it's explicitly doing this atm). 

this also makes an update to the configuration dance, so that the read access actually shoots out metrics (and hits validator, which is pass through for get stuff, and uses the extensions), which helps for debugging things like this... from what I can tell, there are no other behavior changes here since we were previously doing the wrap for any datastore just later down the line than from the config option, this is just adding the same treatment to the read store pretty much.

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -359,7 +359,7 @@ func WithTLS(service string, tlsCfg *tls.Config) Option {
 // WithReadDataAccess overrides the LB read DataAccess for a server
 func WithReadDataAccess(ds agent.ReadDataAccess) Option {
 	return func(ctx context.Context, s *Server) error {
-		s.lbReadAccess = ds
+		s.lbReadAccess = agent.NewMetricReadDataAccess(ds)
 		return nil
 	}
 }
@@ -371,7 +371,7 @@ func WithDatastore(ds models.Datastore) Option {
 		s.datastore = datastore.Wrap(s.datastore)
 		s.datastore = fnext.NewDatastore(s.datastore, s.appListeners, s.fnListeners, s.triggerListeners)
 		if s.lbReadAccess == nil {
-			s.lbReadAccess = agent.NewCachedDataAccess(s.datastore)
+			return WithReadDataAccess(agent.NewCachedDataAccess(s.datastore))(ctx, s)
 		}
 		return nil
 	}
@@ -444,7 +444,10 @@ func WithAgentFromEnv() Option {
 				placer = pool.NewNaivePlacer(&placerCfg)
 			}
 
-			s.lbReadAccess = agent.NewCachedDataAccess(cl)
+			err = WithReadDataAccess(agent.NewCachedDataAccess(cl))(ctx, s)
+			if err != nil {
+				return errors.New("LBAgent creation failed")
+			}
 			s.agent, err = agent.NewLBAgent(runnerPool, placer)
 			if err != nil {
 				return errors.New("LBAgent creation failed")
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -201,7 +201,6 @@ type Server struct {
 	// TODO: extend this to cover gRPC options.
 	svcConfigs map[string]*http.Server
 
-	// Agent enqueue  and read stores
 	lbReadAccess           agent.ReadDataAccess
 	noHTTTPTriggerEndpoint bool
 	noFnInvokeEndpoint     bool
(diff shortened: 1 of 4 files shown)


api/server/server.go
```diff
@@ -430,9 +423,6 @@ func WithAgentFromEnv() Option {
 			if runnerURL == "" {
 				return errors.New("no FN_RUNNER_API_URL provided for an Fn NuLB node")
 			}
-			if s.datastore != nil {
-				return errors.New("lb nodes must not be configured with a datastore (FN_DB_URL)")
-			}
```
rdallman: i removed a few of these, this is being validated elsewhere in a 'more consistent way' still - see right about here: https://github.com/fnproject/fn/pull/1471/files#diff-44a527ea72ea9572e1cf71c7f1e2a0f0L568

rdallman:
here's my hokey load tests for the curious (well, a sample of them, have run this a lot now.. note that i'm far from scientifically precise but still pretty good confidence anyway):

```
this patch:

n 10000 p 20 duration: 9.420491084s tps: 1061.5157862613185 max: 161.733308ms min: 750.382µs mean: 17.725253ms median: 15.882986ms std: 10.653208ms
n 10000 p 20 duration: 9.456716897s tps: 1057.449441377731 max: 133.333246ms min: 665.583µs mean: 17.643815ms median: 15.542724ms std: 10.804356ms

master: 

n 10000 p 20 duration: 13.516036645s tps: 739.8618591123242 max: 281.636281ms min: 905.501µs mean: 25.403707ms median: 21.105557ms std: 18.449826ms
n 10000 p 20 duration: 15.589936024s tps: 641.4394507203527 max: 243.949316ms min: 788.739µs mean: 29.79846ms median: 24.202449ms std: 22.995301ms
```

api/server/server.go
```diff
@@ -454,7 +444,10 @@ func WithAgentFromEnv() Option {
 				placer = pool.NewNaivePlacer(&placerCfg)
 			}
 
-			s.lbReadAccess = agent.NewCachedDataAccess(cl)
+			err = WithReadDataAccess(agent.NewCachedDataAccess(cl))(ctx, s)
+			if err != nil {
+				return errors.New("LBAgent creation failed")
```
skinowski: fyi, the err msg is identical to below, maybe it should be more explicit.

reclaro:
discussion with other colleagues here, the only concern is that the backend database could be something which already uses some in-memory solution so the gain for this change could be not big and there is no way to disable it. In our case we use boltDB which uses a 
> memory-mapped data file 
(https://godoc.org/github.com/boltdb/bolt see Caveats paragraph)

I haven't done any benchmarking but it would be good to have a way to disable caching or maybe the solution is to implement a DataAccess without caching, what do you think?

rdallman:
> I haven't done any benchmarking but it would be good to have a way to disable caching or maybe the solution is to implement a DataAccess without caching, what do you think?

I've linked to how to do this in slack ;) - by default, the data access is not wrapped with the cache when providing a ReadDataAccess layer via `server.WithReadDataAccess` and we're manually wrapping with the cache in the service lb code before passing in the option, which we could delete that line if we're not interested in. agree that it would be useful to benchmark, as noted, with sqlite3 even the gains are pretty nice. i guess the state of things is that it's already explicitly being used for apps, unwittingly or not, since we have some load testing stuff now it may be interesting to see if there's any perf bump when adding in the use of it for functions, that ought to give a good idea.

rdallman:
am gonna merge since all concerns are addressed (andrea out ROW), happy to follow up, thanks all... hoping to get another 10x or so out of the agent before digging into lb_agent/pure_runner

PR #1472, opened 2019-04-13
update tracing stuff

rdallman:
* jaeger deprecated config stuff
* add call_id trace tag, proper app and fn trace span attributes

turns out the tag keys don't get added to the trace span attributes, have to
annotate them differently, kinda sucks but whatever... these are useful for
debugging (particularly call id for me right now, anyway)


rdallman:
it only shows up on the span you add it and not down stream, but still useful and possible to search and find enclosing spans, so it works. pic: 

Screen Shot 2019-04-12 at 6 16 06 PM

rdallman:
ty

PR #1473, opened 2019-04-15
Bump oomer limit to 120GB

hatchan:
- Why I did it

The test kept failing for me. The code in question kept "succeeding" by returning a `200`, instead of the expected `502`. My changes bump the size to `120GB`, though I'm not sure if this is actually triggering the OOM exception, or that there is some other check that already fails the request before actually running the function. 

I did notice this line: https://github.com/fnproject/fn/compare/bump_oomer?expand=1#diff-bb0376183f5d812769b683c93f9b3cc1L262 and https://github.com/fnproject/fn/compare/bump_oomer?expand=1#diff-b1e41947cb0f525a276bbaccc8bd6eb3L209 So this might be a goose chase. 

- What I did

Increase the size of the allocated memory of a test.

- How I did it

Change the `allocateMemory` property of a request to `120000000000`.

- How to verify it

Run the tests

Bump `oomer` to request 120GB

- Docker version

```
Client:
 Version:           18.09.5
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        e8ff056
 Built:             Thu Apr 11 04:43:57 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.5
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.8
  Git commit:       e8ff056
  Built:            Thu Apr 11 04:10:53 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```

https://media.giphy.com/media/d2aIU85OBNgRy/giphy.gif

### DIFF
--- a/api/server/runner_fninvoke_test.go
+++ b/api/server/runner_fninvoke_test.go
@@ -171,8 +171,8 @@ func TestFnInvokeRunnerExecution(t *testing.T) {
 	// Therefore, not testing for EndOfLogs for hot containers (which has complex I/O processing) anymore.
 	multiLogExpectHot := []string{"BeginOfLogs" /*, "EndOfLogs" */}
 
-	crasher := `{"echoContent": "_TRX_ID_", "isDebug": true, "isCrash": true}`           // crash container
-	oomer := `{"echoContent": "_TRX_ID_", "isDebug": true, "allocateMemory": 120000000}` // ask for 120MB
+	crasher := `{"echoContent": "_TRX_ID_", "isDebug": true, "isCrash": true}`              // crash container
+	oomer := `{"echoContent": "_TRX_ID_", "isDebug": true, "allocateMemory": 120000000000}` // ask for 120GB
 	// XXX(reed): do we have an invalid http response? no right?
 	ok := `{"echoContent": "_TRX_ID_", "responseContentType": "application/json; charset=utf-8", "isDebug": true}` // good response / ok
 	respTypeLie := `{"echoContent": "_TRX_ID_", "responseContentType": "foo/bar", "isDebug": true}`                // Content-Type: foo/bar
--- a/api/server/runner_httptrigger_test.go
+++ b/api/server/runner_httptrigger_test.go
@@ -245,7 +245,7 @@ func TestTriggerRunnerExecution(t *testing.T) {
 	multiLogExpectHot := []string{"BeginOfLogs" /*, "EndOfLogs" */}
 
 	crasher := `{"echoContent": "_TRX_ID_", "isDebug": true, "isCrash": true}`                                     // crash container
-	oomer := `{"echoContent": "_TRX_ID_", "isDebug": true, "allocateMemory": 120000000}`                           // ask for 120MB
+	oomer := `{"echoContent": "_TRX_ID_", "isDebug": true, "allocateMemory": 120000000000}`                        // ask for 120GB
 	ok := `{"echoContent": "_TRX_ID_", "responseContentType": "application/json; charset=utf-8", "isDebug": true}` // good response / ok
 	respTypeLie := `{"echoContent": "_TRX_ID_", "responseContentType": "foo/bar", "isDebug": true}`                // Content-Type: foo/bar
 


api/server/runner_fninvoke_test.go
```diff
@@ -171,8 +171,8 @@ func TestFnInvokeRunnerExecution(t *testing.T) {
 	// Therefore, not testing for EndOfLogs for hot containers (which has complex I/O processing) anymore.
 	multiLogExpectHot := []string{"BeginOfLogs" /*, "EndOfLogs" */}
 
-	crasher := `{"echoContent": "_TRX_ID_", "isDebug": true, "isCrash": true}`           // crash container
-	oomer := `{"echoContent": "_TRX_ID_", "isDebug": true, "allocateMemory": 120000000}` // ask for 120MB
+	crasher := `{"echoContent": "_TRX_ID_", "isDebug": true, "isCrash": true}`              // crash container
+	oomer := `{"echoContent": "_TRX_ID_", "isDebug": true, "allocateMemory": 120000000000}` // ask for 120GB
```
rdallman: i wonder if this has to do with oom behavior differences between circleci and a host. these tests fail for me locally, as well, but think we maybe should investigate a little more, this is likely hitting the oom on the host and getting killed rather than testing that we're killing containers that reach their limits [that are well within the host limits] - this is something we should ensure in testing to avoid containers running away with too much ram, and this seems worthy of investigation.

hatchan: Are your referring to it being oom killed because the machine doesn't have anymore memory vs the container reaching the memory limit set by the cgroup?

denismakogon: OOM killer operations based on machine resources. So, I’d recommend to investigate system events related to Docker daemon in first place. Because there’s a case when container can cause spikes on resource consumption in an unusual way that is recognized as the case for OOM killer.

That’s said, can you please check system events when the test fails?

skinowski:
I'm also super curious (and worried) why these pass in my environments (and circleci), but not others. It's worth to investigate it in case there's a bug lurking in somewhere.

api/server/runner_fninvoke_test.go
rdallman: > Are your referring to it being oom killed because the machine doesn't have anymore memory vs the container reaching the memory limit set by the cgroup?

yep. can play with this a little.

rdallman: I did docker stats and the function basically redlines at memory capacity (I'm allocating 200MB on a 128MB max function here) and then eventually execution finishes and I get a valid response, see here:

```
✗: echo '{"allocateMemory":200000000}' | cli invoke testapp fn-test-utils
{"request":{"allocateMemory":200000000},"header":{"Accept-Encoding":["gzip"],"Content-Type":["text/plain"],"Fn-Call-Id":["01D8CHXSEW180043RZJ0000007"],"Fn-Deadline":["2019-04-13T16:54:06-07:00"],"User-Agent":["Go-http-client/1.1"]},"config":{"FN_APP_ID":"01D722GHXY180043RZJ0000001","FN_FN_ID":"01D8CHTVV9180043RZJ0000001","FN_FORMAT":"http-stream","FN_LISTENER":"unix:/tmp/iofs/lsnr.sock","FN_MEMORY":"128","FN_TYPE":"sync","HOME":"/root","HOSTNAME":"debian","PATH":"/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"},"data":{},"trailer":[]}
```

in another term watching this:

```
$ docker stats 01D8CHXSEX180043RZJ0000008
CONTAINER ID        NAME                         CPU %               MEM USAGE / LIMIT   MEM %               NET I/O             BLOCK I/O           PIDS
4a4fd08886e0        01D8CHXSEX180043RZJ0000008   0.00%               127.8MiB / 128MiB   99.87%              836B / 0B           1.29MB / 0B         8
```

the invocation itself takes a few seconds, and then about a minute later I see in the docker events stream the container gets oom killed:

```
$ docker events
2019-04-13T16:57:06.984997580-07:00 container create 95d511cf51213cfef2ca37507b4bb4b07536ea1171831f2b76497e00d11e8029 (image=rdallman/fn-test-utils:0.0.3, name=01D8CJ46Z2180043RZJ000000B)                                                  
2019-04-13T16:57:06.986409141-07:00 container attach 95d511cf51213cfef2ca37507b4bb4b07536ea1171831f2b76497e00d11e8029 (image=rdallman/fn-test-utils:0.0.3, name=01D8CJ46Z2180043RZJ000000B)                                                  
2019-04-13T16:57:07.007376726-07:00 network connect 37a4bdb05fa9740de28c794e25ea4b9d5916c0282f5d407d5191439cac0cf362 (container=95d511cf51213cfef2ca37507b4bb4b07536ea1171831f2b76497e00d11e8029, name=bridge, type=bridge)                  
2019-04-13T16:57:07.348085439-07:00 container start 95d511cf51213cfef2ca37507b4bb4b07536ea1171831f2b76497e00d11e8029 (image=rdallman/fn-test-utils:0.0.3, name=01D8CJ46Z2180043RZJ000000B)                                                   
2019-04-13T16:57:24.437272050-07:00 container pause 95d511cf51213cfef2ca37507b4bb4b07536ea1171831f2b76497e00d11e8029 (image=rdallman/fn-test-utils:0.0.3, name=01D8CJ46Z2180043RZJ000000B)                                                   
2019-04-13T16:57:54.403912763-07:00 container kill 95d511cf51213cfef2ca37507b4bb4b07536ea1171831f2b76497e00d11e8029 (image=rdallman/fn-test-utils:0.0.3, name=01D8CJ46Z2180043RZJ000000B, signal=9)                                          
2019-04-13T16:57:54.412789289-07:00 container unpause 95d511cf51213cfef2ca37507b4bb4b07536ea1171831f2b76497e00d11e8029 (image=rdallman/fn-test-utils:0.0.3, name=01D8CJ46Z2180043RZJ000000B)                                                 
2019-04-13T16:57:54.575229163-07:00 container die 95d511cf51213cfef2ca37507b4bb4b07536ea1171831f2b76497e00d11e8029 (exitCode=137, image=rdallman/fn-test-utils:0.0.3, name=01D8CJ46Z2180043RZJ000000B)                                       
2019-04-13T16:57:54.672831726-07:00 network disconnect 37a4bdb05fa9740de28c794e25ea4b9d5916c0282f5d407d5191439cac0cf362 (container=95d511cf51213cfef2ca37507b4bb4b07536ea1171831f2b76497e00d11e8029, name=bridge, type=bridge)               
2019-04-13T16:57:54.735045976-07:00 container destroy 95d511cf51213cfef2ca37507b4bb4b07536ea1171831f2b76497e00d11e8029 (image=rdallman/fn-test-utils:0.0.3, name=01D8CJ46Z2180043RZJ000000B) 
```

anyway, all that's going on here is the oom guy is lazy and taking his sweet time to come around and kill the guy (see 137 above, it happens eventually). I think there are some settings to fiddle with, can look into it.

rdallman: I think I've confirmed that it's doing swapping on the host, and fun enough there's a bug in the fsouza driver that doesn't allow disabling this I don't think. the value is `omitempty` but we need to set to 0 to turn this off, see https://github.com/fsouza/go-dockerclient/blob/master/container.go#L757 - I've had a play with this and even commenting out the omitempty and rebuilding I still see memory swapping on the host. so, I think we need to figure this out as my understanding (reinforced from our other settings for swap) is that we want to disable container swapping altogether [not relying on host swappiness].

gonna try to dig into this. have wanted to switch to the docker/docker bindings for a time, too, and now that all the error checking is out and we're waiting to build up the error database again, it's a good time...

rdallman: yea, we seem to be making containers set with MemorySwap set to -1 (unlimited swap) when this is not our intention, did a docker inspect on a container. gonna keep digging, seems like there's some bug in bindings possibly but idk yet.

rdallman: I tried updating the fsouza library and docker/docker library and my local docker server and i'm still seeing containers get created with memory swap set to -1, was hoping for a quick fix there. going to do some speculative work to see if the docker/docker bindings fix, and then look at scoping out getting rid of the rest of the fsouza stuff. it may take a day to get something together that just uses the 'real' docker client to create containers (leaving everything else) and test it and everything, bit messy atm.

rdallman: i've confirmed that setting MemorySwappiness to 0 fixes the issue, managed to get it working with the docker/docker bindings. the API seems a bit weird, perhaps the docs are wrong or I'm confused, see https://docs.docker.com/config/containers/resource_constraints/#--kernel-memory-details but what's happening is that we're unable to set the memory swap value (it stays as `-1` no matter what value I try to set it to, I tried some stuff) which from the docs seems to indicate that we'd get 2x the memory we're setting, however now that I've got memory swappiness set to 0 instead of -1/null, when I try to allocate 130MB I can consistently get containers to get oom killed on 128MB functions, which would seem to indicate that we're not getting double the amount of ram via swap as the docs would indicate.

this is a bit confusing, but it seems to plug the hole. i'm assuming our prod nodes disable swap, so this isn't as much of an issue there, however we should plug it up, not least for tests. I'm going to see if I can patch up fsouza to get something working before trying to piecemeal us into the docker/docker bindings, I'd rather the latter be an all or nothing thing instead of us driving around with 2 docker drivers for various operations (though, it works).

rdallman: I managed to get it working with fsouza, I had to fiddle with the go mod stuff to build with the modified vendored version of the package I was testing. 

I've made this thing https://github.com/fsouza/go-dockerclient/pull/776 and can hopefully get merged upstream and then we can update our version of this thing and close this out.

I'll get to work on moving us over to docker/docker bindings anyway, it may take some time, however.

hatchan: @rdallman Thanks for all the investigations and the fix :) that does sound like a better solution than this PR. I'll checkout if these changes will result in better succeeding unit tests for me.

hatchan: Forgot to mention this yesterday; by changing the swappiness of my host machine I was able to run the tests successfully.

rdallman: ah, useful to know, thanks.

PR #1479, opened 2019-04-19
default user logs to send to fn stderr at INFO level

### LINKED ISSUE #1328: race: funcLogger write race
this could be a write-write race when using a func logger as both stdout and stderr, a write-write race when closing a func logger when it may still be writing lines or a write-read race from the same close method with a different configuration. 

the symptom is a panic at this line https://github.com/fnproject/fn/blob/master/api/agent/func_logger.go#L154 when calling Next() to advance the inner buffer, the slice bounds is out of range. we need to fix this, logging has been disabled for the oracle functions service so it's good there, but this is still hanging out if configured, mostly because of the log api, which we also need to remove really.

### LINKED ISSUE #1474: no function logs by default
since we've accomplished #1055 now and the logs API is gone, there is no default logging story for functions for users that run `fn start`. this is kinda bad.

there are 2 ways to get logs now, neither of which are defaulted to:

* add a syslog url to an app
* set `FN_LOG_LEVEL=debug` / `fn start --log-level=debug`

one solution is that we could provide additional configuration to turn off the logs and allow setting the container log level, possibly defaulting to out of the box using 'info' level, but allow disabling this / setting to debug level. then users could see the logs from the `fn start` docker container or from their terminal in attached mode. these are not great, but a definite improvement over nothing. and having them available via `docker logs` allows running easy grep commands against call ids, etc. 

`fn start` itself is a dind container, so we could do "interesting" things like leaving logs laying around in there somewhere. other ideas are welcome here.

rdallman:
closes #1474 

see linked issue above for some details of the saga. this patch does a number of things in order to accomplish the goal of sending user logs to fn stderr at INFO level, while allowing it to easily be configured to off. as pure_runner is explicitly setting these to off for the moment, that configuration is not affected by this patch, however in the future it's possible we may want to change these to align so that lb/runner configs can get logs out of user containers in this manner. this only affects the full node (not just full agent) configuration, where calls are created in the invoke/httptrigger handlers from OSS fn. open to discussion on default behavior, however I think we have to have logs by default in some capacity for OSS, absolutely want to make it not brittle and easily disabled for production services. attempt to cover surface area here:

* moves the logger configuration up to a CallOpt instead of getting defaulted to in GetCall, now GetCall defaults to configuring logs to off, which seems much less brittle for other packages that call into agent and may absolutely not want logs (they exist). now our handlers in OSS configure this setting. let the record show, I'm not perfectly content with some of the munging that had to go on here (there's agent env vars to configure this option which is called from the front end, which needs the config in call opts), however after trying a few things this seems the most reasonable but I am very much open to concrete ideas - this was a particular pain point.
* changed stderr to io.WriteCloser as there's no longer a need to buffer logs since we are not uploading them anywhere, modifies our func logger heavily to remove all the old cruft and adds the ability to configure log writer with a level
* adds ability to configure FN_USER_LOG_LEVEL (i'll update docs) which defaults to 'info'. the level at which this is defaulted to info is open for discussion. removes unused FN_MAX_LOG_SIZE_BYTES
* updates logrus to 1.4.1 for `logger.Log(level, stuff)` method
* deleted a lot of old TODOs from yours truly
* deleted StdErr method from RunnerCall interface, as far as I can tell in all our repos here and in the service, this method is uncalled (can remove at will, since any implementers it's unused and still satisfies the interface)
* fixes #1328 issue by using buffer.ReadBytes('\n')
* fixed bug where we'd log between containers' logs at debug level when logs are configured off
* fixes a couple tests that fail for me now on docker 18.09.5 w/ new 404 message

tested and she seems to work as expected


api/agent/func_logger.go
```diff
@@ -136,29 +103,35 @@ func newLineWriterWithBuffer(b *bytes.Buffer, w io.WriteCloser) io.WriteCloser {
...
 	li.b.Write(ogb) // bytes.Buffer is guaranteed, read it!
 
+	var n int
 	for {
-		b := li.b.Bytes()
-		i := bytes.IndexByte(b, '\n')
-		if i < 0 {
-			break // no more newlines in buffer
+		// read the line and advance buffer past it
+		l, err := li.b.ReadBytes('\n')
```
skinowski: unintended buffer consumption here if the buffer doesn't contain '\n', Close() will miss these bytes in that case.

rdallman: thanks, added test to cover this.

api/agent/func_logger.go
```diff
@@ -148,8 +120,9 @@ func (li *lineWriter) Write(ogb []byte) (int, error) {
 		// write in this line and advance buffer past it
 		l := b[:i+1]
 		ns, err := li.w.Write(l)
+		n += ns
```
skinowski: Write() should return "how much of the original buffer was consumed". Looks like existing bug... In other words, in this state, Write() may return n, which can be larger than len(ogb) which is not really what we want to return.. I think we might want to change a few things such as:

- li.b buffer should be limited. Currently, series of Write() calls without a newline can grow li.b.buffer without limits with each Write() having to parse entire buf for newlines.
- if buffer is at limit (or going to grow over limit with current call to Write()), then we should flush li.b with appended newline ourselves before processing ogb. If that flush fails, then we can return error.

api/agent/func_logger.go
```diff
@@ -148,8 +120,9 @@ func (li *lineWriter) Write(ogb []byte) (int, error) {
...
 		if err != nil {
-			return ns, err
+			return n, err
 		}
 		li.b.Next(len(l))
```
skinowski: this also looks odd. if li.w.Write() return error with partial write, then we do need to execute li.b.Next(ns) regardless of error, no?

api/agent/func_logger.go
rdallman: i guess, in practice, the reality of this is that the inner writer is never going to error (it's a logWriter, which never errors) and writing to/reading from the buffer is never going to error either (we're not setting a capacity on buffers, and many methods of the buffer contract will never error - the runtime will oom before an error gets returned, but this is a panic). it's also the case that this is not to be used for production applications and we can avoid thinking about some of the more fun attack vectors here, but would be happy to note them - I just don't think we need to add any more arbitrary configuration than we already have for something that has a configuration to just turn it off for prod apps (glad to fix correctness issues, of course).

that being said, appreciate the feedback can spend an hour here tidying this up so that it's more obviously correct.

PR #1480, opened 2019-04-22
Fix swappiness

rdallman:
closes #1473 

without this, is vm.swappiness is not turned off, then it allows containers to
use unlimited (up to host max) anonymous swap space on the host. this behavior
is not what we want, particularly for getting tests to pass on everybody's dev
environment, but we don't really want this behavior in any case.

see https://github.com/fsouza/go-dockerclient/pull/776 for 'why fsouza needed to be updated' - this will affect https://github.com/fnproject/fn/pull/1477 as well (it is easy to update, just change pid limit to a pointer, so I'm not sure order matters).

pretty small and straightforward change though


hatchan:
I merged #1477 already. So you need to rebase on master and change the value to a pointer  in `api/agent/drivers/docker/cookie.go`. (Just noticed I never approved this PR, sorry)

rdallman:
thanks!

PR #1481, opened 2019-04-23
remove unused extensions cruft

rdallman:
* changes v1 to v2 for adding an endpoint
* removed the handler funcs for adding handlers onto eg /apps/:app_id/x, we
don't have them for funcs or triggers, and they honestly seem useless as it's
easy to build it with the ability to add a handler and to access the fnext
datastore which we offer, as they were they are really expensive since they
yank the app out of the db even if the operation may not even need it in that
handler. so instead of adding for the rest, remove all of these. none of our
example extensions, which aren't working at the moment, use these either
(that's why I'm here anyway).
* removes dead code helpers and references to app name url param which is no
longer a thing. these were just hanging around bugging me when I ran into
them, so killing them..

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,11 +54,11 @@ fn-test-volume:
 	cd images/fn-test-volume && ./build.sh
 
 .PHONY: test-middleware
-test-middleware: test-basic
+test-middleware:
 	cd examples/middleware && go build
 
 .PHONY: test-extensions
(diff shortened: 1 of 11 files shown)


rdallman:
bump @denismakogon could you review please? nothing is using (not in service), and I'm going to update the extension examples which are already broken with this change.

PR #1482, opened 2019-04-24
fix init logging

rdallman:
we're using json as the log format, and there were various issues in startup
where when setting FN_LOG_FORMAT, a few messages would not be in json format.
this fixes those cases in this repo, but may not for any packages that extend
fn due to initialization order of go (in particular, packages that log
anything in init() may or may not come before this is set, we have little
control over this except for playing with package names, yes, it's mad)

* we were logging a giant ascii fn. i guess this is cool but it doesn't really
matter, just moved it into a logrus line so that we don't have to fight over
it. call me a peace maker
* we were setting the log level/format/destination when initializing the
server, but many packages initialize themselves before calling server.New, and
those may log (particularly if they fail, which we want...)
* particularly hairy were our registration of the various dbs and container
drivers, since these logged in init(). I tried to hack around this but did not
have any luck. I'm proposing that we just get rid of these lines, there is no
easy fix here as setting to debug will result in them not getting logged
either, since we can't set debug level until after they've been logged. users
that attempt to use any of the drivers or dbs without registering them will
still see an error message, so this should be okay overall, we just basically
lose the warning that we may have accidentally pulled in one we don't need.

### DIFF
--- a/api/common/env.go
+++ b/api/common/env.go
@@ -1,0 +1,74 @@
+package common
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+// GetEnv looks up a key under its name in env or name+_FILE to read the value
+// from a file. fallback will be defaulted to if a value is not found.
+func GetEnv(key, fallback string) string {
+	if value, ok := os.LookupEnv(key); ok {
+		return value
+	} else if value, ok := os.LookupEnv(key + "_FILE"); ok {
+		dat, err := ioutil.ReadFile(filepath.Clean(value))
+		if err == nil {
+			return string(dat)
+		}
+	}
+	return fallback
+}
+
+// GetEnvInt looks up a key under its name in env or name+_FILE to read the
+// value from a file. fallback will be defaulted to if a value is not found.
+func GetEnvInt(key string, fallback int) int {
+	if value, ok := os.LookupEnv(key); ok {
+		// linter liked this better than if/else
+		var err error
+		var i int
+		if i, err = strconv.Atoi(value); err != nil {
+			logrus.WithError(err).WithFields(logrus.Fields{"string": value, "environment_key": key}).Fatal("Failed to convert string to int")
+		}
+		return i
+	} else if value, ok := os.LookupEnv(key + "_FILE"); ok {
+		dat, err := ioutil.ReadFile(filepath.Clean(value))
+		if err == nil {
+			var err error
+			var i int
+			if i, err = strconv.Atoi(strings.TrimSpace(string(dat))); err != nil {
+				logrus.WithError(err).WithFields(logrus.Fields{"string": dat, "environment_key": key}).Fatal("Failed to convert string to int")
+			}
+			return i
+		}
+	}
+	return fallback
+}
+
+// GetEnvDuration looks up a key under its name in env or name+_FILE to read
+// the value from a file. fallback will be defaulted to if a value is not
+// found. if an integer is provided, the value will be returned in seconds
+// (value * time.Second)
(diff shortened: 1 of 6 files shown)


rdallman:
tested and looks good now:

```
$ FN_LOG_FORMAT=json ./fnserver
{"fields.level":"info","level":"info","msg":"Setting log level to","time":"2019-04-19T05:33:35.422565001-07:00"}
{"level":"info","msg":"Registering data store provider 'sql'","time":"2019-04-19T05:33:35.422931453-07:00"}
{"level":"info","msg":"Connecting to DB","time":"2019-04-19T05:33:35.423478075-07:00","url":"sqlite3:///home/reed/dev/gocode/src/github.com/fnproject/fn/data/fn.db"}
{"datastore":"sqlite3","level":"info","max_idle_connections":256,"msg":"datastore dialed","time":"2019-04-19T05:33:35.423886251-07:00","url":"sqlite3:///home/reed/dev/gocode/src/github.com/fnproject/fn/data/fn.db"}
{"level":"info","msg":"agent starting cfg={MinDockerVersion:17.10.0-ce ContainerLabelTag: DockerNetworks: DockerLoadFile: FreezeIdle:50ms HotPoll:200ms HotLauncherTimeout:1h0m0s HotPullTimeout:10m0s HotStartTimeout:5s DetachedHeadRoom:6m0
s MaxResponseSize:0 MaxHdrResponseSize:0 MaxLogSize:1048576 MaxTotalCPU:0 MaxTotalMemory:0 MaxFsSize:0 MaxPIDs:0 PreForkPoolSize:0 PreForkImage:busybox PreForkCmd:tail -f /dev/null PreForkUseOnce:0 PreForkNetworks: EnableNBResourceTracker
:false MaxTmpFsInodes:0 DisableReadOnlyRootFs:false DisableDebugUserLogs:false IOFSEnableTmpfs:false IOFSAgentPath: IOFSMountRoot: IOFSOpts: ImageCleanMaxSize:0 ImageCleanExemptTags: ImageEnableVolume:false}","time":"2019-04-19T05:33:35.4
24247253-07:00"}                   
{"cgroup_limit":9223372036854771712,"head_room":325128192,"level":"info","msg":"available memory","time":"2019-04-19T05:33:35.452878583-07:00","total_memory":3251281920}
{"avail_memory":2926153728,"level":"info","msg":"ram reservations","time":"2019-04-19T05:33:35.452934456-07:00"}
{"avail_cpu":2000,"level":"info","msg":"available cpu","time":"2019-04-19T05:33:35.453059583-07:00","total_cpu":2000}
{"cpu":2000,"level":"info","msg":"cpu reservations","time":"2019-04-19T05:33:35.453072463-07:00"}
{"level":"info","msg":"\n        ______\n       / ____/___\n      / /_  / __ \\\n     / __/ / / / /\n    /_/   /_/ /_/\n","time":"2019-04-19T05:33:35.453975873-07:00"}
{"level":"info","msg":"Fn serving on `:8080`","time":"2019-04-19T05:33:35.454014047-07:00","type":0,"version":"0.3.697"}
^C{"level":"info","msg":"Halting...","time":"2019-04-19T05:33:38.457870485-07:00"}
```

master had the cases mentioned in the OP

api/common/env.go
```diff
@@ -0,0 +1,74 @@
...
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+// GetEnv looks up a key under its name in env or name+_FILE to read the value
+// from a file. fallback will be defaulted to if a value is not found.
+func GetEnv(key, fallback string) string {
```
rdallman: we've declared these in a few places now, this is from fiddling, but gonna try to drive this wedge in here ;)

I'll go through in another PR and replace all usage with these.

rdallman:
ty

PR #1486, opened 2019-04-25
make detached writer concurrency safe

### LINKED ISSUE #1484: racy test
```
=== RUN   TestBasicConcurrentDetachedExecution
fatal error: concurrent map read and map write

goroutine 1182 [running]:
runtime.throw(0x1006393, 0x21)
	/usr/local/go/src/runtime/panic.go:617 +0x72 fp=0xc00379f8b0 sp=0xc00379f880 pc=0x439e12
runtime.mapaccess1_faststr(0xef6880, 0xc001b3ea80, 0xff1dd1, 0xe, 0x0)
	/usr/local/go/src/runtime/map_faststr.go:21 +0x469 fp=0xc00379f920 sp=0xc00379f8b0 pc=0x41e9f9
net/textproto.MIMEHeader.Get(0xc001b3ea80, 0xff1dd1, 0xe, 0x0, 0x30)
	/usr/local/go/src/net/textproto/header.go:35 +0x5d fp=0xc00379f958 sp=0xc00379f920 pc=0x6a8add
net/http.Header.Get(...)
	/usr/local/go/src/net/http/header.go:47
github.com/fnproject/fn/api/agent.logCallFinish(0x118be20, 0xc003790780, 0xc0036a4aa8, 0xc001b3ea80, 0xc0000000ca)
	/home/circleci/go/src/github.com/fnproject/fn/api/agent/runner_client.go:446 +0xa8 fp=0xc00379f9e8 sp=0xc00379f958 pc=0xbcb9e8
github.com/fnproject/fn/api/agent.receiveFromRunner(0x1174280, 0xc00372be30, 0x1183ba0, 0xc003701fb0, 0xff1c81, 0xe, 0x1183ae0, 0xc003770120, 0xc003790300)
	/home/circleci/go/src/github.com/fnproject/fn/api/agent/runner_client.go:403 +0x64e fp=0xc00379ff98 sp=0xc00379f9e8 pc=0xbcafde
runtime.goexit()
	/usr/local/go/src/runtime/asm_amd64.s:1337 +0x1 fp=0xc00379ffa0 sp=0xc00379ff98 pc=0x4695c1
created by github.com/fnproject/fn/api/agent.(*gRPCRunner).TryExec
	/home/circleci/go/src/github.com/fnproject/fn/api/agent/runner_client.go:217 +0xbbd
```

https://circleci.com/gh/fnproject/fn/5549 while it lasts

I can't figure out if it's a race in tests or a race in the code. it looks awfully suspicious since it's detached and we write the headers early on that one, but the race appears in the lb, which seems funky.

rdallman:
in detached mode we're handing off execution but not interested in the headers
or body that the function writes at all. at present however, we're handing
over the original response headers for the detached writer to use, which could
happen concurrently when we're writing headers back to the client response.
since we're not interested in them, this just stops passing them from the
client response headers down into the agent for the func to write to, so there
should no longer be a race with the front end at least. the other piece was
that writing the status code in and reading it out is also not safe, exactly,
so made this explicitly safe - the case here would be that the function writes
a response code while we're going to write the response code in the front end.

I am not sure whether I like the way we're doing this or not, overall (prior
to the modifications here even). open to ideas.

updated the makefile/system test script so that I could run these faster to
repro, pretty handy, should add to other stuff too...

closes #1484

### DIFF
--- a/api/agent/lb_agent.go
+++ b/api/agent/lb_agent.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"sync/atomic"
 	"time"
 
 	"github.com/sirupsen/logrus"
@@ -28,14 +29,18 @@ type lbAgent struct {
 	callOpts      []CallOpt
 }
 
+// DetachedResponseWriter is a concurrency safe http.ResponseWriter which
+// implements http.ResponseWriter interface, it is safe for concurrent calls to
+// WriteHeader, Write, Status, while its returned headers are not safe for
+// concurrent use across threads.
 type DetachedResponseWriter struct {
-	Headers http.Header
-	status  int
+	headers http.Header
+	status  uint32
 	acked   chan struct{}
 }
 
 func (w *DetachedResponseWriter) Header() http.Header {
-	return w.Headers
+	return w.headers
 }
 
 func (w *DetachedResponseWriter) Write(data []byte) (int, error) {
@@ -43,18 +48,18 @@ func (w *DetachedResponseWriter) Write(data []byte) (int, error) {
 }
 
 func (w *DetachedResponseWriter) WriteHeader(statusCode int) {
-	w.status = statusCode
+	atomic.StoreUint32(&w.status, uint32(statusCode))
 	w.acked <- struct{}{}
 }
 
 func (w *DetachedResponseWriter) Status() int {
-	return w.status
+	return int(atomic.LoadUint32(&w.status))
 }
 
-func NewDetachedResponseWriter(h http.Header, statusCode int) *DetachedResponseWriter {
+func NewDetachedResponseWriter(statusCode int) *DetachedResponseWriter {
 	return &DetachedResponseWriter{
-		Headers: h,
-		status:  statusCode,
+		headers: make(http.Header), // TODO allocation waste, this is unused but need to satisfy interface
+		status:  uint32(statusCode),
 		acked:   make(chan struct{}, 1),
 	}
 }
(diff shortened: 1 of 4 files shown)


rdallman:
I do think it's possible that we need to add an additional layer in the agent right around https://github.com/fnproject/fn/blob/master/api/agent/agent.go#L684 due to how this works, since we can't pass the context down into the io.Copy there, we're wrapping the whole thing, but the caller could return before writeResp is done and we wind up writing while the front end is reading. that whole situation is frustrating, I decided not to go at it here, it will take a bit of fudging with.

I maintain that we should do this change regardless of changing agent, and that this one does likely close the bug since it's from detached. particularly the part about the headers in the detached writer, at least.

api/agent/lb_agent.go
```diff
@@ -28,33 +29,37 @@ type lbAgent struct {
...
-	return w.status
+	return int(atomic.LoadUint32(&w.status))
 }
 
-func NewDetachedResponseWriter(h http.Header, statusCode int) *DetachedResponseWriter {
+func NewDetachedResponseWriter(statusCode int) *DetachedResponseWriter {
 	return &DetachedResponseWriter{
-		Headers: h,
-		status:  statusCode,
+		headers: make(http.Header), // TODO allocation waste, this is unused but need to satisfy interface
```
rdallman: i guess this comment is confusing. basically, the headers get written to but are never really read by the front end. the lb agent does end up using them though to log some of the headers, so leaving this. it does seem wasteful, happy to remove comment though if it only adds confusion.

rdallman:
gonna WIP this, started adding more crap

PR #1489, opened 2019-04-26
strip transport headers from function response

### LINKED ISSUE #1488: client gets connection, upgrade, etc headers from function response
there are some transport headers we should strip when copying them from the function response to the client response. we're doing this for triggers, but we need to do it for invoke as well, so we likely just need to push some of that down into the agent. we may also be copying down the connection headers from the client request to the container as well, which we need to fix up.

rdallman:
we weren't stripping transport headers from function response before copying
to the client response. this means e.g. a function could send a "Connection:
close" when fn wants to use a keep alive.

this also moved this to common since we're doing it in 3 places. I know that
it's not the ideal ideal big O performance and we can pedant about it, but
we're talking about 7 things here. I'm open to ideas about what this should
look like, I ripped it out of the stdlib and it seems okay. changed the
trigger stuff to use this too, as it was copied. added tests to cover all 3 of
these spots.

the only behavior change for existing behavior is no longer stripping 'authorization' header when
passing headers from the client request into the function. authorization seems
like a useful header for users that want to use auth in their functions, I
think if we need to use this one in the backend we should more carefully strip
out the parts we use before passing them on instead of relying on the agent to
strip the header to not leak anything into a function.

closes #1488

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -605,16 +605,6 @@ func (s *hotSlot) exec(ctx context.Context, call *call) error {
 	return err
 }
 
-var removeHeaders = map[string]bool{
-	"connection":        true,
-	"keep-alive":        true,
-	"trailer":           true,
-	"transfer-encoding": true,
-	"te":                true,
-	"upgrade":           true,
-	"authorization":     true,
-}
-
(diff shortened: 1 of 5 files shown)


rdallman:
thanks!

PR #1490, opened 2019-04-27
safe responsewriter usage in TryExec

### LINKED ISSUE #1484: racy test

rdallman:
inside of TryExec we were writing directly to the response writer inside of a
goroutine, but TryExec can timeout and then get called again to a different
runner or even have the front end writing headers while TryExec is writing
headers.

one way to make this safe is to make a new response writer for TryExec to
write the response into, and only after the goroutine handling the response
has returned, from the TryExec goroutine we can copy the response back up as
the caller will not call TryExec again until it has returned (this is
seemingly part of the placer contract). unfortunately, we're already buffering
the response writer in the front end, too - it's possible we can get rid of
that but it may need further testing.

this adds an optimization when copying the request body from the LB to a
runner, since we're using request.GetBody() and returning a reader we
are familiar with that happens to just wrap a buffer's bytes (which we just
need multiple readers on, but the data doesn't change). anyway, this whole
interaction is unfortunate but kind of necessary due to needing to maneuver
into a protobuf, it seems like a worth it and somewhat ok abstraction wise
optimization.

additionally, this gets rid of passing the client response headers down into
the agent for detached functions. we don't need these since detached functions
are not responding with the functions response to the client, only a 202, this
was leading to races around writing the headers in retries too, but this is
just for posterity/correctness now.

updated the makefile/system test script so that I could run these faster to
repro, pretty handy, should add to other stuff too...

closes #1484

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,9 +70,9 @@ test: checkfmt pull-images test-basic test-middleware test-extensions test-syste
 
 .PHONY: test-system
 test-system:
-	./system_test.sh sqlite3
-	./system_test.sh mysql
-	./system_test.sh postgres
+	./system_test.sh sqlite3 $(run)
+	./system_test.sh mysql $(run)
+	./system_test.sh postgres $(run)
 
 .PHONY: img-busybox
 img-busybox:
--- a/api/agent/lb_agent.go
+++ b/api/agent/lb_agent.go
@@ -28,14 +28,18 @@ type lbAgent struct {
 	callOpts      []CallOpt
 }
 
+// DetachedResponseWriter implements http.ResponseWriter without allowing
+// writes to the body or writing the headers from a call to Write or
+// WriteHeader, it is only intended to allow writing the status code in and
+// being able to fetch it later from Status()
 type DetachedResponseWriter struct {
-	Headers http.Header
+	headers http.Header
 	status  int
 	acked   chan struct{}
 }
 
 func (w *DetachedResponseWriter) Header() http.Header {
-	return w.Headers
+	return w.headers
 }
 
 func (w *DetachedResponseWriter) Write(data []byte) (int, error) {
@@ -51,9 +55,9 @@ func (w *DetachedResponseWriter) Status() int {
 	return w.status
 }
 
-func NewDetachedResponseWriter(h http.Header, statusCode int) *DetachedResponseWriter {
+func NewDetachedResponseWriter(statusCode int) *DetachedResponseWriter {
 	return &DetachedResponseWriter{
-		Headers: h,
+		headers: make(http.Header),
 		status:  statusCode,
 		acked:   make(chan struct{}, 1),
 	}
--- a/api/agent/runner_client.go
+++ b/api/agent/runner_client.go
@@ -1,6 +1,7 @@
 package agent
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"encoding/hex"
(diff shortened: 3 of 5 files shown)


reclaro:
The description of the change is clear and seems a cool thing to have but then reading the implementation I am having some troubles.
 Some questions:
> even have the front end writing headers while TryExec is writing headers.
Is this change addresses this issue as well? Reading the code seems to me that this case (having the front end writing headers) is still possible

The change to the sentToRunner function is about the optimisation for copying RequestBody, right? If so shall it go into a different change? I am not a great fan of merging more changes in a single PR and it confuses me in reviewing as I have to change context, probably it is just me being an old man  so if you don't want to change that's fine.

rdallman:
> even have the front end writing headers while TryExec is writing headers.
Is this change addresses this issue as well? Reading the code seems to me that this case (having the front end writing headers) is still possible

I should have been more clear here. TryExec and the front end will not compete for writing headers anymore as the [client] headers (well, buffered client headers, or the front end response writer, however we want to look at it) are no longer written to from a goroutine in TryExec (whose lifetime may exceed TryExec's lifetime). the front end will only write headers after TryExec has returned, now that we're writing headers to the front end's response writer inside of TryExec and not in a goroutine, this operation is safe as long as the contract there holds (namely, that TryExec is waited on by its calling routine, or at least when submit returns there are no outstanding TryExec's lingering in other goroutines).

> The change to the sentToRunner function is about the optimisation for copying RequestBody, right?

I can move this, no problem. try to slip in some cleanup where I can

rdallman:
> The change to the sentToRunner function is about the optimisation for copying RequestBody, right?

this is removed from this PR

rdallman:
thanks!

PR #1492, opened 2019-04-29
tidy up release file

rdallman:
this file was pushing to deprecated refs we don't really need anymore. we
started using the vX.Y.Z to make mod happy and that's widely tracked in VCS
now, whereas I doubt the naked versions are tracked anywhere without something
analogous existing (we've been on mod for some time now).

fnproject/functions was deprecated, and I was only able to find a couple of
stale references (some merely comments). I suspect we can safely remove this.

reduced git pushes from 5 to 1 (woo)
reduced docker pushes from 4 to 1 (yay)

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -3,7 +3,6 @@ set -exuo pipefail
 
 user="fnproject"
 image="fnserver"
-image_deprecated="functions"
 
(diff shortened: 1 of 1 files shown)


rdallman:
🤞

PR #1494, opened 2019-05-03
Log the root error in HandleFindRunnersFailure, when available

ostrain:


api/runnerpool/placer_tracker.go
```diff
@@ -43,7 +43,12 @@ func (tr *placerTracker) IsDone() bool {
 
 // HandleFindRunnersFailure is a convenience function to record error from runnerpool.Runners()
 func (tr *placerTracker) HandleFindRunnersFailure(err error) {
-	common.Logger(tr.requestCtx).WithError(err).Error("Failed to find runners for call")
+	logger := common.Logger(tr.requestCtx).WithError(err)
+	w, ok := err.(models.APIErrorWrapper)
+	if ok {
+		logger = logger.WithField("root_error", w.RootError())
+	}
+	logger.Error("Failed to find runners for call")
```
rdallman: I have an inkling that this is the case for pretty much any site where we log an APIErrorWrapper, it makes me wonder if we should implement the Error method to print the root error (for anywhere we log) and have another method to get the 'user' error for the front end to spit out as a response (kind of inverting the APIError contract). with that, this would fall out. I just have a feeling this is going to happen all over and it seems like we should avoid that, if not then ok.

ostrain: That's not a bad idea, but it feels riskier to me... I don't feel confident that we wouldn't accidentally end up emitting messages intended just for our logs to the user in some cases.

rdallman: maybe, we do have pretty tight control over the routine(s) that turns APIError into actual HTTP responses, at least. now that we have 3 different front ends though (CP,LB,OSS), I kind of bemoan digging through them as well, to figure out if we're incidentally writing any errors in any other bespoke way (I think all 3 have their own error handling logic) and/or go update them all for the flipped logic of the wrapped.

my only other clever-ish idea is something akin to https://github.com/pkg/errors/blob/master/stack.go#L64 - may require digging into logrus internals. it's tempting to change usage to something like that which is more 'baked', but haven't analyzed extensively our solution vs pkg/errors.

also not sure how wide of an issue this is/will be

PR #1495, opened 2019-05-03
Add ability to limit ulimits

hatchan:
Add the ability to limit the following ulimits: 

- Open files: the max number of files that the process in a function is allowed to open
- Locked memory: maximum number of bytes of memory that may be locked into RAM
- Pending signals: limit on the number of signals that may be queued
- Message queue: limit on the number of bytes that can be allocated for POSIX message queues

Similar to the Limit PID PR: https://github.com/fnproject/fn/pull/1477


api/agent/drivers/docker/cookie.go
```diff
@@ -303,6 +353,17 @@ func (c *cookie) configureSecurity(log logrus.FieldLogger) {
 		"CapDrop": c.opts.HostConfig.CapDrop, "SecurityOpt": c.opts.HostConfig.SecurityOpt, "call_id": c.task.Id()}).Debug("setting security")
 }
 
+// addULimit adds lim to the docker host config ulimits slice, optionally
+// creating the slice if it isn't created yet.
+func (c *cookie) addULimit(lim docker.ULimit) {
+	limits := c.opts.HostConfig.Ulimits
+	if limits == nil {
+		limits = []docker.ULimit{}
+	}
```
rdallman: nit: this block (nil check) is unnecessary, appending to a nil slice is the same as appending to an empty, zero capacity, non-nil slice. I think this renders the `addULimit` method basically unnecessary, as each call site can simply be the one liner `c.opts.HostConfig.Ulimits = append(c.opts.HostConfig.Ulimits, limit)` rather than `addULimit`, but up to you - at least, this method can be a one liner.

api/agent/drivers/docker/cookie.go
```diff
@@ -140,6 +140,56 @@ func (c *cookie) configurePIDs(log logrus.FieldLogger) {
 	c.opts.HostConfig.PidsLimit = &pids64
 }
 
+// configureOpenFiles will set the ULimit for `nofile` on the Docker container
+func (c *cookie) configureOpenFiles(log logrus.FieldLogger) {
```
kmjohansen: Potential idea for reducing the 4 all very similar functions.  What do you think about something like the following?

```
type ulimitSetter func()(*uint64, string)

func (c *cookie) configureUlimit(usf ulimitSetter, log logrus.FieldLogger) {
    ulimitValP, ulimitName := usf()
    if ulimitValP == nil {
        return
    }

    ulimit64 := int64(*ulimitValP)
    // bail out if ulimit is greater that MaxInt64
    if ulimit64 < 0 {
        return
    }

    log.WithFields(logrus.Fields{ulimitName: ulimit64, "call_id": c.task.Id()}).Debugf("setting %s", ulimitName)
    c.addULimit(docker.ULimit{Name: ulimitName, Soft: ulimit64, Hard: ulimit64})
} 
```

This does make the logging slightly more gross, but it means you can do a common set with 1/4th as much code.

If you decided to go this route, you'd also need to change the interfaces in `driver.go`, tweak the functions in `agent.go` to return the ulimit's name in addition to the value, and then  modify the config bits in `docker.go` to look something like:
```
cookie.configureUlimit(task.MessageQueue(), log)
```
As one example.

api/agent/drivers/docker/cookie.go
hatchan: Good point, does look better that way :+1:

api/agent/drivers/docker/cookie.go
hatchan: Didn't go with the setterFunc route but just a single method that gets called multiple times. Let me know what you think of that solution.

hatchan: https://github.com/fnproject/fn/pull/1495/commits/45dbdc66351d56af4629f738529dbf7384550336

hatchan:
I've added the ability to specify the default values (though the values are currently `nil`) and removed some repeated code.

api/agent/drivers/docker/cookie.go
kmjohansen: Thanks; like the change here.  I realized after I posted my comments that the setterFunc was just needless complication, so I'm glad you didn't do that part.

api/agent/config_test.go
```diff
@@ -0,0 +1,79 @@
+package agent
+
+import (
+	"errors"
+	"os"
+	"testing"
+)
+
+func Test_setEnvUintPointer(t *testing.T) {
```
rdallman: all camel case, please

api/agent/config_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+			if *tt.Expected != *val {
+				t.Fatalf("expected %d; got %d", *tt.Expected, *val)
+			}
+		})
+	}
+}
+
+// Test_setEnvUintPointer_error tests the use cases
+func Test_setEnvUintPointer_error(t *testing.T) {
```
rdallman: ditto

api/agent/config_test.go
hatchan: Changed both of them :+1:

hatchan:
Ok I've added the ability to add defaults to `setEnvUint` and added some defaults to the limits. If there are no more objections, then I will merge this tomorrow.

PR #1498, opened 2019-05-07
remove usage of time.After

rdallman:
these little guys leak in the background, it's easy enough to remove usage of
them altogether. this reduces cpu load on fn, the main culprit was the one in
waitHot which polls every 200ms. if you run a load test where functions get a
slot and exec a lot faster than 200ms then these things pile up kinda quickly.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -433,6 +433,9 @@ func (a *agent) checkLaunch(ctx context.Context, call *call, caller slotCaller) 
 	var notifyChans []chan struct{}
 	var tok ResourceToken
 
+	timer := time.NewTimer(a.cfg.HotPoll)
+	defer timer.Stop()
+
 	// WARNING: Tricky flow below. We are here because: isNewContainerNeeded is true,
 	// in other words, we need to launch a new container at this time due to high load.
 	//
@@ -452,7 +455,7 @@ func (a *agent) checkLaunch(ctx context.Context, call *call, caller slotCaller) 
 	// need to start a new container, then waiters will wait.
 	select {
 	case tok = <-a.resources.GetResourceToken(ctx, mem, call.CPUs, isNB):
-	case <-time.After(a.cfg.HotPoll):
+	case <-timer.C:
 		// Request routines are polling us with this a.cfg.HotPoll frequency. We can use this
 		// same timer to assume that we waited for cpu/mem long enough. Let's try to evict an
 		// idle container. We do this by submitting a non-blocking request and evicting required
@@ -520,7 +523,9 @@ func (a *agent) waitHot(ctx context.Context, call *call, caller *slotCaller) (Sl
 	// 1) if we can get a slot immediately, grab it.
 	// 2) if we don't, send a signaller every x msecs until we do.
 
-	sleep := 1 * time.Microsecond // pad, so time.After doesn't send immediately
+	timer := time.NewTimer(1 * time.Microsecond) // pad, so time.After doesn't send immediately
+	defer timer.Stop()
+
 	for {
 		select {
 		case err := <-caller.notify:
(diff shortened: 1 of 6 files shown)


rdallman:
thanks for catching that, it does seem like we need to change uses of reset to do the stop and reset, it's unfortunate that Reset doesn't do this but I guess that it could lead to reset hanging forever if another goroutine is receiving on the t.C channel concurrently - I didn't read the source for why, I think it might be related and not necessarily for timing correctness. I don't think we need to change the defer Stop usage since they're in the same thread in each instance as the resets? i'll update

kmjohansen:
Re adjusting the defer on `timer.Stop()`, I think you're correct.  It's not actually harmful for stop to be called multiple times on a timer that we're finished with.  So probably fine to leave the way it is.  Was overthinking that a bit.

rdallman:
@kmjohansen I've added a timer that gives us the behavior that I was expecting to begin with: https://github.com/fnproject/fn/pull/1498/commits/fefc264210ccd619d7b6c638dca6da94950c4341

that commit message contains links for reading, if interested, about why stdlib couldn't justify the change themselves. this seems like the most seamless solution, rather than polluting all of the call sites with the drain logic.

thanks again for catching, looks like docs were added in 1.8 about this and I completely missed this!

api/agent/agent.go
```diff
@@ -520,7 +523,9 @@ func (a *agent) waitHot(ctx context.Context, call *call, caller *slotCaller) (Sl
 	// 1) if we can get a slot immediately, grab it.
 	// 2) if we don't, send a signaller every x msecs until we do.
 
-	sleep := 1 * time.Microsecond // pad, so time.After doesn't send immediately
+	timer := common.NewTimer(1 * time.Microsecond) // pad, so time.After doesn't send immediately
```
kmjohansen: Nit: `time.After` is no longer used here, making this comment slightly out of date.

kmjohansen:
Apologies; catching up to where you left off.  I agree that `timer.Reset()` is confusing, and the bug reports filed against golang actually muddy the waters more than they clarify.  I spent some more time digging around, and have at least managed to convince myself about which docs are correct.  Let me see if I can also convince you.

```
            t1                                            t2
           ---                                           ---
                                                 #a ->  timer fires
	  t.Stop()
          select {
#1 ->     case <-t.C:
                                                 #b ->  t.C <- timer_event
          default:
          }
          return t.Timer.Reset()
```

The bad thing happens here if the timer fires in a), and then the select on `t.C` at #1 occurs before the timer event is actually queued on the channel in b).

The idiom proposed in the documentation works around this problem.
```
if !t.Stop() {
	<-t.C
}
```
It does so because if the timer has already expired then we are guaranteed delivery of the event in some eventual period of time, which a synchronous wait will catch.  The caveat, as you already observed, is that this doesn't account for any case where there's been another receiver on this channel.

The docs attempt to warn about this, but are a bit more roundabout than I would be under the circumstances:
>For example, assuming the program has not received from t.C already

Which, I think, gets to why there were so may exasperated programmers in the golang bug reports about this.  If there are concurrent receivers on this channel, then there needs to be a separate mechanism to track whether the timer event has been received somewhere, and that does get complicated quickly.

It seems like it would be safe to use the idiom in the documentation, provided that the code calling reset is also the only receiver on the channel. (And it has not already received an event).  If this is a situation where there's the potential for multiple readers/consumers on a single channel, then it may be easier to take the scorched earth approach and simply create a new timer instead of resetting the old one.

rdallman:
thanks for digging in, we're in agreement about behavior. as it currently is, it's racy. we either have to add a method for previous receivers of the token to notify the timer that the token has been received from the fired timer, e.g. `case <-t.C: t.Notify()` or we can add a parameter onto reset that says whether it's been received already or not, e.g. `t.Reset(dur, received)`; either way, then we can say `if !received && !t.Stop() { <-t.C }` safely. 

>  it may be easier to take the scorched earth approach and simply create a new timer instead of resetting the old one.

this is a really good idea, sadly, it might be a lot simpler - thanks. I'll tinker with this today. yea, I hope they don't forget to fix this in 2.0, reset is way too complicated to be useful.

rdallman:
@kmjohansen I think your idea to just switch in a new one worked out pretty well with a wrapped timer, small update https://github.com/fnproject/fn/pull/1498/commits/f13c43fa985857fc0bfcf6a83f0b1d115a75dbcf - by my reasoning, at least for our use cases, it seems to check out. (we need reset to stop the old timer and start it for the given duration, and then the added defers will stop the last assigned [inner] timer). lmk what you think

rdallman:
thanks!

PR #1506, opened 2019-06-06
fix docker security options

kppk:
This changes fixes the following logging message by docker:

`Security options with `:` as a separator are deprecated and will be completely unsupported in 17.04, use `=` instead.`

The message was coming from the error is coming from here:
https://github.com/docker/docker-ce/blob/2ba8f3d83c15d3cb018c80e724efe14e3726512f/components/engine/daemon/daemon_unix.go#L228

using security option as 
`c.opts.HostConfig.SecurityOpt = []string{"no-new-privileges"}`

removes the problem


kmjohansen:
Thanks for tracking this down and fixing.

PR #1512, opened 2019-06-18
Add ability to disable web and admin server

hatchan:
This is useful when you only want to interact with the gRPC endpoints. By default this will still enable both the web and admin server, but it is possible to use the `WithWebEnabled` and `WithAdminEnabled` to disable either one.

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -205,6 +205,8 @@ type Server struct {
 	noHTTTPTriggerEndpoint bool
 	noFnInvokeEndpoint     bool
 	noProfilerEndpoint     bool
+	noWebServer            bool
+	noAdminServer          bool
 	appListeners           *appListeners
 	fnListeners            *fnListeners
 	triggerListeners       *triggerListeners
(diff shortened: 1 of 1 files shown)


api/server/server.go
```diff
@@ -205,6 +205,8 @@ type Server struct {
 	noHTTTPTriggerEndpoint bool
 	noFnInvokeEndpoint     bool
 	noProfilerEndpoint     bool
+	noWebServer            bool
```
kmjohansen: Nit: I find sticking a `!` in front of a `noXXX` variable to be stylistically frustrating.  At least in English this comes out as a double negative.  If the predominant case for `noWebServer` and `noAdminServer` are not checks, can these be changed to something like `webServerEnabled` and `adminServerEnabled` so that you can either check `adminServerEnabled` or `!adminServerEnabled`?

hatchan: I agree with you, but I did go this route to ensure that backwards compatibility was maintained. If there is any place where the server is created, it will have the same behavior as before, due to the bool defaulting to `false`. 

For the code within this project it is not a issue to ensure that the bool will have the correct value set by default. It is however a exported struct, so it could be initialized as a struct in other code.

Let me know if you think having the better names is worth breaking with backwards compatibility.

kmjohansen: It's a fine point, and one that I had neglected to consider.  I'd also prefer not to break people needlessly, even if it does frustrate my OCD.

Issue #1517, opened 2019-06-27
Duplicate agent_submit spans in agent and lb_agent


bharadwr:
There's two spans with the same name, namely in agent.go and lb_agent.go. 

https://github.com/fnproject/fn/blob/4ccd8fa0c63e65aa4b35c1f76cbd7db919f59c7c/api/agent/agent.go#L250
https://github.com/fnproject/fn/blob/1fb78ed8362693105aee48a506129e6776d1a3a8/api/agent/lb_agent.go#L200

Should this be renamed? Someone might find a span on jaeger and go down the wrong rabbit hole if theyre debugging (I certainly did)

rdallman:
yes, please - lb_agent likely has a few duplicates from agent, `lb_agent_submit` is probably sufficient?

bharadwr:
Yep, I'll make that change. Lemme see if there's a few more duplicate traces, and I'll add them all to one ticket.

bharadwr:
![Screen Shot 2019-06-27 at 11 00 22 AM](https://user-images.githubusercontent.com/25177182/60289336-03243b80-98cb-11e9-805c-2bdd3cf5a3f6.png)
It's gonna look like this, added some more span attributes that are probably useful

PR #1518, opened 2019-06-27
Changing lb_agent span name, attrib (#1517)

bharadwr:
https://github.com/fnproject/fn/issues/1517

Added fn appid and fnid (needed for internal tracing), and corrected span name in lb_agent.go to avoid duplicate span conflict with span in agent.go
https://github.com/fnproject/fn/blob/1fb78ed8362693105aee48a506129e6776d1a3a8/api/agent/lb_agent.go#L200
https://github.com/fnproject/fn/blob/4ccd8fa0c63e65aa4b35c1f76cbd7db919f59c7c/api/agent/agent.go#L250

It's gonna look like this
Screen Shot 2019-06-27 at 11 00 22 AM

Changing duplicate span name

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -258,7 +258,7 @@ func (a *agent) Submit(callI Call) error {
 	rid := common.RequestIDFromContext(ctx)
 	if rid != "" {
 		span.AddAttributes(
-			trace.StringAttribute("fn.rid", common.RequestIDFromContext(ctx)),
+			trace.StringAttribute("fn.rid", rid),
 		)
 	}
 
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -250,7 +250,14 @@ func (a *agent) Submit(callI Call) error {
 	ctx, span := trace.StartSpan(ctx, "agent_submit")
 	defer span.End()
 
-	span.AddAttributes(trace.StringAttribute("agent.call_id", call.ID))
+	span.AddAttributes(
+		trace.StringAttribute("fn.id", call.ID),
+		trace.StringAttribute("fn.app_id", call.AppID),
+		trace.StringAttribute("fn.fn_id", call.FnID),
+		trace.StringAttribute("fn.app_name", call.AppName),
+		trace.StringAttribute("fn.fn_id", call.FnID),
+		trace.StringAttribute("fn.rid", common.RequestIDFromContext(ctx)),
+	)
 
 	return a.submit(ctx, call)
 }
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -251,13 +251,16 @@ func (a *agent) Submit(callI Call) error {
 	defer span.End()
 
 	span.AddAttributes(
-		trace.StringAttribute("fn.id", call.ID),
+		trace.StringAttribute("fn.call_id", call.ID),
 		trace.StringAttribute("fn.app_id", call.AppID),
 		trace.StringAttribute("fn.fn_id", call.FnID),
-		trace.StringAttribute("fn.app_name", call.AppName),
-		trace.StringAttribute("fn.fn_id", call.FnID),
-		trace.StringAttribute("fn.rid", common.RequestIDFromContext(ctx)),
 	)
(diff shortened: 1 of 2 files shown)


bharadwr:
Hold off on merge, need to add a couple other things to span attribs

api/agent/agent.go
```diff
@@ -250,7 +250,14 @@ func (a *agent) Submit(callI Call) error {
...
 	defer span.End()
 
-	span.AddAttributes(trace.StringAttribute("agent.call_id", call.ID))
+	span.AddAttributes(
+		trace.StringAttribute("fn.id", call.ID),
+		trace.StringAttribute("fn.app_id", call.AppID),
+		trace.StringAttribute("fn.fn_id", call.FnID),
+		trace.StringAttribute("fn.app_name", call.AppName),
+		trace.StringAttribute("fn.fn_id", call.FnID),
+		trace.StringAttribute("fn.rid", common.RequestIDFromContext(ctx)),
```
rdallman: these won't be present always, and we'll end up with `fn.rid=""` in OSS specifically, could we possibly add a check for empty string here and conditionally add this attribute?

api/agent/agent.go
```diff
@@ -250,7 +250,14 @@ func (a *agent) Submit(callI Call) error {
 	ctx, span := trace.StartSpan(ctx, "agent_submit")
 	defer span.End()
 
-	span.AddAttributes(trace.StringAttribute("agent.call_id", call.ID))
+	span.AddAttributes(
+		trace.StringAttribute("fn.id", call.ID),
+		trace.StringAttribute("fn.app_id", call.AppID),
+		trace.StringAttribute("fn.fn_id", call.FnID),
+		trace.StringAttribute("fn.app_name", call.AppName),
+		trace.StringAttribute("fn.fn_id", call.FnID),
```
rdallman: this is a duplicate

api/agent/agent.go
```diff
@@ -250,7 +250,14 @@ func (a *agent) Submit(callI Call) error {
 	ctx, span := trace.StartSpan(ctx, "agent_submit")
 	defer span.End()
 
-	span.AddAttributes(trace.StringAttribute("agent.call_id", call.ID))
+	span.AddAttributes(
+		trace.StringAttribute("fn.id", call.ID),
```
rdallman: this seems a little ambiguous, is there a reason to change from call_id ?

api/agent/agent.go
```diff
@@ -250,7 +250,14 @@ func (a *agent) Submit(callI Call) error {
...
+		trace.StringAttribute("fn.app_id", call.AppID),
+		trace.StringAttribute("fn.fn_id", call.FnID),
+		trace.StringAttribute("fn.app_name", call.AppName),
```
rdallman: app name is probably irrelevant (or even troublesome, if name changes?) now with app id - not sure, but we should probably try to minimize bytes in these things if we can afford it

api/agent/agent.go
```diff
@@ -250,7 +250,14 @@ func (a *agent) Submit(callI Call) error {
 	ctx, span := trace.StartSpan(ctx, "agent_submit")
 	defer span.End()
 
-	span.AddAttributes(trace.StringAttribute("agent.call_id", call.ID))
+	span.AddAttributes(
```
rdallman: since this span would be an eventual child of the lb_agent span, these attributes are duplicates of the lb_agent span, just double checking that this method will not duplicate entries? from what I can tell, all the values will be the same, so it makes sense for the keys to match (at least, I don't think we want lb_agent's fn.call_id separately from agent's fn.call_id, or any of the other attributes here).

api/agent/lb_agent.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/fnproject/fn/fnext"
 )
 
+const RequestIDContextKey = "fn_request_id"
```
rdallman: this is unused?

bharadwr:
@mattstep branch builder namespace still kicking after 2 days looks like it's good to merge

PR #1519, opened 2019-06-27
use a new slot if call extensions change

gkvijay:
When the same call is invoked with changes in the extensions, we end up using the old cached values if we reuse a slot 

Included call extensions in the slot key so that we match and reuse a slot only if the extensions do not change

### DIFF
--- a/api/agent/drivers/driver.go
+++ b/api/agent/drivers/driver.go
@@ -75,6 +75,10 @@ type Driver interface {
 	// Set image pull retry policy and retriable error checker
 	SetPullImageRetryPolicy(policy common.BackOffConfig, checker RetryErrorChecker) error
 
+	// Get serialized string of fields in call extn that determines if we can use
+	// an existing hot container slot or pull a new image and start a container
+	GetSlotKeyExtensions(extn map[string]string) string
+
 	// close & shutdown the driver
 	Close() error
 }
--- a/api/agent/lb_agent.go
+++ b/api/agent/lb_agent.go
@@ -173,7 +173,6 @@ func (a *lbAgent) GetCall(opts ...CallOpt) (Call, error) {
 
 	c.ct = a
 	c.stderr = common.NoopReadWriteCloser{}
-	c.slotHashId = getSlotQueueKey(&c)
 	return &c, nil
 }
 
--- a/api/agent/slots.go
+++ b/api/agent/slots.go
@@ -281,7 +281,7 @@ var shapool = &sync.Pool{New: func() interface{} { return sha256.New() }}
 
 // TODO do better; once we have app+fn versions this function
 // can be simply app+fn ids & version
-func getSlotQueueKey(call *call) string {
+func getSlotQueueKey(call *call, slotExtns string) string {
 	// return a sha256 hash of a (hopefully) unique string of all the config
 	// values, to make map lookups quicker [than the giant unique string]
 
@@ -349,6 +349,11 @@ func getSlotQueueKey(call *call) string {
 		hash.Write(unsafeBytes("\x00"))
 		v, _ := call.Annotations.Get(k)
 		hash.Write(v)
+		hash.Write(unsafeBytes("\x00"))
+	}
+
+	if slotExtns != "" {
+		hash.Write(unsafeBytes(slotExtns))
 		hash.Write(unsafeBytes("\x00"))
 	}
 
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -342,7 +342,8 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
 	var isNew bool
 
 	if call.slotHashId == "" {
-		call.slotHashId = getSlotQueueKey(call)
+		slotExtns := a.driver.GetSlotKeyExtensions(call.Extensions())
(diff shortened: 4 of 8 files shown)


api/agent/slots.go
```diff
@@ -352,6 +352,13 @@ func getSlotQueueKey(call *call) string {
 		hash.Write(unsafeBytes("\x00"))
 	}
 
+	for k,v := range call.extensions {
```
rdallman: additionally to the prior comment (agreed), this won't work as intended either because map iteration is random, meaning these will have to get sorted (see annotations block above this) or new containers will get launched for no real reason (it is sad, if we had func versioning it may not be needed and we could save some bytes too). it's unclear from context on the PR which specific keys are of interest here, too

gkvijay: Thanks for the review. Updated the change

gkvijay:
@kmjohansen, Changed the approach to use a driver method to get the extns that are required for the slot hash

api/agent/drivers/driver.go
```diff
@@ -75,6 +75,9 @@ type Driver interface {
 	// Set image pull retry policy and retriable error checker
 	SetPullImageRetryPolicy(policy common.BackOffConfig, checker RetryErrorChecker) error
 
+	// Get fields from call extensions that are used in finding a slot to place the call
+	GetSlotKeyExtensions(extn map[string]string) map[string]string
```
skinowski: I think this can simply return a hash (aka string) to keep things simple. Then getSlotQueueKey() wouldn't need to handle sorting, etc.

api/agent/lb_agent.go
```diff
@@ -173,7 +173,7 @@ func (a *lbAgent) GetCall(opts ...CallOpt) (Call, error) {
 
 	c.ct = a
 	c.stderr = common.NoopReadWriteCloser{}
-	c.slotHashId = getSlotQueueKey(&c)
+	c.slotHashId = getSlotQueueKey(&c, nil)
```
skinowski: This line above shows that LB is actually setting the slot key. In the past, we wanted to move slot key generation to LBs which creates some opportunities to pick/optimize runners. But at the end, we didn't utilize this much at all. So, we have to options:
1) remove this line above completely and let runners set the slot hash id
2) or make sure LB processes slot hash id correctly, (aka do not pass "nil" as driver)

gkvijay: Will remove it, if its not used

api/agent/slots.go
kmjohansen: To address the question of context: the original problem observed was that if an auth token passed through in an extension expired before a successful pull could occur for a slot, all subsequent pulls would fail.  There are other things like tokens that get passed through to containers in some cases.  The original concern was that expiry of these kinds of objects would also cause similar problems, but after consulting with those authors it looks like it's just the auth token.  We could potentially simplify here to simply hash the output of the auth token's function, which already allows plug-able implementations.  That said, the next extension that impacted container behavior would need to implement this, so having something more generic earlier may be a good thing.

api/agent/drivers/driver.go
```diff
@@ -75,6 +75,9 @@ type Driver interface {
 	// Set image pull retry policy and retriable error checker
 	SetPullImageRetryPolicy(policy common.BackOffConfig, checker RetryErrorChecker) error
 
+	// Get serialized call extensions to select a slot to place the call
```
kmjohansen: Would you please expand on this comment just a bit further?  In particular, it might be helpful to note that this allows driver implementations to select what change to extension fields warrant creation of a new container, versus reusing an existing one.

api/agent/drivers/driver.go
gkvijay: changed to return a hash (serialized string of the required fileds in extn)

api/agent/slots.go
gkvijay: This was changed to use a hash of the extensions

api/agent/slots.go
```diff
@@ -352,6 +354,11 @@ func getSlotQueueKey(call *call) string {
 		hash.Write(unsafeBytes("\x00"))
 	}
 
+	if driver != nil {
+		slotExtns := driver.GetSlotKeyExtensions(call.Extensions())
```
kmjohansen: Some part of me finds it peculiar to pass in `driver` just to invoke `GetSlotKeyExtensions()` when we have both `driver` and `call` present in the callsite at `agent.getSlot()`.  Would it be cleaner to hoist this out into `getSlot()` and then just pass the filtered string into `getSlotQueueKey()`?  Or are we not guaranteed to have all of the pieces together in all cases?

api/agent/slots.go
```diff
@@ -352,6 +354,11 @@ func getSlotQueueKey(call *call) string {
...
+		hash.Write(unsafeBytes(slotExtns))
```
kmjohansen: Is a `hash.Write(unsafeBytes("\x00"))` needed after this line?

api/agent/drivers/driver.go
gkvijay: updated

api/agent/slots.go
gkvijay: Missed that. Updated the change

PR #1523, opened 2019-07-04
Expose the runner-reported execution duration

jan-g:
We'd like a better picture of how long we spend running actual user
code. The pure_runner has a decent picture of this; pull that out of
the runner response and feed it back into the *models.Call structure.

- What I did

Captured actual function run time

- How I did it

Believed what the runner said and promoted that value to the models.Call

"Expose the runner-reported execution duration"

### DIFF
--- a/api/agent/call.go
+++ b/api/agent/call.go
@@ -331,6 +331,9 @@ func (c *call) AddUserExecutionTime(dur time.Duration) {
 		c.userExecTime = new(time.Duration)
 	}
 	*c.userExecTime += dur
+	// We expose this on the upstream models.Call also.
+	// CallListeners have access to the latter, but not the internals of the agent, so any
+	// reporting or bean-counting that's going on from there will need access to this.
 	c.Model().ExecutionDuration = *c.userExecTime
 }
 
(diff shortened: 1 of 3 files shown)


api/agent/call.go
```diff
@@ -331,6 +331,7 @@ func (c *call) AddUserExecutionTime(dur time.Duration) {
 		c.userExecTime = new(time.Duration)
 	}
 	*c.userExecTime += dur
+	c.Model().ExecutionDuration = *c.userExecTime
```
reclaro: Where do you want to consume this value? we have it in the *call object I guess you want to add it to the Call object (naming...) to use it somewhere where the *call object is not available, can you give some example? Can you add a comment to specify why this will be useful?

jan-g: I've updated that to tat effect, but for the record here:

The point of this is to expose the ExecutionDuration to models.Call, since it's that that any CallListener gets to play with. This gives us a better view of the complete lifecycle of an
invocation on the loadbalancer; in particular, we get to actually meter runtime rather than
(placement wait time plus runtime).

PR #1524, opened 2019-07-04
fix-1484: Fix racy test by fixing concurrent header read/write

srpurani:
The issue 1484 reported racy test due to concurrent read/write of response headers during TryExec. This PR fixes the issue by making a copy of header before we write and use the copy to read later.

This PR also enhances the system test for detached calls by also verifying that the call returns within 30 seconds even if the function execution takes more than 30 seconds.

- Link to issue this resolves
https://github.com/fnproject/fn/issues/1484

- What I did
This PR fixes the issue by making a copy of header before we write and use the copy to read later.

- How I did it
This PR fixes the issue by making a copy of header before we write and use the copy to read later.

- How to verify it
Run the tests and ensure it does not fail.

Fix 1484 by making a copy of header before read/write.

### DIFF
--- a/api/agent/runner_client.go
+++ b/api/agent/runner_client.go
@@ -346,12 +346,24 @@ func recordFinishStats(ctx context.Context, msg *pb.CallFinished, c pool.RunnerC
 	}
 }
 
+func cloneHeaders(src http.Header) http.Header {
+	dst := make(http.Header, len(src))
+	for k, vs := range src {
+		for _, v := range vs {
+			dst.Add(k, v)
+		}
+	}
+	return dst
+}
+
 func receiveFromRunner(ctx context.Context, protocolClient pb.RunnerProtocol_EngageClient, runnerAddress string, c pool.RunnerCall, done chan error) {
 	w := c.ResponseWriter()
 	defer close(done)
 
 	log := common.Logger(ctx).WithField("runner_addr", runnerAddress)
 	statusCode := int32(0)
+	// Make a copy of header to avoid concurrent read/write error when logCallFinish runs.
+	clonedHeaders := cloneHeaders(w.Header())
 	isPartialWrite := false
 
 DataLoop:
@@ -372,6 +384,7 @@ DataLoop:
 			case *pb.CallResultStart_Http:
 				log.Debugf("Received meta http result from runner Status=%v", meta.Http.StatusCode)
 				for _, header := range meta.Http.Headers {
+					clonedHeaders.Set(header.Key, header.Value)
 					w.Header().Set(header.Key, header.Value)
 				}
 				if meta.Http.StatusCode > 0 {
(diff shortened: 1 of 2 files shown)


api/agent/runner_client.go
```diff
@@ -346,12 +346,24 @@ func recordFinishStats(ctx context.Context, msg *pb.CallFinished, c pool.RunnerC
...
+}
+
 func receiveFromRunner(ctx context.Context, protocolClient pb.RunnerProtocol_EngageClient, runnerAddress string, c pool.RunnerCall, done chan error) {
 	w := c.ResponseWriter()
 	defer close(done)
 
 	log := common.Logger(ctx).WithField("runner_addr", runnerAddress)
 	statusCode := int32(0)
+	// Make a copy of header to avoid concurrent read/write error when logCallFinish runs.
+	header := cloneHeaders(w.Header())
```
reclaro: can you please rename this "heade" in "clonedHeader" or something similar? That clashes with the name defined in the loop at 386, it doesn't create any issue because the header at line 386 is a different type but It creates some confusion for the reader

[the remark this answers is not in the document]
srpurani: Concurrent access happens after line 391 is executed in detached case. When this line is executed, it triggers this line https://github.com/fnproject/fn/blob/master/api/server/runner_fninvoke.go#L123-L127, which writes to the header. Next, logCallFinish can happen concurrently and result in concurrent read of header, which this PR will fix.

srpurani: done

api/agent/runner_client.go
```diff
@@ -400,7 +412,7 @@ DataLoop:
 
 		// Finish messages required for finish/finalize the processing.
 		case *pb.RunnerMsg_Finished:
-			logCallFinish(log, body, w.Header(), statusCode)
+			logCallFinish(log, body, clonedHeaders, statusCode)
```
skinowski: we need headers received in line 386 as well...

api/agent/runner_client.go
```diff
@@ -346,12 +346,24 @@ func recordFinishStats(ctx context.Context, msg *pb.CallFinished, c pool.RunnerC
...
+}
+
 func receiveFromRunner(ctx context.Context, protocolClient pb.RunnerProtocol_EngageClient, runnerAddress string, c pool.RunnerCall, done chan error) {
 	w := c.ResponseWriter()
 	defer close(done)
 
 	log := common.Logger(ctx).WithField("runner_addr", runnerAddress)
 	statusCode := int32(0)
+	// Make a copy of header to avoid concurrent read/write error when logCallFinish runs.
+	clonedHeaders := cloneHeaders(w.Header())
```
skinowski: cloning w.Header() is nice, to cover cases where somewhere we already written some headers, but most importantly, we need headers (see line 386) which we receive from runner. In that for-loop, let's add/set headers as well.

srpurani: done

Issue #1525, opened 2019-07-10
Exclude healthcheck & other http spans from ochttp


bharadwr:
The ochttp wrapper in 
https://github.com/fnproject/fn/blob/7e0ee579bb373be324f42a004acceec301402c14/api/server/server.go#L882
https://github.com/fnproject/fn/blob/7e0ee579bb373be324f42a004acceec301402c14/api/server/server.go#L858
creates a ton of spans which we could do without on a jaeger/zipkin ui. 

I'm working on adding StartOptions to the wrapper that can remove these requests (mostly kube-probe and prometheus) from the console. Without it, it will look like this: 

![Screen Shot 2019-07-10 at 10 09 57 AM](https://user-images.githubusercontent.com/25177182/60989779-a4fc4d00-a2fb-11e9-8c6d-3049a4d0d3a2.png).

I'm thinking we could also use this to ratelimit other traces, and perhaps even remove traces from canary tests.

bharadwr:
We could either sample only a set of requests that we need and exclude the rest, or just find a few of the spans that we dont want and keep them out. I've got the latter working like this

https://github.com/fnproject/fn/blob/7c132f3840ffba1dd7568e73a07f52c813d6b317/api/server/server.go#L858:L867

PR #1526, opened 2019-07-10
Adding ochttp startOpts to exclude junk spans

bharadwr:
https://github.com/fnproject/fn/issues/1525

Added startOptions to ochttp wrapper in server.go, excludes http traces with useragents that are blacklisted. Also reduces the ochttp sampler load by ignoring spans that dont need sampling.
Currently removes kube-probe and Prometheus requests

Used ochttp.GetstartOptions in handler https://godoc.org/go.opencensus.io/plugin/ochttp#Handler

Before:
Screen Shot 2019-07-10 at 11 36 41 AM


bharadwr:
Also possibly time to phase out 
https://github.com/fnproject/fn/blob/efc3e3e4a495fe4b0b2693897101b36775726c7d/api/server/server.go#L705 
Will require some extra changes but might as well get it out of the way

PR #1529, opened 2019-07-15
Adding back the change to pull image by digest

gkvijay:
- What I did
Added back the change to pull image using digest. This was reverted temporarily to debug token caching issues in runner

Image name is used in image puller, image cache, image cleaner and slot hash. In most of these place an inspect image is called to get image details from docker daemon. Inspect image api returns the image corresponding to the digest if present in the image name. Image cache code uses the repotags only to identify any special images that should not be cached or expired. Slot Hash will use different containers for functions invoked by tag-only and functions invoked with tag+digest. Overall, the code doesn't seem to use image tags for indexing, but use ID instead. 

Test case summary

Case 1: Pull older digest than repo latest
 Invoked a function (image:tag@digest) with an older digest than the latest in registry. The pulled image in runner will not have RepoTags associated with it. The RepoDigest field will have the digest of the image.  

Case 2: Pull the latest image digest in repo
 Invoked a function (image:tag@digest) with latest digest in registry associated with the tag. The pulled image in runner will not have RepoTags associated with it, as we pull it using the digest only. The RepoDigest field will have the digest of the image.  

Case 3: Update registry image with same tag. Pull using new digest
Invoked a function with the updated digest in registry.  The pulled image in runner will not have RepoTags associated with it. The RepoDigest field will have the digest of the image.  

Case 4: Pull image with tag only
 Invoked a function with the tag in registry. The image will be updated with RepoTags. The RepoDigest field will have the digest of the image.  

Case 5: Update the image after pulling with tag
Invoked a function with the new digest in registry, after image was pulled with tag. The new image will not have RepoTags. The RepoDigest field will have the digest of the image. The older image which was pulled by tag will continue to have the tag. Invoking a function with tag at this point executes the older tagged version of the function

### DIFF
--- a/api/agent/drivers/driver.go
+++ b/api/agent/drivers/driver.go
@@ -269,17 +269,24 @@ type Config struct {
 }
 
 // https://github.com/fsouza/go-dockerclient/blob/master/misc.go#L166
-func parseRepositoryTag(repoTag string) (repository string, tag string) {
-	parts := strings.SplitN(repoTag, "@", 2)
+func parseRepositoryTag(repoTag string) (repository, tag string) {
+	parts := strings.Split(repoTag, "@")
+	var digest string
+	if len(parts) == 2 {
+		digest = parts[1]
+	}
 	repoTag = parts[0]
 	n := strings.LastIndex(repoTag, ":")
 	if n < 0 {
-		return repoTag, ""
+		return repoTag, digest
+	}
+	if digest != "" {
+		return repoTag[:n], digest
 	}
 	if tag := repoTag[n+1:]; !strings.Contains(tag, "/") {
 		return repoTag[:n], tag
 	}
-	return repoTag, ""
+	return repoTag, digest
 }
 
 func ParseImage(image string) (registry, repo, tag string) {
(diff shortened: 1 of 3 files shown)


api/agent/drivers/driver.go
```diff
@@ -269,17 +269,24 @@ type Config struct {
 }
 
 // https://github.com/fsouza/go-dockerclient/blob/master/misc.go#L166
-func parseRepositoryTag(repoTag string) (repository string, tag string) {
+func parseRepositoryTag(repoTag string) (repository, tag string) {
 	parts := strings.SplitN(repoTag, "@", 2)
+	var digest string
+	if len(parts) == 2 {
```
kmjohansen: Is there any validation elsewhere that ensures that there's only a single '@' in this string?  E.g. if this function gets more parts than expected, is it safe to proceed?

gkvijay: Before we reach here, we would have fetched the registry token in LB. If the image is not valid, we will fail fetching token. But I changed the logic here to use Split and decode digest only if there are two parts

PR #1532, opened 2019-07-17
Add SyslogUnavailable as new codified error

reclaro:
We codified the SyslogUnavailable error into a public error exposed by
the models package

### DIFF
--- a/api/agent/drivers/docker/docker_test.go
+++ b/api/agent/drivers/docker/docker_test.go
@@ -574,7 +574,7 @@ func TestRunnerDockerInvalidSyslog(t *testing.T) {
 	if err == nil {
 		t.Fatal("Error expected when running with invalid syslog configuration")
 	}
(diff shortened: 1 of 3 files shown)


api/agent/drivers/docker/docker_test.go
```diff
@@ -574,7 +574,7 @@ func TestRunnerDockerInvalidSyslog(t *testing.T) {
 	if err == nil {
 		t.Fatal("Error expected when running with invalid syslog configuration")
 	}
-	if err.Error() != "Syslog Unavailable" {
+	if err.Error() != models.ErrSyslogUnavailable.Error() {
```
riconnon: Can this not just be `if err != models.ErrSyslogUnavailable`?

reclaro: Done

PR #1533, opened 2019-07-18
Bugfix: use the most recent call for ephemeral data

jan-g:
Things like pull tokens (which can be attached to an incoming call) may
be effectively ephemeral - they have a shorter lifetime than the
slotQueue.

Rather than having hotLauncher use the *original* call that spawned it
for this metadata, we update the slotQueue to hold an atomic reference
to the *most recent* call.

Any tokens on new calls will be used by preference, therefore - meaning
that we won't see a spurious expired token used for image pulls, etc.

Cost here is a single atomic Store/Load, which is probably as good as
it can get. The call extensions should be immutable for this process.

- Link to issue this resolves

- What I did

- How I did it

- How to verify it

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -371,6 +371,9 @@ func (a *agent) getSlot(ctx context.Context, call *call) (Slot, error) {
 		caller.notify = make(chan error)
 	}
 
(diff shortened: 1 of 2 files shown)


reclaro:
as pointed out by @kmjohansen this is superseded by other patches.

PR #1534, opened 2019-07-18
fn: provide update method for hot slot auther

skinowski:


api/agent/agent.go
```diff
@@ -1235,6 +1237,11 @@ func newHotContainer(ctx context.Context, evictor Evictor, call *call, cfg *Conf
 		},
 	}
 
+	auther := call.dockerAuth
```
skinowski: This is to make the tests pass, since TestDockerAuthExtn does not have a slot queue/mgr.

PR #1536, opened 2019-07-18
Enforce more rigorous handling of input data and pipes.

kmjohansen:
Catches and generates function errors for two new cases.  The first
occurs due to a function/FDK error.  If the function closes the read end
of the pipe that the hostagent uses to write data before the hostagent
has finished writing the data, generate an error.  This ensures that
any premature close of the input stream is detected and handled by the
hostagent.

Second, catch any cases where the function attempts to respond before
reading all of the input data from the stream.  This is safe because we
already enforce a maximum upper bound on the request body, so a function
or fdk will not have to read for an unbounded amount of time to consume
the outstanding data.  Since the container contract enforces HTTP-like
semantics, and HTTP expects the server side to wait for the response
body to arrive before responding, this is not unreasonable.  If the
function or fdk attempts to write before the end of the input stream has
been processed by the hostagent, return a different function error
indicating that a premature write has been detected.

### DIFF
--- a/api/agent/pure_runner.go
+++ b/api/agent/pure_runner.go
@@ -79,7 +79,9 @@ var (
 type callHandle struct {
 	engagement runner.RunnerProtocol_EngageServer
 	ctx        context.Context
-	c          *call // the agent's version of call
+	sctx       context.Context    // child context for submit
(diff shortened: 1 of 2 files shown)


jan-g:
Looks good, working in my testing.

Issue #1538, opened 2019-07-30
Opencensus span support for remote runners


bharadwr:
Remote runners (in pure_runner mode for instance) that do not have an exporter defined and cannot communicate due to firewall/egress rules need a way for their metrics to get into an opencensus span. Additionally, we can take this opportunity to add information about image pull times (if applicable), and perhaps image id to the span data. 

Lightest way of doing this is adding extra fields to the callfinished proto, retrieve data as needed 
This is better than a gin server log retriever on the runner, or any fluentd socat stuff. Also, this will run AFTER the function execution, and will not delay the execution. 
Adding spans to runner_client.go, and span attribs will give users an effective means of getting remote runner data.

Formatting incoming invoke requests using ochttp format span name is also needed, floods the jaeger ui operation tab otherwise.

Can add a few attributes here for image pull duration, and imagename
https://github.com/fnproject/fn/blob/6cf4799410b5ea07aceac64e5424885513127ad8/api/agent/grpc/runner.proto#L40:L51

bharadwr:
Proposed changes:

https://github.com/fnproject/fn/blob/77c1e2d36e429b2182ae7bceea71f1afe42b09e7/api/agent/grpc/runner.proto#L40:L53

https://github.com/fnproject/fn/blob/77c1e2d36e429b2182ae7bceea71f1afe42b09e7/api/agent/runner_client.go#L436:L453

PR #1539, opened 2019-07-30
opencensus attribs for remote runner (#1538)

bharadwr:
adding span attrib support for remote runner metrics
adding ochttp filter to group invoke calls in server.go
changing up the grpc caallfinished for display on jaeger
changed call.go, including imgpull time (if applicable)
span annotation in runner_client
add imagepulltime in protobuf
Adding span.setstatus to some spans
moving the image and getusererr bool from annotation to attribs in runner_client for viz
Adding fmtspanname in adminserver

- Link to issue this resolves
https://github.com/fnproject/fn/issues/1538

- What I did - How I did it
Changed CallFinished proto in agent/grpc/runner.proto, adding strings imageid, imgpullduration
Added span.Annotate to dump the results of CallFinished in runner_client.go
Added timestamps for start and end of imgpull if applicable in Call, function to calculate duration of pull (0 if no pull needed) 

- How to verify it
Screen Shot 2019-07-30 at 2 53 37 PM

adding span attrib support for remote runner metrics


api/models/call.go
```diff
@@ -141,6 +141,12 @@ type Call struct {
 
 	// Fn this call belongs to.
 	FnID string `json:"fn_id" db:"fn_id"`
+
+	// Image Pull Duration
+	PullStartedAt common.DateTime `json:"pull_started_at" db:"pull_started_at"`
+
+	// Image Pull Duration
+	PullEndedAt common.DateTime `json:"pull_ended_at" db:"pull_ended_at"`
```
skinowski: we don't need to add these to model/call. We can use the same hackery we have used for userExecTime in agent/call.go (https://github.com/fnproject/fn/blob/master/api/agent/call.go#L295)

api/agent/agent.go
```diff
@@ -880,9 +880,9 @@ func (a *agent) runHot(ctx context.Context, caller slotCaller, call *call, tok R
 	if tryQueueErr(err, errQueue) != nil {
 		return
 	}
-
 	needsPull, err := cookie.ValidateImage(ctx)
 	if needsPull {
+		call.PullStartedAt = common.DateTime(time.Now())
```
skinowski: This is tricky here. If multiple pulls are being executed for the same image, then this metric here will not yield the true docker pull duration, but it will simply show **the amount of wait time for this consumer code.**   Basically, we have 2 metrics at work here, 1) true docker-pull time 2) wait time experienced here in this code-block.

For example, another client might have already triggered a docker-pull a few seconds earlier. If this code blocks on that as well, then the wait time here would show less time than the actual docker-pull duration, since earlier client started the download earlier.

If we are trying to account for how much time this client code have waited, then the metric here as is is OK and good. But in that case, maybe we should rename this metric as 'docker-pull-wait'. I think this is what we should do... I suggest a small rename to clarify that this metric here may not be the true download duration.

bharadwr: The true pull duration is what we need, not the pull-wait as you suggested, let me revise this. I'm thinking maybe the method used to add userexecution time, by incrementing using time.Duration would be better

api/models/call.go
bharadwr: Agreed, I'll change it up

api/agent/agent.go
bharadwr: https://github.com/fnproject/fn/blob/8cef854cccf96da4d78cf1aa6d0dd8b478a8fe8b/api/agent/pure_runner.go#L239

https://github.com/fnproject/fn/blob/8cef854cccf96da4d78cf1aa6d0dd8b478a8fe8b/api/agent/agent.go#L899

Here you go, docker pull wait now

api/models/call.go
bharadwr: https://github.com/fnproject/fn/blob/8cef854cccf96da4d78cf1aa6d0dd8b478a8fe8b/api/models/call.go#L127

api/agent/call.go
```diff
@@ -406,3 +417,15 @@ func GetCallLatencies(c *call) (time.Duration, time.Duration) {
 	}
 	return schedDuration, execDuration
 }
+
```
bharadwr: Follows the same method as GetUserExecTime and adduserexectime

bharadwr:
Splitting into two issues, opencensus remote runner support and docker pull stats

PR #1540, opened 2019-08-07
remote runner support for trace exporters (#1538)

bharadwr:
Handling docker metrics in a different issue

adding span attrib support for remote runner metrics (docker wait time, etc. in runner_client)
adding ochttp filter to group invoke calls in server.go (removes traces from /, /version paths)
changing up the grpc callfinished for display on jaeger (adding image and docker wait fields)
Adding span.setstatus to some spans

- Link to issue this resolves
https://github.com/fnproject/fn/issues/1538

- What I did
- How I did it
Added image name and docker wait duration fields to callfinished grpc response sent to runner_client, span annotations, attributes for visibility/debugging

- How to verify it
Set a jaeger/zipkin url, invoke a function on a remote runner and check attributes in new span receive_from_runner
image

remote runner support for jaeger (#1538)


api/models/call.go
```diff
@@ -123,6 +123,9 @@ type Call struct {
 	// Duration that user code was running for, in nanoseconds.
 	ExecutionDuration time.Duration `json:"execution_duration,omitempty" db:"execution_duration"`
 
+	// Duration that image pull has waited, in nanoseconds.
+	DockerWaitDuration time.Duration `json:"docker_wait_duration,omitempty" db:"-"`
+
```
bharadwr: Questionable, but necessary.

api/server/server.go
```diff
@@ -855,7 +856,18 @@ func (s *Server) startGears(ctx context.Context, cancel context.CancelFunc) {
 
 	server := s.svcConfigs[WebServer]
 	if server.Handler == nil {
-		server.Handler = &ochttp.Handler{Handler: s.Router}
```
bharadwr: Prevents traces of / path from showing up on the exporter/metrics. 
Need to scrap AlwaysSample by default and move to per span sampling eventually.

api/agent/runner_client.go
```diff
@@ -247,14 +252,17 @@ func sendToRunner(ctx context.Context, protocolClient pb.RunnerProtocol_EngageCl
 		// WARNING: blocking read.
 		n, err := bodyReader.Read(writeBuffer)
 		if err != nil && err != io.EOF {
-			log.WithError(err).Error("Failed to receive data from http client body")
+			errorMsg = fmt.Sprintf("Failed to receive data from http client body")
```
skinowski: This is a const-string, no need for 'fmt.Sprintf'

bharadwr: Good eye, will change that back.

bharadwr:
summary of the changes:
stopped tracing / path on webserver, scrapping ochttp on adminserver
imgPullWaitDuration, Ctr create and prep duration, Init_start timestamp all using atomic in api/agent.go, call.go and pure_runner.go
Adding these metrics to proto runnerstatus and callfinished runner.proto
Adding new metrics as spandata in runner_client.go along with spanstatuses for err reporting

PR #1542, opened 2019-08-12
linear and log scale histogram bucket generators

shreyagarge:
- What I did
 added helper functions to generate log scale and linear histogram buckets for metrics


skinowski:
@shreyagarge Could you please perhaps move these to api/common/stats_utils.go?

api/agent/drivers/stats/stats.go
```diff
@@ -130,3 +132,38 @@ func Decimate(maxSamples int, stats []Stat) []Stat {
...
+
+// GenerateLogScaleHistogramBucketsWithRange generates histogram buckets on the log scale between the specified min and max range,
+// such that the min value is in the first bucket and the max in the last.
+func GenerateLogScaleHistogramBucketsWithRange(min, max float64) []float64 {
+	if min <= 0 {
+		logrus.Fatal("cannot generate log scale with non positive domain values")
+	}
+	count := int(math.Ceil(math.Log2(max / min)))
+	var res []float64
+	for i := 0; i <= count; i++ {
```
skinowski: last bucket is bigger than max (eg. min:2 max:10, gets 2,4,8,16) , off by one error.

shreyagarge: true! I'll change this up to divide from max instead of multiply from min.

PR #1543, opened 2019-08-20
Prevent long timeout when invoking a function and the user has misconfigured the environment.

templecloud:
Updates the RetryAllBackOff function to fail early when a request to obtain Runners cannot succeed due to an Fn misconfiguration error. This prevents a user having to wait for the maximum timeout (~180s) when the system is misconfigured.

- Link to issue this resolves

- What I did

    1. Updated the `RetryAllBackOff` to accept the error returned by `rp.Runners(ctx, call)`.

    2. Added check to not re-try when the error returned by `rp.Runners(ctx, call)` returns a `user configuration error`.

        1.  The error is checked to see if it a user configuration error with the ` models.IsFuncError(err)`. This is the same function used to create the `blame user` tag in the existing middleware (with the exception we do not need to check for nullity).

        2. Additionally check for an error status of `502` (which also denotes a user error).

- How I did it

    1. Worked with Jan Grant (already a comitter).

    2. Replicated the issue tests against a development version of the Fn platform.

    3. Dynamically checked that `tr.requestCtx` and `tr.placerCtx` did not have the required information at the point of early return.

    4. Statically examined the Fn code to see how the `user blame` mechanism worked.

    5. Added a small piece of code to replicate the `user blame` functionality in `RetryAllBackOff` and return early if necessary.

- How to verify it

    > NB: This was tested against an OCI backend. 

    1. Create OCI test resources, Application, and Function.

    2. Remove the VCN Policy permissions from the invoking Function compartment.

    3. Attempt to invoke the Function.

    Without the changes it will take around 180s to fail.

    Manual Integration Test

    1. Show invoke fails after 180s when miss-configured.

    ```
    $> time fn invoke test-app test-fn-01
    Error invoking function. status: 502 message: subnet ocid1.subnet.oc1.iad.aaaaaaaa3zi47ktmvvoqjltnw2kksgbedfcdzqrocvuxeonaz4rs3vped6gq does not exist or Oracle Functions is not authorized to use it

    real	3m2.990s
    user	0m0.065s
    sys	0m0.032s
    ```

    2. PR - Invoke fails fast when miss-configured.

    ```
    $> time fn invoke test-app test-fn-01
    Error invoking function. status: 502 message: subnet ocid1.subnet.oc1.iad.aaaaaaaa3zi47ktmvvoqjltnw2kksgbedfcdzqrocvuxeonaz4rs3vped6gq does not exist or Oracle Functions is not authorized to use it

    real	0m1.426s
    user	0m0.065s
    sys	0m0.018s
    ```

    3. PR - Invoke still works when NOT miss-configured..

    ```
    $> time fn invoke test-app test-fn-01
    {"message":"Hello World"}

    real	0m1.121s
    user	0m0.066s
    sys	0m0.017s
    ```

Prevent long timeout when invoking a function and the user has misconfigured the environment.

### DIFF
--- a/api/runnerpool/placer_tracker.go
+++ b/api/runnerpool/placer_tracker.go
@@ -113,7 +113,24 @@ func (tr *placerTracker) HandleDone() {
 
 // RetryAllBackoff blocks until it is time to try the runner list again. Returns
 // false if the placer should stop trying.
-func (tr *placerTracker) RetryAllBackoff(numOfRunners int) bool {
+func (tr *placerTracker) RetryAllBackoff(numOfRunners int, err error) bool {
+
+	// If the last call to obtain runners failed due to a user error or
+	// misconfiguration then do not re-attempt to obtain a runner.
+	if err != nil {
+		tr.HandleFindRunnersFailure(err)
+		// IsFuncError currently synonymous with tag: 'blame == user'
+		// See: runner_fninvoke.handleFnInvokeCall2
+		if models.IsFuncError(err) {
+			// We also santiy check for a 502 before returning.
+			w, ok := err.(models.APIErrorWrapper)
+			if ok {
+				if 502 == w.Code() {
+					return false
+				}
+			}
+		}
+	}
 
 	// This means Placer is operating on an empty list. No runners
 	// available. Record it.
--- a/api/runnerpool/placer_tracker.go
+++ b/api/runnerpool/placer_tracker.go
@@ -121,7 +121,7 @@ func (tr *placerTracker) RetryAllBackoff(numOfRunners int, err error) bool {
 		// See: runner_fninvoke.handleFnInvokeCall2
 		if models.IsFuncError(err) {
 			// We also sanity check for a 502 before returning.
-			w, ok := err.(models.APIErrorWrapper)
+			w, ok := err.(models.APIError)
 			if ok {
 				if 502 == w.Code() {
 					return false
--- a/api/runnerpool/placer_tracker.go
+++ b/api/runnerpool/placer_tracker.go
@@ -114,15 +114,13 @@ func (tr *placerTracker) HandleDone() {
 // RetryAllBackoff blocks until it is time to try the runner list again. Returns
 // false if the placer should stop trying.
 func (tr *placerTracker) RetryAllBackoff(numOfRunners int, err error) bool {
-
-	// If the last call to obtain runners failed due to a user error or
-	// misconfiguration then do not re-attempt to obtain a runner.
-	if err != nil {
-		tr.HandleFindRunnersFailure(err)
+	// If there are no runners and last call to provision runners failed due
+	// to a user error (or misconfiguration) then fail fast.
+	if numOfRunners == 0 && err != nil {
 		// IsFuncError currently synonymous with tag: 'blame == user'
 		// See: runner_fninvoke.handleFnInvokeCall2
 		if models.IsFuncError(err) {
-			// We also santiy check for a 502 before returning.
+			// We also sanity check for a 502 before returning.
 			w, ok := err.(models.APIErrorWrapper)
 			if ok {
 				if 502 == w.Code() {
--- a/api/runnerpool/ch_placer.go
+++ b/api/runnerpool/ch_placer.go
@@ -55,7 +55,7 @@ func (p *chPlacer) PlaceCall(ctx context.Context, rp RunnerPool, call RunnerCall
 			i = (i + 1) % len(runners)
 		}
 
-		if !state.RetryAllBackoff(len(runners)) {
+		if !state.RetryAllBackoff(len(runners), runnerPoolErr) {
 			break
 		}
 	}
--- a/api/runnerpool/naive_placer.go
+++ b/api/runnerpool/naive_placer.go
@@ -47,7 +47,7 @@ func (sp *naivePlacer) PlaceCall(ctx context.Context, rp RunnerPool, call Runner
 			}
 		}
 
-		if !state.RetryAllBackoff(len(runners)) {
+		if !state.RetryAllBackoff(len(runners), runnerPoolErr) {
 			break
 		}
 	}


api/runnerpool/placer_tracker.go
```diff
@@ -113,7 +113,24 @@ func (tr *placerTracker) HandleDone() {
 
 // RetryAllBackoff blocks until it is time to try the runner list again. Returns
 // false if the placer should stop trying.
-func (tr *placerTracker) RetryAllBackoff(numOfRunners int) bool {
+func (tr *placerTracker) RetryAllBackoff(numOfRunners int, err error) bool {
+
+	// If the last call to obtain runners failed due to a user error or
+	// misconfiguration then do not re-attempt to obtain a runner.
+	if err != nil {
+		tr.HandleFindRunnersFailure(err)
```
jan-g: This'll be called if and when we exit the Placer loop - I don't think you want to do it here as well.

api/runnerpool/placer_tracker.go
```diff
@@ -113,7 +113,24 @@ func (tr *placerTracker) HandleDone() {
...
+		// IsFuncError currently synonymous with tag: 'blame == user'
+		// See: runner_fninvoke.handleFnInvokeCall2
+		if models.IsFuncError(err) {
+			// We also santiy check for a 502 before returning.
```
jan-g: Spelling nit: "sanity"

api/runnerpool/placer_tracker.go
```diff
@@ -113,7 +113,24 @@ func (tr *placerTracker) HandleDone() {
 
 // RetryAllBackoff blocks until it is time to try the runner list again. Returns
 // false if the placer should stop trying.
-func (tr *placerTracker) RetryAllBackoff(numOfRunners int) bool {
+func (tr *placerTracker) RetryAllBackoff(numOfRunners int, err error) bool {
+
+	// If the last call to obtain runners failed due to a user error or
+	// misconfiguration then do not re-attempt to obtain a runner.
+	if err != nil {
```
jan-g: The only other thing here I'm not sure on is whether this should be guarded by `numOfRunners == 0` as well as this check.

Open to second opinions here: @ostrain ?

api/runnerpool/placer_tracker.go
templecloud: Thanks. Fixed.

api/runnerpool/placer_tracker.go
templecloud: Jan and I discussed the advantages/disadvantages of adding this clause:

(+)
1. It is a tighter more defensive control over this 'fail-fast' mechanism.
2. If the user had existing runners provisioned before miss-configuring the platform, then, they may have some time to correct the error before their invocation attempts being to fail. So it is more robust.

(-)
1. When (re-)configuring the platform incorrectly, the user may not immediately notice their mistake. Failures will only occur if/when the runners are re-cycled or require an increase capacity.

I have made the change, but, can roll it back if @ostrain has more info.

api/runnerpool/placer_tracker.go
templecloud: Thanks. You are right, I think was seeing double logs in Kibana. Removed it.

templecloud:
Trying to determine how long it takes to recover once a user error has been fixed. I am having network issues here atm, but, it is still returning 502's 10 mins later.

jan-g:
The cached response behaviour is what I was concerned about. Will discuss this face-to-face.

templecloud:
Thanks Jan. 

Just to update the thread, it only gets cached for 60 mins. So not a showstopper, but, we are going to remedy it.

ostrain:
Hmm, we're supposed to clear the error as soon as we successfully provision a runner, so if you're still seeing it after 10 minutes that suggests we have a bug. Plus the fact that we clear it shouldn't generally matter since once the runner is up your invoke ought to succeed in which case you won't see an error regardless.

api/runnerpool/placer_tracker.go
reclaro: I still see this in the code, did you push the latest version of your change?

api/runnerpool/placer_tracker.go
reclaro: ditto

api/runnerpool/placer_tracker.go
```diff
@@ -114,15 +114,13 @@ func (tr *placerTracker) HandleDone() {
...
 // false if the placer should stop trying.
 func (tr *placerTracker) RetryAllBackoff(numOfRunners int, err error) bool {
-
-	// If the last call to obtain runners failed due to a user error or
-	// misconfiguration then do not re-attempt to obtain a runner.
-	if err != nil {
-		tr.HandleFindRunnersFailure(err)
+	// If there are no runners and last call to provision runners failed due
+	// to a user error (or misconfiguration) then fail fast.
+	if numOfRunners == 0 && err != nil {
```
reclaro: Not sure I agree with this check on the number of runners.
Adding this means that the user will be notified of the misconfiguration only if she has not runners or, in case has some old runners, only when they get recycled. 
I'd prefer to immediately notify the users about the error so they can fix it soon. All the point of this change is to fail fast and give a quick feedback to customers, if we check for empty runner list we miss the point of this change.
Let's consider this scenario where I make a change (misconfiguring my env) and deploy it, I test that it works and i managed to invoke a function because I land on an old runner, after "n" hours my functions suddenly start to fail because the old good runners get recycled and the new ones are affected by my misconfiguration. Personally I prefer to find the error soon so i have the right context at the right moment to fix it. What do you think?

reclaro: in any case I think that we still record the stats for numOfRunners==0

templecloud: Yes, I agree. It is a choice between two behaviors, but, my gut feeling is I prefer fast feedback when I am/have done something wrong. I can revert the check for existing runners if we all agree this is most useful/suitable.

reclaro: I like the tradeoff of inspecting the error and fail fast if the error is a policy misconfiguration, that is a specific error for a specific service implementation, not sure how to make it abstract here, maybe some sort of pluggable error handling ? just a random idea not sure the effort to implement that.

reclaro: if it is too difficult or taking too long to implement the inspect of errors I think we can go with this solution as it is now with the numOfRunners check and having a follow up patch.

templecloud: @skinowski  - I was going to use the actual error, but, it was just a string specific to the OCI backend implementation I was using, I assumed we would want a more detailed abstraction here and not leak what is underneath (OCI). The IsFuncError() is what (to me) looked like the current way of determining a 'user blame' error? I also would prefer something better and clearer than using IsFuncError, but, that should not be a unilateral. Wanna help?

I agree with @ostrain . I currently think we should only skip when there are no runners. Looks like lots of intermittent things could go wrong inside the GetRunners call e.g. periodic shonky DNS.

riconnon: To add my $0.02

I think trying to extend this to cover cases where runner count is already greater than 0 may be a good move long term but will introduce complications.
In the short term handling the 0 runners case effectively is a good basis on which to improve.

Using the function error interface seems like the right way to determine blame, or, if we want to consider this tangential to invocation errors we could introduce a new interface representing user blame errors in the runner pool implementation.

templecloud: What actionable changes do I need to make, or, further tests to add/perform? Thanks!

Can I confirm IsFunc() is the mandated way to determine this is a user blame error?

api/runnerpool/placer_tracker.go
```diff
@@ -113,7 +113,22 @@ func (tr *placerTracker) HandleDone() {
...
-func (tr *placerTracker) RetryAllBackoff(numOfRunners int) bool {
+func (tr *placerTracker) RetryAllBackoff(numOfRunners int, err error) bool {
+	// If there are no runners and last call to provision runners failed due
+	// to a user error (or misconfiguration) then fail fast.
+	if numOfRunners == 0 && err != nil {
+		// IsFuncError currently synonymous with tag: 'blame == user'
+		// See: runner_fninvoke.handleFnInvokeCall2
+		if models.IsFuncError(err) {
+			// We also sanity check for a 502 before returning.
+			w, ok := err.(models.APIErrorWrapper)
```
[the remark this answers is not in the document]
jan-g: I've pushed a fix for this.

api/runnerpool/placer_tracker.go
jan-g: This works as written.

api/runnerpool/placer_tracker.go
```diff
@@ -113,7 +113,22 @@ func (tr *placerTracker) HandleDone() {
...
 // false if the placer should stop trying.
-func (tr *placerTracker) RetryAllBackoff(numOfRunners int) bool {
+func (tr *placerTracker) RetryAllBackoff(numOfRunners int, err error) bool {
+	// If there are no runners and last call to provision runners failed due
+	// to a user error (or misconfiguration) then fail fast.
+	if numOfRunners == 0 && err != nil {
+		// IsFuncError currently synonymous with tag: 'blame == user'
+		// See: runner_fninvoke.handleFnInvokeCall2
+		if models.IsFuncError(err) {
+			// We also sanity check for a 502 before returning.
```
skinowski: minor nit: do we need this extra check here? I think IsFuncError() should be good enough.

jan-g: Possibly not; it's hard to envisage a user error that prevents us from scaling up (from zero) where we *don't* want to bail immediately. On the other hand, I'd rather err on the side of being overly cautious in the first instance and then relax this condition.

PR #1546, opened 2019-09-06
fn: do not parse mysql URL via go url parse

skinowski:
With go 1.12.8, url parsing of mysql DB strings will
fail. With this change, we prepare fn for upcoming
go lang changes.


skinowski:
I was hoping just to fixup the tests for this, but looking closer, we do need to review the datastore URL parsing. WithDBURL() option seems to be not tested and should also fail for mysql for go lang >= 1.12.8.

PR #1551, opened 2019-09-25
fn: simplify CPU/Memory acquisition to avoid race condition

skinowski:
Before this change, agent used GetResourceToken() calls
for both non-blocking and blocking mode. However, for
nested calls such as in agent checkLaunch(), a token might
already be queued via go-routine in getResourceTokenNBChan().
If the consuming code in checkLaunch() runs faster, than
it could place another GetResourceToken() call while the
active token is not yet closed. This can result in momentarily
double cpu/mem allocation resulting in 503 or excess wait.

With this PR, resource allocation blocking and non-blocking
interface is changed and ResourceToken is directly returned
without a channel.

### DIFF
--- a/api/agent/agent.go
+++ b/api/agent/agent.go
@@ -405,18 +405,27 @@ func (a *agent) hotLauncher(ctx context.Context, call *call, caller *slotCaller)
 
 	// IMPORTANT: get a context that has a child span / logger but NO timeout
 	// TODO this is a 'FollowsFrom'
-	ctx = common.BackgroundContext(ctx)
+	var cancel func()
+	ctx, cancel = context.WithCancel(common.BackgroundContext(ctx))
+	defer cancel()
+
 	ctx, span := trace.StartSpan(ctx, "agent_hot_launcher")
 	defer span.End()
+
+	// trigger ctx cancel if server shutdown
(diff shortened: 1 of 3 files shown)


api/agent/agent.go
```diff
@@ -453,40 +465,16 @@ func (a *agent) checkLaunch(ctx context.Context, call *call, caller slotCaller)
...
-		case tok = <-a.resources.GetResourceToken(ctx, mem, call.CPUs, true):
-		case <-ctx.Done(): // timeout
-		case <-a.shutWg.Closer(): // server shutdown
-		}
-	case <-ctx.Done(): // timeout
-	case <-a.shutWg.Closer(): // server shutdown
+	// For blocking-mode, we wait on a channel for CPU/MEM for cfg.HotPoll duration.
+	// If the request is not satisfied during this wait, we perform a non-blocking
+	// GetResourceToken()) in an attempt to determine how much mem/cpu we need to evict.
+	if !isNB {
```
kmjohansen: A bit of nit, but would you be willing to change line 452 to `isBlocking := !a.cfg.EnableNBResourceTracker` so that we can just say `if isBlocking` here on line 471?

PR #1556, opened 2019-10-02
Add status v2

dmutreja:
- Link to issue this resolves
      None. 
- What I did
      Added the ability to supply input to status checker 
- How I did it
       Extended runner's grpc interface and added a new method, Status2 that can accept any JSON as a protocol buf struct.  
- How to verify it
    There's a new system test in the test/fn-system-tests/exec_runner_status_test.go that can be run individually. From root of the repo:
` cd  test/fn-system-tests/`
`go test -run TestStatus_verifyIO`


test/fn-system-tests/exec_runner_status_test.go
```diff
@@ -302,7 +304,12 @@ func TestCustomHealthChecker(t *testing.T) {
 		t.Fatal("Timeout")
 	}
 	shouldCustomHealthCheckerFail = true
-	defer func() { shouldCustomHealthCheckerFail = false }()
+	defer func() {
+		// Reset test state
+		// Ensure status cache expires
+		shouldCustomHealthCheckerFail = false
+		time.Sleep(2 * time.Second)
```
skinowski: minor: sleep should also check ctx cancel. Maybe use select{} to block on both?

test/fn-system-tests/exec_runner_status_test.go
```diff
@@ -405,3 +412,42 @@ func TestExampleLogStreamer(t *testing.T) {
...
+	defer func() {
+		if t.Failed() {
+			t.Log(buf.String())
+		}
+	}()
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	// Connect to any node
+	r := runnerGrpcServerAddr(-1)
```
skinowski: for deterministic tests, should we just pick same node? or hit all?

dmutreja: This is a very simple test so targeting all nodes seems wasteful.  I'll pick a node.

test/fn-system-tests/exec_runner_status_test.go
dmutreja: Not sure I understand the reason for checking the context cancellation in this last defer block.  The intent here is that this test doesn't leave any state regardless of what happened on the other end, so as to not impact the any other tests when the whole suite is run.  Or are you saying that if ctx was cancelled  then there is a guarantee that the server would've not cached any state?

skinowski: oh, I see the intent now. 2 sec expiry or delay itself is not robust, but this is OK for time being. FWIW, the select stmt above this change doesn't guarantee a 2 sec delay either. Perhaps we could convert this test to poll the status (eg. every 1 sec) until non-cached results are fetched. In general, any test calling Status() should handle and explicitly check for cached results.

dmutreja: Right, that would be better.  There are other tests in the same file that use the same (unreliable) pattern of waiting for 2 secs in their setup phase to ensure a clean cache.  I think what we want is a  fixture type that sets ups and cleans the state(including cache unless we are testing caching behavior of course) correctly.  And then refactor all these tests to use that fixture.  We should do that as a separate PR though.

PR #1566, opened 2019-10-30
fn: naive placer rr fixup with tests

skinowski:
If naive placer is not instantiated per call/runner group (aka LBG),
then the rr index will not trigger an round-robin behavior since
the index is initialized and stored in the placer configuration.

With this PR, moving rr index to per RunnerPool.Runners() inner loop
to ensure a round robin within that set. Each time we fetch a set,
since the set might be different, we reset our rr index. This means
we rr within that set once, then randomly start from another node
for the next RunnerPool.Runners() iteration. In busy systems,
no significant behavior change is expected (accept the removal of
atomic operations with respect to performance), but in idle systems
round robin behavior should be more observable and simple to follow and can
reduce same hit cases for the given RunnerPool.Runners().

In addition, introducing naive placer tests to ensure we observe
this behavior.

### DIFF
--- a/api/agent/lb_agent_test.go
+++ b/api/agent/lb_agent_test.go
@@ -200,43 +200,6 @@ func TestDetachedPlacerTimeout(t *testing.T) {
 
 }
 
-func TestRRRunner(t *testing.T) {
-	cfg := pool.NewPlacerConfig()
-	placer := pool.NewNaivePlacer(&cfg)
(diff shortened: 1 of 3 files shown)


skinowski:
removed the old TestRRRunner() test case. This was testing the old LBG/group level round-robin behavior which with the new changes we longer have.

PR #1568, opened 2019-12-03
fn: scheduler latency metric and pull/init duration

skinowski:
Image pull and container uds initialization duration is
out of Fn Service control, therefore should not be included
in scheduler latency which tracks service health.

### DIFF
--- a/api/agent/call.go
+++ b/api/agent/call.go
@@ -415,6 +415,23 @@ func GetCallLatencies(c *call) (time.Duration, time.Duration) {
 				}
 			}
 		}
+
+		// latency out of our control: 1) docker pull wait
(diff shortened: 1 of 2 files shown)


kmjohansen:
What's left in this metric with the uds and pull times removed?  Just the container start time?  Wonder if it makes sense to public two metrics here: the total time as we do now, and the time without uds and pull?  Apologies if we're already doing this and I've just overlooked it.

skinowski:
The remaining time is container create & setup and any delay during slot queue wait. However, noticed that this simplistic subtraction will not work as I intended since the container might not necessarily serve the request that triggered it. I'm going to close this PR since converting the request-vs-container processing to 1-1 will alleviate these issues and will make such accounting very trivial.

PR #1576, opened 2020-03-30
Increase maximum page size to 500

tcoupland:
One hundred is a sensible limit, but five hundred gives us a lot more head room. The entities are only small after all.


davidkhala:
Is there anything blocking on the way to get it merged?

PR #1613, opened 2025-04-18
support podman auth json config

marvin659:
Update Fn to support podman config location: /run/user/<uid>/containers/auth.json.

In CloudShell OL8 environment, podman auth config is available but not docker config so this change is necessary so that CloudShell user does not need to set the FN_DOCKER_AUTH variable.

If the file is found, it will take precedence over the docker config. If it is not found or not valid, Fn will fallback to docker config.


yufeifly:
您好，您的邮件已收到，谢谢。——熊宇飞

api/agent/drivers/docker/registry.go
```diff
@@ -19,17 +21,38 @@ func registryFromEnv() (map[string]driverAuthConfig, error) {
...
+	}
+
+	authJson := fmt.Sprintf("/run/user/%s/containers/auth.json", currentUser)
+	_, err = os.Stat(authJson)
+	if err != nil && os.IsNotExist(err) {
+		return nil, err
+	}
+	auths, err := docker.NewAuthConfigurations(strings.NewReader(authJson))
+	if err != nil {
+		logrus.WithError(err).Infof("Failed to parse %s podman config. Fall back to docker config.", authJson)
```
ahjones, line 51: Rather than log from here, return an error that can be logged at a higher level if necessary. It's not for this method to know that there will be a fallback to docker config, this method is just meant to get the config for Podman.

marvin659:
No required. Fixed the Fn Cli instead.

Issue #1610, opened 2024-12-22
Sock Bind error in with Python in local - Mac with Rancher


jmalbarran:
I'm testing Apache Functions and I'm trying to test my function locally

For this, I have started the local server with
```
fn start --log-level DEBUG --port 8080
```

and I'm trying to create, deploy and invoke a test function with

```
fn use context default
fn update context registry mylocalregistry

fn -v init --runtime python fn-local

cd fn-local

fn -v deploy --create-app --app app-local --local
fn invoke app-local fn-local
```

But the invocation (last line) fails with `Error invoking function. status: 502 message: Container failed to initialize, please ensure you are using the latest fdk and check the logs`

In the server logs I see that the problem is a python socket creation

```
...
time="2024-12-21T19:48:40Z" level=debug msg="  File \"/python/fdk/__init__.py\", line 62, in start\n" app_id=01JFN9NJFZNG8G00GZJ0000001 container_id=01JFNC7J7ZNG8G00GZJ0000003 fn_id=01JFNC6GG1NG8G00GZJ0000001 image="mylocalregistry/fn-local:0.0.2" tag=stderr
time="2024-12-21T19:48:40Z" level=debug msg="    sock.bind(phony_socket_path)\n" app_id=01JFN9NJFZNG8G00GZJ0000001 container_id=01JFNC7J7ZNG8G00GZJ0000003 fn_id=01JFNC6GG1NG8G00GZJ0000001 image="mylocalregistry/fn-local:0.0.2" tag=stderr
time="2024-12-21T19:48:40Z" level=debug msg="PermissionError: [Errno 1] Operation not permitted\n" app_id=01JFN9NJFZNG8G00GZJ0000001 container_id=01JFNC7J7ZNG8G00GZJ0000003 fn_id=01JFNC6GG1NG8G00GZJ0000001 image="mylocalregistry/fn-local:0.0.2" tag=stderr
time="2024-12-21T19:48:40Z" level=info msg="hot function terminated" app_id=01JFN9NJFZNG8G00GZJ0000001 container_id=01JFNC7J7ZNG8G00GZJ0000003 cpus= error="container exit code 1" fn_id=01JFNC6GG1NG8G00GZJ0000001 idle_timeout=30 image="mylocalregistry/fn-local:0.0.2" memory=256

...
```

I have already tried to run the starting with root  `sudo fn start` with no change

I'm using Mac Sequioa with Rancher

Issue #1608, opened 2023-12-28
error="docker version is too old. Required: 17.10.0-ce Found: 4.8.1"


MikeyCarter:
Running:
`/usr/local/bin/fn start  --log-level DEBUG --port 8101
`

Getting the error that the latest version of Podman is tool old.  So the issue is podman version reporting is different than docker so even the podman-docker package won't allow me to start.

Need a way to either skip this check or make it podman aware.

Full logs:
```
2023/12/28 13:04:44 ¡¡¡ 'fn start' should NOT be used for PRODUCTION !!! see https://github.com/fnproject/fn-helm/
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
time="2023-12-28T18:04:45Z" level=info msg="Setting log level to" fields.level=DEBUG
time="2023-12-28T18:04:45Z" level=info msg="Registering data store provider 'sql'"
time="2023-12-28T18:04:45Z" level=debug msg="creating new datastore" db=sqlite3
time="2023-12-28T18:04:45Z" level=debug msg="mysql does not support sqlite3"
time="2023-12-28T18:04:45Z" level=debug msg="postgres does not support sqlite3"
time="2023-12-28T18:04:45Z" level=debug msg="mysql does not support sqlite3"
time="2023-12-28T18:04:45Z" level=debug msg="postgres does not support sqlite3"
time="2023-12-28T18:04:45Z" level=info msg="Connecting to DB" url="sqlite3:///app/data/fn.db"
time="2023-12-28T18:04:45Z" level=info msg="datastore dialed" datastore=sqlite3 max_idle_connections=256 url="sqlite3:///app/data/fn.db"
time="2023-12-28T18:04:45Z" level=debug msg="mysql does not support sqlite3"
time="2023-12-28T18:04:45Z" level=debug msg="postgres does not support sqlite3"
[...]

Issue #1595, opened 2022-11-15
External API invocation is failing with 502 error


suneelkandali:
I wrote a Java function with below http client code snippet and deployed to my local Fn then function invocation failed with 502 error

I have Rancher Desktop on my local

HttpClient client = HttpClient.newHttpClient();
 HttpRequest request = HttpRequest.newBuilder()
         .uri(URI.create("http://webcode.me"))
         .GET() // GET is default
         .build();

 HttpResponse<Void> response = client.send(request,
         HttpResponse.BodyHandlers.discarding());

 System.out.println(response.statusCode()); 
Error message in log

time="2022-11-02T21:37:59Z" level=debug msg="Inside Java Hello World function  suneel
n" action="server.handleFnInvokeCall)-fm" app_id=01GGVTFKZYNG8G00GZJ0000001 call_id=01GGX4F2ESNG8G00GZJ000000W fn_id=01GGWWQSQGNG8G00GZJ0000006 image="sunkandalidocker/fn-process-aggregator-stream-consumer:0.0.18" user_log=true

time="2022-11-02T21:38:00Z" level=info msg="hot function terminating" app_id=01GGVTFKZYNG8G00GZJ0000001 container_id=01GGX4F2ETNG8G00GZJ000000X cpus= error="Post http://localhost/call: EOF" fn_id=01GGWWQSQGNG8G00GZJ0000006 idle_timeout=30 image="sunkandalidocker/fn-process-aggregator-stream-consumer:0.0.18" memory=128

Issue #1577, opened 2020-04-01
`fn invoke myapp hello` on Windows: Error invoking function. status: 502 message: Container failed to initialize


ceefour:
**Description**

`fn invoke myapp hello` failed with:

Error invoking function. status: 502 message: Container failed to initialize, please ensure you are using the latest fdk and check the logs

**Steps to reproduce the issue:**

```
D:\project_amanah\fn-tutorial>fn init --runtime node hello
Creating function at: ./hello
Function boilerplate generated.
func.yaml created.

D:\project_amanah\fn-tutorial>cd hello

D:\project_amanah\fn-tutorial\hello>fn create app myapp
Successfully created app:  myapp

D:\project_amanah\fn-tutorial\hello>fn deploy --app myapp --local
Deploying hello to app: myapp
Bumped to version 0.0.2
Building image hello:0.0.2 ......................................
Updating function hello using image hello:0.0.2...
Successfully created function: hello with hello:0.0.2

D:\project_amanah\fn-tutorial\hello>fn invoke myapp hello
Error invoking function. status: 502 message: Container failed to initialize, please ensure you are using the latest fdk and check the logs
```

**Describe the results you received:**

```
Error invoking function. status: 502 message: Container failed to initialize, please ensure you are using the latest fdk and check the logs
```

**Describe the results you expected:**

`hello` function executed.

**Additional information you deem important (e.g. issue happens only occasionally):**

Always reproducible. Also reproducible with go:

```
D:\project_amanah\fn-tutorial>fn init --runtime go hello-go
Creating function at: ./hello-go
Function boilerplate generated.
func.yaml created.

D:\project_amanah\fn-tutorial>cd hello-go

D:\project_amanah\fn-tutorial\hello-go>fn deploy --app myapp --local
Deploying hello-go to app: myapp
Bumped to version 0.0.2
Building image hello-go:0.0.2 .............................................................................................................................
Updating function hello-go using image hello-go:0.0.2...
Successfully created function: hello-go with hello-go:0.0.2

D:\project_amanah\fn-tutorial\hello-go>fn invoke myapp hello-go
Error invoking function. status: 502 message: Container failed to initialize, please ensure you are using the latest fdk and check the logs
```

**Output of `fn version` (CLI command):**

```
Client version is latest version: 0.5.96
Server version:  0.3.749
```

**Additional environment details (OSX, Linux, flags, etc.):**

Windows 10 Pro.

geroe1:
[...]

PR #1623, opened 2026-01-06
Fix permission issue when FDK tried to bind socket (plus other change required to start up FnServer without pain)

marvin659:
Fix permission issue when FDK tried to chmod or bind socket to file and deprecated cgroup v1 mem swappiness api, plus other minor fixes that required for Fn Server to function normally with the latest container engine desktop such as Podman and Rancher.

- Link to issue this resolves
  - https://github.com/fnproject/fn/issues/1577
  - https://github.com/fnproject/fn/issues/1610
  - https://github.com/fnproject/fn/issues/1608
  - https://github.com/fnproject/fn/issues/1595

- What I did
  - Remove MemorySwappiness as it has been deprecated in cgroup v2
    - See: https://github.com/opencontainers/runtime-spec/issues/1005 on why MemorySwappiness is removed.
      - Tejun: "It’s not very clear what swappiness encodes. A lot of it is compared to file-backed IOs, how [un]favorable IOs for anonymous memory are considering their inherently higher randomness. As such, it’s more a function of the underlying hardware than workloads. Also, the implementation wasn’t quite right either – iirc, the behavior would differ depending on who’s reclaiming."
  - Mount the IOFS path from host vm to Fn Container with "Z" options, which is required for host that is on SELinux. "Z" is the option for unshared private access. Private Unshared access is enough since we created a separate iofs path for each instance of FnServer container. The option has no effect on host that has not enabled or do not support SELinux, or if user uses container engine named volume as the iofs path.
  - Allow user to use named volume for Fn Server to create the unix socket file. It resolves the issue that the unix socket file could not be hosted under $HOME/.fn directory (See: https://github.com/fnproject/fn/issues/1577#issuecomment-1297736260) for Podman/Rancher desktop.
  - Change the way the IOFS path mapping (bind mount) from host VM to Fn Container, which is required if we want to allow users to use container engine named volume as iofs path. 
    - bind mount change: 
      - before fix: \<IOFS Path in Host VM>/\<temp directory created by Fn Server>:/tmp/iofs
        - (FN_LISTENER was /tmp/iofs)
      - after fix:      \<IOFS Path in Host VM or named volume>:/tmp/iofs/\<temp directory created by Fn Server>
        - FN_LISTENER is set to /tmp/iofs/\<temp directory created by Fn Server>
  - Disable docker version check since customers could use any container engine desktop other than docker desktop.
  - Update dind to 24.0.9

- How to verify it
  - Verified by running Fn Server with Podman and Rancher Desktop, create and invoke local Functions with Fn CLI

  - Fix a couple of critical issues about starting up Fn Server on latest container engine desktop

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ jobs:
           mkdir tmp
           cd tmp
           sudo rm -rf /usr/local/go
-          wget https://storage.googleapis.com/golang/go$GOVERSION.$OS-$ARCH.tar.gz
+          wget https://go.dev/dl/go$GOVERSION.$OS-$ARCH.tar.gz
           sudo tar -C /usr/local -xzf go$GOVERSION.$OS-$ARCH.tar.gz
       - run: go version
       # install latest Docker
--- a/api/agent/call.go
+++ b/api/agent/call.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"path/filepath"
 	"strings"
 	"time"
 
@@ -251,7 +250,6 @@ func (a *agent) GetCall(opts ...CallOpt) (Call, error) {
 	if c.Call.Config == nil {
 		c.Call.Config = make(models.Config)
 	}
-	c.Call.Config["FN_LISTENER"] = "unix:" + filepath.Join(iofsDockerMountDest, udsFilename)
 	c.Call.Config["FN_FORMAT"] = "http-stream" // TODO: remove this after fdk's forget what it means
 	// TODO we could set type here too, for now, or anything else not based in fn/app/trigger config
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ The fastest way to experience Fn is to follow the quickstart below, or you can j
 
 ### Pre-requisites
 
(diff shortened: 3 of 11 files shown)


.circleci/config.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           mkdir tmp
           cd tmp
           sudo rm -rf /usr/local/go
-          wget https://storage.googleapis.com/golang/go$GOVERSION.$OS-$ARCH.tar.gz
```
marvin659: deprecated path

api/agent/call.go
```diff
@@ -251,7 +250,6 @@ func (a *agent) GetCall(opts ...CallOpt) (Call, error) {
 	if c.Call.Config == nil {
 		c.Call.Config = make(models.Config)
 	}
-	c.Call.Config["FN_LISTENER"] = "unix:" + filepath.Join(iofsDockerMountDest, udsFilename)
```
marvin659: moving it to agent.go as shown above. Changed the location from "iofsDockerMountDest" to "iofsDockerMountDest"/<temp dir created by fn server", which is necessary change if we allow user to use named volume as the location to host unix socket file.

pdenness (approved):

